PR #38, opened 2014-07-17
Initial mavenization, including jar signing. 

azymer:
Had to change one test to make it compatible with newer version of testng.  Tests all pass in intellij and maven.  The only explicit dependency in the pom is htsjdk, which could be great or could couple picard too tightly to htsjdk.


vdauwera:
Closing this PR since it is positively ancient.

PR #51, opened 2014-09-02
Add discard level option to DownsampleSam and ensure non-primary reads a...

mp15:
Add discard level option to DownsampleSam and ensure non-primary reads are retained by default.  PR samtools/htsjdk#91 is a prerequisite for this.


vdauwera:
Closing this PR as per @bradtaylor 's comment of Dec. 18.

PR #87, opened 2014-09-29
Additonal output stats to CollectInsertSizeMetrics.  

travc:
Small addition to the output of CollectInsertSizeMetrics.
Gives 2.5%, 25%, 75%, and 97.5% percentile values useful for making boxplots and such.

Just uses the methods already available in the Histogram class, so it is a very simple addition.
New output is at the end of the current stats fields (though before SAMPLE, LIBRARY, and READ_GROUP), so impact on tools which process the output might exist, but should be minimal.

### DIFF
--- a/src/java/picard/analysis/CollectWgsMetrics.java
+++ b/src/java/picard/analysis/CollectWgsMetrics.java
@@ -14,6 +14,8 @@ import htsjdk.samtools.util.IOUtil;
 import htsjdk.samtools.util.Log;
 import htsjdk.samtools.util.ProgressLogger;
 import htsjdk.samtools.util.SamLocusIterator;
+import htsjdk.samtools.util.Interval;
+import htsjdk.samtools.util.IntervalList;
 import picard.cmdline.CommandLineProgram;
 import picard.cmdline.Option;
 import picard.cmdline.StandardOptionDefinitions;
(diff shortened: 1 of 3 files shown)


vdauwera:
What's the status of this pull request? If still relevant, is it waiting on us (@nh13) or @travc ?

travc:
I haven't been keeping up with the code, but think it was pretty much done from my POV.

yfarjoun:
Thanks for the bug fix and the extra options.

I am wondering what is the use-case for an interval list for WGS metrics?

cheers,

Yossi.

travc:
I'm mapping to a fairly gappy ref.  It also includes some scaffolds which are unplaced as well as a male-only chromosome.  When trying to evaluate the quality of mapping, it really only makes sense to look at the "good" portions of the reference.  It is also a big speedup to just compute metrics for a few characteristic regions.

Generally, the use case will be when you want to metrics for different regions... probably different sorts or regions.  So one might run metrics on a just some relatively conserved regions.  Alternatively, one might want to look at the metrics just for a region (or some regions) which are highly variable to check that reference is close enough to generate reasonable results.

I can think of lots of other cases when you want metrics just on specific regions, but I think you get the idea.  The interval list implementation seemed to be the most general way of doing it as well as fitting in with how other picard tools does it.

PR #118, opened 2014-12-02
Modify GenotypeConcordance to explicitly require that the VCF files be indexed if an interval file is specified

gbggrant:
Modify GenotypeConcordance to explicitly require that the VCF files be indexed if an interval file is specified.  Otherwise it would blow up, somewhat more obscurely, for lack of an index file.

### DIFF
--- a/src/java/picard/vcf/GenotypeConcordance.java
+++ b/src/java/picard/vcf/GenotypeConcordance.java
@@ -31,6 +31,7 @@ import htsjdk.samtools.util.Log;
 import htsjdk.samtools.util.PeekableIterator;
 import htsjdk.samtools.util.ProgressLogger;
 import htsjdk.samtools.util.SequenceUtil;
+import htsjdk.tribble.Tribble;
 import htsjdk.variant.variantcontext.Genotype;
 import htsjdk.variant.variantcontext.VariantContext;
 import htsjdk.variant.variantcontext.VariantContextComparator;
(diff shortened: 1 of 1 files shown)


vdauwera:
Assuming @bradtaylor doesn't mind being unassgined, this sounds like something that could be done by one of our consultants.

gbggrant:
I think all that's left is a final ok.  Kind of fell through the cracks.

vdauwera:
@bradtaylor @gbggrant Sorry for bugging you guys with this, but might this final ok be forthcoming? Would like to clean up the issue tracker, plug the cracks, etc.

src/java/picard/vcf/GenotypeConcordance.java
```diff
@@ -138,7 +138,8 @@ public static void main(final String[] args) {
         IntervalList intervals = null;
         SAMSequenceDictionary intervalsSamSequenceDictionary = null;
         if (usingIntervals) {
-            log.info("Loading up region lists.");
+            USE_VCF_INDEX = true;   // If intervals are specified, the VCF files must be indexed
```
nh13: I would suggest put this in the custom command line validation or emit a log warning stating you are setting ```UCE_VCF_INDEX```.

nh13:
@vdauwera I added a comment here.  @bradtaylor @gbggrant should agree on the final solution before merging.

bradtaylor:
I like the custom command line validation. Thumbs up!

PR #149, opened 2015-01-15
Separate out the actual work for FastqToSam so an outside caller can dir...

geoffjentry:
...ectly pass in FastqReaders

### DIFF
--- a/src/java/picard/sam/FastqToSam.java
+++ b/src/java/picard/sam/FastqToSam.java
@@ -128,6 +128,36 @@ public class FastqToSam extends CommandLineProgram {
 
     private static final SolexaQualityConverter solexaQualityConverter = SolexaQualityConverter.getSingleton();
 
+    /**
+     * Looks at fastq input(s) and attempts to determine the proper quality format
+     *
+     * Closes the reader(s) by side effect
+     *
+     * @param reader1 The first fastq input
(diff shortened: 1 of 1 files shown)


src/java/picard/sam/FastqToSam.java
```diff
@@ -135,41 +165,62 @@ public static void main(final String[] argv) {
...
+
+        QUALITY_FORMAT = FastqToSam.determineQualityFormat(reader, reader2, QUALITY_FORMAT);
+
+        final SAMFileHeader header = createSamFileHeader();
+        final SAMFileWriter writer = new SAMFileWriterFactory().makeSAMOrBAMWriter(header, false, OUTPUT);
+
+        reader = fileToFastqReader(FASTQ);
+        if (FASTQ2 != null) {
+            reader2 = fileToFastqReader(FASTQ2);
+        }
```
jacarey: Its kind of annoying that we have to open the readers to determine quality then close them.. them reopen new ones to start reading.

geoffjentry: I agree, which was part of my problem in the first place as I was sucking the data through on a single named pipe. You could potentially tee the input or something like that but admittedly I was trying to do the minimum amount of change to enable what I needed

geoffjentry:
Is that a :+1: @jacarey ?

jacarey:
Yes sorry. (thumbsup)

PR #142, opened 2014-12-22
Fix overridable option fields.

jacarey:
Added a test and some java reflection to set all hidden fields to the child value.

PR #150, opened 2015-01-16
A recent change to the command line parsing causes problems with scala b...

geoffjentry:
...ased CLPs. In cases where the

new system fails, fall back to the original system


geoffjentry:
Reviewed by @jacarey

nh13:
Which commit was the recent change?

geoffjentry:
It was Pull Request #142

PR #154, opened 2015-01-26
Enable file source for Bam Index creation.

jacarey:


nh13:
:+1:  but can you specify which commit caused the regression originally?

jacarey:
c0cdc5a03715004b4633c1f6a049b426cac6e4a7 
was the commit hash the caused the regression.

Issue #157, opened 2015-02-04
PERCENT_DUPLICATION Calculation Bug


nathanhaigh:
The `METRICS_FILE` generated by `MarkDuplicates` does not show the `PERCENT_DUPLICATION` as expected. It is in fact a fraction and not a percentage. i.e. it is 2-orders of magnitude less than one might expect given the column name.

nh13:
Unfortunately this is a bug that we will not fix but it is documented in the FAQ: http://broadinstitute.github.io/picard/faq.html

Many folks including ourselves rely on the names and changing that would affect other systems.  My apologies in advance.

PR #158, opened 2015-02-04
PERCENT_DUPLICATION is now a percentage, not a fraction

nathanhaigh:
This is a fix for issue #157


nh13:
Wont fix as described on #157.

PR #168, opened 2015-03-13
Kb gcbias take2

kbergin:
Changed CollectGcBiasMetrics to work as a SinglePassSam

### DIFF
--- a/src/java/picard/analysis/CollectGcBiasMetrics.java
+++ b/src/java/picard/analysis/CollectGcBiasMetrics.java
@@ -27,23 +27,12 @@ package picard.analysis;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMReadGroupRecord;
 import htsjdk.samtools.SAMRecord;
-import htsjdk.samtools.SAMSequenceDictionary;
-import htsjdk.samtools.SamReader;
-import htsjdk.samtools.SamReaderFactory;
 import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.reference.ReferenceSequence;
-import htsjdk.samtools.reference.ReferenceSequenceFile;
-import htsjdk.samtools.reference.ReferenceSequenceFileFactory;
-import htsjdk.samtools.util.CloserUtil;
 import htsjdk.samtools.util.IOUtil;
-import htsjdk.samtools.util.Log;
-import htsjdk.samtools.util.PeekableIterator;
-import htsjdk.samtools.util.ProgressLogger;
 import htsjdk.samtools.util.QualityUtil;
 import htsjdk.samtools.util.SequenceUtil;
 import htsjdk.samtools.util.StringUtil;
-import picard.PicardException;
-import picard.cmdline.CommandLineProgram;
 import picard.cmdline.CommandLineProgramProperties;
 import picard.cmdline.Option;
 import picard.cmdline.StandardOptionDefinitions;
@@ -73,21 +62,11 @@ import java.util.List;
         usageShort = "Collects information about GC bias in the reads in the provided SAM or BAM",
         programGroup = Metrics.class
 )
-public class CollectGcBiasMetrics extends CommandLineProgram {
+public class CollectGcBiasMetrics extends SinglePassSamProgram {
     /** The location of the R script to do the plotting. */
     private static final String R_SCRIPT = "picard/analysis/gcBias.R";
 
     // Usage and parameters
-
-    @Option(shortName = StandardOptionDefinitions.REFERENCE_SHORT_NAME, doc = "The reference sequence fasta file.")
-    public File REFERENCE_SEQUENCE;
-
-    @Option(shortName = StandardOptionDefinitions.INPUT_SHORT_NAME, doc = "The BAM or SAM file containing aligned reads.  " +
-            "Must be coordinate-sorted.")
-    public File INPUT;
-
-    @Option(shortName = StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "The text file to write the metrics table to.")
-    public File OUTPUT;
 
     @Option(shortName = "CHART", doc = "The PDF file to render the chart to.")
     public File CHART_OUTPUT;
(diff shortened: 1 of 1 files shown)


src/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -82,13 +71,6 @@
     @Option(shortName = StandardOptionDefinitions.REFERENCE_SHORT_NAME, doc = "The reference sequence fasta file.")
```
yfarjoun: I think you need to remove the reference as it is already a standard argument of command_line program. I'm not sure why it was an explicit field before....perhaps someone else can explain.

src/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -245,19 +195,17 @@ protected int doWork() {
         String title = INPUT.getName().replace(".duplicates_marked", "").replace(".aligned.bam", "");
 
         // Qualify the title with the library name iff it's for a single sample
-        final List<SAMReadGroupRecord> readGroups = sam.getFileHeader().getReadGroups();
+        final List<SAMReadGroupRecord> readGroups = save_header.getReadGroups();
         if (readGroups.size() == 1) {
```
yfarjoun: given that this is your only usage of the header, I'd pull out the information you need (during the initialize) and store it (it's a String) rather than holding on to an old header.

src/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -101,107 +83,75 @@
...
+    // Histograms to track the number of windows at each GC, and the number of read starts
+    // at windows of each GC
+    final int[] windowsByGc = new int[101];
+    final int[] readsByGc = new int[101];
+    final long[] basesByGc = new long[101];
+    final long[] errorsByGc = new long[101];
+    static int lastContig = -1;
+    byte[] gc;
+    byte[] refBases;
+    private SAMFileHeader save_header;
```
yfarjoun: naming convention for variables: "camelCase" with first letter lower-case

src/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -101,107 +83,75 @@
...
     // Used to keep track of the total clusters as this is kinda important for bias
     private int totalClusters = 0;
     private int totalAlignedReads = 0;
+    // Histograms to track the number of windows at each GC, and the number of read starts
+    // at windows of each GC
+    final int[] windowsByGc = new int[101];
+    final int[] readsByGc = new int[101];
+    final long[] basesByGc = new long[101];
+    final long[] errorsByGc = new long[101];
+    static int lastContig = -1;
```
yfarjoun: there's a constant in SamRecord called NO_ALIGNMENT_REFERENCE_INDEX. Use this instead of defining your own.

yfarjoun:
My review is possibly not the official review, but I wanted to take a peek...you should not be checking in the picard.iml & picard.ipr files. So you'll need to revert these and re-commit.

src/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -101,107 +83,75 @@
...
+        save_header = header;
+    }
+    ////////////////////////////////////////////////////////////////////////////
+    // Loop over the reference and the reads and calculate the basic metrics
+    ////////////////////////////////////////////////////////////////////////////
+    @Override
+    protected void acceptRead(final SAMRecord rec, final ReferenceSequence ref) {
+        //if read is unaligned then ref is passed in as null
+        if (ref == null) {
+            if (!rec.getReadPairedFlag() || rec.getFirstOfPairFlag()) ++this.totalClusters;
```
mcovarr: This statement is on both sides of the `if (ref == null)` conditional.  Can this be pulled above the `if (ref == null)` and then that `if` branch eliminated?  I'm not 100% sure it can, there's some intervening logic in the `else` branch before this statement appears, so this is really a question.  :smile:

src/java/picard/analysis/CollectGcBiasMetrics.java
nh13: Also, use camel case.

src/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -101,107 +83,75 @@
...
-
     // Used to keep track of the total clusters as this is kinda important for bias
     private int totalClusters = 0;
     private int totalAlignedReads = 0;
+    // Histograms to track the number of windows at each GC, and the number of read starts
+    // at windows of each GC
+    final int[] windowsByGc = new int[101];
+    final int[] readsByGc = new int[101];
+    final long[] basesByGc = new long[101];
+    final long[] errorsByGc = new long[101];
```
mcovarr: A symbolic constant for the 101 would be good.

kbergin:
Ok thank you! I will fix the things you and others have mentioned and then I'm not sure what the Picard.iml and Picard.ipr files are I don't remember touching them but I will figure out where I went wrong and fix it. 

Thanks!
Kylee

src/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -101,107 +83,75 @@
...
+    @Override
+    protected void acceptRead(final SAMRecord rec, final ReferenceSequence ref) {
+        //if read is unaligned then ref is passed in as null
+        if (ref == null) {
+            if (!rec.getReadPairedFlag() || rec.getFirstOfPairFlag()) ++this.totalClusters;
+        } else {
+            //only do the recalculation of gc if current ref is different from last ref
+            if (ref.getContigIndex() != lastContig) {
+                refBases = ref.getBases();
+                StringUtil.toUpperCase(refBases);
```
mcovarr: I think you want
```
refBase = StringUtil.toUpperCase(refBases);
```
Java strings are immutable. `StringUtil.toUppercase` returns the uppercase version of its argument, but that's being discarded here.

src/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -101,107 +83,75 @@
...
     public boolean IS_BISULFITE_SEQUENCED = false;
 
-    private static final Log log = Log.getInstance(CollectGcBiasMetrics.class);
-
     // Used to keep track of the total clusters as this is kinda important for bias
     private int totalClusters = 0;
     private int totalAlignedReads = 0;
+    // Histograms to track the number of windows at each GC, and the number of read starts
+    // at windows of each GC
+    final int[] windowsByGc = new int[101];
```
mcovarr: It looks like most of the new fields could be `private`.

src/java/picard/analysis/CollectGcBiasMetrics.java
kbergin: You're right, I'm not sure why it was before. It gave me an error when I had others that were duplicates but didn't have a problem with that one. I'll remove it.

src/java/picard/analysis/CollectGcBiasMetrics.java
kbergin: Good idea!

src/java/picard/analysis/CollectGcBiasMetrics.java
yfarjoun: @mcovarr I think that this is a verbatim copy of the original code behavior. I agree that as written it is effectively a noop, but since @kbergin is not fixing bugs here we need the results to agree. It might make sense in a subsequent fix to change this, but right now I think she shouldn't change any functional aspect of the code. (my opinion)

src/java/picard/analysis/CollectGcBiasMetrics.java
kbergin: I think you're right, it could be moved outside of the if and the else could become 
if(ref!=null)

src/java/picard/analysis/CollectGcBiasMetrics.java
tfenne: It's a little harder to see now that the declaration has moved up the class level, but refBases is a byte[], not a String.  So in this case StringUtil.toUpperCase(byte[]) is being called, which does in fact mutate the byte[] so as not to allocate a second copy.

yfarjoun: evil. :-)

kbergin: Cool! Hi Tim!

src/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -211,7 +161,7 @@ protected int doWork() {
         for (int i = 0; i < windowsByGc.length; ++i) {
```
kbergin: IntelliJ is warning me that the variable minimumWindowsToCountInSummary is never used despite being calculated. Is that something we need? @yfarjoun

yfarjoun: This is also true in the original code...feel free to leave it despite the warning.

src/java/picard/analysis/CollectGcBiasMetrics.java
mcovarr: :open_mouth: 

Hi Tim, thanks for the correction and explanation.  I probably should have realized that `StringUtils.toUpperCase` wasn't operating on a `String` since `String#toUpperCase` could do that job.

yfarjoun:
when you are done responding to the reviewer's comments you should comment so that they will not to take a final look and give you thumbs-up!

kbergin:
Oh ok, thanks @yfarjoun! Hey Everyone - I think I fixed everything. Reverted the picard files I accidentally pushed. Updated camelCase and changed that variable to a string to get the sample name. I was a little concerned about that because in the original code it was only pulled if the sample being run was a single sample. But I can't think of an instance where more than one sample would be run through this together. Correct me if I'm wrong about that, please! Made variables private and changed the if else to be just an if statement.

The only thing I didn't change and am unsure about still is the "symbolic constant" that @mcovarr suggested. As you mentioned I didn't write that part of the code so I am not sure why 101 was chosen or what the constant should be set to instead. Anyone have an input here?

yfarjoun:
Argh!! sorry. found another untested bug. Need to add unit tests!!!

yfarjoun:
oops. wrong PR. don't worry @kbergin all is well!

kbergin:
Phew!

mcovarr:
Regarding the symbolic constant, `private static final int NUM_HISTOGRAM_BUCKETS = 101;` would work for me.  Ideas for better names are welcome!  I'm definitely not suggesting a different value for this constant, I have no idea why this was chosen either.  :smile:

yfarjoun:
To allow for nice round numbers including 0 and 100, you need 101 buckets.

mcovarr:
That would be a great Javadoc comment above the constant!  :smile:

kbergin:
Hi All - Changed the 101's to constants and pushed again.

kbergin:
@mcovarr if it's all good, can I get a thumbs up on this so I can merge and make a new branch to work on something else? :)

mcovarr:
Purely a nitpick, but the convention for constants in Java is `ALL_CAPS`, otherwise :+1:

kbergin:
@mcovarr Ah thanks! Sorry about that, I'm learning Java as I go. Fixed it!

kbergin:
@mcovarr What's the next step to have this merged? Is that something you do?

kbergin:
Oh no I did something!! What happened?

kbergin:
Phew

mattsooknah:
All that's left is to squash it into one commit + rebase on top of master. Then we can merge it.

kbergin:
Ok I think I did that @mattsooknah

kbergin:
YAY!

yfarjoun:
:-)

PR #170, opened 2015-03-16
Fix to the build file to expand lib jars before packaging the picard CLP...

tfenne:
... jar.  Fixes snappy integration.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -32,6 +32,7 @@
     <property name="src.www" value="src/www"/>
     <property name="lib" value="lib"/>
     <property name="dist" value="dist"/>
+    <property name="dist.tmp" value="dist/tmp"/>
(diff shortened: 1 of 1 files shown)


tfenne:
Hey @nh13 I just discovered that, probably as part of the transition to the single CLP, the picard CLP jar was being built with other lib jars nested inside of it, which was causing those libraries to not load properly.  I noticed this because I was running out of tmp space doing a sort because Snappy wasn't loading correctly.  The lib jars needs to be un-jarred first and then the resulting .class files included into the picard CLP jar.  This pull request is a small change to the build.xml to accomplish that.

nh13:
:+1: thanks for the fix and sorry for the embarrassing mistake.

PR #174, opened 2015-03-18
use ssh url for htsjdk git.

jacarey:
First pass picard sbt build.

Change git url to https on the fly

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ before_install:
   - curl -OL http://raw.github.com/craigcitro/r-travis/master/scripts/travis-tool.sh
   - chmod 755 ./travis-tool.sh
   - ./travis-tool.sh bootstrap
+  - sed -i 's/git@github.com:/https:\/\/github.com\//' build.xml
 script:
   - ant -lib lib/ant clone-htsjdk 
   - ant test
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ before_install:
   - curl -OL http://raw.github.com/craigcitro/r-travis/master/scripts/travis-tool.sh
   - chmod 755 ./travis-tool.sh
   - ./travis-tool.sh bootstrap
-  - sed -e 's_git@github.com:_https://github.com/_' build.xml
+  - sed -i -e 's_git@github.com:_https://github.com/_' build.xml
 script:
   - ant -lib lib/ant clone-htsjdk 
   - ant test
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ before_install:
   - curl -OL http://raw.github.com/craigcitro/r-travis/master/scripts/travis-tool.sh
   - chmod 755 ./travis-tool.sh
   - ./travis-tool.sh bootstrap
-  - sed -i 's/git@github.com:/https:\/\/github.com\//' build.xml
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -6,6 +6,7 @@ before_install:
   - curl -OL http://raw.github.com/craigcitro/r-travis/master/scripts/travis-tool.sh
   - chmod 755 ./travis-tool.sh
   - ./travis-tool.sh bootstrap
+  - sed -i 's/git@github.com:/https:\/\/github.com\//' build.xml
```
nh13: gorgeous

src/tests/java/picard/illumina/CheckIlluminaDirectoryTest.java
```diff
@@ -51,28 +51,30 @@ private void setUp() throws Exception {
         illuminaDir = IOUtil.createTempDir("ift_test", "IlluminaDir");
 
         interopDir = new File(illuminaDir, "InterOp");
-        if (!interopDir.mkdir()) {
+        if (!interopDir.exists() && !interopDir.mkdir()) {
```
nh13: why would these directories exist?  Do other tests create an ```InterOp``` directory?

jacarey: yes they do and since we are using parellell execution of tests now it is important to check. Alternatively we could single thread test like we currently do

.travis.yml
jacarey: that should work. I'll test it definitely prettier.

.travis.yml
```diff
@@ -6,6 +6,7 @@ before_install:
   - curl -OL http://raw.github.com/craigcitro/r-travis/master/scripts/travis-tool.sh
   - chmod 755 ./travis-tool.sh
   - ./travis-tool.sh bootstrap
+  - sed -e 's_git@github.com:_https://github.com/_' build.xml
```
mcovarr: Does this mean to write to stdout or is there a `-i` flag missing?

jacarey: whooops good catch.

src/tests/java/picard/illumina/CheckIlluminaDirectoryTest.java
nh13: I don't understand why this is needed still.  If this ```setup``` method is called for this class and this directory is specific for this set of tests, why would it already exist?  Is there another class that creates the same directory?  If so, we should modify the ```IOUtil.createTempDir``` call, even though I think it should create a unique tmp dir?  Basically, we should not need this check *and* be able to do parallel execution of tests.

jacarey: Because it is parallelizing on the data provider you have multiple instances of the same method potentially running at the same time. The temp dir is set up before the method not before each invocation of the method with different data providers.

 Perhaps we should turn of data provider parallel execution until we know are tests are constructed in a manner that makes them more easily parallelizable?

mcovarr:
The parts that aren't `build.sbt` are :+1: for me.  I don't know SBT yet so `build.sbt` is :question:

src/tests/java/picard/illumina/CheckIlluminaDirectoryTest.java
nh13: I understand now, so this fix should be fine for this test.  :+1:

PR #175, opened 2015-03-18
Add ability to split interval_list file on band multiples

gbggrant:
Added ability to taken an interval list and split into finer intervals based on band multiples in order to allow for finer granularity of scattering.


src/java/picard/util/IntervalListTools.java
```diff
@@ -74,6 +74,9 @@
     @Option(doc = "Whether to include filtered variants in the vcf when generating an interval list from vcf", optional = true)
     public boolean INCLUDE_FILTERED=false;
 
+    @Option(shortName = "BRK", doc = "Break up intervals at multiples of this value.  Set to 0 to not break up intervals", optional = true)
```
yfarjoun: Specify that the intervalList to be broken up is the resulting intervallist, not the input(s)

gbggrant:
Tests pass and addressed Yossi's comment on documenting the option

PR #177, opened 2015-03-19
Modified IntervalListTools so that if scattering we do the sorting and uniquification up front

gbggrant:
Modified IntervalListTools so that if scattering we do the sorting and uniquification up front (and not in the scatter method).
This was important for BREAK_BANDS_AT_MULTIPLES_OF where the unique in scatter was re-merging broken adjacent intervals.
Also made interval list file in scattered directories be NOT broken up into multiples.

### DIFF
--- a/src/java/picard/util/IntervalListScatterer.java
+++ b/src/java/picard/util/IntervalListScatterer.java
@@ -60,10 +60,14 @@ public class IntervalListScatterer {
         }
     }
 
-    public List<IntervalList> scatter(final IntervalList sourceIntervalList, final int scatterCount) {
+    public List<IntervalList> scatter(final IntervalList uniquedIntervalList, final int scatterCount) {
+        return scatter(uniquedIntervalList, scatterCount, false);
+    }
+
+    public List<IntervalList> scatter(final IntervalList sourceIntervalList, final int scatterCount, final boolean isUniqued) {
         if (scatterCount < 1) throw new IllegalArgumentException("scatterCount < 1");
 
-        final IntervalList uniquedList = sourceIntervalList.uniqued();
+        final IntervalList uniquedList = isUniqued ? sourceIntervalList : sourceIntervalList.uniqued();
         final long idealSplitLength = deduceIdealSplitLength(uniquedList, scatterCount);
 
         final List<IntervalList> accumulatedIntervalLists = new ArrayList<IntervalList>();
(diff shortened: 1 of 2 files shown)


src/java/picard/util/IntervalListScatterer.java
```diff
@@ -60,10 +60,14 @@ public int compare(final Interval o1, final Interval o2) {
         }
     }
 
-    public List<IntervalList> scatter(final IntervalList sourceIntervalList, final int scatterCount) {
+    public List<IntervalList> scatter(final IntervalList uniquedIntervalList, final int scatterCount) {
+        return scatter(uniquedIntervalList, false, scatterCount);
+    }
+
+    public List<IntervalList> scatter(final IntervalList sourceIntervalList, final boolean isUniqued, final int scatterCount) {
```
nh13: Could the ```isUniqued``` parameter be the third argument so that it looks like you appended to line 63?

src/java/picard/util/IntervalListScatterer.java
```diff
@@ -107,7 +110,7 @@ public int compare(final Interval o1, final Interval o2) {
                         } else {
                             // Push this interval into the next scatter; re-inject it into the queue, then advance the scatter.
                             intervalQueue.addFirst(interval);
-                            accumulatedIntervalLists.add(runningIntervalList);
+                            accumulatedIntervalLists.add(runningIntervalList.uniqued());
```
nh13: The ```uniqued``` method causes a sort each time through the loop.  Is there any performance issues here?

src/java/picard/util/IntervalListTools.java
```diff
@@ -68,7 +68,7 @@
     @Option(doc = "One or more lines of comment to add to the header of the output file.", optional = true)
     public List<String> COMMENT = null;
 
-    @Option(doc = "The number of files into which to scatter the resulting list by locus; in some situations, fewer intervals may be emitted.")
+    @Option(doc = "The number of files into which to scatter the resulting list by locus; in some situations, fewer intervals may be emitted.  Implies SORT=true, UNIQUE=true.")
```
nh13: it implies it if it is greater than one, but not if it is one.

src/java/picard/util/IntervalListScatterer.java
yfarjoun: George and I read the code carefully and concluded that .uniqued() it is only called once per output IntervalList. So, no.

src/java/picard/util/IntervalListScatterer.java
gbggrant: Yes.  That's better form.

src/java/picard/util/IntervalListTools.java
gbggrant: I'll update the wording here.

PR #183, opened 2015-03-24
Moved test so that we can get rid of testng as a compile time dependency...

jacarey:
Added maven dependency on htsjdk instead of cloning it out.


pgrosu:
Thank you :)

PR #187, opened 2015-03-30
Extracted a DownsamplingIterator from DownsampleSam - behaviour is...

tfenne:
unchanged.


tfenne:
@nh13, @yfarjoun - could one of you please take a peek at this and give me a thumbs up?  Thanks.

src/java/picard/sam/DownsamplingIterator.java
```diff
@@ -0,0 +1,125 @@
+package picard.sam;
```
nh13: why is this not in htsjdk?

tfenne: I guess it could be; I just extracted it from DownsampleSam and put it into the same package.  Want me to shift it over?

nh13: I like the idea of iterators being available in htsjdk, so folks do not need to pull in Picard.  This tool is *really* useful as so many folks downsample.

PR #188, opened 2015-03-30
change for overriding reference

yfarjoun:


nh13:
If it is required, then why do we have the code linked below here?  It would seem someone wanted to allow this to run without a reference.
https://github.com/broadinstitute/picard/pull/188/files#diff-0d1f7e06cdd1e13a2d0f7453706ef3c4R119

yfarjoun:
You are right @nh13. I changed it back to optional, but now it appears in the default help and the doc is different.

nh13:
:+1:

PR #189, opened 2015-03-31
Fix NPE when nonstandard bases are found

mattsooknah:
Deal with an edge case when non-standard (i.e. not ACGTN) bases are found in the reference.


gbggrant:
Looks good.  :+1:

PR #190, opened 2015-04-03
Add a new method for scattering intervals whereby we try to estimate the

nh13:
sizes of interval lists that have not been created.  This should produce
a more balanced set of interval lists, while fixing the maximum number
of scatters.

### DIFF
--- a/src/java/picard/util/IntervalListScatterer.java
+++ b/src/java/picard/util/IntervalListScatterer.java
@@ -2,6 +2,7 @@ package picard.util;
 
 import htsjdk.samtools.util.Interval;
 import htsjdk.samtools.util.IntervalList;
+import picard.PicardException;
 
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -33,7 +34,16 @@ public class IntervalListScatterer {
          * which is one of the objectives of scattering.</li>
          * </ol>
          */
-        BALANCING_WITHOUT_INTERVAL_SUBDIVISION
+        BALANCING_WITHOUT_INTERVAL_SUBDIVISION,
+        /**
+         * A scatter approach that differs from {@link Mode#BALANCING_WITHOUT_INTERVAL_SUBDIVISION}.
+         * <ol>
+         * <li>We try to balance the number of unique bases in each interval list by estimating the remaining interval lists sizes.  This is 
+         * computed from the total number of unique bases and the bases we have consumed.  This means that the interval list with the most
+         * number of unique bases is at most the ideal split length larger than the smallest interval list (unique # of bases).</li>
+         * </ol>         
+         */
+        BALANCING_WITHOUT_INTERVAL_SUBDIVISION_WITH_OVERFLOW
     }
 
     private final Mode mode;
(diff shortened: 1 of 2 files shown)


src/java/picard/util/IntervalListScatterer.java
```diff
@@ -33,7 +34,16 @@
          * which is one of the objectives of scattering.</li>
          * </ol>
          */
-        BALANCING_WITHOUT_INTERVAL_SUBDIVISION
+        BALANCING_WITHOUT_INTERVAL_SUBDIVISION,
+        /**
+         * A scatter approach that differs from {@link Mode#BALANCING_WITHOUT_INTERVAL_SUBDIVISION} in a few ways.
```
mcovarr: This seems like only one way. :smile:  So the ol probably should go, or it could be fleshed out with more li's describing how this differs.

mcovarr:
Tests for this new scattering method?

src/java/picard/util/IntervalListScatterer.java
```diff
@@ -63,6 +74,20 @@ public int compare(final Interval o1, final Interval o2) {
...
+    /** Helper for the scatter method */
+    private boolean shouldAddToRunningIntervalList(final long idealSplitLength, final long projectedSize, final double projectedSizeOfRemainingDivisions) {
+        switch (mode) {
+            case INTERVAL_SUBDIVISION:
+            case BALANCING_WITHOUT_INTERVAL_SUBDIVISION:
+                return (projectedSize <= idealSplitLength);
+            case BALANCING_WITHOUT_INTERVAL_SUBDIVISION_WITH_OVERFLOW:
+                return (projectedSize <= idealSplitLength || idealSplitLength < projectedSizeOfRemainingDivisions);
+            default:
+                throw new PicardException("Unknown mode: " + mode.name());
```
jacarey: This default really would never happen. Might be better to structure it with 
      case BALANCING_WITHOUT_INTERVAL_SUBDIVISION_WITH_OVERFLOW:
                return (projectedSize <= idealSplitLength || idealSplitLength < projectedSizeOfRemainingDivisions);
      default:
                return (projectedSize <= idealSplitLength);

src/java/picard/util/IntervalListScatterer.java
```diff
@@ -130,6 +169,20 @@ public int compare(final Interval o1, final Interval o2) {
         if (!runningIntervalList.getIntervals().isEmpty()) {
             accumulatedIntervalLists.add(runningIntervalList.uniqued());
         }
+        
+        long maximumIntervalSize = -1, minimumIntervalSize = Integer.MAX_VALUE;
+        for (final IntervalList intervalList : accumulatedIntervalLists) {
+            final long baseCount = intervalList.getBaseCount();
+            if (baseCount < minimumIntervalSize) minimumIntervalSize = baseCount;
+            if (maximumIntervalSize < baseCount) maximumIntervalSize = baseCount;
+        }
```
jacarey: Man java seems so verbose after doing nothing but scala for so long.

jacarey:
:+1:

nh13:
@mcovarr the reason for this new scatter method is to improve scattering when using the break bands option in ```IntervalListTools```.  Given the current urgency of this feature, I have chosen not to produce a test case that differentiates this scatter method with the others.  It would be useful to do so, and I would be happy if it were included in the future.

Issue #143, opened 2014-12-23
Allow BedToIntervalList to accept a variety of files for its SD argument


yfarjoun:
Can use the spiffy SAMSequenceDictionaryExtractor class for this.

yfarjoun:
solved.

PR #192, opened 2015-04-06
Modify BedToIntervalList to accept other dictionary files

abaumann:
Added ability to use any of the file formats specified within
SAMSequenceDictionaryExtractor as dictionary files for
BedToIntervalList.  Updated test at BedToIntervalListTest to test
each file format that is appropriate.  Each test has associated
dictionary that matches the requirement for producting the expected
seq_dict_test.bed.interval_list output from the input file
seq_dict_test.bed.  We also test for bad dictionary input to make
sure that fails properly - both the case of a non-dictionary passed
and a dictionary that is not sufficient passed.

Resolves: #143


abaumann:
One part I am not sure about how the old code used to take the header from a SAM as is and associate that with the IntervalList, and whether or not these changes will affect anything. 

The new code creates a new header using the dictionary returned by the SAMSequenceDictionaryExtractor.  What gets lost from the old code is any other additional header info that may have been in the SAM other than the dictionary mappings.  For example the sort order.  Right now it sets the sort order to coordinate (BedToIntervalList line 68:             `header.setSortOrder(SAMFileHeader.SortOrder.coordinate);`) - which I think is right because the written out IntervalList is the `uniqued()` form.  

One potential modification is to try to get the SAMFileHeader for the input file if it exists and use that, otherwise create a new one.

nh13:
The old method was to read just the the header of the SAM file only, and use that in the new interval list.  I am not sure why the call to ```referenceSequence``` was needed when getting the header from the SAM file.

:+1:  this looks good to me.

abaumann:
should be all set

PR #193, opened 2015-04-07
A little program to match what happens in DataQualityCheck from Zamboni ...

geoffjentry:
...except that this one happens as a command line program so we can run more easily in a cloud environment.


geoffjentry:
Going to handle this differently than pushing to picard

PR #194, opened 2015-04-07
Fix bug introduced in PR #188 that causes CollectAlignmentSummaryMetrics...

tfenne:
... to not calculate most fields when run as part of CollectMultipleMetrics even when a reference sequence is provided.

### DIFF
--- a/src/java/picard/analysis/CollectAlignmentSummaryMetrics.java
+++ b/src/java/picard/analysis/CollectAlignmentSummaryMetrics.java
@@ -116,6 +116,10 @@ public class CollectAlignmentSummaryMetrics extends SinglePassSamProgram {
                     "in the file are aligned then alignment summary metrics collection will fail.");
         }
 
+        if (super.REFERENCE_SEQUENCE != null && this.REFERENCE_SEQUENCE == null) {
+            this.REFERENCE_SEQUENCE = super.REFERENCE_SEQUENCE;
+        }
+
         final boolean doRefMetrics = REFERENCE_SEQUENCE != null;
         collector = new AlignmentSummaryMetricsCollector(METRIC_ACCUMULATION_LEVEL, header.getReadGroups(), doRefMetrics,
                 ADAPTER_SEQUENCE, MAX_INSERT_SIZE,  IS_BISULFITE_SEQUENCED);


tfenne:
@yfarjoun @ktibbett - PR #188 which overrode the REFERENCE_SEQUENCE parameter in CollectAlignmentSummaryMetrics has an unintended consequence which causes most metrics not to be calculated when run as part of CollectMultipleMetrics.  This PR fixes that - can you please take a look?

yfarjoun:
EEK!

src/java/picard/analysis/CollectAlignmentSummaryMetrics.java
```diff
@@ -116,6 +116,10 @@ public static void main(final String[] argv) {
                     "in the file are aligned then alignment summary metrics collection will fail.");
         }
 
+        if (super.REFERENCE_SEQUENCE != null && this.REFERENCE_SEQUENCE == null) {
```
yfarjoun: I'm surprised that one needs to do this. the logic in CommanlineProgram was supposed to take care of this exact issue. @jacarey can you comment?

jacarey: Unless I'm mistaken this.REFERENCE_SEQUENCE would only be null if someone was setting it to NULL explicitly since there is a default value.
 In that case we would be faithful to the fact that you want to override the Default as well as the super.REFERENCE_SEQUENCE to null. If that is not desired behavior we might want to consider what should be. In the mean time this will fix the issue.

tfenne: @jacarey CollectAlignmentSummaryMetrics defaults REFERENCE_SEQUENCE to Defaults.REFERENCE_FASTA, but that itself is null unless you supply one with -Dsamjdk.reference_fasta= as far as I can tell.

jacarey: Ah ok. We might want to add some logic in the command line parser to handle an overriden option that is being set to null. Of course that is assuming you would never want to override and option and set it to null.  In any case this fix seems good to me for this particular case.

yfarjoun:
This is a fine fix for this case, but I think it brings up a greater issue that needs to be resolved: how do we override options to "null" in the derived class and not bubble up the default value from the super class?

tfenne:
Ok, well I think I’m going to merge this PR then.  @yfarjoun I think what you did is effectively that - override and not have the superclass value bubble up.  But I think what we want in this case is to override to add documentation, but not change the behaviour.

yfarjoun:
I agree with that. I'm wondering how this can be done without specifically asking for super.blah. give me another day on this please.

vdauwera:
I recommend reverting @yfarjoun's original commit for now, and following up with a complete solution that achieves what he wanted without requiring a specific-cased patch once all the areas of doubt are cleared up. This will solve the issue for the production pipeline without risking further unforeseen ramifications.

yfarjoun:
The problem is (as far as I can tell) that collectMultipleMetrics is rather
sneaky and it updates the arguments directly, but due to the way it was
implemented it updated the fields in the super-class. in the case where we
override these arguments there is a field in both the super and the derived
classes and the value in the super class is shadowed by the derived class.

I think a better change would be to fix collect multiple metrics and make
it change both the super and derived fields...and add tests :-). I did that
in PR #195

tfenne:
Closing in favor of @yfarjoun's PR #195.

PR #195, opened 2015-04-07
I propose this as an alternative

yfarjoun:
to @tfenne 's PR #194

The difference is that the change is done in collectMultipleMetrics and tests are included for 
greater good!

Comments welcome: @jacarey @ktibbett

### DIFF
--- a/src/tests/java/picard/analysis/CollectMultipleMetricsTest.java
+++ b/src/tests/java/picard/analysis/CollectMultipleMetricsTest.java
@@ -1,0 +1,249 @@
+package picard.analysis;
+
+import htsjdk.samtools.metrics.MetricsFile;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+import picard.cmdline.CommandLineProgramTest;
+
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+
+/**
+ * Tests the two default "programs" that have tests in CollectMultipleMetrics
+ *
+ *
+ * @author Yossi farjoun
+ */
+
+public class CollectMultipleMetricsTest extends CommandLineProgramTest {
+
+    private static final File TEST_DATA_DIR = new File("testdata/picard/sam");
+
+    public String getCommandLineProgramName() {
+        return CollectMultipleMetrics.class.getSimpleName();
+    }
+
+
+    @Test
+    public void testAlignmentSummaryViaMultipleMetrics() throws IOException {
+        final File input = new File(TEST_DATA_DIR, "summary_alignment_stats_test.sam");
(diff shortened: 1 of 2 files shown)


tfenne:
Looks good to me!  I almost went down that path, but when I found that makeInstance() method didn't pass through enough, I didn't want to go changing that without more testing.  I think this is the better approach - will close my PR.

yfarjoun:
Thanks for your input @tfenne. can I consider that a code review, @nh13 ?

src/tests/java/picard/analysis/CollectMultipleMetricsTest.java
```diff
@@ -0,0 +1,249 @@
...
+                "PROGRAM=null",
+                "PROGRAM="+CollectMultipleMetrics.Program.CollectAlignmentSummaryMetrics.name(),
+                "PROGRAM="+CollectMultipleMetrics.Program.CollectInsertSizeMetrics.name()
+        };
+        Assert.assertEquals(runPicardCommandLine(args), 0);
+
+        final MetricsFile<AlignmentSummaryMetrics, Comparable<?>> output = new MetricsFile<AlignmentSummaryMetrics, Comparable<?>>();
+        output.read(new FileReader(outfile));
+
+        for (final AlignmentSummaryMetrics metrics : output.getMetrics()) {
```
nh13: duplicate code in your tests.

yfarjoun: not really. the input files are different. it's true that I run the same two programs in both tests, but I only look at the result of the relevant one in each. the reason I run two programs is to make sure that running two programs doesn't somehow change the results.

nh13: I disagree. The asserts are the same, and the ```input``` and ```reference``` files are the same.

yfarjoun: We seem to be not looking at the same lines of code.

the lines of code you marked are a duplicate of what?

nh13: See: https://github.com/broadinstitute/picard/pull/195/files#diff-95ae0051bfea869e3e0d29ebad1bda73R47
See: https://github.com/broadinstitute/picard/pull/195/files#diff-a9a05ca5beab622293e1050cace330f9R263

yfarjoun:
@nh13: thanks for catching that. A relic of my misunderstanding regarding what needs fixing (and thus what needs testing).

yfarjoun:
updated commit.

mattsooknah:
Looks good. Merge away!

PR #196, opened 2015-04-14
New CLP for downsampling in a way that tries to keep neighboring reads together

yfarjoun:
This is intended to mimic to a higher degree the results of sequencing less. The idea is to create a periodic mask of reads based on their location within their tile. The mask is setup in such a way so that the same percentage of reads is removed from the bulk and from the boundaries of each tile, and so that neighboring tiles agree on which part of their shared boundary to keep or remove. 

Since this is based on location, mates (and secondary alignments) are treated equally.

### DIFF
--- a/src/java/picard/sam/PositionBasedDownsampleSam.java
+++ b/src/java/picard/sam/PositionBasedDownsampleSam.java
@@ -1,0 +1,380 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015 The Broad Institute
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package picard.sam;
+
+import htsjdk.samtools.SAMFileHeader;
+import htsjdk.samtools.SAMFileWriter;
+import htsjdk.samtools.SAMFileWriterFactory;
+import htsjdk.samtools.SAMProgramRecord;
+import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.SamReader;
+import htsjdk.samtools.SamReaderFactory;
+import htsjdk.samtools.util.CloserUtil;
+import htsjdk.samtools.util.CollectionUtil;
+import htsjdk.samtools.util.Histogram;
+import htsjdk.samtools.util.IOUtil;
+import htsjdk.samtools.util.Log;
+import htsjdk.samtools.util.ProgressLogger;
+import picard.PicardException;
+import picard.cmdline.CommandLineProgram;
+import picard.cmdline.CommandLineProgramProperties;
+import picard.cmdline.Option;
+import picard.cmdline.StandardOptionDefinitions;
+import picard.cmdline.programgroups.SamOrBam;
+import picard.sam.util.PhysicalLocation;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+
+/**
+ * Class to downsample a BAM file while respecting that we should either get rid
+ * of both ends of a pair or neither end of the pair. In addition, this program uses the read-name
+ * and extracts the position within the tile whence the read came from. The downsampling is based on this position.
+ * <p/>
+ * Note 1: This is technology and read-name dependent. If your read-names do not have coordinate information, or if your
+ * BAM contains reads from multiple technologies (flowcell versions, sequencing machines) this will not work properly.
+ * This has been designed with Illumina MiSeq/HiSeq in mind.
+ * <p/>
+ * Note 2: The downsampling is _not_ random. It is deterministically dependent on the position of the read within its tile. Specifically,
+ * it draws out an ellipse that covers a FRACTION fraction of the area and each of the edges and uses this to determine whether to keep the
+ * record. Since reads with the same name have the same position (mates, secondary and supplemental alignments), the decision will be the same for all of them.
+ * <p/>
+ * Finally, the code has been designed to simulate sequencing less as accurately as possible, not for getting an exact downsample fraction.
+ * In particular, since the reads may be distributed non-evenly within the lanes/tiles, the resulting downsampling percentage will not be accurately
+ * determined by the input argument FRACTION. One should re-MarkDuplicates after downsampling in order to "expose" the duplicates whose representative has
+ * been downsampled away.
+ *
+ * @author Yossi Farjoun
+ */
+@CommandLineProgramProperties(
+        usage = "Class to downsample a BAM file while respecting that we should either get rid of both ends of a pair or neither \n" +
+                "end of the pair. In addition, this program uses the read-name and extracts the position within the tile whence \n" +
+                "the read came from. The downsampling is based on this position. Results with the exact same input will produce the \n" +
+                "same results.\n" +
+                "\n" +
+                "Note 1: This is technology and read-name dependent. If your read-names do not have coordinate information, or if your\n" +
+                "BAM contains reads from multiple technologies (flowcell versions, sequencing machines) this will not work properly. \n" +
+                "This has been designed with Illumina MiSeq/HiSeq in mind.\n" +
+                "Note 2: The downsampling is not random. It is deterministically dependent on the position of the read within its tile.\n" +
+                "Note 3: Downsampling twice with this program is not supported.\n" +
+                "Note 4: You should call MarkDuplicates after downsampling.\n" +
+                "\n" +
+                "Finally, the code has been designed to simulate sequencing less as accurately as possible, not for getting an exact downsample \n" +
+                "fraction. In particular, since the reads may be distributed non-evenly within the lanes/tiles, the resulting downsampling \n" +
+                "percentage will not be accurately determined by the input argument FRACTION.",
+        usageShort = "Downsample a SAM or BAM file to retain a subset of the reads based on the reads location in each tile in the flowcell.",
+        programGroup = SamOrBam.class
+)
+public class PositionBasedDownsampleSam extends CommandLineProgram {
+
+    @Option(shortName = StandardOptionDefinitions.INPUT_SHORT_NAME, doc = "The input SAM or BAM file to downsample.")
+    public File INPUT;
+
+    @Option(shortName = StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "The output, downsampled, SAM or BAM file to write.")
+    public File OUTPUT;
+
+    @Option(shortName = "F", doc = "The (approximate) fraction of reads to be kept, between 0 and 1.", optional = false)
+    public Double FRACTION = null;
+
+    @Option(doc = "Stop after processing N reads, mainly for debugging.", optional = true)
+    public Long STOP_AFTER = null;
+
+    @Option(doc = "Allow Downsampling again despite this being a bad idea with possibly unexpected results.", optional = true)
+    public boolean ALLOW_MULTIPLE_DOWNSAMPLING_DESPITE_WARNINGS = false;
+
+    @Option(doc = "Determines whether the duplicate tag should be reset since the downsampling requires re-marking duplicates.")
+    public boolean REMOVE_DUPLICATE_INFORMATION = true;
+
+    private final Log log = Log.getInstance(PositionBasedDownsampleSam.class);
+
+    private PhysicalLocation opticalDuplicateFinder;
+    private long total = 0;
+    private long kept = 0;
+    public static String PG_PROGRAM_NAME = "PositionBasedDownsampleSam";
+    private final static double ACCEPTABLE_FUDGE_FACTOR = 0.2;
+
+    /* max-position in tile as a function of tile. We might need to
+       look per-readgroup, but at this point I'm making the assumptions that I need to downsample a
+       sample where all the readgroups came from the same type of flowcell. */
+
+    CollectionUtil.DefaultingMap.Factory<Coord, Short> defaultingMapFactory = new CollectionUtil.DefaultingMap.Factory<Coord, Short>() {
+        @Override
+        public Coord make(final Short aShort) {
+            return new Coord();
+        }
+    };
+
+    final private Map<Short, Coord> tileCoord = new CollectionUtil.DefaultingMap<Short, Coord>(defaultingMapFactory, true);
+
+
+    final Map<Short, Histogram<Short>> xPositions = new HashMap<Short, Histogram<Short>>();
+    final Map<Short, Histogram<Short>> yPositions = new HashMap<Short, Histogram<Short>>();
+
+    public static void main(final String[] args) {
+        new PositionBasedDownsampleSam().instanceMainWithExit(args);
+    }
+
+    @Override
+    protected String[] customCommandLineValidation() {
+        final List<String> errors = new ArrayList<String>();
+
+        if (FRACTION < 0 || FRACTION > 1) {
+            errors.add("FRACTION must be a value between 0 and 1, found: " + FRACTION);
+        }
+
+        if (errors.isEmpty()) {
+            return null;
+        } else {
+            return errors.toArray(new String[errors.size()]);
+        }
+    }
+
+    @Override
+    protected int doWork() {
+
+        IOUtil.assertFileIsReadable(INPUT);
+        IOUtil.assertFileIsWritable(OUTPUT);
+
+        log.info("Checking to see if input file has been downsampled with this program before.");
+        checkProgramRecords();
+
+        opticalDuplicateFinder = new PhysicalLocation();
+
+        log.info("Starting first pass. Examining read distribution in tiles.");
+        fillTileMinMaxCoord();
+        log.info("First pass done.");
+
+        log.info("Starting second pass. Outputting reads.");
+        outputSamRecords();
+        log.info("Second pass done.");
+
+        final double finalP = kept / (double) total;
+        if (Math.abs(finalP - FRACTION) / (Math.min(finalP, FRACTION) + 1e-10) > ACCEPTABLE_FUDGE_FACTOR) {
+            log.warn(String.format("You've requested FRACTION=%g, the resulting downsampling resulted in a rate of %f.", FRACTION, finalP));
+        }
+        log.info(String.format("Finished! Kept %d out of %d reads (P=%g).", kept, total, finalP));
+
+        return 0;
+    }
+
+    private void outputSamRecords() {
+
+        final ProgressLogger progress = new ProgressLogger(log, (int) 1e7);
+        final SamReader in = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(INPUT);
+
+        final SAMFileHeader header = in.getFileHeader().clone();
+        final SAMFileHeader.PgIdGenerator pgIdGenerator = new SAMFileHeader.PgIdGenerator(header);
+        final SAMProgramRecord programRecord = new SAMProgramRecord(pgIdGenerator.getNonCollidingId(PG_PROGRAM_NAME));
+
+        programRecord.setProgramName(PG_PROGRAM_NAME);
+        programRecord.setCommandLine(getCommandLine());
+        programRecord.setProgramVersion(getVersion());
+        header.addProgramRecord(programRecord);
+
+        final SAMFileWriter out = new SAMFileWriterFactory().makeSAMOrBAMWriter(header, true, OUTPUT);
+
+        final CircleSelector selector = new CircleSelector(FRACTION);
+
+        for (final SAMRecord rec : in) {
+            if (STOP_AFTER != null && total >= STOP_AFTER) break;
+
+            total++;
+
+            final PhysicalLocation pos = getSamRecordLocation(rec);
+
+            if (!xPositions.containsKey(pos.getTile())) {
+                xPositions.put(pos.getTile(), new Histogram<Short>(pos.getTile() + "-xpos", "count"));
+            }
+            if (!yPositions.containsKey(pos.getTile())) {
+                yPositions.put(pos.getTile(), new Histogram<Short>(pos.getTile() + "-ypos", "count"));
+            }
+
+            final boolean keepRecord = selector.select(pos, tileCoord.get(pos.getTile()));
+
+            if (keepRecord) {
+                if (REMOVE_DUPLICATE_INFORMATION) rec.setDuplicateReadFlag(false);
+                out.addAlignment(rec);
+                kept++;
+            }
+            progress.record(rec);
+        }
+
+        out.close();
+
+        CloserUtil.close(in);
+    }
+
+    private void checkProgramRecords() {
+        final SamReader in = SamReaderFactory
+                .makeDefault()
+                .referenceSequence(REFERENCE_SEQUENCE)
+                .open(INPUT);
+
+        for (final SAMProgramRecord pg : in.getFileHeader().getProgramRecords()) {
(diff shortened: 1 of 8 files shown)


src/java/picard/cmdline/StandardOptionDefinitions.java
```diff
@@ -45,4 +45,5 @@
     public static final String MINIMUM_LOD_SHORT_NAME = "LOD";
     public static final String SORT_ORDER_SHORT_NAME = "SO";
     public static final String USE_ORIGINAL_QUALITIES_SHORT_NAME = "OQ";
+    public static final String PROBABILITY_SHORT_NAME = "P";
```
nh13: how many places is this used?  Is it used so much it should only get one letter?

src/java/picard/sam/PhysicalLocation.java
```diff
@@ -0,0 +1,120 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2014 The Broad Institute
```
nh13: 2015

src/java/picard/sam/PhysicalLocation.java
```diff
@@ -0,0 +1,120 @@
...
+
+package picard.sam;
+
+import picard.PicardException;
+import picard.sam.markduplicates.util.OpticalDuplicateFinder;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static picard.sam.markduplicates.util.ReadNameParsingUtils.getRapidDefaultReadNameRegexSplit;
```
nh13: should ```ReadNameParsingUtils``` be moved out of ```picard.sam.markduplicates.util``` too?

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,295 @@
...
+    public File OUTPUT;
+
+    @Option(shortName = StandardOptionDefinitions.PROBABILITY_SHORT_NAME, doc = "The (approximate) probability of keeping a read, between 0 and 1.")
+    public double PROBABILITY = 1;
+
+    @Option(doc = "Stop after processing N reads, mainly for debugging.")
+    public long STOP_AFTER = 0;
+
+    @Option(doc = "Determines whether the duplicate tag should be reset since the downsampling requires re-marking duplicates.")
+    public boolean RESET_DUPLICATE_FLAG = true;
```
nh13: call this the same as the similar option in ```RevertSam```.

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,295 @@
...
+
+    private final Log log = Log.getInstance(PositionBasedDownsampleSam.class);
+
+    private CircleSelector selector = new CircleSelector(PROBABILITY);
```
nh13: why is the construction/assignment needed here?

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,295 @@
...
+
+    private PhysicalLocation opticalDuplicateFinder;
+    private long total = 0;
+    private long kept = 0;
+
+    //max-position in tile as a function of tile. We might need to
```
nh13: ```/* */``` is your friend.

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,295 @@
...
+
+    @Override
+    protected String[] customCommandLineValidation() {
+        final List<String> errors = new ArrayList<String>();
+
+        if (PROBABILITY < 0 || PROBABILITY > 1) {
+            errors.add("PROBABILITY must be a value between 0 and 1, found: " + PROBABILITY);
+        }
+
+        IOUtil.assertFileIsReadable(INPUT);
```
nh13: Usually put in the ```doWork``` method, since others may want to make an instance of this class and do work.

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,295 @@
...
+
+        final double finalP = kept / (double) total;
+        if (abs(finalP - PROBABILITY) / (Math.min(finalP, PROBABILITY) + 1e-10) > .2) {
+            log.warn(String.format("You've requested PROBABILITY=%g, the resulting downsampling resulted in a rate of %f. ", PROBABILITY, finalP));
+        }
+        log.info(String.format("Finished! Kept %d out of %d reads (P=%g).", kept, total, finalP));
+        return 0;
+    }
+
+    private void outputRecs() {
```
nh13: ```outoutSAMRecords```?

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,295 @@
...
+        final SAMFileWriter out = new SAMFileWriterFactory().makeSAMOrBAMWriter(in.getFileHeader(), true, OUTPUT);
+
+        for (final SAMRecord rec : in) {
+            if (STOP_AFTER != 0 && total >= STOP_AFTER) break;
+
+            total++;
+
+            final PhysicalLocation pos = getSamRecordLocation(rec);
+
+            if (!xPositions.containsKey(pos.getTile()))
```
nh13: brackets: ```{ }```

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,295 @@
...
+                xPositions.put(pos.getTile(), new Histogram<Short>(pos.getTile() + "-xpos", "count"));
+            if (!yPositions.containsKey(pos.getTile()))
+                yPositions.put(pos.getTile(), new Histogram<Short>(pos.getTile() + "-ypos", "count"));
```
nh13: same as above: ```{}```

src/java/picard/cmdline/StandardOptionDefinitions.java
yfarjoun: Including this, twice (in downsampleSAM), would be appropriate for "padhopping"-Metric and HiSeqXPFFail as well, I think

src/java/picard/sam/PositionBasedDownsampleSam.java
yfarjoun: No problem. could you explain what this method should be used for then?

nh13: Validate the values of the command line parameters, like you do, especially if there are dependencies between them.  It is not wrong to put file existence checking their, it is just more useful if they are in the ```doWork``` method due to my reasons above, and for consistency in the code (legal precedence?).

geoffjentry: Agree with Nils here. IMO a Picard CLP should be the thinnest possible wrapper around Real Code which could be called in a very general fashion. It's happened more times than I'd like that I was hamstrung due to logic being too tightly coupled with the CLP-ness.

yfarjoun:
This is failing since it is waiting for a pull-request in hts-jdk, but it can still be reviewed, I think. @nh13 care to take another look?

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,361 @@
...
+                "of both ends of a pair or neither end of the pair. In addition, this program uses the read-name \n" +
+                "and extracts the position within the tile from whence the read came from. THe downsampling is based on this position. \n" +
+                "results with the exact same input will produce the same results.\n" +
+                "\n" +
+                "<it>Note 1</it>: This is technology and read-name dependent. if your read-names do not have coordinate information, or if your\n" +
+                "BAM contains reads from multiple technologies (flowcell versions, sequencing machines) this will not work properly. \n" +
+                "This has been designed with Illumina MiSeq/HiSeq in mind.\n" +
+                "<it>Note 2</it>: The downsampling is _not_ random. It is deterministically dependent on the position of the read within its tile." +
+                "<it>Note 3</it>: Downsampling twice with this program is not supported." +
+                "<it>Note 4</it>: You should MarkDuplicates after downsampling." +
```
nh13: will ```<it>``` be properly displayed on the terminal?

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,361 @@
...
+                "\n" +
+                "Finally, the code has been designed to simulate sequencing less as accurately as possible, not for getting an exact downsample fraction. " +
+                "In particular, since the reads may be distributed non-evenly within the lanes/tiles, the resulting downsampling percentage will not be accurately" +
+                "determined by the input argument PROBABILITY.",
+        usageShort = "Down-sample a SAM or BAM file to retain a subset of the reads based on the reads location in each tile in the flowcell.",
```
nh13: no dash in downsample, see link 96.

nh13:
:+1:  from me, but you need the others on the team too.

yfarjoun:
"the others"? not sure what you mean? methods or engineering? anyone in
particular?

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,361 @@
...
+        programGroup = SamOrBam.class
+)
+public class PositionBasedDownsampleSam extends CommandLineProgram {
```
mcovarr: This doesn't seem like an intuitive name for a CLP that downsamples while keeping pairs intact.

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,361 @@
...
+        usage = "Class to downsample a BAM file while respecting that we should either get rid\n" +
+                "of both ends of a pair or neither end of the pair. In addition, this program uses the read-name \n" +
+                "and extracts the position within the tile from whence the read came from. THe downsampling is based on this position. \n" +
+                "results with the exact same input will produce the same results.\n" +
+                "\n" +
+                "<it>Note 1</it>: This is technology and read-name dependent. if your read-names do not have coordinate information, or if your\n" +
+                "BAM contains reads from multiple technologies (flowcell versions, sequencing machines) this will not work properly. \n" +
+                "This has been designed with Illumina MiSeq/HiSeq in mind.\n" +
+                "<it>Note 2</it>: The downsampling is _not_ random. It is deterministically dependent on the position of the read within its tile." +
+                "<it>Note 3</it>: Downsampling twice with this program is not supported." +
```
mcovarr: It looks like downsampling twice is possible if `ALLOW..._DESPITE_WARNINGS=true` is specified.

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,361 @@
...
+ * In particular, since the reads may be distributed non-evenly within the lanes/tiles, the resulting downsampling percentage will not be accurately
+ * determined by the input argument PROBABILITY. One should re-MarkDuplicates after downsampling in order to "expose" the duplicates who's representative has
+ * been downsampled away.
+ *
+ * @author Yossi Farjoun
+ */
+@CommandLineProgramProperties(
+        usage = "Class to downsample a BAM file while respecting that we should either get rid\n" +
+                "of both ends of a pair or neither end of the pair. In addition, this program uses the read-name \n" +
+                "and extracts the position within the tile from whence the read came from. THe downsampling is based on this position. \n" +
```
mcovarr: "from whence the read came from" -> "whence the read came"
THe -> The

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,361 @@
...
+                "results with the exact same input will produce the same results.\n" +
```
mcovarr: missing capitalization here and below

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,361 @@
...
+        log.info("Starting first pass. Examining read distribution in tiles.");
+        fillTileMinMaxCoord();
+        log.info("First pass done.");
+
+        log.info("Starting second pass. Outputting reads.");
+        outputSamRecords();
+        log.info("Second pass done.");
+
+        final double finalP = kept / (double) total;
+        if (Math.abs(finalP - PROBABILITY) / (Math.min(finalP, PROBABILITY) + 1e-10) > .2) {
```
mcovarr: What's the significance of the .2 magic number?  Could this be a symbolic constant?

src/java/picard/sam/PositionBasedDownsampleSam.java
yfarjoun: hmmm. Any downsampler that respects itself should keep read-mates together. (DownsampleSam does the same).

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,361 @@
...
+                .referenceSequence(REFERENCE_SEQUENCE)
+                .open(INPUT);
+
+        for (final SAMProgramRecord pg : in.getFileHeader().getProgramRecords()) {
+            if (pg.getProgramName() != null && pg.getProgramName().equals(PG_PROGRAM_NAME)) {
+
+                String outText = "Found previous Program Record that indicates that this BAM has been downsampled already with this program. Operation not supported! Previous PG: " + pg.toString();
+
+                if (ALLOW_MULTIPLE_DOWNSAMPLING_DESPITE_WARNINGS) {
+                    log.warn(outText);
```
mcovarr: Despite the warning that will say downsampling again is impossible, isn't this going to downsample again if `ALLOW...DESPITE_WARNINGS=true` is specified?

src/java/picard/sam/PositionBasedDownsampleSam.java
yfarjoun: It is _possible_ but it's not supported...if you tell me that it does strange things when you downsampled a second time, I'll say "sorry, downsampling twice is not supported".

src/java/picard/sam/util/PhysicalLocation.java
```diff
@@ -0,0 +1,118 @@
...
+     * Method used to extract tile/x/y from the read name and add it to the PhysicalLocation so that it
+     * can be used later to determine optical duplication
+     *
+     * @param readName the name of the read/cluster
+     * @param loc      the object to add tile/x/y to
+     * @return true if the read name contained the information in parsable form, false otherwise
+     */
+    public boolean addLocationInformation(final String readName, final PhysicalLocation loc) {
+        // Optimized version if using the default read name regex (== used on purpose):
+        if (this.readNameRegex == this.DEFAULT_READ_NAME_REGEX) {
```
mcovarr: Lots of unnecessary `this` in this file, but the one on the RHS of the expression here is particularly undesirable for static member access.

src/java/picard/sam/util/PhysicalLocation.java
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Small interface that provides access to the physical location information about a cluster.
+ * All values should be defaulted to -1 if unavailable.  Tile should only allow
+ * non-zero positive integers, x and y coordinates must be non-negative.
+ * This is different from OpticalDuplicateFinder.PhysicalLocation in that the x and y positions are ints, not shorts
+ * thus, they do not overflow within a HiSeqX tile.
+ */
+public class PhysicalLocation {
+
+    public static final String DEFAULT_READ_NAME_REGEX = OpticalDuplicateFinder.DEFAULT_READ_NAME_REGEX;
```
mcovarr: I'm not sure I see the value of aliasing the static member in `OpticalDuplicateFinder`.

src/java/picard/sam/util/ReadNameParsingUtils.java
```diff
@@ -0,0 +1,83 @@
...
+            tokensIdx++;
+        }
+        return tokensIdx;
+    }
+
+    /**
+     * Very specialized method to rapidly parse a sequence of digits from a String up until the first
+     * non-digit character.
+     */
+    static public int rapidParseInt(final String input) {
```
mcovarr: Does the use of this method result in significantly more performant code than a `-?\d+` regex and `Integer.valueOf`?  Same question for the method above.

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,361 @@
...
+
+    private class CircleSelector {
+
+        private final double radiusSquared;
+        private final double offset;
+        private final boolean positiveSelection;
+
+        CircleSelector(final double fraction) {
+            assert (fraction >= 0);
+            assert (fraction <= 1);
```
mcovarr: I think these `assert`s probably shouldn't be here.

src/java/picard/sam/PositionBasedDownsampleSam.java
mcovarr: :smile: OK, I wasn't aware we allowed such a distinction.

src/java/picard/sam/PositionBasedDownsampleSam.java
mcovarr: OK, I was warned about that sort of behavior from some downsamplers, but I'll defer to you on this.

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,361 @@
...
+public class PositionBasedDownsampleSam extends CommandLineProgram {
+
+    @Option(shortName = StandardOptionDefinitions.INPUT_SHORT_NAME, doc = "The input SAM or BAM file to downsample.")
+    public File INPUT;
+
+    @Option(shortName = StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "The output, downsampled, SAM or BAM file to write.")
+    public File OUTPUT;
+
+    @Option(shortName = StandardOptionDefinitions.PROBABILITY_SHORT_NAME, doc = "The (approximate) probability of keeping a read, between 0 and 1.")
+    public double PROBABILITY = 1;
```
mcovarr: Is `PROBABILITY` really the right terminology for what seems to be a deterministic fraction of the tile area?

src/java/picard/sam/util/ReadNameParsingUtils.java
yfarjoun: Not sure, though I imagine it does. This file was simply moved into the current package, not changed. Given how optimized MarkDuplicates is, I didn't want to change anything.

src/java/picard/sam/PositionBasedDownsampleSam.java
yfarjoun: it is true that GATK's downsampling doesn't do this....

yfarjoun:
@mcovarr I responded to your comments. Care to take another glance?

src/java/picard/sam/util/PhysicalLocation.java
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Small interface that provides access to the physical location information about a cluster.
+ * All values should be defaulted to -1 if unavailable.  Tile should only allow
+ * non-zero positive integers, x and y coordinates must be non-negative.
+ * This is different from OpticalDuplicateFinder.PhysicalLocation in that the x and y positions are ints, not shorts
+ * thus, they do not overflow within a HiSeqX tile.
+ */
+public class PhysicalLocation {
+
+    public static final String DEFAULT_READ_NAME_REGEX = "[a-zA-Z0-9]+:[0-9]:([0-9]+):([0-9]+):([0-9]+).*".intern();
```
mcovarr: unnecessary intern

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,364 @@
...
+        CloserUtil.close(in);
+    }
+
+    private PhysicalLocation getSamRecordLocation(final SAMRecord rec) {
+        final PhysicalLocation pos = new PhysicalLocation();
+        opticalDuplicateFinder.addLocationInformation(rec.getReadName(), pos);
+        return pos;
+    }
+
+    private class CircleSelector {
```
mcovarr: Is it necessary that the mask for the rectangular tile be circular?  It feels a bit round peg/square hole.  :smile:  A rectangle of full width and height proportional to `fraction` would seem more straightforward, assuming that's not violating any assumptions about read distribution within the tile I don't know about.

src/java/picard/sam/util/PhysicalLocation.java
mcovarr: Not a big deal, but you might use `\d` instead of `[0-9]`.  Is that first `[0-9]` really not supposed to have a `+` quantifier?

src/java/picard/sam/util/ReadNameParsingUtils.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Common functions for quickly parsing strings. Used for parsing the tile and coordinates from the read names
+ */
+public class ReadNameParsingUtils {
+
+    /**
+     * Single pass method to parse the read name for the default regex.  This will only insert the 2nd to the 4th
+     * tokens (inclusive).  It will also stop after the fifth token has been successfully parsed.
+     */
+    static public int getRapidDefaultReadNameRegexSplit(final String readName, final char delim, final int[] tokens) {
```
mcovarr: Nitpick: The Java Language Specification [says](http://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.1.1) "it is customary, though not required" that the order of these class modifiers would be `public static int`.

src/java/picard/sam/util/PhysicalLocation.java
```diff
@@ -0,0 +1,118 @@
...
+
+    public PhysicalLocation() {this(DEFAULT_READ_NAME_REGEX);}
+
+    public PhysicalLocation(final String readNameRegExp) {this.readNameRegex = readNameRegExp;}
+
+    private Pattern readNamePattern;
+
+    public short tile = -1;
+    public int x = -1, y = -1;
+
```
mcovarr: kind of strange these fields are public *and* have setters

src/java/picard/sam/PositionBasedDownsampleSam.java
yfarjoun: The problem is that you need to take a fraction of the area, AND a fraction of both the top and bottom edges of the tile. AND you have to be symmetric top/bottom and left/right (on the edges). This can still be done with an offset rectangle, but there are _more_ degrees of freedom, and thus more decisions to make. Also, the circle has the highest ratio of area to circumference, which is also being optimized here...happy to discuss this further if you like.

src/java/picard/sam/util/PhysicalLocation.java
yfarjoun: again, I copied this verbatim from markduplicates...so I'm pretty sure it's correct. Happy to discuss the logic though. the first [0-9] is the lane. We have less than 10 lanes. Not sure about the choice of [0-9] over \d perhaps, again, speed?

src/java/picard/sam/util/PhysicalLocation.java
yfarjoun: fixed.

src/java/picard/sam/util/PhysicalLocation.java
yfarjoun: fixed the intern.

src/java/picard/sam/util/ReadNameParsingUtils.java
yfarjoun: I only moved the code....but I'll fix it.

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,364 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+
+/**
+ * Class to downsample a BAM file while respecting that we should either get rid
+ * of both ends of a pair or neither end of the pair. In addition, this program uses the read-name
+ * and extracts the position within the tile whence the read came from. The downsampling is based on this position.
```
mcovarr: "whence the read came from" -> "whence the read came"

src/java/picard/sam/PositionBasedDownsampleSam.java
mcovarr: Thanks Yossi, I'll stop by today and we can discuss.  In the meantime I might suggest adding a comment similar to your explanation to discourage a future maintainer from "fixing" the code as I suggested. :smile: It wasn't clear to me at least why this particular type of downsampling was required to simulate the effect of sequencing less.  And actually I don't see it documented here the motivation for this CLP is to simulate the effect of sequencing less.

yfarjoun: good point.

nh13:
@yfarjoun the folks who were doing the code sampling.

src/java/picard/sam/PositionBasedDownsampleSam.java
```diff
@@ -0,0 +1,364 @@
...
+                "results with the exact same input will produce the same results.\n" +
+                "\n" +
+                "Note 1: This is technology and read-name dependent. If your read-names do not have coordinate information, or if your\n" +
+                "BAM contains reads from multiple technologies (flowcell versions, sequencing machines) this will not work properly. \n" +
+                "This has been designed with Illumina MiSeq/HiSeq in mind.\n" +
+                "Note 2: The downsampling is not random. It is deterministically dependent on the position of the read within its tile.\n" +
+                "Note 3: Downsampling twice with this program is not supported.\n" +
+                "Note 4: You should call MarkDuplicates after downsampling.\n" +
+                "\n" +
+                "Finally, the code has been designed to simulate sequencing less as accurately as possible, not for getting an exact downsample fraction. " +
```
mcovarr: missing newlines

yfarjoun:
@mcovarr this is a mess. do you want me to rebase and give you a clean slate to look at?

mcovarr:
LGTM :+1:

yfarjoun:
sweet! thanks @mcovarr

I'll rebase, wait for tests to pass again and merge.

PR #197, opened 2015-04-15
Adding a tool that collects per-sample and aggregate (spanning all sampl...

nh13:
...es) metrics from the provided VCF file.


nh13:
@tfenne we are adding guava

tfenne:
makeitso

PR #198, opened 2015-04-21
Changes to enable MergeBamAlignment to output in queryname order without...

tfenne:
... sorting first into coordinate and then back again into coordinate. The only caveat is that if outputting querynaem order then neither NM or UQ are calculated (similar to the behaviour if no reference sequence is provided).


tfenne:
@ktibbett Would you mind taking a look at this please?  You're really the only other person who understands this code I think ;)

ktibbett:
Looks good!

PR #200, opened 2015-04-24
Adding a tool SimpleMarkDuplicatesWithMateCigar that uses the new HTSJDK

nh13:
DuplicateSetIterator.

Depends on: https://github.com/samtools/htsjdk/pull/225.

### DIFF
--- a/src/java/picard/sam/markduplicates/util/AbstractMarkDuplicatesCommandLineProgram.java
+++ b/src/java/picard/sam/markduplicates/util/AbstractMarkDuplicatesCommandLineProgram.java
@@ -266,6 +266,12 @@ public abstract class AbstractMarkDuplicatesCommandLineProgram extends AbstractO
     /**
      * Looks through the set of reads and identifies how many of the duplicates are
      * in fact optical duplicates, and stores the data in the instance level histogram.
+     * 
+     * We expect only reads with FR or RF orientations, not a mixture of both. 
+     * 
+     * In PCR duplicate detection, a duplicates can be a have FR and RF when fixing the orientation order to the first end of the mate.  In
+     * optical duplicate detection, we do not consider them duplicates if one read as FR ann the other RF when we order orientation by the
+     * first mate sequenced (read #1 of the pair).
      */
     private static void trackOpticalDuplicates(final List<? extends OpticalDuplicateFinder.PhysicalLocation> list,
                                                final OpticalDuplicateFinder opticalDuplicateFinder,
--- a/src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
+++ b/src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
@@ -1,0 +1,222 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015 The Broad Institute
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package picard.sam.markduplicates;
+
+import htsjdk.samtools.DuplicateSet;
+import htsjdk.samtools.DuplicateSetIterator;
+import htsjdk.samtools.SAMFileHeader;
+import htsjdk.samtools.SAMFileWriter;
+import htsjdk.samtools.SAMFileWriterFactory;
+import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.SAMTag;
+import htsjdk.samtools.util.IOUtil;
+import htsjdk.samtools.util.IterableAdapter;
+import htsjdk.samtools.util.Log;
+import htsjdk.samtools.util.ProgressLogger;
+import picard.cmdline.CommandLineProgramProperties;
+import picard.cmdline.programgroups.Testing;
+import picard.sam.DuplicationMetrics;
+import picard.sam.markduplicates.util.AbstractMarkDuplicatesCommandLineProgram;
+import picard.sam.markduplicates.util.LibraryIdGenerator;
+import picard.sam.markduplicates.util.ReadEnds;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * This is a simple tool to mark duplicates using the DuplicateSetIterator, DuplicateSet, and SAMRecordDuplicateComparator.
+ * 
+ * Users should continue to use MarkDuplicates in general.  The main motivation of this tool was the fact that 
+ * MarkDuplicates has many, many, many useful test cases, but few unit tests for validating individual duplicate sets. To
+ * test the DuplicateSetIterator, DuplicateSet, and SAMRecordDuplicateComparator, the most expedient method was to write
+ * this tool and make sure it behaves similarly to MarkDuplicates.  Not the best, I know, but good enough.  NH 06/25/2015.
+ *  
+ * 
+ * See MarkDuplicates for more details.
+ *
+ * @author nhomer
+ */
+@CommandLineProgramProperties(
+        usage = "Examines aligned records in the supplied SAM or BAM file to locate duplicate molecules. " +
+                "All records are then written to the output file with the duplicate records flagged.",
+        usageShort = "Examines aligned records in the supplied SAM or BAM file to locate duplicate molecules.",
+        programGroup = Testing.class
+)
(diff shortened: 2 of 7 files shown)


nh13:
@yfarjoun do you want to champion the code review and inclusion into picard with the team over the next sprint?

yfarjoun:
sure. not sure what that requires from me though...

src/java/picard/sam/markduplicates/util/AbstractMarkDuplicatesCommandLineProgram.java
```diff
@@ -267,7 +267,7 @@ public static void trackOpticalDuplicates(List<? extends ReadEnds> ends,
      * Looks through the set of reads and identifies how many of the duplicates are
      * in fact optical duplicates, and stores the data in the instance level histogram.
      */
-    private static void trackOpticalDuplicates(final List<? extends OpticalDuplicateFinder.PhysicalLocation> list,
+    protected static void trackOpticalDuplicates(final List<? extends OpticalDuplicateFinder.PhysicalLocation> list,
                                                final OpticalDuplicateFinder opticalDuplicateFinder,
```
tovanadler: While you are here, perhaps add a comment about when one would use each version of trackOpticalDuplicates? I assume you plan to have your code call this version, as you are changing it to protected, but I am not sure why you want to do that, as you are setting orientationForOpticalDuplicates, which is only used by the other version of this function.

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
```diff
@@ -0,0 +1,211 @@
...
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * This is a simple tool to mark duplicates using the DuplicateSetIterator, DuplicateSet, and SAMRecordDuplicateComparator.
+ *
```
tovanadler: Would you mind explaining it's realtionship to the original mark duplicates? Would one use this instead? Do we expect the same outcomes?

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
```diff
@@ -0,0 +1,211 @@
...
+                        }
+                        numDuplicates++;
+                    }
+                    
+                    // To track optical duplicates, store a set of locations for mapped pairs, first end only
+                    if (doOpticalDuplicateTracking &&
+                            record.getReadPairedFlag() &&
+                            !record.getReadUnmappedFlag() &&
+                            !record.getMateUnmappedFlag() &&
+                            !readEndsSeen.contains(record.getReadName())) {
```
tovanadler: Making sure I understand correctly: doOpticalDuplicateTracking is set to true if the representative (which I assume is the "best" of the set of duplicates) is paired, the first one in the pair, and not unmapped. Once we know that about the representative, we compare it to all other duplicate records that are paired and not unmapped. What this (and the code that follows) implies is that duplicate pairs could be in either order. Is that correct?

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
```diff
@@ -0,0 +1,211 @@
...
+        
+        for (final DuplicateSet duplicateSet : new IterableAdapter<DuplicateSet>(iterator)) {
+            final SAMRecord representative = duplicateSet.getRepresentative();
+            final boolean doOpticalDuplicateTracking = representative.getReadPairedFlag() &&
+                    representative.getFirstOfPairFlag() &&
+                    !representative.getReadUnmappedFlag() &&
+                    !representative.getMateUnmappedFlag();
+            final Set<String> readEndsSeen = new HashSet<String>();
+            
+            final List<ReadEnds> readEnds = new ArrayList<ReadEnds>();
```
tovanadler: It appears that this is populated with optical duplicates, correct? If so, consider renaming to make that clear?

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
```diff
@@ -0,0 +1,211 @@
...
+            for (final SAMRecord record : duplicateSet.getRecords()) {
+
```
tovanadler: I assume that getRecords returns all the duplicates of the representative (but not the representative).

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
```diff
@@ -0,0 +1,211 @@
...
+        int numDuplicates = 0;
+        
+        libraryIdGenerator = new LibraryIdGenerator(headerAndIterator.header);
+        
+        for (final DuplicateSet duplicateSet : new IterableAdapter<DuplicateSet>(iterator)) {
+            final SAMRecord representative = duplicateSet.getRepresentative();
+            final boolean doOpticalDuplicateTracking = representative.getReadPairedFlag() &&
+                    representative.getFirstOfPairFlag() &&
+                    !representative.getReadUnmappedFlag() &&
+                    !representative.getMateUnmappedFlag();
```
tovanadler: It might make sense to make a named function for getReadPairedFlag, !getReadUnmappedFlag and !getMateUnmappedFlag, as you use it below as well and it might be nice to label what concept those three together capture.

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
```diff
@@ -0,0 +1,211 @@
...
+            final Set<String> readEndsSeen = new HashSet<String>();
+            
+            final List<ReadEnds> readEnds = new ArrayList<ReadEnds>();
+            for (final SAMRecord record : duplicateSet.getRecords()) {
+
+                if (!record.isSecondaryOrSupplementary()) {
+                    final String library = LibraryIdGenerator.getLibraryName(header, record);
```
tovanadler: consider inverting the boolean and exiting early, just to make it a little bit more readable.

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
```diff
@@ -0,0 +1,211 @@
...
+                    }
+                    out.addAlignment(record);
+                    progress.record(record);
+                }
+            }
+
+            // Track the optical duplicates
+            if (1 < readEnds.size()) {
+                AbstractMarkDuplicatesCommandLineProgram.trackOpticalDuplicates(readEnds, opticalDuplicateFinder, libraryIdGenerator);
+            }
```
tovanadler: Do you want to use the result of this and update the optical duplicate metric?

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        // remember to close the inputs
+        iterator.close();
+
+        out.close();
+
+        if (this.READ_NAME_REGEX == null) {
+            log.warn("Skipped optical duplicate cluster discovery; library size estimation may be inaccurate!");
```
tovanadler: This is slightly confusing - In the base program trackOpticalDuplicates is only called if READ_NAME_REGEX is not null, but in this version you are not checking that condition (like 187 does not seem to depend on it). Could you unify the conditions? Also, in general, is markDuploicates from the base MarkDuplicates class ever called?

src/java/picard/sam/markduplicates/util/AbstractMarkDuplicatesCommandLineProgram.java
nh13: Agreed, update.

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
nh13: PCR duplicates yes, optical duplicates not fully specified (see ```AbstractMarkDuplicatesCommandLineProgram.trackOpticalDuplicates```).

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
nh13: No, it returns all the records in the duplicate set.  Otherwise it would be something like ```getDulicateRecordsOnly``` or the like.

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
nh13: But then we would miss the code on https://github.com/broadinstitute/picard/pull/200/files#diff-b53ccb5378397aee3630490ae0092165R176

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
nh13: We also care that the representative's mate is also mapped (not unmapped).  

I am assuming you mean ```duplicate pairs could be in either order`` to be in terms of orientation.  For PCR duplicates yes, for optical duplicates no.

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
nh13: It's updated in the call, see ```opticalDuplicatesByLibraryId.increment```.

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
nh13: Good catch!

src/tests/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.Set;
+
+/**
+ * This is a simple tool to mark duplicates using the DuplicateSetIterator, DuplicateSet, and SAMRecordDuplicateComparator.
+ * 
+ * Users should continue to use MarkDuplicates in general.  The main motivation of this tool was the fact that 
+ * MarkDuplicates has many, many, many useful test cases, but few unit tests for validating individual duplicate sets. To
+ * test the DuplicateSetIterator, DuplicateSet, and SAMRecordDuplicateComparator, the most expedient method was to write
+ * this tool and make sure it behaves similarly to MarkDuplicates.  Not the best, I know, but good enough.  NH 06/25/2015.
+ *
```
tovanadler: Thanks!

PR #202, opened 2015-04-29
Make method public so that a sequence dictionary can be created programm...

alecw:
...atically

without having to write, then read a file.


nh13:
:+1:

alecw:
image

PR #203, opened 2015-04-29
adding CollectRnaSeqMetrics to CollectMultipleMetrics

nh13:


kbergin:
Besides the one question I had about the semi colon, it looks good to me! :+1:

nh13:
In Java enums are treated as classes, so the semi-colon can be omitted.

PR #213, opened 2015-05-21
Print default value for option that is a collection.

alecw:
I don't know why this wasn't done originally.  The code dates back to 2009.  Who wrote this anyway???


tfenne:
image

alecw:
image

PR #218, opened 2015-05-28
Don't ignore strand when comparing Genes.

alecw:
E.g. if you read the following with RefFlatReader, before this bug fix you would end up with only one Gene:

MIR1253	MIR1253-201	17	-	2651371	2651476	2651371	2651476	1	2651371,	2651476,
hsa-mir-1253	hsa-mir-1253.1-001	17	+	2651371	2651476	2651371	2651476	1	2651371,	2651476,

### DIFF
--- a/src/java/picard/annotation/Gene.java
+++ b/src/java/picard/annotation/Gene.java
@@ -57,6 +57,23 @@ public class Gene extends Interval implements Iterable<Gene.Transcript>  {
         return transcripts.values().iterator();
     }
 
+    public int compareTo(final Gene that) {
+        final int ret = super.compareTo(that);
+        if (ret != 0) return ret;
+        return Boolean.valueOf(this.isPositiveStrand()).compareTo(that.isPositiveStrand());
+    }
+
+    public boolean equals(final Object other) {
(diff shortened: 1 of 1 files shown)


alecw:
Can this be merged? In general it's a little lame that RefFlatReader collapses entries from the input (who wrote that code anyway???) and this reduces the possibility.

vdauwera:
Makes sense to me. @geoffjentry can we get this reviewed?

geoffjentry:
@alecw can you rebase this on top of master?

I did try to verify that this doesn't break picard-private. I had 2 test failures, although I had 15 when I compared w/ master picard so YMMV :) I did see @yfarjoun say that picard is currently breaking picard-private.

yfarjoun:
Yes, picard-private tests are not passing due to (I think) a recent commit
that modified some metrics files (which then cannot be uploaded to the DB
in the test)

alecw:
OK, rebased.

geoffjentry:
Ok - this appears to have the same set of test failures on picard-private as master does so I'll call this complete. Thanks @alecw

PR #221, opened 2015-06-02
CollectGcBiasMetrics as a MultiLevelCollector

kbergin:
Changed CollectGcBiasMetrics to work as a multilevel collector.

Added test - Fixed test to create its own temporary sam file from chrM for testing

### DIFF
--- a/src/java/picard/analysis/CollectGcBiasMetrics.java
+++ b/src/java/picard/analysis/CollectGcBiasMetrics.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2009 The Broad Institute
+ * Copyright (c) 2015 The Broad Institute
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -25,24 +25,28 @@
 package picard.analysis;
 
 import htsjdk.samtools.SAMFileHeader;
-import htsjdk.samtools.SAMReadGroupRecord;
 import htsjdk.samtools.SAMRecord;
 import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.reference.ReferenceSequence;
+import htsjdk.samtools.reference.ReferenceSequenceFile;
+import htsjdk.samtools.reference.ReferenceSequenceFileFactory;
+import htsjdk.samtools.util.CollectionUtil;
 import htsjdk.samtools.util.IOUtil;
-import htsjdk.samtools.util.QualityUtil;
-import htsjdk.samtools.util.SequenceUtil;
 import htsjdk.samtools.util.StringUtil;
+import picard.PicardException;
+import picard.analysis.directed.GcBiasMetricsCollector;
 import picard.cmdline.CommandLineProgramProperties;
 import picard.cmdline.Option;
-import picard.cmdline.StandardOptionDefinitions;
 import picard.cmdline.programgroups.Metrics;
+import picard.metrics.GcBiasMetrics;
 import picard.util.RExecutor;
 
 import java.io.File;
 import java.text.NumberFormat;
-import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 /**
  * Tool to collect information about GC bias in the reads in a given BAM file. Computes
@@ -52,6 +56,7 @@ import java.util.List;
  * to the average number of reads per window across the whole genome.
  *
  * @author Tim Fennell
+ * edited by Kylee Bergin
  */
 @CommandLineProgramProperties(
         usage = "Tool to collect information about GC bias in the reads in a given BAM file. Computes" +
@@ -64,7 +69,7 @@ import java.util.List;
 )
 public class CollectGcBiasMetrics extends SinglePassSamProgram {
     /** The location of the R script to do the plotting. */
-    private static final String R_SCRIPT = "picard/analysis/gcBias.R";
+    private static final File R_SCRIPT = new File("src/scripts/picard/analysis/","gcBias.R");
 
     // Usage and parameters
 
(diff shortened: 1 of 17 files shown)


src/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -83,187 +87,105 @@
...
+    @Option(shortName = "LEVEL", doc = "The level(s) at which to accumulate metrics.  ")
+    private final Set<MetricAccumulationLevel> METRIC_ACCUMULATION_LEVEL = CollectionUtil.makeSet(MetricAccumulationLevel.ALL_READS);
 
-    /** Stock main method. */
+    // Calculates GcBiasMetrics for all METRIC_ACCUMULATION_LEVELs provided
+    private GcBiasMetricsCollector multiCollector;
+    public byte[] refBases;
+    public final int windowSize = WINDOW_SIZE;
+    final int[] windowsByGc = new int[WINDOWS];
+    public static final int WINDOWS = 101;
```
gbggrant: Why is this public now?  Also, you might want to leave the previous comment 'Need 101 to get from 0-100' (or not)

src/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -83,187 +87,105 @@
...
-    private byte[] gc;
-    private byte[] refBases;
-    private String saveHeader;
+    @Option(shortName = "LEVEL", doc = "The level(s) at which to accumulate metrics.  ")
+    private final Set<MetricAccumulationLevel> METRIC_ACCUMULATION_LEVEL = CollectionUtil.makeSet(MetricAccumulationLevel.ALL_READS);
 
-    /** Stock main method. */
+    // Calculates GcBiasMetrics for all METRIC_ACCUMULATION_LEVELs provided
+    private GcBiasMetricsCollector multiCollector;
+    public byte[] refBases;
```
gbggrant: Was going to ask why this is public now, but I don't think it's even used.

src/java/picard/analysis/CollectGcBiasMetrics.java
gbggrant: Maybe
private static final int WINDOWS = WINDOW_SIZE+1

src/java/picard/analysis/CollectGcBiasMetrics.java
kbergin: I think you're right, after I moved some code around I put it in setup instead. Will delete, thanks!

src/java/picard/analysis/CollectGcBiasMetrics.java
kbergin: This is a little confusing, but the number of WINDOWS should always be 101 because we want gcbias calculated from 0% gc content to 100% gc content. However, WINDOW_SIZE is a command line input that corresponds to the size of the window that moves along the reference genome. WINDOW_SIZE doesn't always need to be 100 but WINDOWS should always be 101.

src/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -83,187 +87,105 @@
...
     }
 
+    /////////////////////////////////////////////////////////////////////////////
+    // Setup calculates gc[] for the reference. Must be done at startup to avoid
+    // missing reference sequences in the case of small files that may
+    // not have reads aligning to every reference sequence
+    /////////////////////////////////////////////////////////////////////////////
     @Override
-    protected void setup(final SAMFileHeader header, final File samFile) {
+    protected void setup(final SAMFileHeader header, final File samFile, final File referenceFile) {
```
gbggrant: referenceFile is passed to this method but not used.  Looks like the makeItSo method in SinglePassSamProgram sets referenceFile from REFERENCE_SEQUENCE, so you should use that.

src/java/picard/analysis/directed/GcBiasMetricsCollector.java
```diff
@@ -0,0 +1,286 @@
...
+
+        public void finish() {}
+
+        /////////////////////////////////////////////////////////////////////////////
+        // Sums the values in an int[].
+        /////////////////////////////////////////////////////////////////////////////
+        private double sum(final int[] values) {
+            final int length = values.length;
+            double total = 0;
+            for (int i = 0; i < length; ++i) {
```
gbggrant: i++ is generally the convention unless you explicitly need to preincrement.

kbergin: Wow I didn't even notice that, that's old code from the original but I will change it. I've never seen ++i before!

src/java/picard/analysis/SinglePassSamProgram.java
```diff
@@ -138,8 +138,7 @@ public static void makeItSo(final File input,
     protected boolean usesNoRefReads() { return true; }
 
     /** Should be implemented by subclasses to do one-time initialization work. */
-    protected abstract void setup(final SAMFileHeader header, final File samFile);
-
+    protected abstract void setup(final SAMFileHeader header, final File samFile, final File referenceSequence);
```
gbggrant: I don't think you really need to pass along referenceSequence here.  The only method that calls setup 'makeItSo', has just passes the global 'REFERENCE_SEQUENCE' directly to setup.

It would probably simplify the code if you just used REFERENCE_SEQUENCE itself within the setup method of CollectGcBiasMetrics

src/java/picard/analysis/CollectGcBiasMetrics.java
gbggrant: Second comment here.  Looking further into SinglePassSamProgram, you are just passing along the global REFERENCE_SEQUENCE, now as a parameter.  Probably better to just use it directly here and not parameterize it.

src/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -83,187 +87,105 @@
...
         IOUtil.assertFileIsWritable(CHART_OUTPUT);
         if (SUMMARY_OUTPUT != null) IOUtil.assertFileIsWritable(SUMMARY_OUTPUT);
-
-        try{
-            saveHeader = header.getReadGroups().get(0).getLibrary();
-        } catch (Exception ignored){
-            saveHeader = "LibraryNameUnavailable";
+        final ReferenceSequenceFile refFile = ReferenceSequenceFileFactory.getReferenceSequenceFile(REFERENCE_SEQUENCE);
```
gbggrant: Is it possible for REFERENCE_SEQUENCE to be null?  Should handle this if so (i.e. in a CustomCommandLineValidation to prevent an ugly exception).

src/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
```diff
@@ -138,7 +138,7 @@ public static void main(final String[] args) {
     }
 
     @Override
-    protected void setup(final SAMFileHeader header, final File samFile) {
+    protected void setup(final SAMFileHeader header, final File samFile, final File referenceSequence) {
```
gbggrant: Note that this change also breaks the compile for personal tools in picard private (which are not built by the ant scripts).  Not sure what protocol is for that.

src/java/picard/metrics/GcBiasMetrics.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009 The Broad Institute
```
gbggrant: 2009???

src/tests/java/picard/analysis/CollectGcBiasMetricsTest.java
```diff
@@ -0,0 +1,217 @@
...
+    final static String readGroupId3 = "TestReadGroup3";
+    final static String platform = "ILLUMINA";
+    final static String library1 = "TestLibrary1";
+    final static String library2 = "TestLibrary2";
+    final static String library3 = "TestLibrary3";
+
+    final static File TEST_DIR = new File("testdata/picard/sam/CollectGcBiasMetrics/");
+    final File dict = new File(TEST_DIR, "Mheader.dict");
+
+    File tempDir;
```
gbggrant: Not used

src/tests/java/picard/analysis/CollectGcBiasMetricsTest.java
```diff
@@ -0,0 +1,217 @@
...
+    File tempSamFile;
+    File samFile;
```
gbggrant: Not used.  (intelliJ subtly colors these light grey to indicate their lack of use in the code).

src/tests/java/picard/analysis/CollectGcBiasMetricsTest.java
```diff
@@ -0,0 +1,217 @@
...
+
+    //create a samfile with different samples, read groups and libraries that overlap for testing.
+    @BeforeTest
+    void setupBuilder() throws IOException {
+        final int numReads = 100;
+        final String flowCellBarcode = "TESTBARCODE";
+        final String readName = flowCellBarcode;
+
+        tempSamFile = File.createTempFile("CollectGcBias", ".bam", TEST_DIR);
+        File tempSamFileUnsorted = File.createTempFile("CollectGcBias", ".bam", TEST_DIR);
```
gbggrant: Minor comment.  You can use outfile.deleteOnExit(); to set these up for deletion now - kind of nice.

src/java/picard/analysis/directed/GcBiasMetricsCollector.java
```diff
@@ -0,0 +1,286 @@
...
+import java.util.Set;
+import java.util.HashMap;
+
+/** Calculates GC Bias Metrics on multiple levels
+ * Created by kbergin on 3/23/15.
+ */
+public class GcBiasMetricsCollector extends MultiLevelCollector<GcBiasMetrics, Integer, GcBiasCollectorArgs> {
+    // Histograms to track the number of windows at each GC, and the number of read starts
+    // at windows of each GC. Need 101 to get from 0-100.
+    public final int windowSize;
```
gbggrant: These can all be private

src/java/picard/analysis/directed/GcBiasMetricsCollector.java
```diff
@@ -0,0 +1,286 @@
...
+                prefix = this.readGroup;
+                gcData.put(prefix, new GcObject());
+            } else if (this.library != null) {
+                prefix = this.library;
+                gcData.put(prefix, new GcObject());
+            } else if (this.sample != null) {
+                prefix = this.sample;
+                gcData.put(prefix, new GcObject());
+            } else {
+                prefix = "All_Reads";
```
gbggrant: Might want to make this a symbol since you are looking up by it later.

src/java/picard/analysis/directed/GcBiasMetricsCollector.java
```diff
@@ -0,0 +1,286 @@
...
+                final long[] errorsByGc = gcCur.errorsByGc;
+                final long[] basesByGc = gcCur.basesByGc;
+                final int totalClusters = gcCur.totalClusters;
+                final int totalAlignedReads = gcCur.totalAlignedReads;
+                final String group = gcCur.group;
+                final GcBiasMetrics metrics = new GcBiasMetrics();
+                final double totalWindows = sum(windowsByGc);
+                final double totalReads = sum(readsByGc);
+                final double meanReadsPerWindow = totalReads / totalWindows;
+                if(totalAlignedReads>0) {
```
gbggrant: add a space between if and '('

src/java/picard/analysis/directed/GcBiasMetricsCollector.java
```diff
@@ -0,0 +1,286 @@
...
+                    for (int i = 0; i < windowsByGc.length; ++i) {
+                        final GcBiasDetailMetrics m = new GcBiasDetailMetrics();
+                        m.GC = i;
+                        m.WINDOWS = windowsByGc[i];
+                        m.READ_STARTS = readsByGc[i];
+                        if (errorsByGc[i] > 0) m.MEAN_BASE_QUALITY = QualityUtil.getPhredScoreFromObsAndErrors(basesByGc[i], errorsByGc[i]);
+                        if(windowsByGc[i]!=0) {
```
gbggrant: spacing 'if (windowsByGc[i] != 0) {'
(I know this is just copy and pasting, but it hurts my eyes to see such spacing)

src/java/picard/analysis/directed/GcBiasMetricsCollector.java
```diff
@@ -0,0 +1,286 @@
...
+        double gcDropout = 0;
+
+        for (final GcBiasDetailMetrics detail : details) {
+            final double relativeReads = detail.READ_STARTS / totalReads;
+            final double relativeWindows = detail.WINDOWS / totalWindows;
+            final double dropout = (relativeWindows - relativeReads) * 100;
+
+            if (dropout > 0) {
+                if (detail.GC <= 50) atDropout += dropout;
+                if (detail.GC >= 50) gcDropout += dropout;
```
gbggrant: This could be an if / else.

src/java/picard/analysis/directed/GcBiasMetricsCollector.java
```diff
@@ -0,0 +1,286 @@
...
+                        if(windowsByGc[i]!=0) {
+                            m.NORMALIZED_COVERAGE = (m.READ_STARTS / (double) m.WINDOWS) / meanReadsPerWindow;
+                            m.ERROR_BAR_WIDTH = (Math.sqrt(m.READ_STARTS) / (double) m.WINDOWS) / meanReadsPerWindow;
+                        }
+                        else{
+                            m.NORMALIZED_COVERAGE = 0;
+                            m.ERROR_BAR_WIDTH = 0;
+                        }
+                        m.ACCUMULATION_LEVEL = group;
+                        if(group.equals("Read Group")) {m.READ_GROUP = gcType;}
```
gbggrant: add appropritate white spacing here and on the next couple of lines

src/java/picard/analysis/directed/GcBiasMetricsCollector.java
```diff
@@ -0,0 +1,286 @@
...
+                        else if(group.equals("Sample")) {m.SAMPLE = gcType;}
+                        else if(group.equals("Library")) {m.LIBRARY = gcType;}
+                        metrics.DETAILS.addMetric(m);
+                    }
+                    // Synthesize the high level summary metrics
+                    final GcBiasSummaryMetrics s = new GcBiasSummaryMetrics();
```
gbggrant: Even though it's only used briefly, better to use a longer/more informative name than 's' for a variable.

gbggrant:
All set.  Mostly stylistic and protection level comments.  And the probably unnecessary change of adding referenceFile to SinglePassSamProgram.setup.  Once corrected then :+1:

src/java/picard/analysis/CollectGcBiasMetrics.java
kbergin: You are right!

src/java/picard/metrics/GcBiasMetrics.java
kbergin: :) the original code is very old, I didn't change the calculations it does, just so that it works with single pass sam and the multi level collector. Should I just change this to 2015 or is there a new copyright statement I should paste in?

src/tests/java/picard/analysis/CollectGcBiasMetricsTest.java
kbergin: Would this delete when the entire test exits or when setupBuilder exits?

src/java/picard/metrics/GcBiasMetrics.java
gbggrant: Probably best to update it to 2015.  I don't know if the copyright wording has updated since 2009.  Maybe search the code base for a more recent copyright?

src/tests/java/picard/analysis/CollectGcBiasMetricsTest.java
gbggrant: When the VM terminates, so should be when the test exits.

kbergin: Awesome just making sure. I saw that online but I wanted to make sure I was understanding what it meant correctly :)

src/java/picard/analysis/directed/GcBiasMetricsCollector.java
kbergin: Is a symbol a variable type that I don't know of? I tried googling but am not sure I am understanding what you mean..

kbergin: @gbggrant, @yfarjoun

gbggrant: Sorry - didn't see your comment.  I was thinking just a private static final string in the class.  If you use a string more than once like this, (i.e. for storing in a hash and then later looking up) I tend to do that sort of thing to make it clear it's a special purpose string and avoiding errors when you make a typo in one.

kbergin: Np! Ah ok thanks! good call

yfarjoun:
Kylee, make sure that you are aware of the next "picard release" and that
you see that the gc metrics are produced as are all the other metrics etc.

kbergin:
Ok, who should I talk to about that- I saw your hip chat about picard private breaking :( what broke and how do I fix it?

yfarjoun:
Can you follow the link?
http://prodinfobuild.broadinstitute.org:8085/browse/PICARD-TRUNK-3823
It shows that the breakage happens when some test tries inserting your new
metric files into the database.

I think it's going to be a bit difficult to figure it out on your own. I
recommend asking Matt Sooknah or Miguel since the problem is that the
database is missing the "ACCUMULATION_LEVEL" field, which is now added
since you are collecting using the multi-level collector.

Does this make sense?

Yossi.

kbergin:
Oh that makes sense. Yes I will need help, who volunteers as tribute? :P

Thanks!

yfarjoun:
I'm sure the former picard team will be happy to assist.

Y.

PR #222, opened 2015-06-10
Minor change to BedToIntervalList to make sorting and uniquing option…

tfenne:
…al vs. the prior behaviour of always sorting and uniquing the intervals.  Makes is easier to create an interval list that exactly matches a BED file when the BED file is out of order or contains overlapping regions.

### DIFF
--- a/src/java/picard/util/BedToIntervalList.java
+++ b/src/java/picard/util/BedToIntervalList.java
@@ -44,6 +44,12 @@ public class BedToIntervalList extends CommandLineProgram {
     @Option(shortName = StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "The output Picard Interval List")
     public File OUTPUT;
 
+    @Option(doc="If true, sort the output interval list before writing it.")
+    public boolean SORT = true;
+
+    @Option(doc="If true, unique the output interval list by merging overlapping regions, before writing it (implies sort=true).")
+    public boolean UNIQUE = true;
+
     final Log LOG = Log.getInstance(getClass());
 
     // Stock main method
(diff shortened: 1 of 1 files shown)


src/java/picard/util/BedToIntervalList.java
```diff
@@ -44,6 +44,12 @@
     @Option(shortName = StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "The output Picard Interval List")
     public File OUTPUT;
 
+    @Option(doc="If true, sort the output interval list before writing it.")
+    public boolean SORT = true;
+
+    @Option(doc="If true, unique the output interval list by merging overlapping regions, before writing it (implies sort=true).")
```
akiezun: the code does not follow this. nothing i see makes unique imply sort=true

akiezun:
no tests?

tfenne:
No - no tests.  All it is doing is turning on/off well tested functionality in IntervalList.

src/java/picard/util/BedToIntervalList.java
tfenne: Take a look at IntervalList.java:137 - the implementation of IntervalList.uniqued() is documented that it will sort the interval list prior to merging adjacent intervals.

akiezun: That would indicate that the method is misnamed then: it should be called `sortedAndUniqued()`

nh13:
:+1:  from me, @akiezun since you chimed in?

akiezun:
not up to hellbender standards but i'll let this one go :) :+1:

PR #224, opened 2015-06-16
Disable escaping html of preamble text in doc generation

vdauwera:
Done to enable use of HTML formatting in the documentation. Includes an example of what this allows us to do; see CollectAlignmentSummaryMetrics.java

For a preview of what the result looks like, build from source and see the raw html page in dist/html/program_usage

### DIFF
--- a/src/java/picard/analysis/CollectAlignmentSummaryMetrics.java
+++ b/src/java/picard/analysis/CollectAlignmentSummaryMetrics.java
@@ -64,12 +64,21 @@ import java.util.Set;
  * @author Doug Voet (dvoet at broadinstitute dot org)
  */
 @CommandLineProgramProperties(
-        usage = CollectAlignmentSummaryMetrics.USAGE,
-        usageShort = CollectAlignmentSummaryMetrics.USAGE,
+        usage = CollectAlignmentSummaryMetrics.USAGE_SUMMARY + CollectAlignmentSummaryMetrics.USAGE_DETAILS,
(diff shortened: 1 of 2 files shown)


src/java/picard/analysis/CollectAlignmentSummaryMetrics.java
```diff
@@ -69,7 +69,14 @@
         programGroup = Metrics.class
 )
 public class CollectAlignmentSummaryMetrics extends SinglePassSamProgram {
-    static final String USAGE = "Produces from a SAM or BAM a file containing summary alignment metrics";
+    static final String USAGE = "Produces from a SAM or BAM a file containing summary alignment metrics." +
```
nh13: See line 68.  This USAGE is used in the short usage part of the command line, and so it will go over multiple lines (not good).  You will have to split this into a short and long usage.

vdauwera: Oh right -- can you point me to an example if the short/long usage syntax?  If you know one off the top of your head, otherwise I'll look for one.

nh13: There should be ton, take a look at other clps.

vdauwera:
@nh13 Fixed (split into short and long usage) and rebased.

vdauwera:
@nh13 Is it ok to merge now? David needs this to move ahead with his work.

nh13:
:+1:

vdauwera:
Thanks

Issue #225, opened 2015-06-17
CollectMultipleMetrics IllegalArgumentException in 1.134


jeremiahsavage:
I can run 1.133 on a particular BAM without error.

```
$ java -d64 -jar picard-tools-1.133/picard.jar CollectMultipleMetrics INPUT=2817851434.bam OUTPUT=2817851434.bam.metrics VALIDATION_STRINGENCY=LENIENT
[Wed Jun 17 16:58:19 UTC 2015] picard.analysis.CollectMultipleMetrics INPUT=2817851434.bam OUTPUT=2817851434.bam.metrics VALIDATION_STRINGENCY=LENIENT    ASSUME_SORTED=true STOP_AFTER=0 PROGRAM=[CollectAlignmentSummaryMetrics, CollectInsertSizeMetrics, QualityScoreDistribution, MeanQualityByCycle, CollectBaseDistributionByCycle] VERBOSITY=INFO QUIET=false COMPRESSION_LEVEL=5 MAX_RECORDS_IN_RAM=500000 CREATE_INDEX=false CREATE_MD5_FILE=false GA4GH_CLIENT_SECRETS=client_secrets.json
[Wed Jun 17 16:58:19 UTC 2015] Executing as ubuntu@frontend001 on Linux 3.13.0-53-generic amd64; OpenJDK 64-Bit Server VM 1.7.0_79-b14; Picard version: 1.133(95818b6e611725775e29d3b4acbaa23747eb3e94_1433169311) IntelDeflater
INFO	2015-06-17 16:58:23	SinglePassSamProgram	Processed     1,000,000 records.  Elapsed time: 00:00:04s.  Time for last 1,000,000:    4s.  Last read position: chr2:121,708,696
.
.
.
```
 However, with 1.134, I get
```
$ java -d64 -jar picard-tools-1.134/picard.jar CollectMultipleMetrics INPUT=2817851434.bam OUTPUT=2817851434.bam.metrics VALIDATION_STRINGENCY=LENIENT
[...]

PR #226, opened 2015-06-18
Fixing bugs in CollectMultipleMetrics

kbergin:
Needs to notify a user if reference sequence is not specified but is needed. Also removes CollectGcBiasMetrics from the list of default programs. In response to #225.

### DIFF
--- a/src/java/picard/analysis/CollectMultipleMetrics.java
+++ b/src/java/picard/analysis/CollectMultipleMetrics.java
@@ -1,6 +1,7 @@
 package picard.analysis;
 
 import htsjdk.samtools.util.CollectionUtil;
+import picard.PicardException;
 import picard.cmdline.CommandLineProgram;
 import picard.cmdline.CommandLineProgramProperties;
 import picard.cmdline.Option;
@@ -9,7 +10,6 @@ import picard.cmdline.StandardOptionDefinitions;
 
 import java.io.File;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


src/java/picard/analysis/CollectMultipleMetrics.java
```diff
@@ -184,7 +217,10 @@ public int doWork() {
 
         final List<SinglePassSamProgram> programs = new ArrayList<SinglePassSamProgram>();
         for (final ProgramInterface program : new HashSet<ProgramInterface>(programsToRun)) {
-            final SinglePassSamProgram instance = program.makeInstance(OUTPUT , INPUT, REFERENCE_SEQUENCE);
+            if (program.needsReferenceSequence() && REFERENCE_SEQUENCE==null) {
+                throw new PicardException("A program you are trying to run needs a Reference Sequence, please set REFERENCE_SEQUENCE in the command line");
```
nh13: Instead of ```a program``` can you print the program's name?

kbergin: I was hoping to be able to but couldn't figure it out right away. Let me look into it further.

yfarjoun: won't program.toString() do it?

kbergin: Yes it does, I figured that out finally and added a new commit where it does that a little while ago. Thanks Yossi!

bradtaylor:
Looks good to me!

kbergin:
Thank you!

PR #234, opened 2015-07-15
fixed a small bug in IntervalListTool and remove spurious exception

yfarjoun:
- Fixed a small bug in IntervalListTool where in some cases bases were not accounted for in the total count (this can lead to non-uniform intervalLists in the scatter)

- Removed an erroneous Exception that was thrown when the intervals were not uniformly enough sized. We cannot guarantee that they will be uniformly sized...

- Fixed bug that was causing the Last intervalList to be double the size of the rest.


ktibbett:
image

yfarjoun:
Note: the last interval generated here might still be too big. This needs a careful look.

yfarjoun:
@ktibbett found -1 bug we discussed yesterday. Please re-review.
@nh13 you might enjoy this one too.

nh13:
:+1:

PR #235, opened 2015-07-16
the output of scattering intervals should be .interval_list

nh13:


yfarjoun:
:+1: need to coordinate the release.

PR #236, opened 2015-07-17
Changes to picard Dockerfile

bradtaylor:
Build it from source directly, instead of relying on jar to exist. Should allow dockerhub to build the image


Dockerfile
```diff
@@ -1,7 +1,24 @@
...
+
+# Assumes Dockerfile lives in root of the git repo. Pull source files into container
+COPY build.xml /usr/picard/build.xml
+COPY src /usr/picard/src
+COPY lib /usr/picard/lib
 WORKDIR /usr/picard
 
-ENTRYPOINT ["java", "-jar", "picard.jar"]
+# Clone out htsjdk. First turn off git ssl verification
+RUN git config --global http.sslVerify false && git clone https://github.com/samtools/htsjdk.git
```
gbggrant: There is an htsjdk-clone target in picard-public's build.xml, but perhaps you are doing it this way to get around the ssl verification?

Dockerfile
```diff
@@ -1,7 +1,24 @@
...
+
+# Build the distribution jar
+RUN ant clean all
+
```
gbggrant: Do you want to clean up /usr/picard/src and /usr/picard/lib?  I think you only don't need them once picard is compiled.

gbggrant:
:+1:

Dockerfile
bradtaylor: Yah, clone-htdjdk uses the ssh clone, and root user in a docker container doesn't have keys.

bradtaylor:
DO NOT MERGE. Going to add a dockerignore and a helper script so we can deal with JVM args

bradtaylor:
@gbggrant I added a helper script to facilitate passing in JVM args. Can you re-review?

Dockerfile
bradtaylor: Also, for some reason the cache on this layer gets invalidated and the clone is re-run whenever a change is made to picard source files. No idea why. I had them as separate run command explicitly so I could take advantage of caching.

PR #237, opened 2015-07-20
better error messaging for corrupt bcl.gz files

davidangb:
Track the files on disk that are used to create InputStreams. Then, if a stream fails, we can output the offending file within our exception.


nh13:
:+1:

PR #229, opened 2015-07-01
Editing genotypeConcordance to handle a new scheme

kbergin:
-Working on generalizing the scheme code to work in a factory design pattern for multiple scheme options.

-Removed duplicate scheme validation and edited new scheme to match Erics suggestion for testing. This change to the code was brought on by a need to count missing sites in the truth set in a different way due to a certain type of truth set that has hom ref sites as 'missing'.

-Attempting to refactor tests to correspond to scheme factory refactor done in code base.
-Added new interval list for new scheme testing purposes. Updating tests accordingly

PR #238, opened 2015-07-21
Fixing previous mistake with scheme logic

kbergin:
I had a mistake in the scheme logic in the last pull request where I had set most of the MISSING truth states to call like NO_CALLS, when in fact they should be called identical to HOM_REF. This pull request fixes that scheme logic, and the corresponding tests.


kbergin:
@nh13 I made this new PR to fix the scheme settings in #229. I changed the MISSING column and the hard coded test comparisons in the corresponding tests. If you get a chance, could you :+1: if it looks good?

yfarjoun:
did you calculate the test results by hand?

kbergin:
Yes I did. I did last time as well but I calculated them off of a scheme that wasn't valid so they still came out as correct then.

yfarjoun:
just checking! :-)

kbergin:
Thanks!!

nh13:
:+1:

kbergin:
Thanks!

PR #243, opened 2015-07-28
More updates to Genotype Concordance

kbergin:
Updating Genotype Concordance to track MISSING-MISSING states, calculate specificity only for VAR and not for HET and HOM, and adding new metrics Genotype Concordance and Non-Ref Genotype Concordance

Forgot to make a variable final


src/java/picard/vcf/GenotypeConcordance.java
```diff
@@ -220,6 +221,10 @@ public static void main(final String[] args) {
         snpCounter   = new GenotypeConcordanceCounts();
         indelCounter = new GenotypeConcordanceCounts();
 
+        //need to know size of intervals to add missing-missing sites for NIST schema.
+        long intervalBaseCount = 0L;
+        if(usingIntervals && MISSING_SITES_HOM_REF) {intervalBaseCount = intervals.getBaseCount();}
```
geoffjentry: add a space after `if`

src/java/picard/vcf/GenotypeConcordance.java
```diff
@@ -281,6 +286,17 @@ else if (callVariantContextType == INDEL) {
             progress.record(variantContextForLogging.getContig(), variantContextForLogging.getStart());
         }
 
+        //snp counter add in X number of missing-missing hom ref's (truth and call state)
+        //missing missing is total interval size minus number of iterations in while loop
+
+        if(MISSING_SITES_HOM_REF && usingIntervals){
```
geoffjentry: space after the `if` and before the `{`

src/java/picard/vcf/GenotypeConcordance.java
```diff
@@ -281,6 +286,17 @@ else if (callVariantContextType == INDEL) {
...
+            addMissingMissingSites(snpCounter.getCounterSize(), intervalBaseCount, snpCounter);
```
geoffjentry: I see it throughout so guessing the answer is 'yes' but double checking that `Missing`

kbergin:
Sorry for the travis fails; I forgot to update one of the test files to include the new metrics so it's calling them failures. @geoffjentry the changes would be to new files that aren't changed in the current commits, so don't worry, the comments you made are still relevant :) Thank you!

src/java/picard/vcf/GenotypeConcordance.java
nh13: beat me too it

src/java/picard/vcf/GenotypeConcordance.java
nh13: Perhaps ```MissingTruthAndMissingCall```

src/java/picard/vcf/GenotypeConcordance.java
```diff
@@ -311,6 +327,18 @@ else if (callVariantContextType == INDEL) {
     }
 
     /**
+     * Method to add missing sites that are KNOWN to be HOM_REF in the case of the NIST truth data set.
+     * @param numVariants
+     * @param intervalBaseCount
+     * @param counter
```
nh13: The ```@param``` tags do not add value here.

src/java/picard/vcf/GenotypeConcordanceCounts.java
```diff
@@ -49,6 +53,14 @@ public void increment(final TruthAndCallStates truthAndCallStates) {
         this.counter.increment(truthAndCallStates);
     }
 
+    public void increment(final TruthAndCallStates truthAndCallStates, double count){
```
nh13: ```final double count```

src/java/picard/vcf/GenotypeConcordance.java
```diff
@@ -311,6 +327,18 @@ else if (callVariantContextType == INDEL) {
...
+     */
+    private void addMissingMissingSites(double numVariants , long intervalBaseCount, GenotypeConcordanceCounts counter){
```
nh13: missing finals here.  Check IntelliJ's yellow marks and fix them

src/java/picard/vcf/GenotypeConcordanceCounts.java
```diff
@@ -49,6 +53,14 @@ public void increment(final TruthAndCallStates truthAndCallStates) {
...
+        this.counter.increment(truthAndCallStates, count);
+    }
+
+    public double getCounterSize(){
```
nh13: spaces

src/java/picard/vcf/GenotypeConcordanceCounts.java
```diff
@@ -74,6 +86,72 @@ public void validateCountsAgainstScheme(final GenotypeConcordanceScheme scheme)
...
+
+        for (final TruthState truthState : allTruthStates) {
+            for (final CallState callState : allCallStates) {
+                if(!missingSitesFlag && isMissing(truthState, callState)) {
+                    continue;
+                }
+                else {
+                    final TruthAndCallStates truthAndCallStates = new TruthAndCallStates(truthState, callState);
+                    final int count = getCount(truthAndCallStates);
+                    if (truthState.toString().equals(callState.toString())) {//we are on the diagonal
```
nh13: space before ```//```

nh13: Is there a better way to do the comparison than testing string equality?

src/java/picard/vcf/GenotypeConcordance.java
eitanbanks: I'd recommend putting this all in one line:
final long intervalBaseCount = intervals.getBaseCount();
You don't need the if statement here to protect the variable initialization since it's only used below within the exact same if statement.  (And, come to think about it, you could just move this line down below to where it is used)

src/java/picard/vcf/GenotypeConcordance.java
```diff
@@ -281,6 +286,17 @@ else if (callVariantContextType == INDEL) {
...
 
+        //snp counter add in X number of missing-missing hom ref's (truth and call state)
+        //missing missing is total interval size minus number of iterations in while loop
+
+        if(MISSING_SITES_HOM_REF && usingIntervals){
+            addMissingMissingSites(snpCounter.getCounterSize(), intervalBaseCount, snpCounter);
+            addMissingMissingSites(indelCounter.getCounterSize(), intervalBaseCount, indelCounter);
+        }
+        else if(MISSING_SITES_HOM_REF && !usingIntervals) {
+            throw new PicardException("You cannot use the MISSING_SITES_HOM_REF flag without using a high confidence interval list");
```
eitanbanks: This check should be moved further up in this method (to where the other fail fast checks are).  It would be a shame to do tons of work only to find out at the end that the user neglected to provide high confidence intervals.

Then, given that if MISSING_SITES_HOM_REF is true here we know that usingIntervals must be true (otherwise you would have errored out earlier in the method), the if statement a few lines earlier (line 292) now only needs to be:
if ( MISSING_SITES_HOM_REF )

Finally, I would change the error message to something like:
"You cannot use the MISSING_HOM option without also supplying an interval list over which missing sites are considered confident homozygous reference calls."

src/java/picard/vcf/GenotypeConcordanceCounts.java
kbergin: The enum has a .equals() method but intelliJ doesn't like 
```truthState.equals(callState)```
saying they aren't comparable.

But if I do this: 
```truthState.name().equals(callState.name())```

it's fine with it. I think this would still be a string comparison though...

Comment (nh13) [inline src/java/picard/vcf/GenotypeConcordanceCounts.java]: No there isn't a better way since they are different enum classes, which are not inherently comparable.

File: src/java/picard/vcf/GenotypeConcordance.java (hunk 9 above)
Comment (nh13) [inline src/java/picard/vcf/GenotypeConcordance.java]: :+1:

File: src/java/picard/vcf/GenotypeConcordanceCounts.java (hunk 8 above)
Comment (kbergin) [inline src/java/picard/vcf/GenotypeConcordanceCounts.java]: If I put the enum constants in the same order in truth and call I can use the truthState.ordinal() (returns an int) to compare instead.

File: src/java/picard/vcf/GenotypeConcordanceCounts.java [hunk 10]
```diff
@@ -74,6 +86,76 @@ public void validateCountsAgainstScheme(final GenotypeConcordanceScheme scheme)
...
+
+        for (final TruthState truthState : allTruthStates) {
+            for (final CallState callState : allCallStates) {
+                if(!missingSitesFlag && isMissing(truthState, callState)) {
+                    continue;
+                }
+                else {
+                    final TruthAndCallStates truthAndCallStates = new TruthAndCallStates(truthState, callState);
+                    final int count = getCount(truthAndCallStates);
+                    if (truthState.ordinal()==callState.ordinal()) {
```
kbergin: @nh13 @eitanbanks is it better to compare this way, instead of a string comparison? As long as the CallState and TruthState enum's have the entries that can be paired in the same order then their ordinals will match. The order of the entries should never be changed so I think it will work.

eitanbanks: My opinion is that this is a dangerous way of doing it.  There's nothing to stop someone from adding another state in one of the enums and throwing off the ordering.  Here's a potentially better way (but @nh13 should weigh in too):

    public enum TruthState {
        MISSING (MISSING_CODE),
        HOM_REF (HOM_REF_CODE),
        HET_REF_VAR1 (HET_REF_CODE),
        etc.;

        private final int code;

        private TruthState(final int code) {
            this.code = code;
        }

        public int getCode() { return code; }
    }

where we have e.g. private final static int MISSING_CODE = 0;

eitanbanks: should have added:
then the comparison would be
if (truthState.getCode()==callState.getCode())

kbergin: That would look nice! I like that.

nh13: :+1:

src/java/picard/vcf/GenotypeConcordanceStates.java
```diff
@@ -18,21 +20,39 @@
...
+    private static final int HET_REF_VAR1_CODE = 2;
+    private static final int HET_VAR1_VAR2_CODE = 3;
+    private static final int HOM_VAR1_CODE = 4;
+    private static final int NO_CALL_CODE = 5;
+    private static final int LOW_GQ_CODE = 6;
+    private static final int LOW_DP_CODE = 7;
+    private static final int VC_FILTERED_CODE = 8;
+    private static final int GT_FILTERED_CODE = 9;
+    private static final int IS_MIXED_CODE = 10;
+    private static final int INCOMPARABLE_CODE = 11;
```
nh13: make an enum?

nh13:
:+1: but you need to rebase onto master and squash onto master.

src/java/picard/vcf/GenotypeConcordanceCounts.java
```diff
@@ -74,6 +85,76 @@ public void validateCountsAgainstScheme(final GenotypeConcordanceScheme scheme)
...
+        for (final TruthState truthState : allTruthStates) {
+            for (final CallState callState : allCallStates) {
+                if(!missingSitesFlag && isMissing(truthState, callState)) {
+                    continue;
+                }
+                else {
+                    final TruthAndCallStates truthAndCallStates = new TruthAndCallStates(truthState, callState);
+                    final int count = getCount(truthAndCallStates);
+                    if (truthState.getCode()==callState.getCode()) {
+                        //Because the TruthState and CallState enums are in the same order, the ordinals correspond to the same constants
```
eitanbanks: 2 minor points:
1. This comment no longer applies (since you aren't using ordinals anymore).
2. You can pull "denominator += count;" outside the if statement (since it's applied in either case) and get rid of the else block.

src/java/picard/vcf/GenotypeConcordanceCounts.java
```diff
@@ -74,6 +85,76 @@ public void validateCountsAgainstScheme(final GenotypeConcordanceScheme scheme)
...
+                        //If we enter this, we are 'on the diagonal
+                        numerator += count;
+                        denominator += count;
+                    } else {
+                        denominator += count;
+                    }
+                }
+            }
+        }
+        return (numerator / denominator);
```
eitanbanks: Is it possible to get to this line with a denominator that has never been incremented (e.g. if there are only missing-missing states)?  If so, the denominator will be zero and you'll exception out (division by zero is bad).
You could protect yourself by doing this instead:
    return (denominator > 0.0 ? (numerator / denominator) : 0.0);

kbergin: Even if there were only missing missing states, these would be in the numerator and the denominator so it should be 1. Unless it wasn't NIST then it would be 0/0.. So we could protect for that.

src/java/picard/vcf/GenotypeConcordanceCounts.java
```diff
@@ -74,6 +85,76 @@ public void validateCountsAgainstScheme(final GenotypeConcordanceScheme scheme)
...
+
+        final TruthState[] allTruthStates = TruthState.values();
+        final CallState[] allCallStates = CallState.values();
+
+        for (final TruthState truthState : allTruthStates) {
+            for (final CallState callState : allCallStates) {
+                if(!missingSitesFlag && isMissing(truthState, callState)) {
+                    continue;
+                }
+                else if (isVar(truthState, callState)) {
```
eitanbanks: This entire method is identical to calculateGenotypeConcordance() except for this one line.
Perhaps we could break it out into a utility method that the 2 public methods call into.

src/java/picard/vcf/GenotypeConcordanceCounts.java
nh13: Or return ```Double.NA```?

src/java/picard/vcf/GenotypeConcordanceCounts.java
nh13: @kbergin @eitanbanks I agree with Eric's point here.  Given that both are calculating concordance it makes sense to have a private helper method.  If one was sensitivity and the other specificity, there would be more value in keeping the separate to be explicit.  It's a judgement call.

nh13:
@gbggrant @kbergin  needs database migration in picard-private before release, since the metric definition has changed.

kbergin:
@nh13 @gbggrant if someone can review the last change refactoring the genotype concordance function into a utility function I can merge! :)

src/java/picard/vcf/GenotypeConcordanceCounts.java
```diff
@@ -74,6 +85,53 @@ public void validateCountsAgainstScheme(final GenotypeConcordanceScheme scheme)
...
+                    final int count = getCount(truthAndCallStates);
+                    if (truthState.getCode()==callState.getCode()) {
+                        //If we enter this, we are 'on the diagonal'
+                        numerator += count;
+                    }
+                    denominator += count;
+                }
+            }
+        }
+        return (denominator > 0.0 ? (numerator / denominator): Double.NaN);
```
nh13: space after ```:``` (optional)

nh13:
:+1:

PR #245, opened 2015-07-29
Make docker file compatible with cromwell

dshiga:
The existing docker file doesn't work with cromwell.

When Cromwell tries to start up the container, this nonsensical command gets issued:
java -jar picard.jar /bin/bash

and it errors out.

This happens because the picard docker file specifies an entry point of "java -jar picard.jar" but cromwell tries to call /bin/bash immediately when starting the container. The result is a concatenation of the two commands, which makes no sense and causes an error.

This PR removes the entry point in the docker file. Cromwell can now successfully start the container with /bin/bash and then issue whatever command you specify in the wdl file.


geoffjentry:
This raises the question of if cromwell should be doing that.

dshiga:
I personally think it's clearer to not have a custom entry point, because if you do then you have to look in two places, WDL + Dockerfile, to figure out what the full command is. So even if you add support for custom entry points, I personally don't want to use that feature!

mcovarr:
Per earlier conversation with @bradtaylor I believe the use of an entry point is considered a Docker best practice.  As discussed with @dshiga and @geoffjentry, it might be possible to default Cromwell to not use entry points by default via the `--entrypoint=/bin/bash` option to the `docker` command.

scottfrazer:
@mcovarr @bradtaylor @dshiga 

There are other subtleties that cause issues.  For example, I noticed that the your Dockerfile also specifies a starting directory, but this is something we cannot honor as the working directory of the process we load. Cromwell needs to have control over that.  Also, the script referenced in the entry point requires `/usr/share` to be the current working directory.  I'm not 100% sure but I'm pretty sure something in here would subtly break.

I also challenge the idea that `ENTRYPOINT` in Dockerfile is considered best practice.  I think it's context dependent.  If you're writing a service in a container, then I think `ENTRYPOINT` makes a lot of sense (a way of saying "this is how you run the service that's contained in this container").  However if your container is just an environment to run commands in it, then an `ENTRYPOINT` doesn't make a lot of sense to me.  Could you not create a container that could run both `picard.jar` style command and also other kinds of commands?  This would be pragmatic because we don't want to be downloading a 500MB Docker container for each job,  If we had a common container that could run lots of things, we could pre-cache it on new nodes.  So considering a docker container that could run other tools as well, would the `ENTRYPOINT` not be misleading?

I agree with @dshiga in that I think it's much more clear what's actually being run if the whole command is in one place instead of split between two disparate pieces.  Then there's also a clear separation between "environment" and "command to run".

geoffjentry:
We should probably move this convo off of the picard repo for now as it's more a Q about cromwell. Perhaps somewhere in the cromwell repo or in an internal mailing list (or in person)

scottfrazer:
That's probably a good idea :+1:

nh13:
Closing until a decision is made elsewhere.

PR #246, opened 2015-07-30
Fixed bug that caused BedToIntervalList to flip strand of all feature…

tfenne:
…s with a strand field. Also fixed the test case so that it correctly tests that the strand doesn't flip!


tfenne:
@nh13 Can you please review and merge this one?  It's a silly little bug fix, but it made me laugh that the unit test is actually testing the wrong behaviour!!

PR #252, opened 2015-08-06
Create TMP_DIR as owner writable instead of world writable

supernifty:
The TMP_DIR is by default created world writable which is a security issue in a shared environment. This change creates the TMP_DIR as owner writable.

### DIFF
--- a/src/java/picard/cmdline/CommandLineProgram.java
+++ b/src/java/picard/cmdline/CommandLineProgram.java
@@ -180,8 +180,8 @@ public abstract class CommandLineProgram {
             // Intentially not checking the return values, because it may be that the program does not
             // need a tmp_dir. If this fails, the problem will be discovered downstream.
             if (!f.exists()) f.mkdirs();
-            f.setReadable(true, false);
-            f.setWritable(true, false);
+            f.setReadable(true, true);
+            f.setWritable(true, true);
             System.setProperty("java.io.tmpdir", f.getAbsolutePath()); // in loop so that last one takes effect
         }
 


ktibbett:
This would cause unexpected issues when more than one user specifies the same temp directory (a not-necessarily uncommon occurrence).  Our workaround is to specify the desired permissions on the files themselves.

supernifty:
OK, a few points:
* The temp directory is TMP/USERNAME so different users shouldn't interfere with each other (?)
* Having a world writable directory means that anyone can add/remove any file from that directory, regardless of file permissions.
* The tmp dir is also not cleaned up, leaving behind a bunch of world writable directories.

yfarjoun:
The problem is (I think) that the temp directory is an input and doesn't
have to be given as TMP/USER. So different users could be using the same
TMP direrctory. Individual programs are responsible for setting the temp
file to deleteOnExit() and if they want, only user Readable/Writable.

Another (user) workaround would be to place the TMP_DIR inside another 700
directory.
Would that work for you?

I think I see where the TMP/user confusion comes from: in
htsJDK getDefaultTmpDir returns tmp/user, but in picard TMP is a user
input. This user input is used to update the java.io.tmpdir variable which
will then be used by htsjdk (together with a /user subdir).

Perhaps the desired behavior is for modify htsjdk to set the return value
of getDefaultTmpDir to user-only readable?

Yossi.

sooheelee:
A question on this topic:

Can someone please explain when one should be setting -Djava.io.tmpdir
or/and TMP_DIR?
I found setting both allowed commands to run but based on this Biostars
<https://www.biostars.org/p/42613/> forum went on to use just TMP_DIR.

I see in my /broad/hptmp/user folder that I have a number of tmp files from
the time I used both settings. Although whether the files correspond to
those commands or other I used--I don't know how to tell. They are titled
for example "-rw-r--r-- 1 shlee root  76M Aug 17 22:02
sortingcollection.4812568950038997399.tmp".

Thanks.

Soo Hee Lee, Ph.D.
Science Technical Writer

PR #254, opened 2015-08-07
fix HistogramWidth in InsertSizeMetricCollector

### LINKED ISSUE #253: InsertSizeMetricsCollector HistogramWidth width is not updated properly
`HistogramWidth` should be recalculated for each histogram if an explicit width isn't specified.

Instead it's being set by the first histogram and then never recalculated.  This means that a first histogram with a small stdev will cause the table to be trimmed to tightly.

lbergelson:
this wasn't being set properly resulting in a bug
also renamed HistogramWidth -> histogramWidth and made it final

fixes #253


eitanbanks:
@yfarjoun are you reviewing this?
The PR gets a +1 from me.

yfarjoun:
:+1:

PR #255, opened 2015-08-10
Fixed LiftoverVcf to properly handle Genotypes when the Alleles have …

eitanbanks:
…been reverse complemented.

Apparently no one ever used this tool to lift over a VCF that had genotypes in it.
This fix is something that I pretty much just copied from the GATK version of the liftover.
However, it's important to fix it here in Picard since we are planning on deprecating the
GATK version in favor of maintaining just the Picard one.

### DIFF
--- a/src/java/picard/vcf/LiftoverVcf.java
+++ b/src/java/picard/vcf/LiftoverVcf.java
@@ -13,9 +13,7 @@ import htsjdk.samtools.util.ProgressLogger;
 import htsjdk.samtools.util.SequenceUtil;
 import htsjdk.samtools.util.SortingCollection;
 import htsjdk.samtools.util.StringUtil;
-import htsjdk.variant.variantcontext.Allele;
-import htsjdk.variant.variantcontext.VariantContext;
-import htsjdk.variant.variantcontext.VariantContextBuilder;
+import htsjdk.variant.variantcontext.*;
 import htsjdk.variant.variantcontext.writer.Options;
 import htsjdk.variant.variantcontext.writer.VariantContextWriter;
 import htsjdk.variant.variantcontext.writer.VariantContextWriterBuilder;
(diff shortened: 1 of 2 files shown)


yfarjoun:
channelling keizun: can you add a test or two please?

akiezun:
exactly! tests

eitanbanks:
@yfarjoun @droazen added tests.

src/java/picard/vcf/LiftoverVcf.java
```diff
@@ -153,13 +153,17 @@ public static void main(final String[] args) {
...
             else {
                 // Fix the alleles if we went from positive to negative strand
+                reverseComplementAlleleMap.clear();
                 final List<Allele> alleles = new ArrayList<Allele>();
+
                 for (final Allele oldAllele : ctx.getAlleles()) {
                     if (target.isPositiveStrand() || oldAllele.isSymbolic()) {
                         alleles.add(oldAllele);
                     }
                     else {
```
yfarjoun: what if it's symbolic?

src/tests/java/picard/vcf/LiftoverVcfTest.java
```diff
@@ -0,0 +1,48 @@
...
+ * Test class for LiftoverVcf.
+ *
+ * Created by ebanks on 8/11/15.
+ */
+public class LiftoverVcfTest {
+
+    @Test
+    public void testFixReverseComplementedGenotypes() {
+
+        final Allele refA = Allele.create("A", true);
```
yfarjoun: please add a symbolic test case

yfarjoun: please add indel test

src/java/picard/vcf/LiftoverVcf.java
yfarjoun: This doesn't work for indels. should we emit a warning?

yfarjoun:
review done.

src/java/picard/vcf/LiftoverVcf.java
eitanbanks: @yfarjoun I don't understand these comments.
1. If it's symbolic we don't try to reverse complement.  That behavior is no different than it was before my change.
2. Why wouldn't this work for indels?  That behavior is also no different than it was before my change.
My change is simply about matching up the Alleles in the Genotypes with the Alleles in the VariantContext.  But I'm not touching the logic of the VariantContext at all.

yfarjoun:
I retract both issues, as they are indeed non-issues. :+1:

PR #257, opened 2015-08-12
Fix the liftover tool to NOT try to lift over indels when the target …

eitanbanks:
…region in the chain has been reverse complemented.

This was done for you, @yfarjoun...


tfenne:
Why can't we lift over indels when lifting to the opposite strand?  Is this just the problem with the alleles not necessarily being left-aligned any more?  If so, I'd be happy to take a shot at re-working the alleles to be left-aligned post-liftover rather than failing to lift all those indels.

eitanbanks:
I spoke to Yossi about this.  There are so many edge cases to handle (e.g. the chain block ends in the middle of a deletion) and it's unclear how to handle insertions.  If you want to take a stab then go for it!  But we've got to have tons of tests added.

For now, though, we are definitely doing the wrong thing.  So better to reject now and wait for the proper fix when it's ready.

yfarjoun:
@tfenne: I think I'll review and merge this in for now and await your PR
eagerly! OK?

tfenne:
I think, without looking at the code, that edge cases like chains ending in the middle of deletions should already be handled, and those records not lifted over.  I think the only real problem is the re-left-aligning.  FWIW merging the change as is will discard indels in ~3% of the genome for b37 -> hg38 liftover.

src/java/picard/vcf/LiftoverVcf.java
```diff
@@ -147,7 +147,8 @@ public static void main(final String[] args) {
             final Interval source = new Interval(ctx.getContig(), ctx.getStart(), ctx.getEnd(), false, ctx.getContig() + ":" + ctx.getStart() + "-" + ctx.getEnd());
             final Interval target = liftOver.liftOver(source, 1.0);
 
-            if (target == null) {
+            // if the target is null OR (the target is reverse complemented AND the variant is an indel), then we cannot lift it over
```
yfarjoun: Update comment to say "indel or mixed"

src/java/picard/vcf/LiftoverVcf.java
```diff
@@ -147,7 +147,8 @@ public static void main(final String[] args) {
...
+            if (target == null || (target.isNegativeStrand() && (ctx.isMixed() || ctx.isIndel()))) {
                 rejects.add(new VariantContextBuilder(ctx).filter(FILTER_CANNOT_LIFTOVER).make());
```
yfarjoun: I think it would be good to have a better description of why we cannot liftover this variant, to separate it from the other reasons.

yfarjoun:
minor comments. 

:+1: please rebase onto master and merge. 

Please open an issue describing the left align problem, and also the fact that mixed variants will have to be split into two variants and nearby variants might need to be merged into one. 

ping @tfenne to see if he is still interested in giving it a shot.

PR #259, opened 2015-08-13
Tool to collect summary metrics from query-name sorted reads.

eitanbanks:
This is my first ever attempt at writing a picard tool, so I'd love some constructive feedback.  Also, I can close this PR if you don't think it belongs in the public picard repo.

Caveat: this is for work being done on my own time, so please treat accordingly.


tfenne:
Looks very in keeping with the style of the Picard tools.  A few suggestions:
1) Most of the Picard metrics collection tools are named for the kinds of metrics they collect, less so how they do it.  So perhaps CollectWgsMetricsFromQuerySorted or CollectWgsRapidQcMetrics?

2) The SeqMetrics class name is also very generic.  Actually I wonder if it wouldn't be better to simply use the metrics class from CollectWgsMetrics and leave the uncalculated fields blank?  That way it's clear they are calculating the same things different ways?

3) I would probably be tempted to change `getNextReadPair()` and `getNextGoodRead()` by wrapping the Iterator in a `PeekableIterator` so that you can scan forward in the iterator without having to cache state outside the method.

src/java/picard/analysis/CollectNameSortedMetrics.java
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /**
+     * Get the count of overlapping bases for the given reads
+     *
+     * @param read1  the 1st read
+     * @param read2  the 2nd read
+     * @return non-negative int
+     */
+    private int getOverlappingBaseCount(final SAMRecord read1, final SAMRecord read2) {
```
tfenne: I think technically this method should examine the bases in the overlap and ensure that in both R1 and R2 the bases in question are above the low quality base threshold.  Otherwise you'll take a double-hit on those bases.

yfarjoun:
@tfenne: can you complete the review? a truly volunteered commit and a volunteered review?

tfenne:
@yfarjoun: what does it mean to "complete" the review?

yfarjoun:
to be the one who :+1: it eventually :-)

eitanbanks:
@tfenne no pressure to "complete the review" but I've incorporated all of your great suggestions.

eitanbanks:
I just added another commit to error out if the reads are not sorted in queryname order and to include reads with unmapped mates as excluded.

yfarjoun:
other than the soft clipped bases (which I would ask that you document since the current behavior is different than CollectWGSMetrics) :+1:

eitanbanks:
@yfarjoun actually I realized that there was a bug in the previous implementation as I wasn't taking into account the softclips for overhanging bases (e.g. the first base of read2 is not part of the overhang if it is softclipped).  I've fixed that as part of the overall accounting for softclips (which I count under "low quality bases", since there's no other better exclusion for it).  I've kept these changes as a separate commit for now; could you please re-review?
Also, how would you like me to document the difference in softclips (which you mentioned in your previous comment)?
Thanks!

src/java/picard/analysis/CollectWgsMetricsFromQuerySorted.java
```diff
@@ -0,0 +1,300 @@
...
+
+        // finalize metrics
+        metrics.GENOME_TERRITORY = reader.getFileHeader().getSequenceDictionary().getReferenceLength();
+        metrics.MEAN_COVERAGE = metrics.TOTAL_USABLE_BASES / (double)metrics.GENOME_TERRITORY;
+        metrics.PCT_EXC_DUPE = basesExcludedByDupes / (double)metrics.TOTAL_BASES;
+        metrics.PCT_EXC_MAPQ = basesExcludedByMapq / (double)metrics.TOTAL_BASES;
+        metrics.PCT_EXC_UNPAIRED = basesExcludedByPairing / (double)metrics.TOTAL_BASES;
+        metrics.PCT_EXC_BASEQ = basesExcludedByBaseq / (double)metrics.TOTAL_BASES;
+        metrics.PCT_EXC_OVERLAP = basesExcludedByOverlap / (double)metrics.TOTAL_BASES;
+        final double totalExcludedBases = basesExcludedByDupes + basesExcludedByMapq + basesExcludedByPairing + basesExcludedByBaseq + basesExcludedByOverlap;
```
yfarjoun: Isn't this also TOTAL_BASES - TOTAL_USABLE_BASES ? perhaps a bit clearer as one doesn't need to check you remembered all the exclusions.

src/java/picard/analysis/CollectWgsMetricsFromQuerySorted.java
```diff
@@ -0,0 +1,300 @@
...
+        public int lowBQcount = 0;
+
+        public BaseExclusionHelper(final SAMRecord read) {
+            this.read = read;
+
+            final List<CigarElement> cigarElements = read.getCigar().getCigarElements();
+            for (final CigarElement element : cigarElements) {
+                final CigarOperator op = element.getOperator();
+
+                if (op == CigarOperator.SOFT_CLIP) {
```
yfarjoun: This will not work when the read is hardclipped at the end. I think you need to switch on the op and break in case of a non-clip operator. also break after a softclip.

I'd add a test case or two 
1H1S1M -> 1
1S1M1S1H -> 1 // This currently fails, I think
1M1S -> 0 // this also currently fails, I think.

The lower part seems to have similar issues, in reverse.

src/java/picard/analysis/CollectWgsMetricsFromQuerySorted.java
```diff
@@ -0,0 +1,300 @@
...
+        }
+
+        // finalize metrics
+        metrics.GENOME_TERRITORY = reader.getFileHeader().getSequenceDictionary().getReferenceLength();
+        metrics.MEAN_COVERAGE = metrics.TOTAL_USABLE_BASES / (double)metrics.GENOME_TERRITORY;
+        metrics.PCT_EXC_DUPE = basesExcludedByDupes / (double)metrics.TOTAL_BASES;
+        metrics.PCT_EXC_MAPQ = basesExcludedByMapq / (double)metrics.TOTAL_BASES;
+        metrics.PCT_EXC_UNPAIRED = basesExcludedByPairing / (double)metrics.TOTAL_BASES;
+        metrics.PCT_EXC_BASEQ = basesExcludedByBaseq / (double)metrics.TOTAL_BASES;
+        metrics.PCT_EXC_OVERLAP = basesExcludedByOverlap / (double)metrics.TOTAL_BASES;
```
yfarjoun: You should probably give an invalid value (-1? NaN? ) to the fields you are not filling since otherwise they will have the valid value of zero, and it could be confusing.

src/java/picard/analysis/CollectWgsMetricsFromQuerySorted.java
```diff
@@ -0,0 +1,300 @@
...
+            return 0;
+
+        final byte[] read1quals = read1exclusions.read.getBaseQualities();
+        final byte[] read2quals = read2exclusions.read.getBaseQualities();
+        final int overlap = read1exclusions.read.getAlignmentEnd() - read2exclusions.read.getAlignmentStart() + 1;
+        int numHighQualityOverlappingBases = 0;
+
+        for (int i = 0; i < overlap; i++) {
+            // we only want to count it if they are both high quality (i.e. not already counted among bad bases)
+            if (read1quals[read1exclusions.firstTrailingClippedBaseIndex - overlap + i] >= MINIMUM_BASE_QUALITY && read2quals[read2exclusions.firstUnclippedBaseIndex + i] >= MINIMUM_BASE_QUALITY) {
```
yfarjoun: I am having a hard time understanding why this is comparing bases that are over the same genomic position....

yfarjoun:
back to you @eitanbanks

src/java/picard/analysis/CollectWgsMetricsFromQuerySorted.java
eitanbanks: No, this will work just fine with hard clips.  The logic in the loop is:
    if it's a softclip
      do work
    else if it's not a hardclip
      end
So if there is a hardclip at the end it will continue to iterate through the loop...

eitanbanks:
Other comments other than the hard-clips (which I believe is a non-issue) addressed (and rebased).
I can add some tests later.

eitanbanks:
@yfarjoun I've added a test that generally shows the tool working and in particular that it handles H operators in overlaps just fine.

src/java/picard/analysis/CollectWgsMetricsFromQuerySorted.java
```diff
@@ -0,0 +1,327 @@
...
+        }
+
+        return readPair;
+    }
+
+    /**
+     * Return the next usable read in the iterator
+     *
+     * @param iterator the iterator to pull from
+     * @param justPeek if true, just peek the next read rather than pulling it
```
yfarjoun: just peek the next _usable_ read (will remove unusable reads from the iterator)

yfarjoun:
tiny change to the @param comment. 

after that :+1:  and thanks for a nice tool!

eitanbanks:
@yfarjoun I changed the doc as you requested and rebased that in.
Also, as per our in-person discussion, I've added another commit that addresses the problem of genomic vs. read positions.  For now I've just copied your method from hts-jdk with a note to remove it once yours is merged.
Please advise on how to proceed with this PR.

yfarjoun:
any idea on the failed checks?

yfarjoun:
test is failing: 

picard.analysis.CollectWgsMetricsFromQuerySortedTest.testMetricsFromClippedOverhangs(CollectWgsMetricsFromQuerySortedTest.java:42)

eitanbanks:
Sorry about that @yfarjoun - the test was capturing the behavior that I had just changed.
Fixed and rebased.

yfarjoun:
:+1:

PR #260, opened 2015-08-17
Added capability to CollectWgsMetrics to restrict to certain intervals.

eitanbanks:
@yfarjoun it turns out that this feature is incredibly useful for rapid QC because we're going to end up with whole genome bams that have been sampled down to a small number of sites and (critically) we only want them assessed at those exact positions (i.e. we do not want to include the rest of the bases in the reads).


akiezun:
needs tests

nh13:
@akiezun what needs testing here?  Presumbably ```SamLocusIterator``` is tested.  Perhaps you mean "while you are at it, could you add regression tests since we are missing those"?  Seems like a big burden for a fairly understandable and trivial change.

akiezun:
Presumably this code implements some feature. That feature needs to be tested to demonstrate that it works properly.

eitanbanks:
@akiezun done.
Please re-review.

akiezun:
:+1:

PR #262, opened 2015-08-19
fixing histogram count bug in mark illumina adapters

nh13:


nh13:
The histogram counts were being updated based on the first record of a pair, which is problematic when the second record is clipped at a different offset/index.

akiezun:
i have 1 comment: this needs a regression test

nh13:
I don't see any existing regression tests for "MarkIlluminaAdapters" to which to update; also is a bug fix not a new feature.  Thanks for reviewing.

akiezun:
bug fixes need regression tests

nh13:
closing as there are no tests.

yfarjoun:
hey boys. can we please stop this.

yfarjoun:
@nh13 Thank you very much for spotting this bug, and suggesting a fix. Would you be able to supply two reads that cause an error so that **we** could write a test for this? That would be very much appreciated!

nh13:
Moving this to https://github.com/broadinstitute/picard/pull/341.

PR #266, opened 2015-08-21
Separated CollectWgsMetrics into 2 separate tools:

eitanbanks:
1. Maintain previous behavior by reverting my last change to allow for analysis over intervals.
2. Create a new type of collector that requires intervals and expects data to come in as sampled sites.

The motivation here is that this allows us to cleanly compute filter metrics differently (the whole genome
version wants it by base vs. the site sampled one which wants it by read) instead of mucking up the code too much.


yfarjoun:
There's a similar inherited class called CollectRawWgsMetrics. There we "rename" the metrics collection class (with no modification) by overriding generateWgsMetrics() with a derived class. The effect is that when the metrics file is written to disk the header clearly states a different class which might make it easier to see/realize that it was generated with a different program. Consider if you'd like doing that here as well.

src/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -219,9 +209,9 @@ protected int doWork() {
         metrics.MEDIAN_COVERAGE = histo.getMedian();
         metrics.MAD_COVERAGE = histo.getMedianAbsoluteDeviation();
 
-        final long basesExcludedByDupes = dupeFilter.getFilteredBases();
-        final long basesExcludedByMapq = mapqFilter.getFilteredBases();
-        final long basesExcludedByPairing = pairFilter.getFilteredBases();
+        final long basesExcludedByDupes = getBasesExcludedBy(dupeFilter);
```
yfarjoun: Why the extra function call? what's it good for?

yfarjoun: nevermind. got it.

src/java/picard/analysis/CollectWgsMetricsFromSampledSites.java
```diff
@@ -0,0 +1,44 @@
...
+    @Option(shortName = "INTERVALS", doc = "An interval list file that contains the locations of the positions to assess.", optional = false)
+    public File INTERVALS = null;
+
+
+    public static void main(final String[] args) {
+        new CollectWgsMetricsFromSampledSites().instanceMainWithExit(args);
+    }
+
+    @Override
+    protected SamLocusIterator getLocusIterator(final SamReader in) {
```
yfarjoun: This needs a comment explaining why the number of records equals the number of bases.

src/java/picard/analysis/CollectWgsMetricsFromSampledSites.java
```diff
@@ -0,0 +1,44 @@
...
+import htsjdk.samtools.util.*;
+import picard.cmdline.CommandLineProgramProperties;
+import picard.cmdline.Option;
+import picard.cmdline.programgroups.Metrics;
+
+import java.io.File;
+
+/**
+ * Computes a number of metrics that are useful for evaluating coverage and performance of whole genome sequencing experiments,
+ * but only at a set of sampled positions.
```
yfarjoun: It assumes that the positions are samples in such a way so that every record may have *at most* one sampled position. this needs to be made clear if this is to be useful for others.

src/java/picard/analysis/CollectWgsMetricsFromSampledSites.java
```diff
@@ -0,0 +1,44 @@
...
+public class CollectWgsMetricsFromSampledSites extends CollectWgsMetrics {
+
+    @Option(shortName = "INTERVALS", doc = "An interval list file that contains the locations of the positions to assess.", optional = false)
+    public File INTERVALS = null;
+
+
+    public static void main(final String[] args) {
+        new CollectWgsMetricsFromSampledSites().instanceMainWithExit(args);
+    }
+
```
yfarjoun: Should you validateinputs and check that the input intervalList indeed only has 1-sized intervals in it?

src/java/picard/analysis/CollectWgsMetricsFromSampledSites.java
```diff
@@ -0,0 +1,44 @@
...
+                "whole genome sequencing experiments, but only at a set of sampled positions.",
+        usageShort = "Writes whole genome sequencing-related metrics for a SAM or BAM file",
+        programGroup = Metrics.class
+)
+public class CollectWgsMetricsFromSampledSites extends CollectWgsMetrics {
+
+    @Option(shortName = "INTERVALS", doc = "An interval list file that contains the locations of the positions to assess.", optional = false)
+    public File INTERVALS = null;
+
+
```
yfarjoun: spurious empty line

yfarjoun:
A few comments for you @eitanbanks

src/java/picard/analysis/CollectWgsMetricsFromSampledSites.java
eitanbanks: I don't think that check is helpful.  What if you sample every other base?  That's still not desired behavior but it would pass this check.
I'll add very clear docs up top.

eitanbanks:
@yfarjoun I've just pushed in another commit (not rebased) that addresses all of your reviewer comments.
I've also added a proper test that confirms that the right metrics are produced when sampled correctly, plus modified the existing test to check some metrics on contiguous intervals.

eitanbanks:
Done and rebased.  Will merge after I am sure tests pass.

PR #267, opened 2015-08-23
Added new metrics to the query-sorted wgs collector.

eitanbanks:
1. We now capture total read pairs and total duplicate read pairs.
These are needed so that we can estimate library size downstream in the process.

2. We now capture mean insert size.


ktibbett:
Looks good.

PR #269, opened 2015-08-25
Removed temporary method in CollectWgsMetricsFromQuerySorted

eitanbanks:
No longer necessary now that the function was added to htsjdk.
Also, renamed a variable as requested by Jay.


yfarjoun:
:+1:

PR #271, opened 2015-08-27
Fixed an interaction between optional and minElements.

cjllanwarne:


cjllanwarne:
Closed because lots of tests don't specify arguments which would (should?) now be required.

Suggest someone decides how optional and minElements should be interacting then maybe this change  can be reintroduced (alongside the required test updates).

lbergelson:
We dealt with this same issue in hellbender since we use a modified picard arg parser.

PR #273, opened 2015-08-28
docker container as executable

chrishah:
Pull request as discussed with Sheila on the GATK mailing list.

Made small changes to Dockerfile and docker_helper.sh to enable the usage of the Picard docker container as an executable. With the image built from this context (`docker build -t custompicard .`) I can run, e.g.:
`docker run -i -t -v $(pwd):/usr/working custompicard MergeSamFiles I=first.bam I=second.bam O=merged.bam` 
successfully if the bam files are present in the current working directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,4 +28,7 @@ RUN ant clean all && \
     rm -rf lib && \
     rm build.xml
 
-ENTRYPOINT ["./docker_helper.sh"]
+RUN mkdir /usr/working
+WORKDIR /usr/working
+
+ENTRYPOINT ["/usr/picard/docker_helper.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,3 +32,4 @@ RUN mkdir /usr/working
 WORKDIR /usr/working
 
 ENTRYPOINT ["/usr/picard/docker_helper.sh"]
+CMD [""]
--- a/src/scripts/picard/docker_helper.sh
+++ b/src/scripts/picard/docker_helper.sh
@@ -24,4 +24,4 @@ done
 shift $(expr $OPTIND - 1)
 TOOL_WITH_ARGS=$@
 
-java ${JVM_ARGS} -jar picard.jar ${TOOL_WITH_ARGS}
+java ${JVM_ARGS} -jar /usr/picard/picard.jar ${TOOL_WITH_ARGS}


Dockerfile
```diff
@@ -28,4 +28,8 @@ RUN ant clean all && \
     rm -rf lib && \
     rm build.xml
 
-ENTRYPOINT ["./docker_helper.sh"]
\ No newline at end of file
+RUN mkdir /usr/working
+WORKDIR /usr/working
```
bradtaylor: This was a good catch!

Dockerfile
```diff
@@ -28,4 +28,8 @@ RUN ant clean all && \
     rm -rf lib && \
     rm build.xml
 
-ENTRYPOINT ["./docker_helper.sh"]
\ No newline at end of file
+RUN mkdir /usr/working
```
bradtaylor: Should this mkdir be moved to a previous run command in order to avoid an additional layer, and for more efficient caching?

bradtaylor:
Thank you for this PR!  And for your interest in Docker, which is new to picard. A couple little comments/questions, otherwise it looks great.

One thing I do notice is that, looking at the example docker run command you supplied, I don't believe the -i and -t flags are necessary. My understanding is they are used for interactive sessions, as when running /bin/bash. For example, I was able to successfully execute the following command:
```docker run -v $(pwd):/usr/working custompicard SamToFastq I=ubams_testing/newnewnew.bam F=test.fastq F2=test2.fastq```

One other thing to note:
The ENTRYPOINT command does not play nice with some systems we are building internally. So we have an agenda item to remove it. At that point, you would specify the full java invocation as an argument to the container run (the helper script would go away). In place of the entrypoint command will probably be a CMD simply echoing a usage message with instructions on composing an appropriate run command. This will likely happen soon. In the interim, I will merge this PR once I here back from you regarding the questions I mentioned.

We here at Broad really appreciate your interest in Picard. Thanks for your contribution!

chrishah:
Hi Brad,
Glad I could contribute something useful!
I agree with you, the `-i` ant `-t` flags in my command suggestion are not actually necessary. 
The `CMD [""]` is a solution for the behaviour discussed a bit in [this](https://github.com/docker/docker/issues/3753) Docker issue - without it I got an error message "/bin/bash is not a valid command" when running the container without any arguments. I hadn't thought about efficient caching at all, to be honest.
Sure, I've read elsewhere that `ENTRYPOINT` can be problematic in some cases. Looking forward to seeing your solution then! 
Thanks!
Christoph

bradtaylor:
Only now seeing Christoph's response. Sorry for the delay!  Going to merge as-is.

PR #276, opened 2015-09-01
Clear the AS (alignment score) tag when REMOVE_ALIGNMENT_INFORMATION …

ktibbett:
…is true.

### DIFF
--- a/src/java/picard/sam/RevertSam.java
+++ b/src/java/picard/sam/RevertSam.java
@@ -103,8 +103,9 @@ public class RevertSam extends CommandLineProgram {
         add(SAMTag.MQ.name());
         add(SAMTag.SA.name()); // Supplementary alignment metadata
         add(SAMTag.MC.name());      // Mate Cigar
+        add(SAMTag.AS.name());
     }};
-
+    
     @Option(doc = "WARNING: This option is potentially destructive. If enabled will discard reads in order to produce " +
             "a consistent output BAM. Reads discarded include (but are not limited to) paired reads with missing " +
             "mates, duplicated records, records with mismatches in length of bases and qualities. This option can " +


src/java/picard/sam/RevertSam.java
```diff
@@ -374,6 +374,7 @@ public void revertSamRecord(final SAMRecord rec) {
             rec.setAlignmentStart(SAMRecord.NO_ALIGNMENT_START);
             rec.setCigarString(SAMRecord.NO_ALIGNMENT_CIGAR);
             rec.setMappingQuality(SAMRecord.NO_MAPPING_QUALITY);
+            rec.setAttribute(SAMTag.AS.name(), null);
```
nh13: What about other alignment based tags, such as ```MD```, ```MC```, or ```NM```?   Why not just add ```AS``` to ```ATTRIBUTE_TO_CLEAR```?

ktibbett:
@nh13 -- thanks.  Ok now?

sooheelee:
FWI, here are all the tags BWA uses:

NM, MD, AS, BC, X0, X1, XN, XM, XO, XG, XT, XA, XS, XF, XE

Soo Hee Lee, Ph.D.
Science Technical Writer

sooheelee:
So better yet, could we get an ALL option that clears all queryname-associated tags?

nh13:
@ktibbett looks great. @sooheelee I would recommend not adding all the ```X?``` tags to the default list as those are not standardized across aligners and tools.  The user can use the ```ATTRIBUTE_TO_CLEAR ``` command line option to add those.

ktibbett:
@sooheelee , BC is not alignment-releated, and the other tags are custom to the tool vendor.  I agree with @nh13  that the ATTRIBUTE_TO_CLEAR argument gives users sufficient control over what other tags they do and do not retain.

sooheelee:
@nh13 , I will disagree. I think in the least we should provide an option to clear all tags that will subsequently be produced by our GATK variant discovery workflow. This way users can start with a clean file without having to learn what I spent a considerable amount of time on yesterday, that is the command to comprehensively list all tags present in a BAM file. It turns out the BAM given to me by aggregation also contained OC and OP tags that were then misapplied back to the BAM during the MergeBamAlignment step. I am assuming here that the OC and OP tags are applied to a BAM during one of our processing steps downstream of where I am.

nh13:
Picard is a community set of tools, and not just for Broad's Picard+GATK variant discovery workflow.  For tags introduced by that latter workflow, perhaps the Broad can document those in the GATK forums and list the tags needed to clean the BAM.

sooheelee:
@nh13 I'll keep that in mind for our documentation. As for my comment, I am doing my bit as a voice of dissension on behalf of users.

nh13:
@sooheelee I completely understand and keep it up!

PR #278, opened 2015-09-04
Have FilterVcf use the input VCF's SequenceDictionary if output is to

gbggrant:
… an uncompressed VCF or BCF.


ktibbett:
Looks good, @gbggrant !

PR #280, opened 2015-09-08
Fix a bug in path handling

vdanjean:
This patch fixes the convertion from URL to path so that paths containing the '+' character are not converted to ' '.
Without it, picard.jar can't be placed in a path where one directory has a '+' in its name (and that happens on Debian builders)


chandrans:
@yfarjoun Hey Yossi. The user actually made a fix for this. Can you review it? Thanks.

ktibbett:
@chandrans , I have already assigned this to @gbggrant  for review.

chandrans:
Thanks Kathleen. @ktibbett

gbggrant:
:+1:

PR #282, opened 2015-09-10
- added a test that fails if any CLP is missing its @CommandLineProgr…

yfarjoun:
…amProperties annotation.


ktibbett:
Looks good!

PR #291, opened 2015-09-21
Fix unit tests that fail as a result of HTSJDK base comparison changes

mattsooknah:
Various unit test fixes given changes to base comparison method in HTSJDK.

Do not merge until after https://github.com/samtools/htsjdk/pull/325 is merged!


src/java/picard/util/ClippingUtility.java
```diff
@@ -264,12 +265,10 @@ public static int findIndexOfClipSequence(final byte[] read, final byte[] adapte
             int mismatches = 0;
 
             for (int i = 0; i < length; ++i) {
-                if (!SequenceUtil.isNoCall(adapterSequence[i]) && !SequenceUtil.basesEqual(adapterSequence[i], read[start + i])) {
+                if (!SequenceUtil.basesEqual(adapterSequence[i], read[start + i])) {
```
tfenne: This is going to do the wrong thing in a couple of different cases:
1) Where one read is fine and the other has lots of N's at the end, this will cause IlluminaBasecallsToSam and ClipIlluminaAdapters to incorrectly trim both reads.
2) If a read has adapter sequence in it, but then goes to all Ns at the end, it will trim at the wrong position.  E.g. a read that has 60 good bases, then 60 bases of adapter, then 30 Ns, will get trimmed around base 138 instead of base 60.

tfenne:
@gbggrant @mattsooknah I'm not sure what the motivations behind the IUPAC handling changes are, but I think there needs to be a more careful review of the consequences.  I've flagged one problem above where treating N as matching anything is going to cause incorrect behaviour.

In addition, I think it is correct to count N in a read as mismatching any base in the reference - that's how AlignmentSummaryMetrics has always worked - i.e. the metrics are worse when the reads are full of no-calls.  With this change, going from low-quality called bases (A,C,G,T) to Ns will actually make all the metrics look better.  I don't think that's desirable.

Lastly, I think this also effects how various SAM tags are calculated.  E.g. NM is defined in the spec as "Edit distance to the reference, including ambiguous bases but excluding clipping", but this change will break that in HTSJDK.

eitanbanks:
Hey @tfenne, speaking for Kathleen (who is out) the motivation here was just for the NM tag I believe.  If you look at the IUPAC ambiguity codes, N does in fact *match* (A,C,G,T) so we've been incorrectly calculating NM forever.

Maybe the right thing to do then would be to have 2 implementations: the classic one where N's mismatch anything but another N, and a IUPAC supporting version where they match everything.  And then only the code to calculate NM would use the IUPAC version.

Thoughts?

tfenne:
@eitanbanks Two things - firstly, yes I think the best option is to introduce the new functionality as a `basesEqualsWithAmbiguity` or something like that and then selectively change places over in the code where that behaviour is desirable.

Secondly though, I think the current implementation of the NM calculation in htsjdk is spec compliant already.  At least, it matches the way I read the spec.  And it also matches the way that samtools calculates the NM tag (just took a perfect read, introduced some Ns and then re-ran samtools calmd to confirm).

eitanbanks:
Glad we agree on point 1!

For point 2, I don't think that the current implementation is correct.  As you quoted, NM = "Edit distance to the reference, including ambiguous bases..."  But the bases that should be matching the ambiguous IUPAC reference bases are counted as mismatches (e.g. an A in the read counts as mismatching an M or W in the reference).  Looking at the ambiguity encoding, since N does match other bases then according to the strict definition of NM it should not be counted as mismatching either.  Just because samtools is wrong, it doesn't mean we should be.  :)
(Then again, I also don't personally care about the NM tag.  I just like to argue!)

tfenne:
@eitanbanks Ah, I see what you're saying.  What you're saying (at least, I think so) is that non-ambiguous bases in reads that are compatible with an ambiguous base in the reference should not count as mismatches.  And what I'm saying is that Ns in reads should continue to be mismatches vs. less ambiguous bases in the reference.  The two are compatible.

But what the change does causes ambiguous bases in either (ref or read) to count as matches instead of mismatches.

eitanbanks:
Right, well phrased.  So what we really want here (for the IUPAC-adherent version only) is a directionality to the calculation (i.e. to know which is the read and which is the reference).  Unless it's possible to have ambiguous bases (e.g. W) in the reads?  @tfenne have you ever seen anything like that?

tfenne:
Totally agree about the one-way acceptance of ambiguity codes @eitanbanks.  I've personally never seen any ambiguity codes other than `N`s in reads.  BAM does use a four bit encoding to allow for ambiguity codes, but I've never seen then used.

mattsooknah:
@tfenne @eitanbanks sounds good, I will change the implementation accordingly.

Would it suffice to replace the current code:
```java
(bases[lhs] & bases[rhs]) > 0
```
with the following:
```java
(bases[readBase] & bases[refBase]) == bases[readBase]
```

I *think* that should cover all cases, including ambiguous read bases  (assuming I didn't mess up the bitwise arithmetic).

eitanbanks:
@tfenne are you cool being the official reviewer of this PR?  Will assign to you; assign to me if you don't want it.

eitanbanks:
(@tfenne, I can't assign you formally, but you're it unless I hear otherwise)

tfenne:
@eitanbanks No worries - happy to review.

tfenne:
@mattsooknah That sounds right to me.  I'd still suggest the plan of creating a separate method for the weaker comparison and then selectively using it.

mattsooknah:
closing this, since the revised HTSJDK code (https://github.com/samtools/htsjdk/pull/328) does not affect Picard unit tests

Issue #289, opened 2015-09-18
CollectGcBiasMetrics requires a lot more memory


nh13:
Running the current version of CollectGcBiasMetrics requires a lot more memory to complete than the pre-SinglePassProgram version.  Old version is 46330cbbc330ce0925849a2aa2574d99d0cba727.  This is likely due to storing all the gc bias windows for every chromosome, instead of one at a time:  https://github.com/broadinstitute/picard/blob/da43b4f03cdc911f0708a9522516231da874bb86/src/java/picard/analysis/CollectGcBiasMetrics.java#L138

This is a big deal to use so much more memory, could you please fix this?

kbergin:
Oh hey nils! I will look into it. I think there was a reason I had to put that calculation in setup all at once to be able to work with the multi level collector though..

yfarjoun:
There are only 100 windows and about 100 references, so 10K...I doubt that
is the problem. more likely is that the loading of the whole reference at
setup time requires too much memory....would be nice if there were a way to
garbageCollect the unneeded references...

kbergin:
Wouldn't most scenarios need all of the references though anyways? It seemed like it was mostly the smaller test files that didn't use all of them. Would it help to 'close' the ref file after it's read through?

kbergin:
I think the pre-SPS version also loaded the whole reference

yfarjoun:
Not  sure if it will help this issue, but it's probably a good idea.

yfarjoun:
It went through the entire thing, but only as the reads came by, not at the
start. So perhaps there was more time to do garbage collection...

kbergin:
Oh that makes sense. hmm

tfenne:
@kbergin If you look at the old implementation it only ever kept everything in memory for a single chromosome at a time.  You can see the old version [here](https://github.com/broadinstitute/picard/blob/705f349fde4aa9a6ab4a0fcfedba102dcc080e6a/src/java/picard/analysis/CollectGcBiasMetrics.java).  It iterated over the sequences in the references, and then used a PeekableIterator to iterate over the reads, and roll onto the next chromosome when it needed to.

@yfarjoun Unless I've read it wrong, the `gcByRef` map contains an entry for each chromosome, and the byte array stored is the same length as the chromosome - holding the calculated GC for each window on the chromosome.

kbergin:
You're right that it holds the gc for each window on the chromosome. I need to look more into how to avoid doing that with the multi level collector.

[...]

PR #295, opened 2015-09-25
Kb gc bias memory fix

kbergin:
Updates to CollectGcBias to use less memory. 

On tests with current picard CollectGcBias and 1G memory (less than size of human reference) the program fails due to java.lang.OutOfMemoryError: Java heap space while trying to create the large map of gc x ref. With this new version CollectGcBias is able to complete without issue at -Xmx1G. 

GcBiasUtils created to extract out calculating GC content and store it as a state because this is now used once in CollectGcBiasMetrics to calculate the 'windowsByGc' array (size=100 bins) and then again in GcBiasMetricsCollector to calculate the gc[] for each reference contig when it is needed. The GC content of one contig is stored at a time to prevent recalculating it every time a new record comes in.

windowsByGc must be calculated separately from single pass sam passing in records because it must include windows across all reference contigs to get accurate GC bias metrics calculations, regardless of whether or not there is a record in the input sam file that corresponds to a reference contig. 

New test added to test that there is not a bug for small sam files where reference sequences are skipped.


kbergin:
PR in response to #289.

src/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -172,64 +153,40 @@ protected void finish() {
 
         final NumberFormat fmt = NumberFormat.getIntegerInstance();
         fmt.setGroupingUsed(true);
-        RExecutor.executeFromClasspath(R_SCRIPT,
+        /*RExecutor.executeFromClasspath(R_SCRIPT,
                 OUTPUT.getAbsolutePath(),
                 SUMMARY_OUTPUT.getAbsolutePath(),
                 CHART_OUTPUT.getAbsolutePath(),
-                String.valueOf(WINDOW_SIZE));
+                String.valueOf(WINDOW_SIZE));*/
```
mattsooknah: Did you mean to un-comment this?

kbergin: AGH yes sorry. The R script doesn't work with my intelliJ tests for some reason but works in tests when it's actually in github. I'll fix it now.

src/java/picard/analysis/GcBiasUtils.java
```diff
@@ -0,0 +1,44 @@
...
+                if (base == 'G' || base == 'C') ++state.gcCount;
+                else if (base == 'N') ++state.nCount;
+            }
+        } else {
+            final byte newBase = bases[endIndex - 1];
+            if (newBase == 'G' || newBase == 'C') ++state.gcCount;
+            else if (newBase == 'N') ++state.nCount;
+
+            if (state.priorBase == 'G' || state.priorBase == 'C') --state.gcCount;
+            else if (state.priorBase == 'N') --state.nCount;
```
mattsooknah: This is sort of tangential, but could this method be modified to use ```SequenceUtil.basesEqual``` instead of direct ```==``` ? It would be cleaner, and you wouldn't have to call ```StringUtil.toUpperCase(refBases)``` elsewhere (which might speed things up, though perhaps not enough to be noticeable).

mattsooknah:
I keep getting confused between the different "window size" variables ... could they be renamed to more easily differentiate between the # of histogram bins and the size of the scanning window?

mattsooknah:
Also, perhaps ```CollectGcBiasMetrics.calculateRefWindowsByGc``` and ```GcBiasMetricsCollector.calculateAllGcs``` could be moved into ```GcBiasUtils``` ? Then all the GC helper methods would be in one place, and the other two classes would be cleaner.

kbergin:
@mattsooknah I've responded to your review comments if you could take another look!

src/java/picard/analysis/GcBiasUtils.java
```diff
@@ -0,0 +1,98 @@
+package picard.analysis;
```
nh13: license

kbergin:
Thanks @nh13 added the license to two files that were missing it. @mattsooknah should be ready for you again. Thanks!

src/tests/java/picard/analysis/CollectGcBiasMetricsTest.java
```diff
@@ -176,36 +124,193 @@ public void runTest(final File input) throws IOException {
...
+        Assert.assertEquals(runPicardCommandLine(args), 0);
+    }
+
+    /////////////////////////////////////////////////////////////////////////////
+    //Used to generate the Sam Record Sets with SamRecordSetBuilder.addPair().
+    //testNumber 1: runGcBiasMultiLevelTest, generates records aligning to chrM and chrO
+    //testNumber 2: runWindowsComparisonTest, generates records aligning to chrM,N,O.
+    /////////////////////////////////////////////////////////////////////////////
+    public void setup(final int ID, final String readGroupId, final SAMReadGroupRecord readGroupRecord, final String sample,
+                      final String library, final SAMFileHeader header, final SAMRecordSetBuilder setBuilder, final int testNumber)
```
mattsooknah: Instead of passing in ```testNumber```, I would just make a separate setup function for each test.

src/tests/java/picard/analysis/CollectGcBiasMetricsTest.java
```diff
@@ -176,36 +124,193 @@ public void runTest(final File input) throws IOException {
...
+
+        //add records that align to all 3 chr in reference file
+        else if(testNumber==2){
+            for (int i = 0; i < NUM_READS; i++) {
+                final int start = rg.nextInt(max) + min;
+                final String newReadName = READ_NAME + separator + ID + separator + i;
+
+                if (i<=NUM_READS/3) {
+                    setBuilder.addPair(newReadName, contig1, start + ID, start + ID + LENGTH);
+                } else if (i>NUM_READS/3 && i< (NUM_READS - (NUM_READS/3))) {
```
mattsooknah: The ```i>NUM_READS/3``` check is not needed here, since it's already ruled out by the previous clause.

src/tests/java/picard/analysis/CollectGcBiasMetricsTest.java
kbergin: Thanks for the suggestion, I wasn't feeling comfortable with passing in the testNumber but wasn't sure the ideal way to make it better! Fixed in most recent commit.

src/tests/java/picard/analysis/CollectGcBiasMetricsTest.java
```diff
@@ -38,136 +35,87 @@
...
+    private final File dict = new File(TEST_DIR, "MNOheader.dict");
 
-    File tempSamFile;
+    File tempSamFileChrM_O;
+    File tempSamFileAllChr;
 
     SAMRecordSetBuilder setBuilder1 = new SAMRecordSetBuilder(true, SAMFileHeader.SortOrder.coordinate);
     SAMRecordSetBuilder setBuilder2 = new SAMRecordSetBuilder(true, SAMFileHeader.SortOrder.coordinate);
     SAMRecordSetBuilder setBuilder3 = new SAMRecordSetBuilder(true, SAMFileHeader.SortOrder.coordinate);
+    SAMRecordSetBuilder setBuilder4 = new SAMRecordSetBuilder(true, SAMFileHeader.SortOrder.coordinate);
```
mattsooknah: Doesn't ```SAMRecordSetBuilder``` fill in the bases randomly? Will the test results still be deterministic despite this?

kbergin: the setBuilder.addPair takes in a contig number, and a start and end location for each read pair to the reference and the start and end location is determined from a 'random' integer but it always uses the same ones because of `final Random rg = new Random(5);` in setup below.

mattsooknah:
Ok, looks good to me. Just remember to squash and rebase.

kbergin:
Merging! Thanks!

PR #296, opened 2015-09-29
GenotypeConcordance could overflow if there were too many sites for an

nh13:
integer, which could happen for true negative sites for whole genomes
when using MISSING_SITES_HOM_REF=true.

### DIFF
--- a/src/java/picard/vcf/GenotypeConcordanceCounts.java
+++ b/src/java/picard/vcf/GenotypeConcordanceCounts.java
@@ -103,7 +103,7 @@ public class GenotypeConcordanceCounts {
                 }
                 else if (includeHomRef || isVar(truthState, callState)) {
                     final TruthAndCallStates truthAndCallStates = new TruthAndCallStates(truthState, callState);
-                    final int count = getCount(truthAndCallStates);
+                    final long count = getCount(truthAndCallStates);
                     if (truthState.getCode()==callState.getCode()) {
                         //If we enter this, we are 'on the diagonal'
                         numerator += count;
@@ -146,7 +146,7 @@ public class GenotypeConcordanceCounts {
         for (final TruthState truthState : truthStateArray) {
             for (final CallState callState : CallState.values()) {
                 final TruthAndCallStates truthAndCallStates = new TruthAndCallStates(truthState, callState);
-                final int count = getCount(truthAndCallStates);
+                final long count = getCount(truthAndCallStates);
                 for (final ContingencyState contingencyState : scheme.getConcordanceStateArray(truthAndCallStates)) {
                     if (ContingencyState.TP == contingencyState) {
                         numerator += count;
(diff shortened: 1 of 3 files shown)


nh13:
Testing this on a large dataset to validate it works, but here nonetheless in case someone wants to review or others hit this bug.

yfarjoun:
@kbergin, would you like to take a look?

Y.

nh13:
Looks like it gave non-negative numbers on a WGS run for me.

kbergin:
@nh13 it looks good to me!

Just curious, for casting is it typical to put it like `(long)0` or `(long) 0` (with or without space)?

Otherwise :+1:.

lbergelson:
@kbergin for constants you should probably just use a long literal instead of casting like that.  `0l`

src/java/picard/vcf/GenotypeConcordanceCounts.java
```diff
@@ -272,19 +272,19 @@ public int getSum(final Set<TruthState> truthStateSet, final Set<CallState> call
...
      */
-    public Map<ContingencyState, Integer> getContingencyStateCounts(final GenotypeConcordanceScheme scheme) {
+    public Map<ContingencyState, Long> getContingencyStateCounts(final GenotypeConcordanceScheme scheme) {
         scheme.validateScheme();
 
-        final Map<ContingencyState, Integer> counts = new HashMap<ContingencyState, Integer>();
+        final Map<ContingencyState, Long> counts = new HashMap<ContingencyState, Long>();
         for (final ContingencyState contingencyState : ContingencyState.values()) {
-            counts.put(contingencyState, 0);
+            counts.put(contingencyState, (long)0);
```
kbergin: @lletourn @nh13 in that case, would it be best to put `0L` here?

lletourn: It shouldn't change much but yes using literals vs casting should be better.

Louis

nh13: Done, good idea.

kbergin:
Great! :+1: 

Thanks for spotting this bug!

src/java/picard/vcf/GenotypeConcordanceCounts.java
lbergelson: Who's this alternate Louis? There can be only 1!

nh13:
With a :+1: from @kbergin will merge.

src/java/picard/vcf/GenotypeConcordanceCounts.java
lletourn: Haha I worked at McGill University Genome center and I did send pull
requests to picard a few times.

I thought this was about a pull request I had sent awhile ago.

In anycase, still good form to use literals.

Louis

PR #303, opened 2015-10-08
Modified CommandLineProgram to, when the VALIDATION_STRINGENCY is eit…

tfenne:
…her LENIENT or SILENT, set the default option on VariantContextWriterBuilder to allow fields that are not defined in the VCFHeader. The result being that it is now possible to use all VCF modifying programs (e.g. FilterVcf, LiftoverVcf, etc.) with input VCFs that have incomplete headers, by specifying VALIDATION_STRINGENCY=LENIENT on the command line.

### DIFF
--- a/src/java/picard/vcf/LiftoverVcf.java
+++ b/src/java/picard/vcf/LiftoverVcf.java
@@ -2,6 +2,7 @@ package picard.vcf;
 
 import htsjdk.samtools.Defaults;
 import htsjdk.samtools.SAMSequenceRecord;
+import htsjdk.samtools.ValidationStringency;
 import htsjdk.samtools.liftover.LiftOver;
 import htsjdk.samtools.reference.ReferenceSequenceFileWalker;
 import htsjdk.samtools.util.CloserUtil;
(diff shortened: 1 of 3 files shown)


tfenne:
@yfarjoun This is the matching PR to 342 in HTSJDK.  Would you mind taking a quick look please?

src/java/picard/vcf/LiftoverVcf.java
```diff
@@ -137,7 +138,7 @@ public static void main(final String[] args) {
         log.info("Lifting variants over and sorting.");
 
         final SortingCollection<VariantContext> sorter = SortingCollection.newInstance(VariantContext.class,
-                new VCFRecordCodec(outHeader),
+                new VCFRecordCodec(outHeader, VALIDATION_STRINGENCY != ValidationStringency.STRICT),
```
yfarjoun: Why are you not using the default option here (and below)?

tfenne: The default is handled by `VariantContextWriterFactory`, which creates `VariantContextWriter`s, which internally uses a `VCFEncoder`.  The following was new to me when I was doing this - `VCFRecordCodec` is basically a custom wrapper around `VCFEncoder`, for `SortingCollection` to use so that it can write VCF records to files without having to write the header to each of them.  It didn't seem worth it to also add the ability to change the default behaviour of `VCFRecordCodec` .

yfarjoun: got it. thanks for the explanation!

yfarjoun:
@ktibbett do you want another pair of eyes? This looks fine to me!

ktibbett:
@yfarjoun I trust you and Tim!  Thumbs up.

PR #304, opened 2015-10-08
Unmap foreign contaminants in MergeBamAlignment

mattsooknah:
* Add the ability to identify reads that may have originated from cross-species contamination as part of MergeBamAlignment. The typical signature is fewer than N unclipped bases on the primary alignment (N=30 by default), with both sides clipped. 
* For reads identified as contaminant, we un-map all alignments corresponding to that read name,  possibly saving the original mapping info in the OP/OC/OF tags, and flag them using the FT tag.
* Also handle the case where un-mapping alignments results in multiple unmapped records with the same read name and pair flags. If this happens, only write records that originally corresponded to the primary alignment.

Note that this will be turned off by default (for now), so the default behavior of MergeBamAlignment should not be affected.


mattsooknah:
@yfarjoun please review

src/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -349,6 +357,10 @@ public void mergeAlignment(final File referenceFasta) {
                 final boolean clone = nextAligned.numHits() > 1 || nextAligned.hasSupplementalHits();
                 SAMRecord r1Primary = null, r2Primary = null;
 
+                // by this point there should be a single chosen primary alignment, which we will use to determine whether the read is contaminant.
+                // this must be done before the main iteration, since secondary / supplementary alignments will be affected by the primary.
+                final boolean isContaminant = this.unmapContaminantReads && isContaminant(nextAligned);
```
yfarjoun: can you rename this variable to that it includes the unmapContaminantReads somehow?

perhaps removeDueToContaminant, or removeSinceContaminant

mattsooknah:
@yfarjoun comments addressed

mattsooknah:
@yfarjoun ready for re-review

src/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -349,6 +357,10 @@ public void mergeAlignment(final File referenceFasta) {
                 final boolean clone = nextAligned.numHits() > 1 || nextAligned.hasSupplementalHits();
                 SAMRecord r1Primary = null, r2Primary = null;
 
+                // by this point there should be a single chosen primary alignment, which we will use to determine whether the read is contaminant.
+                // this must be done before the main iteration, since secondary / supplementary alignments will be affected by the primary.
+                final boolean removeDueToContaminant = this.unmapContaminantReads && isContaminant(nextAligned);
```
yfarjoun: rename s/remove/unmap/

src/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -505,18 +527,29 @@ private HitsForInsert nextAligned() {
 
     /**
      * Copies alignment info from aligned to unaligned read, clips as appropriate, and sets PG ID.
+     * May also un-map the resulting read if the alignment is bad (e.g. no unclipped bases).
```
yfarjoun: "few" unclipped bases (no unclipped bases is actually illegal)

src/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -505,18 +527,29 @@ private HitsForInsert nextAligned() {
...
         } else if (SAMUtils.recordMapsEntirelyBeyondEndOfReference(aligned)) {
             log.warn("Record mapped off end of reference; making unmapped: " + aligned);
             SAMUtils.makeReadUnmapped(unaligned);
+        } else if (isContaminant) {
+            log.warn("Record looks like foreign contamination; making unmapped: " + aligned);
+            // NB: for reads that look like contamination, just set unmapped flag and zero MQ but keep other flags as-is.
+            // this maintains the sort order so that downstream analyses can use them for calculating evidence
+            // of contamination vs other causes (e.g. structural variants)
+            unaligned.setReadUnmappedFlag(true);
+            unaligned.setMappingQuality(SAMRecord.NO_MAPPING_QUALITY);
```
yfarjoun: how come the previous calls of makeReadUnmapped didn't require the subsequent setMappingQuality(SAMRecord.NO_MAPPING_QUALITY)?

src/java/picard/sam/SamAlignmentMerger.java
```diff
@@ -333,6 +342,28 @@ protected boolean ignoreAlignment(final SAMRecord sam) {
...
     }
 
+    /**
+     * Criteria for contaminant reads:
+     * 1. primary alignment has fewer than minUnclippedBases unclipped bases
+     * 2. primary alignment has both ends clipped
+     * 3. for pairs, at least one end of primary alignment meets above criteria
+     */
+    protected boolean isContaminant(final HitsForInsert hits) {
+        final OverclippedReadFilter contaminationFilter = new OverclippedReadFilter(minUnclippedBases, false);
```
yfarjoun: shoudl this be static? seems silly to be new-ing a object for every read.

src/java/picard/sam/AbstractAlignmentMerger.java
mattsooknah: Because ```SAMUtils.makeReadUnmapped``` takes care of that (plus changing lots of other fields). Whereas ```isContaminant``` is a special case that only affects 0x4 and MQ. I suppose we could make a separate ```SAMUtils``` method for the special case, if that makes things cleaner.

src/java/picard/sam/SamAlignmentMerger.java
mattsooknah: Good point.

src/java/picard/sam/AbstractAlignmentMerger.java
mattsooknah: Illegal in what sense? Zero unclipped bases is a condition we check for here (in ```cigarMapsNoBasesToRef```), so I assume it shows up in raw aligner output from time to time.

src/tests/java/picard/sam/MergeBamAlignmentTest.java
```diff
@@ -178,8 +177,8 @@ public void testMerger() throws Exception {
                 false, true, false, 1,
                 "0", "1.0", "align!", "myAligner",
                 true, fasta, output,
-                SamPairUtil.PairOrientation.FR, null, null, null
-        );
+                SamPairUtil.PairOrientation.FR, null, null, null,
```
yfarjoun: can you fix the code alignment here and below?

src/tests/java/picard/sam/MergeBamAlignmentTest.java
```diff
@@ -106,8 +105,8 @@ public void testMergerWithSupplemental() throws Exception {
                 false, true, false, 1,
                 "0", "1.0", "align!", "myAligner",
                 true, fasta, outputWithSupplemental,
-                SamPairUtil.PairOrientation.FR, null, null, null
-        );
+                SamPairUtil.PairOrientation.FR, null, null, null,
```
yfarjoun: instead of changing so many lines, coudl you add another version of doMergeAlignment that takes one argument less?

yfarjoun:
back to you @mattsooknah

src/java/picard/sam/AbstractAlignmentMerger.java
yfarjoun: yes...I got confused about something else.

mattsooknah:
aand back to you @yfarjoun

yfarjoun:
:+1:

PR #306, opened 2015-10-11
After editing CollectGcBias to avoid using too much memory a bug was …

kbergin:
…found that was causing the program to run for longer than expected in production. The cause was unnecesssarily pulling down the reference when it could be doing it less frequently. The call to pull down the reference was moved inside of an if statement, but to allow it to be used for the 'addRead' call I had to store the refBases byte[] in memory, it needed to be accessible outside of the if statement. The reference contig is used in the addRead function to calculate the error rate. 

@eitanbanks @yfarjoun Please let me know if you think storing the reference bases of one contig is too much memory.

### DIFF
--- a/src/java/picard/analysis/GcBiasMetricsCollector.java
+++ b/src/java/picard/analysis/GcBiasMetricsCollector.java
@@ -55,6 +55,7 @@ public class GcBiasMetricsCollector extends MultiLevelCollector<GcBiasMetrics, I
     //will hold the relevant gc information per contig
     private byte [] gc = null;
     private int referenceIndex = -1;
+    private byte [] refBases = null;
 
     public GcBiasMetricsCollector(final Set<MetricAccumulationLevel> accumulationLevels, final int[] windowsByGc,
                                   final List<SAMReadGroupRecord> samRgRecords, final int scanWindowSize, final boolean bisulfite) {
@@ -125,13 +126,12 @@ public class GcBiasMetricsCollector extends MultiLevelCollector<GcBiasMetrics, I
             final SAMRecord rec = args.getRec();
             final String type;
             if (!rec.getReadUnmappedFlag()) {
-                final ReferenceSequence ref = args.getRef();
-                final byte[] refBases = ref.getBases();
-                StringUtil.toUpperCase(refBases);
-                final int refLength = refBases.length;
-                final int lastWindowStart = refLength - scanWindowSize;
-
                 if(referenceIndex != rec.getReferenceIndex() || gc == null){
+                    final ReferenceSequence ref = args.getRef();
+                    refBases = ref.getBases();
+                    StringUtil.toUpperCase(refBases);
+                    final int refLength = refBases.length;
+                    final int lastWindowStart = refLength - scanWindowSize;
                     gc = GcBiasUtils.calculateAllGcs(refBases, lastWindowStart, scanWindowSize);
                     referenceIndex=rec.getReferenceIndex();
                 }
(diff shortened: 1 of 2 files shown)


eitanbanks:
Looks good.
I believe that the original version from Tim kept one contig's worth of bases in memory, so this should be equivalent to that (and should therefore be fine).

eitanbanks:
@kbergin did you confirm that this runs much more quickly on a (large enough) test BAM?

src/java/picard/analysis/GcBiasMetricsCollector.java
```diff
@@ -125,13 +126,12 @@ public void acceptRecord(final GcBiasCollectorArgs args) {
...
             final String type;
             if (!rec.getReadUnmappedFlag()) {
-                final ReferenceSequence ref = args.getRef();
-                final byte[] refBases = ref.getBases();
-                StringUtil.toUpperCase(refBases);
-                final int refLength = refBases.length;
-                final int lastWindowStart = refLength - scanWindowSize;
-
-                if(referenceIndex != rec.getReferenceIndex() || gc == null){
+                if(referenceIndex != rec.getReferenceIndex() || gc == null || refBases == null){
```
yfarjoun: Why did you make refBases a class non-local? I think it could remain local, as you only need it here for the calculation of the gc (which should remain non-local)..

eitanbanks: good point

src/java/picard/analysis/GcBiasMetricsCollector.java
```diff
@@ -125,13 +126,12 @@ public void acceptRecord(final GcBiasCollectorArgs args) {
             final SAMRecord rec = args.getRec();
```
yfarjoun: since we are already here, could you add a @Override to this function signature?

src/java/picard/analysis/GcBiasMetricsCollector.java
yfarjoun: nm. this is fine. I get it.

yfarjoun: no, I was mistaken.

yfarjoun: we need it in order to calculate the error rate...which is also part of the GC calculation.

yfarjoun:
:100:

src/java/picard/analysis/GcBiasMetricsCollector.java
eitanbanks: I was reading too fast.  I was thinking you meant that the refBases == null check wasn't necessary (since it's not).  But that doesn't really matter I guess.

kbergin:
I double checked the new version and it runs the initial run through the reference in about 90 seconds and the first 1 million records were processed in 7 seconds on -Xmx1G in picard02 on a full size NA12878 wgs bam.

I put in a second commit to remove the unnecessary check for refBases == null and to change the short name of SCAN_WINDOW_SIZE to "WINDOW_SIZE", which is not much shorter but is what the old name was for that command line argument. I know it didn't come up from Kathleen as an issue in production, but anyone who has any commands saved or worked into a pipeline would have an issue with the name change and it seemed unnecessary to make everyone change it. But at least the longer name is more clear per matt's suggestion in my code review. win-win.

kbergin:
@yfarjoun would you look at what the other thing Eric was mentioning in the emails with me on Tuesday? It wasn't holding up production and I want to make sure I do it correctly.

Given the :100: I'm going to rebase and merge

Also sorry for the delay in response, I was on a plane! Back in Boston :)

eitanbanks:
Yes, merge!

Issue #305, opened 2015-10-08
Supporting METRICS_ACCUMULATION_LEVEL in CollectMultipleMetrics


ktibbett:
@davidangb and I would like to support METRICS_ACCUMULTATION_LEVEL in CollectMultipleMetrics.  Not all SinglePassSam programs support this option, so our proposed approach is to add it as a parameter and, if it is specified, to validate that all arguments to PROGRAM support accumulation levels (and fail with an exception if they do not).  This will involve making METRICS_ACCUMULTATION_LEVEL, which is currently private, public on the command-line programs that do support it.

Questions, comments, or concerns (@tfenne/@nh13 )?

tfenne:
Seems reasonable to me.  It might be nice to create a simple interface that all the relevant metrics collection programs can implement, which would a) enforce consistency around how it's handled and b) give you a clearer check than checking a parameter value.

ktibbett:
There's actually a ProgramInterface in CollectMultipleMetrics (currently definesneedsReferenceSequence()) and we'll just add supportsMetricsAccumulationLevels, defaulting to true since most programs do, there.

PR #307, opened 2015-10-13
Define Sequencing ArtifactMetrics in CollectMultipleMetrics, support metric accumulation level

davidangb:
Add support for passing METRIC_ACCUMULATION_LEVEL in CollectMultipleMetrics (see bug #305). Add a definition for CollectSequencingArtifactMetrics.

For reviewers:

* should METRIC_ACCUMULATION_LEVEL default to ALL_READS instead of empty? Defaulting to empty allows us to validate easily - we can check if the caller was explicit.
* the ProgramInterface interface, and its implementing classes, now need to override two methods (needsReferenceSequence and supportsMetricAccumulationLevel). Would it be better to make this an abstract class and specify concrete default implementations?

### DIFF
--- a/src/java/picard/analysis/CollectMultipleMetrics.java
+++ b/src/java/picard/analysis/CollectMultipleMetrics.java
@@ -1,7 +1,9 @@
 package picard.analysis;
 
 import htsjdk.samtools.util.CollectionUtil;
+import htsjdk.samtools.util.Log;
 import picard.PicardException;
+import picard.analysis.artifacts.CollectSequencingArtifactMetrics;
 import picard.cmdline.CommandLineProgram;
 import picard.cmdline.CommandLineProgramProperties;
 import picard.cmdline.Option;
@@ -9,9 +11,7 @@ import picard.cmdline.programgroups.Metrics;
 import picard.cmdline.StandardOptionDefinitions;
 
 import java.io.File;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
+import java.util.*;
 
 /**
  * Class that is designed to instantiate and execute multiple metrics programs that extend
@@ -23,7 +23,7 @@ import java.util.List;
 @CommandLineProgramProperties(
         usage = "Takes an input BAM and reference sequence and runs one or more Picard " +
                 "metrics modules at the same time to cut down on I/O. Currently all programs are run with " +
-                "default options and fixed output extesions, but this may become more flexible in future.",
+                "default options and fixed output extensions, but this may become more flexible in future.",
         usageShort = "A \"meta-metrics\" calculating program that produces multiple metrics for the provided SAM/BAM",
         programGroup = Metrics.class
 )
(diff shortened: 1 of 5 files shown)


ktibbett:
@mattsooknah , could you please review this PR carefully?  Thanks!

mattsooknah:
@davidangb 

* I agree that defaulting METRIC_ACCUMULATION_LEVEL to empty is a good idea, to force the caller to be explicit
* I think it's fine to keep ProgramInterface as an interface, again to force implementing classes to be explicit

src/java/picard/analysis/CollectMultipleMetrics.java
```diff
@@ -197,6 +249,9 @@ public SinglePassSamProgram makeInstance(final String outbase, final File input,
     @Option(shortName = StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "Base name of output files.")
     public String OUTPUT;
 
+    @Option(shortName="LEVEL", doc="The level(s) at which to accumulate metrics.")
+    public Set<MetricAccumulationLevel> METRIC_ACCUMULATION_LEVEL = new HashSet<MetricAccumulationLevel>();
```
mattsooknah: We may want different sub-programs to get called with different values of METRIC_ACCUMULATION_LEVEL, so it doesn't work to have it be a global command line argument. I'm not sure what the best solution to this is, though...

mattsooknah:
Also, CollectSequencingArtifactMetrics has two additional file arguments that we'll probably want to set here - INTERVALS and DB_SNP. Which means they'll probably need to be added to the CollectMultipleMetrics argument list.

mattsooknah:
@davidangb see my comments above

davidangb:
Do INTERVALS and DB_SNP need to be arguments to CollectMultipleMetrics, or can they be overridden with static values, similar to how CollectGcBiasMetrics works inside CollectMultipleMetrics? In the current architecture, making it completely dynamic would require more changes to ProgramInterface, which I'm hesitant to do. And allowing arbitrary arguments to be passed from CollectMultipleMetrics' commandline to the individual programs would be an architectural change.

mattsooknah:
@davidangb They'll have potentially different values each time the code is called, so they need to be passed in dynamically. 

As you pointed out, this requires changing ProgramInterface, which is sort of clunky. But it's probably easier to just add these two arguments to the list than allow passing of arbitrary arguments (since I don't foresee needing any more parameters at this time).

davidangb:
Updated to include DB_SNP and INTERVALS via ProgramInterface. I completely agree this is a clunky change; if there are any more arguments, this really needs to be refactored. But with @mattsooknah 's feedback, we'll leave this as-is, assuming no other PR feedback.

ktibbett:
@mattsooknah , can you please review these changes?

@davidangb , I messaged you offline about one test that we should run locally before merging this -- can you please take a  look?

src/java/picard/analysis/CollectMultipleMetrics.java
```diff
@@ -218,7 +282,20 @@ public static void main(final String[] args) {
...
+            if (DB_SNP == null) {
+                errs.add("DB_SNP must be specified for CollectSequencingArtifactMetrics.");
+            }
+            if (INTERVALS == null) {
+                errs.add("INTERVALS must be specified for CollectSequencingArtifactMetrics.");
+            }
+            if (!errs.isEmpty()) {
+                return errs.toArray(new String[errs.size()]);
+            }
+        }
```
mattsooknah: One caveat: *sometimes* we'll want to pass in null values, so it would be better to not have the validation (yeah, pretty sketchy...). Sorry for not being clear about that.

davidangb:
@ktibbett the manual test that you requested passed; see PO-3063 for details.

@mattsooknah I removed the validation (and its associated test) - is this more along the lines of what you're looking for?

davidangb:
okay, @ktibbett @mattsooknah ready for re-review. By request, I've changed functionality so that:
* METRIC_ACCUMULATION_LEVEL now defaults to ALL_READS. With this change, we don't require any calling scripts to be updated.
* We log a warning if the user has overridden the default for METRIC_ACCUMULATION_LEVEL and is also trying to run a program that does not support METRIC_ACCUMULATION_LEVEL.

mattsooknah:
@davidangb looks good, just squash + rebase and then I'll merge.

davidangb:
@mattsooknah squashed and ready, thanks for all the help!

PR #312, opened 2015-10-22
still produce a plot when we have multiple orientations for all reads

nh13:
@yfarjoun @ktibbett wondering if you could take a look since you were mentioned in the original comment.  Not producing a PDF has bit me multiple times now, since if we run this on all the reads and have multiple orientations detected it barfs.


ktibbett:
@nh13 , it's been a long time since I looked at this, but I think this arose when the library and sample name were identical.  Is that the case you're trying to address?

nh13:
The histogram looks like
```
## HISTOGRAM    java.lang.Integer
insert_size All_Reads.fr_count  All_Reads.rf_count
1   86  86
2   1024    1290
...
```

so the code

```
## The histogram has a fr_count/rf_count/tandem_count for each metric "level"
## This code parses out the distinct levels so we can output one graph per level
headers <- sapply(sub(".fr_count","",names(histogram),fixed=TRUE), "[[" ,1)
headers <- sapply(sub(".rf_count","",headers,fixed=TRUE), "[[" ,1)
headers <- sapply(sub(".tandem_count","",headers,fixed=TRUE), "[[" ,1)

## Duplicated header names cause this to barf. KT & Yossi report that this is going to be extremely difficult to
## resolve and it's unlikely that anyone cares anyways. Trap this situation and avoid the PDF so it won't cause
## the workflow to fail
if (any(duplicated(headers))) {
```
will produce duplicate headers, namely ```All_Reads```.

Interestingly, the sample and library are the same, but they are not present in the metrics file's InsertSizeMetrics.

yfarjoun:
Since you already have a metrics file in hand, do you mind adding it to the test-files and throwing a test on this? You can just check that a reasonable pdf is produced. It would greatly simplify the review process.

nh13:
Good idea, will do.

nh13:
@yfarjoun back to you.

src/scripts/picard/analysis/insertSizeHistogram.R
```diff
@@ -33,11 +33,10 @@ headers <- sapply(sub(".fr_count","",names(histogram),fixed=TRUE), "[[" ,1)
 headers <- sapply(sub(".rf_count","",headers,fixed=TRUE), "[[" ,1)
 headers <- sapply(sub(".tandem_count","",headers,fixed=TRUE), "[[" ,1)
 
-## Duplicated header names cause this to barf. KT & Yossi report that this is going to be extremely difficult to
-## resolve and it's unlikely that anyone cares anyways. Trap this situation and avoid the PDF so it won't cause
-## the workflow to fail
+## Duplicate header names could cause this to barf.  But it really shouldn't when we have "All_reads.fr_count" and 
+## "All_reads.rf_count" for example.  Not sure why this would fail, but I care.
 if (any(duplicated(headers))) {
```
yfarjoun: it seems that your code will work fine in the "FALSE" case as well...do you want to simplify the code?

src/scripts/picard/analysis/insertSizeHistogram.R
```diff
@@ -45,54 +44,53 @@ if (any(duplicated(headers))) {
...
-    }
+  colors <- c()
+  labels <- c()
 
-    if (tandem %in% names(histogram)) {
-      lines(histogram$insert_size, as.matrix(histogram[tandem]),  type="h", col="orange")
-      colors <- c(colors, "orange")
-      labels <- c(labels, "TANDEM")
-    }
+  if (fr %in% names(histogram) ) {
```
yfarjoun: if I were you, I'd complain about the inconsistent spaces in this if statement and the following ones...:-)

src/tests/java/picard/analysis/CollectInsertSizeMetricsTest.java
```diff
@@ -223,4 +225,20 @@ public void testHistogramWidthIsSetProperly() throws IOException {
...
+        final File output = new File("testdata/picard/analysis/directed/CollectInsertSizeMetrics", "multiple_orientation.sam.insert_size_metrics");
+        final File pdf = File.createTempFile("test", ".pdf");
+        pdf.deleteOnExit();
+
+        final int rResult;
+        rResult = RExecutor.executeFromClasspath(
+                CollectInsertSizeMetrics.Histogram_R_SCRIPT,
+                output.getAbsolutePath(),
+                pdf.getAbsolutePath(),
+                "Flags of Chad and Romania");
```
yfarjoun: pretty colors!

yfarjoun:
:+1:

yfarjoun:
@nh13 coudl you please rebase and we can then merge.

nh13:
@yfarjoun rebased and squashed back to you.

PR #313, opened 2015-10-26
changed the unknown parents in the sexMap output to dots (.)

yfarjoun:


yfarjoun:
no /0 here, commit message spilled over from picard-private...

eitanbanks:
Looks reasonable

ktibbett:
Looks good.

PR #314, opened 2015-10-27
merge sam file with interval

lindenb:
The following pull request add an optional parameter INTERVALS to merge the sorted+index bam files in a defined list of regions.

```java
@Option(shortName = "INTERVALS"...)
public File INTERVALS = null;
```
if **INTERVALS** is defined, it checks that the bam is sorted+index, fills a map **samReaderToIterator**   `Map<SamReader, CloseableIterator<SAMRecord> >`   ; The values of the map are used to create a MergingSamRecordIterator.

```java
 if( intervalList == null) {
  iterator = new MergingSamRecordIterator(headerMerger, readers, mergingSamRecordIteratorAssumeSorted);
  }
  else {
  iterator = new MergingSamRecordIterator(headerMerger, samReaderToIterator, true);
  }
```

### DIFF
--- a/src/java/picard/sam/MergeSamFiles.java
+++ b/src/java/picard/sam/MergeSamFiles.java
@@ -32,10 +32,15 @@ import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.SamFileHeaderMerger;
 import htsjdk.samtools.SamReader;
 import htsjdk.samtools.SamReaderFactory;
+import htsjdk.samtools.util.CloseableIterator;
 import htsjdk.samtools.util.CloserUtil;
 import htsjdk.samtools.util.IOUtil;
+import htsjdk.samtools.util.Interval;
+import htsjdk.samtools.util.IntervalList;
 import htsjdk.samtools.util.Log;
 import htsjdk.samtools.util.ProgressLogger;
+import htsjdk.samtools.util.SamRecordIntervalIteratorFactory;
+import picard.PicardException;
 import picard.cmdline.CommandLineProgram;
 import picard.cmdline.CommandLineProgramProperties;
 import picard.cmdline.Option;
@@ -44,7 +49,9 @@ import picard.cmdline.programgroups.SamOrBam;
 
 import java.io.File;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Reads a SAM or BAM file and combines the output to one file
@@ -83,6 +90,13 @@ public class MergeSamFiles extends CommandLineProgram {
     @Option(doc = "Comment(s) to include in the merged output file's header.", optional = true, shortName = "CO")
     public List<String> COMMENT = new ArrayList<String>();
 
+    @Option(shortName = "RGN", doc = "An interval list file that contains the locations of the positions to merge. "+
+            "Assume bam are sorted and indexed. "+
+            "The resulting file will contain alignments that may overlap with genomic regions outside the requested region. "+
+            "Unmapped reads are discarded.",
+            optional = true)
+    public File INTERVALS = null;
+
     private static final int PROGRESS_INTERVAL = 1000000;
 
     /** Required main method implementation. */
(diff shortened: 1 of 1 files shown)


src/java/picard/sam/MergeSamFiles.java
```diff
@@ -83,6 +90,9 @@
     @Option(doc = "Comment(s) to include in the merged output file's header.", optional = true, shortName = "CO")
     public List<String> COMMENT = new ArrayList<String>();
 
+    @Option(shortName = "INTERVALS", doc = "An interval list file that contains the locations of the positions to merge. Assume bam are sorted and indexed.", optional = true)
```
yfarjoun: Please comment on how this should be used. In particular I'm worried of people using this to scatter merge bam alignment and the resulting mess that would arise from that (since a read can overlaps more than one locus...)

lindenb: My initial idea was to quickly create a merged bam for a given (set of) regions before calling SNPs

> and the resulting mess that would arise from that

Doesn't MergingSamRecordIterator handle this problem ? if not, you're absolutely right, that would be a problem.  :-(

feel free to close that pull request if this is the case.

yfarjoun: What I meant was that one could think about scatter/gathering the merge
process by creating a bunch of processes that do merge on non-overlapping
intervals, and then concatenate all the resulting bams together (not sure
if it's worth it, but one could try...) the problem is that then the reads
that over lap two interval-lists will get included twice in the
concatenation step...

What are you starting with that you want this feature? (sorry to be nosy,
but I'm trying to understand the use-case.)

On Tue, Oct 27, 2015 at 3:59 PM, Pierre Lindenbaum <notifications@github.com

lindenb:
> What are you starting with that you want this feature? 

for variant calling I split my capture.bed into a set of smaller BED regions for a few undred BAMS. My pipeline looks like (pseudocode);

```
for each RGN in ${SUB}.bed
   call mpileup in ${RGN} using bam.list > ${RGN}.vcf
   echo  ${RGN}.vcf >  ${SUB}.vcf.list
end
merge-vcf  ${SUB}.vcf.list  >  ${SUB}.vcf
```

this is slow because mpileup needs to load all the BAMs for each regions. I would have liked something like

```
picard MergeSAM -B ${SUB}.bed -I bam.list -OUT /dev/stdout |   samtools mpileup - >  ${SUB}.vcf
```

yfarjoun:
I see. Well for this it should work fine. Could you please make a note in
the usage that the resulting files should not be merged? and that the
resulting file will contain alignments that overlap with regions _outside_
the requested region. Also, it should be mensioned that unmapped reads will
not be handled (right?)

lindenb:
i've added a few comments in @ options 

```java
@Option(shortName = "INTERVALS", doc =
"An interval list file that contains the locations of the positions to merge. "+
"Assume bam are sorted and indexed. "+
"Reads should not overlap two intervals. "+
"The resulting file will contain alignments that may overlap with genomic regions outside the requested region. "+
"Unmapped reads are discarded.",
optional = true)
public File INTERVALS = null;
```

src/java/picard/sam/MergeSamFiles.java
```diff
@@ -83,6 +90,14 @@
     @Option(doc = "Comment(s) to include in the merged output file's header.", optional = true, shortName = "CO")
     public List<String> COMMENT = new ArrayList<String>();
 
+    @Option(shortName = "INTERVALS", doc = "An interval list file that contains the locations of the positions to merge. "+
+            "Assume bam are sorted and indexed. "+
+            "Reads should not overlap two intervals. "+
```
yfarjoun: this comment is unneed, what I meant is that the results of multiple runs of this program may contain the same read even if the regions that were used are disjoint, therefore the results should not be combined downstream without careful validation.

lindenb: so what do you suggest ? checking that the regions in the bed file are distanst from XX bases ? or checking that each read only overlap one unique region ?

yfarjoun: Sorry, my comments are confusing.

I mean to say that the comment "Reads should not overlap two intervals." is unneeded, but that instead there should be a comment warning users from merging the resulting sequence files since they may contain the same read in both files. 

I do not mean that you should add any further checking.

lindenb:
I've added a warning+comment.

yfarjoun:
Thanks @lindenb :+1:  (please rebase-squash and I'll merge)

PR #315, opened 2015-10-27
Throw exception if no read group specified in CollectOxoGMetrics 

brendanofallon:
CollectOxoGMetrics will happily run on alignments without a read group line, but it won't emit any informative output or a warning. This will at least alert the user that a read group line is required.


ktibbett:
Thank you @brendanofallon -- looks good.  Will merge this once tests pass.

PR #321, opened 2015-11-02
AlignedReads int overflow bug in GcBias

kbergin:
Ran into a bug in CollectGcBias with a 100X coverage genome, where the number of aligned reads was greater than MAX_INT. In the latest version, the gc detail and summary metrics were only calculated and written out if totalAlignedReads was >0. Because of the overflow, totalAlignedReads was a random negative number, and thus no metrics were output, but empty (except header) metrics files still were written and no errors were thrown.

To address this, I've changed totalAlignedReads to a long and I've changed the if statement to skip only if totalAlignedReads==0.

### DIFF
--- a/src/java/picard/analysis/GcBiasMetricsCollector.java
+++ b/src/java/picard/analysis/GcBiasMetricsCollector.java
@@ -191,7 +191,7 @@ public class GcBiasMetricsCollector extends MultiLevelCollector<GcBiasMetrics, I
                 final long[] errorsByGc = gcCur.errorsByGc;
                 final long[] basesByGc = gcCur.basesByGc;
                 final int totalClusters = gcCur.totalClusters;
-                final int totalAlignedReads = gcCur.totalAlignedReads;
+                final long totalAlignedReads = gcCur.totalAlignedReads;
                 final String group = gcCur.group;
 
                 final GcBiasMetrics metrics = new GcBiasMetrics();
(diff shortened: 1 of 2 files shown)


src/java/picard/analysis/GcBiasMetricsCollector.java
```diff
@@ -191,7 +191,7 @@ public void addMetricsToFile(final MetricsFile<GcBiasMetrics, Integer> file) {
                 final long[] errorsByGc = gcCur.errorsByGc;
                 final long[] basesByGc = gcCur.basesByGc;
                 final int totalClusters = gcCur.totalClusters;
-                final int totalAlignedReads = gcCur.totalAlignedReads;
+                final long totalAlignedReads = gcCur.totalAlignedReads;
```
yfarjoun: This change seems to solve the problem, why the change below?

kbergin: This may not be the best decision, but the reason I was able to figure out the bug was because the old version (aka pre-single pass sam) of gc bias didn't have this check, so it calculated all gc bias metrics as normal and the only anomaly in the output was the overflowed totalAlignedReads value. I thought it'd be better to have the gc bias program functioning as closely to the old version as possible.

I tried running it without the check altogether but when there are no aligned reads to a specific accumulation level the 'group' ends up being null and it threw errors from that. 

I decided on the skip only if it equaled zero as a compromise, so that when there's really no aligned reads it doesn't have an issue but it would still function similarly to the old version. 

What do you think is best?

yfarjoun: I guess I was wondering why change from the if {} to the if {continue}
construct which resulted in many changed lines, but same behaviour. The
bug-fix (as far as I could tell) was done once the int was changed to a
long..am I incorrect?

Y.

kbergin: Yes you're right, that part fixes the bug. I'll just implement the int->long change.

kbergin:
@yfarjoun I changed the code back per your review comment, there are some lines that were autoformatted by intelliJ and I think they look clearer now so I'll just leave them. 

Let me know if everything looks good and I can merge? Thanks!

yfarjoun:
:+1:

PR #330, opened 2015-11-06
Changes CollectQualityYieldMetrics to exclude secondary and supplemen…

tfenne:
…tal alignments by default.  Since the intended use is to calculate the yield in reads/bases/quality bases of a sequencing experiment, it seems wrong to include secondary and supplemental alignments as that would lead to double counting some bases.  Shouldn't affect any computations done on unmapped BAMs, or on BAMs that don't contain secondary or supplemental alignment records.

### DIFF
--- a/src/java/picard/analysis/CollectQualityYieldMetrics.java
+++ b/src/java/picard/analysis/CollectQualityYieldMetrics.java
@@ -47,7 +47,8 @@ import java.io.File;
  */
 @CommandLineProgramProperties(
         usage = "Collects quality yield metrics, a set of metrics that quantify the quality and yield of sequence data from a " +
-                "SAM/BAM input file.",
+                "SAM/BAM input file. Note that the default behaviour of this program changed as of November 6th 2015 to no longer " +
+                "include secondary and supplemental alignments in the computation.",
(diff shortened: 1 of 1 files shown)


tfenne:
@ktibbett Super simple change here - could you take a look and let me know what you think?

ktibbett:
@tfenne my only concern here is that having the default set to false on these new options will change the behavior of the program for anyone using it on BAMs with secondary/supplementary reads.  I agree that false is the correct default, so I'm not sure there's anything to do, except perhaps ask you to note this change in the default behavior explicitly in the commit comment and note it prominently in the notes for the next release.  

image

tfenne:
@ktibbett I'll update the commit comment, and perhaps also put something in the usage for the program about the default behaviour having changed as of this date.

tfenne:
@ktibbett Am taking the thumbs up pic as approval to merge, I hope that's ok.  I've updated the commit message and the usage.  If that's not what you meant, let me know and I'll unwind the change.

ktibbett:
@tfenne , yep, that was a thumbs-up.   Thanks for the changes.

PR #331, opened 2015-11-06
Nh improve hs metrics

nh13:


nh13:
@yfarjoun @ktibbett @tfenne can you take a look at this one?  There are some significant changes.

To preempt, I added a method to always clips overlapping reads. I could add a command line option to set this to false (and set it to default).  I was unsure about the desire for such an option as not to change the current behavior, although new metrics are being added anyhow.

src/java/picard/analysis/directed/HsMetrics.java
```diff
@@ -83,6 +83,9 @@
     /** The number of PF aligned bases that mapped to a targeted region of the genome. */
     public long ON_TARGET_BASES;
 
+    /** The fraction of aligned bases that were filtered out because they were of low base quality (default is < 20). */
```
nh13: NB: default is not 20.

src/java/picard/analysis/directed/HsMetrics.java
```diff
@@ -98,6 +101,9 @@
     /** The mean coverage of targets that received at least coverage depth = 2 at one base. */
     public double MEAN_TARGET_COVERAGE;
 
+    /** The mean coverage of targets that received at least coverage depth = 2 at one base. */
```
nh13: TODO: verify that it is at least depth 2 we are computing upon.

src/java/picard/analysis/directed/HsMetrics.java
```diff
@@ -110,12 +116,17 @@
     /** The number of targets that did not reach coverage=2 over any base. */
     public double ZERO_CVG_TARGETS_PCT;
 
+    /** The fraction of aligned bases that were filtered out because they were in reads with low mapping quality (default is < 20). */
```
nh13: check the default.

nh13:
Closing since I want to review.

PR #332, opened 2015-11-07
Modernizing Targeted Metrics Calculations

nh13:
@ktibbett @tfenne @yfarjoun there are a lot of updates to how we calculate targeted metrics, so likely a discussion is needed for this pull request.

Some of the things to notice are:
* now count bases in supplemental reads.
* relax the minimum coverage all the way to zero for some of the metric values; previously the minimum coverage was two. @tfenne for a historical explanation.
* deprecate ```CalculateHsMetrics``` in favor of ```CollectHsMetrics```, since this has always annoyed me.
* added some new metrics.  This should hopefully not affect your production use of ```CalculateHsMetrics```
* do not count bases that overlap across reads.  The default mode is to clone the ```SAMRecord``` so we can modify the cigar for overlapping reads, which is thus a side-effect.  Folks who use these collectors and do not want side effects are given a sensible default, but folks who don't care about side effects (I am looking at you ```CalculateHsMetrics```, ```CollectHsMetrics```, and ```CollectTargetedPcrMetrics```) can enable them.  The code should be somewhat faster with side effects, since we avoid the ```SAMRecord.clone``` method.

Thank-you in advance for reviewing this PR.

image


src/java/picard/analysis/directed/HsMetrics.java
```diff
@@ -107,15 +113,20 @@
     /** The fold by which the baited region has been amplified above genomic background. */
     public double FOLD_ENRICHMENT;
 
-    /** The number of targets that did not reach coverage=2 over any base. */
+    /** The number of targets that did not reach coverage1 over any base. */
```
tfenne: coverage1 -> coverage=1

src/java/picard/analysis/directed/HsMetrics.java
```diff
@@ -107,15 +113,20 @@
...
     public double ZERO_CVG_TARGETS_PCT;
 
+    /** The fraction of aligned bases that were filtered out because they were in reads with low mapping quality (default is < 0). */
+    public double PCT_EXC_MAPQ;
```
tfenne: Why include a new metric for this, but not for the pct excluded for base q or overlapping reads?

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -109,16 +116,35 @@
 
     private final long genomeSize;
 
+    private final int minimumMappingQuality;
+
```
tfenne: No need for all the extra empty lines between member variable declarations.

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -228,6 +276,15 @@ public TargetMetricsCollector(final Set<MetricAccumulationLevel> accumulationLev
             }
         }
 
+
+        this.minimumMappingQuality = minimumMappingQuality;
+
```
tfenne: Double spacing

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -301,12 +380,77 @@ public void setBaitSetName(final String name) {
             this.metrics.PROBE_SET = name;
         }
 
+        private int getBasesAtMinimumQualityInBlock(final SAMRecord record, final AlignmentBlock block) {
+            int basesInBLockAtMinimumQuality = 0;
```
tfenne: BLock -> Block

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -301,12 +380,77 @@ public void setBaitSetName(final String name) {
             this.metrics.PROBE_SET = name;
         }
 
+        private int getBasesAtMinimumQualityInBlock(final SAMRecord record, final AlignmentBlock block) {
```
tfenne: Two things: 1) I think the name for this could be clearer, since it's bases at or above minimum quality.  Perhaps getCountOfBasesPassingQualityThreshold()?  2) A one line javadoc would help too.

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -301,12 +380,77 @@ public void setBaitSetName(final String name) {
...
+            }
+            return basesInBLockAtMinimumQuality;
+        }
+
+        private SAMRecord clipOverlappingBases(final SAMRecord rec) {
+            // NB: ignores how to handle supplemental records when present for both ends by just using the mate information in the record.
+
+            if (!rec.getReadPairedFlag()) return rec;
+
+            // Only clip records that are left-most in genomic order and overlapping and are the first end
```
tfenne: I think this is wrong.  If you have an FR read pair where the R2 is earlier on the genome on the F strand and R1 is later on the genome on the R strand, and they overlap, won't this erroneously skip clipping?  I think the clipping logic shouldn't care if it's R1 or R2.

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -301,12 +380,77 @@ public void setBaitSetName(final String name) {
...
-            // Just plain avoid records that are marked as not-primary
-            if (rec.isSecondaryOrSupplementary()) return;
+        public void acceptRecord(final SAMRecord record) {
+            // We should not "Just plain avoid records that are marked as not-primary" since:
+            // 1. We should disregard secondary alignments in for both record and base counting.
+            // 2. We should disregard supplementary alignments for ONLY record counting
+            // This means that we need to consider supplemental alignments for base counting but not record counting.
+
+            // NB: this could modify the record.  See noSideEffects
+            final SAMRecord rec;
```
tfenne: I think a ternary is clearer for cases like this:
`final SAMRecord rec = clipOverlappingReads ? clipOverlappingBases(record) : rec;`

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -284,11 +354,20 @@ public PerUnitTargetMetricCollector(final String probeSetName, final Set<Interva
             metrics.GENOME_SIZE      = genomeSize;
 
             this.coverageByTarget = new LinkedHashMap<Interval, Coverage>(coverageTargets.size() * 2, 0.5f);
-            for (Interval target : coverageTargets) {
+            for (final Interval target : coverageTargets) {
                 this.coverageByTarget.put(target, new Coverage(target,0));
             }
 
+            this.mapQFilter = new CountingMapQFilter(minimumMappingQuality);
```
tfenne: Slightly confusing that this is declared out-of-order from it's adding to the filter list.  Could you move it so that the creation and add are adjacent?

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -301,12 +380,77 @@ public void setBaitSetName(final String name) {
...
+            if (clipOverlappingReads) rec = clipOverlappingBases(record);
+            else rec = record;
+
+            // apply any filters
+            for (final SamRecordFilter filter : this.filters) {
+                if (filter.filterOut(rec)) return;
+            }
 
-            this.metrics.TOTAL_READS += 1;
+            if (!rec.getSupplementaryAlignmentFlag()) this.metrics.TOTAL_READS += 1;
```
tfenne: I'm not sure how I feel about TOTAL_READS getting incremented only after all the filters.  I tend to think it's better to include all reads (except secondary and supplementary) so that in general TOTAL_READS will match other computations of the total number of reads in other tools.  Same goes for PF_READS.

That said, since those are used as denominators in many other metrics, it's less clear to me whether the filtered out reads should be included in the denominators or not.

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -365,8 +520,10 @@ public void acceptRecord(final SAMRecord rec) {
 
                     for (final AlignmentBlock block : rec.getAlignmentBlocks()) {
                         final int end = CoordMath.getEnd(block.getReferenceStart(), block.getLength());
-                        for (int pos=block.getReferenceStart(); pos<=end; ++ pos) {
+                        int readPos = block.getReadStart(); // one-based
```
tfenne: ++ pos -> ++pos

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -365,8 +520,10 @@ public void acceptRecord(final SAMRecord rec) {
...
+                        for (int pos=block.getReferenceStart(); pos<=end; ++ pos, ++readPos) {
```
tfenne: ++ pos -> ++pos

src/java/picard/analysis/directed/TargetMetricsCollector.java
tfenne: why increment readPos in the loop control, but not declare it there?

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -378,15 +535,16 @@ public void acceptRecord(final SAMRecord rec) {
...
+            for (final AlignmentBlock block : rec.getAlignmentBlocks()) mappedBases += getBasesAtMinimumQualityInBlock(rec, block);
             int onBaitBases = 0;
 
             if (probes != null && !probes.isEmpty()) {
                 for (final Interval bait : probes) {
                     for (final AlignmentBlock block : rec.getAlignmentBlocks()) {
                         final int end = CoordMath.getEnd(block.getReferenceStart(), block.getLength());
-
-                        for (int pos=block.getReferenceStart(); pos<=end; ++pos) {
+                        int readPos = block.getReadStart(); // one-based
```
tfenne: Same thing. Declare readPos as a second variable in the for loop control

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -378,15 +535,16 @@ public void acceptRecord(final SAMRecord rec) {
...
+                        for (int pos=block.getReferenceStart(); pos<=end; ++pos, ++readPos) {
+                            if (baseQualities[readPos-1] < minimumBaseQuality) continue; // ignore low-quality bases
```
tfenne: Why not just add this as a third condition to the if() directly below instead of a separate if() and continue?

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -425,32 +585,47 @@ private void calculateTargetCoverageMetrics() {
             int zeroCoverageTargets = 0;
             int depthIndex = 0;
             double totalCoverage = 0;
-            int basesConsidered = 0;
+            long numTargetBasesAtZeroCoverage = 0; // should equal (targetBasesDepth[1] - targetBasesDepth[0]) below
+            final Histogram<Integer> targetCoverageHistogram = new Histogram<Integer>();
```
tfenne: I feel like this is redundant with depths[] above.  Could you just move the Arrays.sort(depths) above the line where you compute median coverage, and then get the median from depths[]?

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -724,32 +867,40 @@ public String toString() {
     /** The fold by which the probed region has been amplified above genomic background. */
     public double FOLD_ENRICHMENT;
 
-    /** The mean coverage of targets that recieved at least coverage depth = 2 at one base. */
+    /** The mean coverage of targets. */
     public double MEAN_TARGET_COVERAGE;
 
-    /** The number of targets that did not reach coverage=2 over any base. */
+    /** The mean coverage of targets. */
```
tfenne: mean -> median

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -724,32 +867,40 @@ public String toString() {
...
+    public double MEDIAN_TARGET_COVERAGE;
+
+    /** The number of targets that did not reach coverage1 over any base. */
```
tfenne: coverage1 -> coverage>=1

src/java/picard/analysis/directed/TargetedPcrMetrics.java
```diff
@@ -77,21 +80,29 @@
...
     public double MEAN_TARGET_COVERAGE;
 
+    /** The mean coverage of targets. */
+    public double MEDIAN_TARGET_COVERAGE;
+
     /** The fold by which the amplicon region has been amplified above genomic background. */
     public double FOLD_ENRICHMENT;
 
-    /** The number of targets that did not reach coverage=2 over any base. */
+    /** The number of targets that did not reach coverage1 over any base. */
```
tfenne: Copy/paste ftw (coverage1 again).

src/java/picard/analysis/directed/TargetedPcrMetrics.java
```diff
@@ -77,21 +80,29 @@
     /** The mean coverage of all amplicons in the experiment. */
     public double MEAN_AMPLICON_COVERAGE;
 
-    /** The mean coverage of targets that recieved at least coverage depth = 2 at one base. */
+    /** The mean coverage of targets. */
     public double MEAN_TARGET_COVERAGE;
 
+    /** The mean coverage of targets. */
```
tfenne: mean -> median

src/java/picard/filter/CountingFilter.java
```diff
@@ -0,0 +1,39 @@
...
+        if (filteredOut) {
+            ++filteredRecords;
+            for (final AlignmentBlock block : record.getAlignmentBlocks()) {
+                this.filteredBases += block.getLength();
+            }
+        }
+        return filteredOut;
+    }
+
+    abstract public boolean reallyFilterOut(final SAMRecord record);
```
tfenne: With this being hoisted out into picard.filters, it would be nice to add a little more javadoc, especially to this method.

src/tests/java/picard/analysis/directed/CollectHsMetricsTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+import picard.cmdline.CommandLineProgramTest;
+
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+
+public class CollectHsMetricsTest extends CommandLineProgramTest {
+
```
tfenne: Extra newline.

src/tests/java/picard/analysis/directed/CollectHsMetricsTest.java
```diff
@@ -0,0 +1,76 @@
...
+    private final static File TEST_DIR = new File("testdata/picard/analysis/directed/CollectHsMetrics");
+
+
```
tfenne: Extra newline.

tfenne:
@nh13 Finished my review.  Looks really good!  I think this update is long overdue!

A handful of typos etc. and two bigger things.  I think the clipping logic should ignore whether it's looking at R1 or R2, and clip in both cases.  And I'm not sure how I feel about filtered out reads dropping out from TOTAL/PF READS.

tfenne:
One more thought - given the deprecation of CalculateHsMetrics and renaming to CollectHsMetrics perhaps this is an opportunity to:
a) Keep CalculateHsMetrics' defaults backward compatible with prior behaviour
b) Have some more sensible defaults on CollectHsMetrics (i.e. read clipping on, MAPQ >= 20 etc)

src/java/picard/analysis/directed/HsMetrics.java
nh13: ```PCT_EXC_BASEQ``` is there but not overlapping.  I will add the latter.

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: I try to follow the style in the rest of the file.  In the diff here it doesn't show the extra lines elsewhere.

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: changed to ```getNumBasesPassingMinimumBaseQuality```

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: :+1:

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: Yup, but now need to store the number of overlapping bases that were clipped, so a ternarary wont work :/

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: This is really a great point.  It would really annoy me if, for example, ```TOTAL_READS``` and ```PF_READS``` did not match the values computed found ```AlignmentSummaryMetrics```.  However, the latter ignores all secondary and supplementary records, with is inappropriate for counting how many bases aligned, so perhaps I need to update that?

nh13:
@tfenne responded to your comments, thank-you.  I will squash the commits once you have approved the final changes.

A few things:
- `AlignmentSummaryMetricsCollector` also ignores supplementary reads when computing the '# of bases' metrics.  I will make a separate pull request for that fix.  `TargetMetricsCollector` now does the right thing.
- I also updated the usage for `CalculateHsMetrics` to say to use `CollectHsMetrics`.

src/java/picard/analysis/directed/CalculateHsMetrics.java
```diff
@@ -24,75 +24,22 @@
...
-                                              final List<SAMReadGroupRecord> samRgRecords,
-                                              final ReferenceSequenceFile refFile,
-                                              final File perTargetCoverage,
-                                              final IntervalList targetIntervals,
-                                              final IntervalList probeIntervals,
-                                              final String probeSetName) {
-        return new HsMetricCollector(accumulationLevels, samRgRecords, refFile, perTargetCoverage, targetIntervals, probeIntervals, probeSetName);
-    }
-}
+@Deprecated
```
yfarjoun: I think it would be good to override the parameters needed to make the default behavior of CalculateHsMetrics remain unchanged. Is that possible?

src/java/picard/analysis/directed/HsMetrics.java
```diff
@@ -107,15 +110,26 @@
     /** The fold by which the baited region has been amplified above genomic background. */
     public double FOLD_ENRICHMENT;
 
-    /** The number of targets that did not reach coverage=2 over any base. */
+    /** The number of targets that did not reach coverage=1 over any base. */
```
yfarjoun: fraction?

src/java/picard/analysis/directed/HsMetrics.java
```diff
@@ -107,15 +110,26 @@
...
     public double ZERO_CVG_TARGETS_PCT;
 
+    /** The fraction of aligned bases that were filtered out because they were in reads with low mapping quality (default is < 0). */
```
yfarjoun: I think that the "default is < 0 "comment is confusing. the metric file should have a header with the parameters used, so perhaps you can refer to that instead of stating the default which may be irrelevant. similarly below.

src/java/picard/analysis/directed/HsMetrics.java
```diff
@@ -107,15 +110,26 @@
...
+    public double PCT_EXC_OVERLAP;
+
     /**
      * The fold over-coverage necessary to raise 80% of bases in "non-zero-cvg" targets to
      * the mean coverage level in those targets.
      */
     public double FOLD_80_BASE_PENALTY;
 
+    /** The percentage of ALL target bases achieving 1X or greater coverage. */
+    public double PCT_TARGET_BASES_1X;
```
yfarjoun: All-> all. 

same below.

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -109,16 +118,32 @@
...
 
     //Converts a targetMetric into a more specific metric of METRIC_TYPE
     public abstract METRIC_TYPE convertMetric(final TargetMetrics targetMetrics);
 
+
+    /**
+     * In the case of ignoring bases in overlapping reads from the same template,
+     * we can choose whether to modify the existing record, or modify a clone.  The
+     * latter may significantly slow down the performance, but also has no side effects.
+     * @param value the boolean value to set.
```
yfarjoun: could you expand here on the side effects that would happen if this is false?

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -301,17 +374,113 @@ public void setBaitSetName(final String name) {
...
+
+        private SAMRecord clipOverlappingBases(final SAMRecord rec, final int numOverlappingBasesToClip) {
+            // NB: ignores how to handle supplemental records when present for both ends by just using the mate information in the record.
+
+            if (!clipOverlappingReads || numOverlappingBasesToClip <= 0) return rec;
+
+            try {
+                // watch out for when the second read overlaps all of the first read
+                if (rec.getMateAlignmentStart() <= rec.getAlignmentStart()) { // make it unmapped
+                    if (noSideEffects) {
```
yfarjoun: factor out the cloning logic to the beginning of this function to avoid much code duplication below

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -301,17 +374,113 @@ public void setBaitSetName(final String name) {
...
+                        final SAMRecord recClone = ((SAMRecord)rec.clone());
+                        recClone.setReadUnmappedFlag(true);
+                        return recClone;
+                    }
+                    else rec.setReadUnmappedFlag(true);
+                    return rec;
+                }
+
+                // clip it, clip it good
+                final Cigar cigar = CigarUtil.addSoftClippedBasesToEndsOfCigar(
```
yfarjoun: I don't understand how this does what you want. don't you want to clip bases that are in the read?

This will turn 101M to 101M30S (if you ask to clip 30 bases) don't you want this to turn into 71M30S? 

Confused.

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -301,17 +374,113 @@ public void setBaitSetName(final String name) {
...
+            }
+        }
+
         /** Adds information about an individual SAMRecord to the statistics. */
-        public void acceptRecord(final SAMRecord rec) {
-            // Just plain avoid records that are marked as not-primary
-            if (rec.isSecondaryOrSupplementary()) return;
+        public void acceptRecord(final SAMRecord record) {
+            /*
+            Count metrics related to # of base and reads. Consider supplemental alignments for base counting but not record counting.
```
yfarjoun: I understand (kindof) counting secondary alignments for base information, but it seems that counting supplemental alignments leads to double counting. Is this already done in practice or are you changing something here?

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -301,17 +374,113 @@ public void setBaitSetName(final String name) {
...
+            Do this counting *prior* to applying filters to ensure we match other metrics' computation for these values, such as AlignmentSummaryMetrics.
+            */
+            if (!record.getNotPrimaryAlignmentFlag()) { // ignore secondary alignments for both # of read and # of base metrics
+                // # of bases
+                if (!record.getReadFailsVendorQualityCheckFlag()) { // only reads that pass vendor's filters
+                    this.metrics.PF_BASES += record.getReadLength();
+                    if (!record.getDuplicateReadFlag() && !record.getReadUnmappedFlag()) { // ignore duplicates and unmapped reads
+                        for (final AlignmentBlock block : record.getAlignmentBlocks()) {
+                            this.metrics.PF_UQ_BASES_ALIGNED += block.getLength();
```
yfarjoun: shouldn't you only consider some types of blocks for this? Presumably you should only consider X=MI operators, not DHSNP, right?

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -365,12 +527,12 @@ public void acceptRecord(final SAMRecord rec) {
 
                     for (final AlignmentBlock block : rec.getAlignmentBlocks()) {
                         final int end = CoordMath.getEnd(block.getReferenceStart(), block.getLength());
-                        for (int pos=block.getReferenceStart(); pos<=end; ++ pos) {
-                            if (pos >= target.getStart() && pos <= target.getEnd()) {
-                                ++this.metrics.ON_TARGET_BASES;
-                                if (mappedInPair) ++this.metrics.ON_TARGET_FROM_PAIR_BASES;
-                                coverage.addBase(pos - target.getStart());
-                            }
+                        for (int pos=block.getReferenceStart(), readPos = block.getReadStart();
```
yfarjoun: same comment as above here, seems strange to include deletions and softclips as bases that are "on target", no?

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: I'll try to rephrase in my next commit, but if we pass `false` to `setNoSideEffects`, then it implies we do not want any side-effects, so there really isn't anything to expand upon other than "there will be none".

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -459,51 +643,15 @@ private void calculateTargetCoverageMetrics() {
...
-                            if (depth >= 20) {
-                                ++targetBases20x;
-                                if (depth >=30) {
-                                    ++targetBases30x;
-	                                if (depth >=40) {
-		                                ++targetBases40x;
-		                                if (depth >=50) {
-			                                ++targetBases50x;
-			                                if (depth >=100) {
-				                                ++targetBases100x;
```
yfarjoun: Thanks for removing this blight. I alway hated it.

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -425,32 +590,51 @@ private void calculateTargetCoverageMetrics() {
...
+
+            this.metrics.MEAN_TARGET_COVERAGE = totalCoverage / this.metrics.TARGET_TERRITORY;
+            this.metrics.MEDIAN_TARGET_COVERAGE =
+                    (0 == (depths.length % 2)) ? ((depths[(depths.length/2) - 1] + depths[(depths.length/2)]) / 2.0) : depths[(depths.length-1)/2];
 
-            // Sort the array (ASCENDING) and then find the base the coverage value that lies at the 80%
+
+            // get the number of bases at zero coverage, and subtract that from the total coverage, then divide that by the total coverage
+            // to get the percent of target bases at 1x coverage or greater.
+            this.metrics.PCT_TARGET_BASES_1X = (this.metrics.TARGET_TERRITORY - numTargetBasesAtZeroCoverage) / this.metrics.TARGET_TERRITORY;
```
yfarjoun: aren't you overwriting this number below?

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: You are not confused, in fact you are correct, great catch.  Much embarrassment on this end.

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -425,32 +590,51 @@ private void calculateTargetCoverageMetrics() {
...
+
+            // find the base the coverage value that lies at the 80%
             // line, which is actually at 20% into the array now
-            Arrays.sort(depths);
-            // Note.  basesConsidered can be between 0 and depths.length inclusive.  indexOf80thPercentile will be -1 in the latter case
-            final int indexOf80thPercentile = Math.max((depths.length - 1 - basesConsidered) + (int) (basesConsidered * 0.2), 0);
+            final int indexOf80thPercentile = Math.max((int) (this.metrics.TARGET_TERRITORY * 0.2 ) - 1, 0);
```
yfarjoun: Given that you didn't include the depth from all the !hasCoverage targets in this calculation (nor do you want to) I think you should use depthIndex rather than TARGET_TERRITORY for this. Am I missing something?

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -724,32 +873,46 @@ public String toString() {
...
-    /** The percentage of ALL target bases acheiving 2X or greater coverage. */
+    /** The percentage of ALL target bases achieving 1X or greater coverage. */
+    public double PCT_TARGET_BASES_1X;
+    /** The percentage of ALL target bases achieving 2X or greater coverage. */
     public double PCT_TARGET_BASES_2X;
-    /** The percentage of ALL target bases acheiving 10X or greater coverage. */
+    /** The percentage of ALL target bases achieving 10X or greater coverage. */
     public double PCT_TARGET_BASES_10X;
-    /** The percentage of ALL target bases acheiving 20X or greater coverage. */
+    /** The percentage of ALL target bases achieving 20X or greater coverage. */
```
yfarjoun: All->all

src/java/picard/filter/CountingPairedFilter.java
```diff
@@ -0,0 +1,9 @@
+package picard.filter;
```
yfarjoun: license

src/java/picard/filter/CountingMapQFilter.java
```diff
@@ -0,0 +1,13 @@
+package picard.filter;
```
yfarjoun: license

src/java/picard/filter/CountingFilter.java
```diff
@@ -0,0 +1,43 @@
+package picard.filter;
```
yfarjoun: license

src/java/picard/filter/CountingDuplicateFilter.java
```diff
@@ -0,0 +1,9 @@
+package picard.filter;
```
yfarjoun: license

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: Re-read the SAM spec 1.2: "Chimeric alignment" and "Multiple mapping".  Please verify my interpretation matches yours.

Supplementary alignments for a read are a set of linear alignments that do not have large overlaps.  Notice they *could* have overlaps, but likely they do not.  In the case of overlaps we are double-counting, but otherwise the prior implementation was under-counting.

Secondary alignments are when you want to represent a sub-optimal alignment, for example a less likely alignment to another location, say a ortholog or paralog.  These should neither be counted for counting reads nor counting bases.

src/java/picard/filter/CountingPairedFilter.java
```diff
@@ -0,0 +1,9 @@
...
+
+import htsjdk.samtools.SAMRecord;
+
+/** Counting filter that discards reads that are unpaired in sequencing and paired reads who's mates are not mapped. */
```
yfarjoun: whose

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: See the definition of an alignment block.  It is a stretch of matching or mismatching bases.  No clipping, padding, skipping, or otherwise allowed.

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: Not included by the definition of AlignmentBlock.

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: I could have had fun and made some type of art out of it.

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: Shouldn't `depthIndex == TARGET_TERRITORY` now?

nh13: Ah, only non-zero depths.

src/java/picard/analysis/directed/TargetMetricsCollector.java
yfarjoun: I even looked (too quickly) at the alignmentBlock code! Thanks for
educating me :-)

src/java/picard/analysis/directed/TargetMetricsCollector.java
yfarjoun: of-course, I had these reverse for a second in my head. must be too late.
I'll continue on a different day. thanks for this important cleanup and
improvements!

src/java/picard/filter/CountingFilter.java
nh13: yeah, yeah, so many of those files have no license.

nh13:
@yfarjoun responded to comments and added a commit (or two).  Only thing I would like you to re-check is the clipping logic that you previously found was incorrect, and a :+1: or :-1:  on the secondary vs. supplementary discussion.

tfenne:
Thinking about this overnight, I think there is one more small problem with the clipping logic.  I think the use of `rec.getAlignmentEnd() - rec.getMateAlignmentStart()  + 1` only works when you assumed that rec has only matching operators in it's CIGAR past the mate alignment start position.  E.g. if the read already has some soft-clipping at the end, I don't think this does the right thing.  Might be better to do:

    if (rec.getMateAlignmentStart() <= rec.getAlignmentEnd()) {
        final int readPos = rec.getReadPositionAtReferencePosition(rec.getMateAlignmentStart(), true);
        if (readPos > 0) numBasesToClip = rec.getReadLength() - readPos + 1;
    }

Either way, it might be good to introduce a handful of testcases for the clipping logic that include reads that have clipping, insertions and deletions at the end and verify that the clipping does what's expected.

yfarjoun:
:+1: on what @tfenne said :-)

yfarjoun:
:+1: on the secondary vs. supplementary.

nh13:
@tfenne see da08f3a.  Does that test what you are talking about?

nh13:
@tfenne after some git commit hickups, can you take a look at 3142843 to see if the test addresses the situation you talked about?

tfenne:
@nh13 - that looks good to me. Thanks!

nh13:
@yfarjoun do the honors of merging?

yfarjoun:
I was specifically asked by @kathleenTibbets to wait with the merge for a
more in depth discussion...given that I think these changes alter the way
CalculateHsMetrics works, I think that's reasonable...

nh13:
@ktibbett could you please review.

nh13:
@yfarjoun I moved the overlapping bases clipping utilities to htsjdk, as they be useful to others and are not specific to the `TargetedMetricsCollector` class: https://github.com/samtools/htsjdk/pull/381

yfarjoun:
good idea.

ktibbett:
@nh13 @tfenne some regression testing is required on our side.  I am starting this now.

yfarjoun:
a bit late in the discussion perhaps, but would it make sense to measure the percent of the reads/bases that were supplementary?

tfenne:
@nh13 A few very minor comments on your latest addition, that once addressed, get a :+1: from me.

@yfarjoun I'm not really sure why you'd want to do that, particularly in the context of HsMetrics.  If anything I would think something like that would make more sense in CollectAlignmentSummaryMetrics, since I think what you're asking is what fraction of read-bases got aligned in a split alignment?

yfarjoun:
yes. I agree. it would be better in CollectAlignmentSummaryMetrics.

nh13:
@tfenne I made the changes you suggested.

@yfarjoun I will leave that to a future improvement.

tfenne:
@ktibbett Any thoughts on when you'll be ready to rule on this PR?  Anything @nh13 or I can do to help with the review?

nh13:
@ktibbett @tfenne I have rebased onto master.

nh13:
@ktibbett rebased onto master

PR #334, opened 2015-11-09
Modified DownsampleSam to to match the changes in DownsamplingIterato…

tfenne:
…r and friends from htsjdk. Notable changes:

  a) Use of multiple downsampling strategies
  b) Now always include secondary and supplemental records with other records for a template
  c) Removed the option to discard unmapped reads (better handled in other programs)

### DIFF
--- a/src/java/picard/sam/DownsampleSam.java
+++ b/src/java/picard/sam/DownsampleSam.java
@@ -1,5 +1,7 @@
 package picard.sam;
 
+import htsjdk.samtools.DownsamplingIteratorFactory;
+import htsjdk.samtools.DownsamplingIteratorFactory.Strategy;
 import htsjdk.samtools.SAMFileWriter;
 import htsjdk.samtools.SAMFileWriterFactory;
 import htsjdk.samtools.SAMRecord;
@@ -17,6 +19,8 @@ import picard.cmdline.StandardOptionDefinitions;
 import picard.cmdline.programgroups.SamOrBam;
 
 import java.io.File;
+import java.text.DecimalFormat;
+import java.text.NumberFormat;
 import java.util.Random;
 
 /**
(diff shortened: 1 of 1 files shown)


tfenne:
Build will fail until htsjdk PR372 (https://github.com/samtools/htsjdk/pull/372) is merged.  Queueing this up so it can be reviewed and pulled at the same time.

tfenne:
@yfarjoun @nh13 Want to take a look?  This is a fairly trivial PR compared to the one in HTSJDK.

src/java/picard/sam/DownsampleSam.java
```diff
@@ -24,10 +28,15 @@
  * of both ends of a pair or neither end of the pair!
  */
 @CommandLineProgramProperties(
-        usage = "Randomly down-sample a SAM or BAM file to retain " +
-                "a random subset of the reads. Mate-pairs are either both kept or both discarded. Reads marked as not primary " +
-                "alignments are all discarded. Each read is given a probability P of being retained - results with the exact " +
-                "same input in the same order and with the same value for RANDOM_SEED will produce the same results.",
+        usage = "Randomly down-sample a SAM or BAM file to retain only a subset of the reads in the file. " +
+                "All reads for a templates are kept or discarded as a unit, with the goal or retaining reads" +
```
yfarjoun: s/or/of/

src/java/picard/sam/DownsampleSam.java
```diff
@@ -24,10 +28,15 @@
  * of both ends of a pair or neither end of the pair!
```
yfarjoun: expand javadoc to state that also extra alignments (not only paired ends) will be included or excluded

yfarjoun:
minor comments, fine be me once tests pass.

src/java/picard/sam/DownsampleSam.java
```diff
@@ -39,13 +48,21 @@
...
+    @Option(shortName="S", doc="The downsampling strategy to use. See usage for discussion.")
+    public Strategy STRATEGY = Strategy.ConstantMemory;
+
     @Option(shortName = "R", doc = "Random seed to use if reproducibilty is desired.  " +
             "Setting to null will cause multiple invocations to produce different results.")
-    public Long RANDOM_SEED = 1L;
+    public Integer RANDOM_SEED = 1;
 
     @Option(shortName = "P", doc = "The probability of keeping any individual read, between 0 and 1.")
     public double PROBABILITY = 1;
```
nh13: Should we enforce that `PROBABILITY < 1` since why run this program otherwise?  Or at least warn?

tfenne: The DownsamplingIteratorFactory enforces that it is >=0 and <=1, and I didn't want to replicate here.

yfarjoun: I think that @nh13 is talking about not running for p=1. I think that while
he is right that it's pointless to run in that case, I think that it should
be up to the user to optimize that away, not for the program to fail on
something it could (uselessly) do.

Y.

tfenne: Yeah. I added a log.warn() when running with P=1 to keep @nh13 happy.  It'll run just fine, but the user will get yelled at.

yfarjoun:
posthumus :+1:

PR #337, opened 2015-11-10
adding cumulative plot to insert size

nh13:


yfarjoun:
:+1:

tfenne:
Very nice.  Can i suggest though that instead of having all the cumulative lines be black, that they be color-coordinated with the original series?  So if there is both a FR and RF distribution shown you can tell which is which by color for the cumulative?  Maybe just use "dark[color]"?

tfenne:
Doh!  @yfarjoun merged while I was typing.

yfarjoun:
should I revert?

nh13:
@yfarjoun I will submit a patch.

nh13:
https://github.com/broadinstitute/picard/pull/339

PR #338, opened 2015-11-10
adding a file extension to CollectMultipleMetrics is now possible

nh13:
With other metric tools, we can specify the output (and thus file extension) directly.  With `CollectMultipleMetrics` we cannot, but in some pipelines specifying the extension would be very handy.  For example, Illumina's BaseSpace doesn't support creating output files with extensions greater than 20 letters.

### DIFF
--- a/src/java/picard/analysis/CollectMultipleMetrics.java
+++ b/src/java/picard/analysis/CollectMultipleMetrics.java
@@ -255,6 +255,9 @@ public class CollectMultipleMetrics extends CommandLineProgram {
     @Option(shortName="LEVEL", doc="The level(s) at which to accumulate metrics.")
     public Set<MetricAccumulationLevel> METRIC_ACCUMULATION_LEVEL = new HashSet<MetricAccumulationLevel>(accumLevelDefault);
 
+    @Option(shortName = "EXT", doc="Append the given file extension to all metric file names (ex. OUTPUT.insert_size_metrics.EXT). None if null", optional=true)
+    public String FILE_EXTENSION = null;
+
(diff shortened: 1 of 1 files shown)


src/java/picard/analysis/CollectMultipleMetrics.java
```diff
@@ -255,6 +255,9 @@ public SinglePassSamProgram makeInstance(final String outbase, final File input,
     @Option(shortName="LEVEL", doc="The level(s) at which to accumulate metrics.")
     public Set<MetricAccumulationLevel> METRIC_ACCUMULATION_LEVEL = new HashSet<MetricAccumulationLevel>(accumLevelDefault);
 
+    @Option(shortName = "EXT", doc="The file extension. None if null", optional=true)
```
tfenne: Would be nice to explain the use of this a bit more - that the extension will be added to all filenames generated, not replace existing extensions.  I.e. you get OUTPUT.insert_size_metrics.EXT, not OUTPUT.EXT.

tfenne:
One little comment, otherwise :clap:

nh13:
@ktibbett @tfenne this should have no effect on existing users who don't use this option, so I see no danger in merging.

tfenne:
Agreed @nh13.

ktibbett:
@nh13 I can't seem to make a comment on the code itself, but I don't understand why the first character is being trimmed from the extension in the else branch (where it's not ".")....  [line 323]

nh13:
@ktibbett I can't for the life of me remember why, so I removed it, and rebased on to master while I was at it.

PR #340, opened 2015-11-11
Plot all insert size orientations in CollectInsertMetrics when MIN_PCT=0

nh13:
The minimum percentage for a paired orientation should be greater than or equal, not greater than.  We have to be careful if there are no inserts, which `testHistogramWidthIsSetProperly` captures in `CollectInsertSizeMetricsTest`.

### DIFF
--- a/src/java/picard/analysis/directed/InsertSizeMetricsCollector.java
+++ b/src/java/picard/analysis/directed/InsertSizeMetricsCollector.java
@@ -6,7 +6,6 @@ import htsjdk.samtools.SamPairUtil;
 import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.reference.ReferenceSequence;
 import htsjdk.samtools.util.Histogram;
-import picard.PicardException;
 import picard.analysis.InsertSizeMetrics;
 import picard.analysis.MetricAccumulationLevel;
 import picard.metrics.MultiLevelCollector;
@@ -36,9 +35,9 @@ public class InsertSizeMetricsCollector extends MultiLevelCollector<InsertSizeMe
     private final Integer histogramWidth;
 
     public InsertSizeMetricsCollector(final Set<MetricAccumulationLevel> accumulationLevels, final List<SAMReadGroupRecord> samRgRecords,
-                                      final double minimumPct, final Integer HistogramWidth, final double deviations) {
+                                      final double minimumPct, final Integer histogramWidth, final double deviations) {
         this.minimumPct = minimumPct;
-        this.histogramWidth = HistogramWidth;
+        this.histogramWidth = histogramWidth;
         this.deviations = deviations;
         setup(accumulationLevels, samRgRecords);
     }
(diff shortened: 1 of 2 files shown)


nh13:
@bradtaylor just checking in on this PR.

nh13:
@bradtaylor @ktibbett @tfenne any chance this can get reviewed?

src/java/picard/analysis/directed/InsertSizeMetricsCollector.java
```diff
@@ -121,7 +133,13 @@ public void addMetricsToFile(final MetricsFile<InsertSizeMetrics,Integer> file)
                 final double total = Histogram.getCount();
```
tfenne: Would you mind fixing the `Histogram` variable name to `histogram` while you're in here please?

src/java/picard/analysis/directed/InsertSizeMetricsCollector.java
```diff
@@ -121,7 +133,13 @@ public void addMetricsToFile(final MetricsFile<InsertSizeMetrics,Integer> file)
...
 
                 // Only include a category if it has a sufficient percentage of the data in it
-                if( total > totalInserts * minimumPct ) {
+                if( total >= totalInserts * minimumPct ) {
+                    // in the case that minimumPct == 0 and total == 0, just make an empty histogram with zero values for the entire range
+                    if (Histogram.isEmpty()) {
```
tfenne: Can you comment in the code on why this is necessary?  I assume it's to make the code below for the `WIDTH_OF...` not blow up, but I'm not sure.

src/java/picard/analysis/directed/InsertSizeMetricsCollector.java
```diff
@@ -113,7 +115,17 @@ public double getTotalInserts() {
         }
 
         public void addMetricsToFile(final MetricsFile<InsertSizeMetrics,Integer> file) {
-            for (final Histogram<Integer> h : this.Histograms.values()) totalInserts += h.getCount();
+            // get the number of inserts, and the maximum and minimum keys across, across all orientations
+            int minKey = Integer.MAX_VALUE, maxKey = Integer.MIN_VALUE;
+            for (final Histogram<Integer> h : this.Histograms.values()) {
+                totalInserts += h.getCount();
+                if (!h.isEmpty()) {
+                    int curMin = Collections.min(h.keySet()), curMax = Collections.max(h.keySet());
```
tfenne: I think this would be clearer written as:
```java
minKey = Math.min(minKey, Collections.min(h.keySet())
maxKey = Math.max(maxKey, Collections.max(h.keySet())
```

tfenne:
@nh13 I'm not sure keen on the filling empty histograms with 0s to avoid code blowing up below.  Is that really the best way to do it? Or could lines 148-190 just be wrapped in a `if (!histogram.isEmpty()) { ... }`?

src/java/picard/analysis/directed/InsertSizeMetricsCollector.java
```diff
@@ -121,7 +133,13 @@ public void addMetricsToFile(final MetricsFile<InsertSizeMetrics,Integer> file)
                 final double total = Histogram.getCount();
 
                 // Only include a category if it has a sufficient percentage of the data in it
-                if( total > totalInserts * minimumPct ) {
+                if( total >= totalInserts * minimumPct ) {
+                    // in the case that minimumPct == 0 and total == 0, just make an empty histogram with zero values for the entire range
+                    if (Histogram.isEmpty()) {
+                        for (int key = minKey; key <= maxKey; key++) Histogram.increment(key, 0);
+                        if (Histogram.isEmpty()) throw new PicardException("BUG: minKey=" + minKey + " maxKey=" + maxKey);
```
mattsooknah: Is this to catch the case where ```minKey > maxKey```? If so, the error message should be more specific.

mattsooknah:
@nh13 Echoing Tim's comments above, plus one of my own.

nh13:
@tfenne @mattsooknah thanks for reviewing.  It is nasty that calling `getMin` or `getMax` may cause a `NullPointerException` to be thrown (see `TreeMap`'s `lastEntry` method).  I have tried to work around that given your suggestions.  I also rebased onto master to get it working with the latest htsjdk.

mattsooknah:
:+1:

PR #341, opened 2015-11-16
fix a bug in clipping utility with pairs that have ends of differing

nh13:
lengths not reported as matching.  We should track the number of bases
clipped, not where relative to the start of the read, as the latter may
differ if the ends are of a different length for a pair.  Tests
included.


src/java/picard/illumina/MarkIlluminaAdapters.java
```diff
@@ -224,8 +226,8 @@ protected int doWork() {
                     progress.record(r);
                     if (out != null) out.addAlignment(r);
 
-                    final Integer clip = rec.getIntegerAttribute(ReservedTagConstants.XT);
-                    if (clip != null) histo.increment(rec.getReadLength() - clip + 1);
+                    final Integer clip = r.getIntegerAttribute(ReservedTagConstants.XT);
+                    if (clip != null) histo.increment(r.getReadLength() - clip + 1);
```
nh13: This doesn't matter so much in the cases where the reads are of the same length, since the `XT` tags will be set to the same value for both ends.  This should only affect the cases where one end is longer than the other, which happens infrequently by design for Illumina data.

src/java/picard/util/IlluminaUtil.java
```diff
@@ -186,6 +186,40 @@ private IlluminaAdapterPair(final String fivePrime, final String threePrime) {
         public String getName() { return this.name(); }
     }
 
+    public static class CustomAdapterPair implements AdapterPair {
```
nh13: Moved this here so `ClippingUtilityTest` could access it.

nh13:
@tfenne @yfarjoun @ktibbett I would be grateful for a review.

src/java/picard/illumina/MarkIlluminaAdapters.java
tfenne: Wow, I'm pretty sure that was just a bug.  Thanks for catching/fixing this.

src/java/picard/util/ClippingUtility.java
```diff
@@ -170,12 +170,15 @@ public static AdapterPair adapterTrimIlluminaPairedReads(final SAMRecord read1,
                     getReadBases(read1), adapterPair.get3PrimeAdapterBytes(), minMatchBases, maxErrorRate);
             final int index2 = findIndexOfClipSequence(
                     getReadBases(read2), adapterPair.get5PrimeAdapterBytesInReadOrder(), minMatchBases, maxErrorRate);
+            final int numBasesClipped1 = (NO_MATCH == index1) ? 0 : read1.getReadLength() - index1;
+            final int numBasesClipped2 = (NO_MATCH == index2) ? 0 : read2.getReadLength() - index2;
 
-            if (index1 == index2) {
+            // Be careful to count the number of bases clipped, rather than where in the read, since the read lengths may differ.
+            if (numBasesClipped1 == numBasesClipped2) {
```
tfenne: I think this change is wrong.  The point here is that if you have two reads that are reading towards each other's start point, through the template, that they should both encounter adapter sequence after the same number of bases (i.e. the template or insert length), and so finding the adapter after the same number of bases in both reads is a very strong signal that you've got it right.

src/java/picard/util/ClippingUtility.java
```diff
@@ -201,28 +204,26 @@ public static AdapterPair adapterTrimIlluminaPairedReads(final SAMRecord read1,
...
+        final int numBasesClipped = numBasesClipped1 == 0 ? numBasesClipped2 : numBasesClipped1;
 
         // If it still matches with a stricter minimum matched bases, then
         // clip both reads
-        if (matchedRead.getReadLength() - matchedIndex >= stricterMinMatchBases) {
-            if (read1.getReadBases().length > matchedIndex) {
-                read1.setAttribute(ReservedTagConstants.XT, matchedIndex + 1);
+        if (stricterMinMatchBases <= numBasesClipped) {
+            if (read1.getReadLength() > numBasesClipped) {
+                read1.setAttribute(ReservedTagConstants.XT, read1.getReadLength() - numBasesClipped + 1);
```
tfenne: I think this is problematic too.  The addition of logic to ensure that each read is long enough that the clip point is within the read is a good idea.  But this changes the clip point to be "number of clipped bases from the end of the read", vs. being at the same position/offset in each read, and I think that's wrong.  We leverage the knowledge that in Illumina sequencing we should see the adapter sequence after the same number of read bases - which when reads are of fixed length, will be the same distance from the end.  But when the reads are of different lengths, it's the distance from the start that is important, not the end.

tfenne:
@nh13 I reviewed the code, but not the tests as I think there are problems with the code that need to be fixed first.

nh13:
@tfenne bump

tfenne:
Looks good.  Merging.

PR #353, opened 2015-11-30
Turned the 250bp hard-coded "near bait" distance into a parameter tha…

tfenne:
…t can be set in both CollectHsMetrics and CollectTargetedPcrMetrics.


tfenne:
@yfarjoun up for a quick review?  This is just taking a hardcoded constant and turning it into a parameter.  Default value is the same as the prior constant, so default behaviour is unchanged.

yfarjoun:
:+1:

PR #357, opened 2015-12-08
adding two command line options to EstimateLibraryComplexity

nh13:
1. `MAX_READ_LENGTH` to limit the number of bases considered during
calculation of the difference rate.
2. `MIN_GROUP_COUNT` to allow the user to specify the minimum count of
duplicates of a given size, which is useful for testing, as well as
seeing the effect of technical artifacts.


nh13:
Note that the default values make the behavior the same, so users should not be affected by this change.   Also, added tests for the new options and verify the old behavior with the defaults.  

The reason for `MAX_READ_LENGTH` is when using this tool on noisy long reads (even as short as a few hundred bases), such that the error rate becomes too large, and so we go above the `MAX_DIFF_RATE` value.  The `MIN_GROUP_COUNT` option was added for testing, but could also be useful for custom QC analysis, where we want to include technical artifacts in the metric calculations (of library size).


@ktibbett @tfenne @yfarjoun thanks for the review!

src/java/picard/sam/markduplicates/EstimateLibraryComplexity.java
```diff
@@ -136,8 +136,30 @@
     @Option(doc = "Read two barcode SAM tag (ex. BX for 10X Genomics)", optional = true)
     public String READ_TWO_BARCODE_TAG = null;
 
+    @Option(doc = "The maximum number of bases to consider when comparing reads.", optional = true)
```
mattsooknah: Add a note indicating that 0 means "no limit"

src/java/picard/sam/markduplicates/EstimateLibraryComplexity.java
```diff
@@ -670,8 +693,10 @@ boolean passesQualityCheck(final byte[] bases, final byte[] quals, final int see
             if (SequenceUtil.isNoCall(bases[i])) return false;
         }
 
+        final int maxReadLength = (MAX_READ_LENGTH <= 0) ? Integer.MAX_VALUE : MAX_READ_LENGTH;
+        final int readLength = Math.min(bases.length, maxReadLength);
         int total = 0;
-        for (final byte b : quals) total += b;
-        return total / quals.length >= minQuality;
+        for (int i = 0; i < readLength; i++) total += bases[i];
```
mattsooknah: Shouldn't this be summing the ```quals```, not the ```bases``` ?

mattsooknah:
Back to you @nh13

nh13:
@mattsooknah thanks, rebased and update the code based on those two comments.  Please feel free to merge when you are happy.

PR #360, opened 2015-12-09
fixup read name parsing for CASAVA 1.8+ which have seven fields versus

nh13:
the previous five.  Broken since 30c6d97272e2d8ead674666dd1c0f551785ef2b6.

@yfarjoun I could really use a detailed eye on this.


nh13:
Previously, `getRapidDefaultReadNameRegexSplit` would try to parse only the first five tokens and so for reads with seven fields (Casava 1.8+) we would never be getting the correct tile, x, and y locations.

nh13:
Also, the unit test failures are not related to my fix.

src/java/picard/sam/util/ReadNameParsingUtils.java
```diff
@@ -29,24 +29,29 @@
 public class ReadNameParsingUtils {
 
     /**
-     * Single pass method to parse the read name for the default regex.  This will only insert the 2nd to the 4th
-     * tokens (inclusive).  It will also stop after the fifth token has been successfully parsed.
+     * Single pass method to parse the read name for the default regex.  Examines the last three fields as split by the delimiter.
      */
     public static int getRapidDefaultReadNameRegexSplit(final String readName, final char delim, final int[] tokens) {
```
tfenne: Perhaps this ought to be renamed to just "getLastThreeFIelds" ?  It's not clear that it's all that related to the default regex any more, other than that in the specific case of Illumina read names, it will mostly function the same way.

src/java/picard/sam/util/ReadNameParsingUtils.java
```diff
@@ -29,24 +29,29 @@
 public class ReadNameParsingUtils {
 
     /**
-     * Single pass method to parse the read name for the default regex.  This will only insert the 2nd to the 4th
-     * tokens (inclusive).  It will also stop after the fifth token has been successfully parsed.
+     * Single pass method to parse the read name for the default regex.  Examines the last three fields as split by the delimiter.
```
tfenne: Also, it's no longer single-pass.

src/java/picard/sam/util/ReadNameParsingUtils.java
```diff
@@ -29,24 +29,29 @@
...
      */
     public static int getRapidDefaultReadNameRegexSplit(final String readName, final char delim, final int[] tokens) {
         int tokensIdx = 0;
         int prevIdx = 0;
+        int numFields = 1;
+        for (int i = 0; i < readName.length(); i++) {
```
tfenne: This is all getting very complicated.  I wonder how much slower it would be to use either the non-regex `StringUtil.split(readName, fieldArray, ":")` and then use `rapidParseToInt` on the last three fields.

yfarjoun:
I doubt that the failures are unrelated as the position-based downsampling use the coordinate parsing just like optical duplicates....

nh13:
@yfarjoun I'll try to fix that up in a day or so.

src/java/picard/sam/util/ReadNameParsingUtils.java
nh13: challenge excepted

nh13:
@tfenne @yfarjoun passes tests locally, I re-organized some code since I didn't realize we now had two versions of `PhysicalLocation`, one with the old "overflow" behavior, and one without.

src/tests/java/picard/sam/markduplicates/util/OpticalDuplicateFinderTest.java
```diff
@@ -17,53 +22,68 @@
...
-            doTestGetRapidDefaultReadNameRegexSplit((i <= 5) ? i : 5, opticalDuplicateFinder);
+            doTestGetRapidDefaultReadNameRegexSplit(i);
         }
     }
 
     // NB: these tests fails due to overflow in the duplicate finder test.  This has been the behavior previously, so keep it for now.
-    @Test(dataProvider = "testParseReadNameDataProvider", enabled = false)
-    public void testParseReadName(final String readName, final int tile, final int x, final int y) {
+    @Test(dataProvider = "testParseReadNameDataProvider", enabled = true)
+    public void testParseReadNameOverflow(final String readName, final int tile, final int x, final int y) {
```
nh13: tests that we overflow with shorts

src/tests/java/picard/sam/markduplicates/util/OpticalDuplicateFinderTest.java
```diff
@@ -17,53 +22,68 @@
...
         Assert.assertEquals(loc.getTile(), tile);
-        Assert.assertEquals(loc.getX(), x);
-        Assert.assertEquals(loc.getY(), y);
+        Assert.assertEquals(loc.getX(), (short)x); // casting to short for the overflow
+        Assert.assertEquals(loc.getY(), (short)y); // casting to short for the overflow
+    }
+
+    // NB: these tests the case wher we do not overflow in the duplicate finder test.
+    @Test(dataProvider = "testParseReadNameDataProvider", enabled = true)
+    public void testParseReadNameOK(final String readName, final int tile, final int x, final int y) {
```
nh13: tests that we don't overflow with ints

src/tests/java/picard/sam/markduplicates/util/OpticalDuplicateFinderTest.java
```diff
@@ -73,4 +93,26 @@ public void testParseReadName(final String readName, final int tile, final int x
                 {"RUNID:7:1203:2884:16834", 1203, 2884, 16834}
         };
     }
+
+    @Test
+    public void testVeryLongReadNames() {
```
nh13: tests 7 fields

yfarjoun:
At the time when I wanted a physicalLocation that didn't overflow, I added the int version. the logic was (if my memory serves me) to keep the short version intact so that I do not break it....at this point, given all the changes, I'm not sure what the utility of keeping the short version around. I remember some optimization was performed that I didn't want to disturb, but I would assume all that went away with the changes you made here...should we just do away with the short version and only have the int version? what do others think? @tfenne @ktibbett

nh13:
@yfarjoun the output of `Mark Duplicates` would change, and so you would need to weight the cost and benefit for your own internal deployment.  My thought was not to change behavior.

Nonetheless, this is a fairly urgent fix since the library size estimate is 2x higher without it.  Can we table this discussion about short vs. int and create a new issue?  That way this fix can go in ASAP.

yfarjoun:
OK. sure.

src/java/picard/sam/markduplicates/util/AbstractMarkDuplicatesCommandLineProgram.java
```diff
@@ -274,7 +275,7 @@ public static void trackOpticalDuplicates(List<? extends ReadEnds> ends,
      * optical duplicate detection, we do not consider them duplicates if one read as FR ann the other RF when we order orientation by the
```
yfarjoun: s/ann/and/ in the comment...since you were here.

src/java/picard/sam/util/PhysicalLocationShort.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package picard.sam.util;
+
+
+/**
+ * Small class that provides access to the physical location information about a cluster.
+ * All values should be defaulted to -1 if unavailable.  Tile should only allow
+ * non-zero positive integers, x and y coordinates must be non-negative.
+ * This is different from OpticalDuplicateFinder.PhysicalLocationInt in that the x and y positions are shorts, not ints
```
yfarjoun: s/OpticalDuplicateFinder.// in comment

src/java/picard/sam/util/ReadNameParser.java
```diff
@@ -0,0 +1,144 @@
+package picard.sam.util;
+
+import htsjdk.samtools.util.Log;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Small class that provides access to the physical location information about a cluster.
```
yfarjoun: no longer a small class...

src/java/picard/sam/util/ReadNameParser.java
```diff
@@ -0,0 +1,144 @@
...
+                            this.readNameRegex, readName));
+                    warnedAboutRegexNotMatching = true;
+                }
+                return false;
+            }
+        }
+    }
+
+    /**
+     * Single pass method to parse the read name for the default regex.  Examines the last three fields as split by the delimiter.
```
yfarjoun: Could you add a javadoc to this as it is public?

yfarjoun: especially since it does some strange things (like change the input array!)

src/java/picard/sam/util/ReadNameParser.java
```diff
@@ -0,0 +1,144 @@
...
+     * can be used later to determine optical duplication
+     *
+     * @param readName the name of the read/cluster
+     * @param loc the object to add tile/x/y to
+     * @return true if the read name contained the information in parsable form, false otherwise
+     */
+    public boolean addLocationInformation(final String readName, final PhysicalLocation loc) {
+        // Optimized version if using the default read name regex (== used on purpose):
+        if (this.readNameRegex == this.DEFAULT_READ_NAME_REGEX) {
+            final int fields = getLastThreeFields(readName, ':', tmpLocationFields);
```
yfarjoun: I am troubled by the fact that when `this.readNameRegex == this.DEFAULT_READ_NAME_REGEX` we use getLastThreeFields which does NOT mimmic the `DEFAULT_READ_NAME_REGEX`...I find this extremely confusing. 

in particular the tests that you have for testVeryLongReadNames will not pass if you provide a non-interned version of `[a-zA-Z0-9]+:[0-9]:([0-9]+):([0-9]+):([0-9]+).*` as this only allows for 5 fields

getLastThreeFields seems to encapsulate the following regex:
```code
.*[^0-9]|^([0-9]+)[^:]*:([0-9]+)[^:]*:([0-9]+)[^:]*
```
Though my regex-fu is not super, so I might be missing something.

src/tests/java/picard/sam/markduplicates/util/OpticalDuplicateFinderTest.java
yfarjoun: Could you also test that this works with the regexp version, i.e. a non-interned copy of DEFAULT_READ_NAME_REGEX?

src/java/picard/sam/util/ReadNameParser.java
nh13: It is concerning, but that is the current behavior.  I could change the default regex, is this what you are suggesting?  Again, I am trying not to perturb behavior.

yfarjoun: Given that the default regex isn't actually used anywhere, I do not see how it would perturb behaviour. The only way in which it could be used is if someone inputed the same string on the commandline, in that case, with 7 fields, the behavior would be broken, so I think that the regex should represent something that actually works, if not we might as well have it be something like "DEFAULT_THIS_DOES_THE_RIGHT_THING"  :smile:

src/tests/java/picard/sam/markduplicates/util/OpticalDuplicateFinderTest.java
```diff
@@ -17,53 +22,68 @@
...
+        PhysicalLocation loc = new PhysicalLocationShort();
         Assert.assertTrue(opticalDuplicateFinder.addLocationInformation(readName, loc));
         Assert.assertEquals(loc.getTile(), tile);
-        Assert.assertEquals(loc.getX(), x);
-        Assert.assertEquals(loc.getY(), y);
+        Assert.assertEquals(loc.getX(), (short)x); // casting to short for the overflow
+        Assert.assertEquals(loc.getY(), (short)y); // casting to short for the overflow
+    }
+
+    // NB: these tests the case wher we do not overflow in the duplicate finder test.
```
yfarjoun: s/wher/where/

src/tests/java/picard/sam/markduplicates/util/OpticalDuplicateFinderTest.java
```diff
@@ -17,53 +22,68 @@
...
         for (int i = 1; i < 10; i++) {
-            doTestGetRapidDefaultReadNameRegexSplit((i <= 5) ? i : 5, opticalDuplicateFinder);
+            doTestGetRapidDefaultReadNameRegexSplit(i);
         }
     }
 
     // NB: these tests fails due to overflow in the duplicate finder test.  This has been the behavior previously, so keep it for now.
-    @Test(dataProvider = "testParseReadNameDataProvider", enabled = false)
-    public void testParseReadName(final String readName, final int tile, final int x, final int y) {
+    @Test(dataProvider = "testParseReadNameDataProvider", enabled = true)
```
yfarjoun: hurray for enabling tests!

src/java/picard/sam/util/ReadNameParser.java
tfenne: I actually kind of like the idea that the default regex should be instructive, rather than looking like an actual regular expression.  E.g. `<optimized capture of last three ':' separated fields>`.  Then it's really clear that a) if you don't touch it you get an optimized version, and b) it's really much clearer what the default does without starting at symbol soup for 5 minutes.  Since the default is treated symbolically anyway, I don't think that would change behaviour, right?

src/tests/java/picard/sam/markduplicates/util/OpticalDuplicateFinderTest.java
tfenne: I'm going to push back on this ... the default is there really only to show what the code will do.  I think the best answer is to have a symbolic default and call it a day.

src/java/picard/sam/util/ReadNameParser.java
nh13: I really like this idea, I will add this.

yfarjoun: I'm happy either way:
- either a regex that would work if it was used
- or a symbolic string that explains what's happening
(but not the current situation)

nh13:
@yfarjoun @tfenne made a few changes in the last commit.  Once :+1: I will squash and make sure the tests pass for you to be ready to merge.

src/java/picard/sam/util/ReadNameParser.java
yfarjoun: if we change the default to being symbolic, we need the comment to be very clear about how to make a regex that works. in particular we need to explain that the three "groups" will be used to denote tile, x-, and y- coordinates and that the user supplied regex should do that...perhaps provide the old default as an example.

yfarjoun:
please fix the doc for READ_NAME_REGEX in AbstractOpticalDuplicateFinderCommandLineProgram
```code
"For 5 element names, the 3rd, 4th and 5th elements are assumed to be tile, x and y values. " +
"For 7 element names (CASAVA 1.8), the 5th, 6th, and 7th elements are assumed to be tile, x and y values.",
```
This is not correct anymore. The information about the groups is there so you can ignore my comment from before.

yfarjoun:
:+1:

yfarjoun:
@nh13 still waiting for you to respond to the lastest comments and then rebase/squash so we can merge.

nh13:
@yfarjoun the comment in AbstractOpticalDuplicateFinderCommandLineProgram is actually correct, due to the check that we have 5 or 7 fields, no more, no less.

nh13:
@yfarjoun back to you.  I added a few more tests and added docs as you requested.

src/java/picard/sam/util/ReadNameParser.java
```diff
@@ -0,0 +1,177 @@
...
+                    warnedAboutRegexNotMatching = true;
+                }
+                return false;
+            }
+        }
+    }
+
+    /**
+     * Given a string, splits the string by the delimiter, and returns the the last three fields parsed as integers.  Parsing a field
+     * considers only a sequence of digits up until the first non-digit character.  The three values are stored in the passed-in array.
```
yfarjoun: and the return value is...

yfarjoun:
:+1:  right. state the return value and then rebase/squash please. 

thanks!

src/java/picard/sam/util/ReadNameParser.java
```diff
@@ -0,0 +1,177 @@
...
+        else {
+            return numFields;
+        }
+    }
+
+    /**
+     * Very specialized method to rapidly parse a sequence of digits from a String up until the first
+     * non-digit character.
+     */
+    public static int rapidParseInt(final String input) {
```
tfenne: Can I make one more suggestion here?  I think this method should fail if it cannot parse at least one digit from the string.  Looking at how this failed, it will in some instances have been parsing flowcell names and other non-numeric Strings, and happily returning 0.

tfenne:
One last comment from me (sorry) about making the field->int parsing fail faster if something changes again in future. Otherwise :+1: from me.

yfarjoun:
Good catch. please test against this failure mode (and others if not
already done...)

tfenne:
@yfarjoun  - I just added code for that failure mode, and test cases.  Will squash and merge once things pass unless you object.

yfarjoun:
au contraire! please do. :+1:

PR #363, opened 2015-12-21
Performance improvements to CollectSequencingArtifactMetrics to make it faster.

eitanbanks:
DO NOT MERGE YET.  NEEDS SIGNIFICANT IN-PERSON REVIEW WITH YOSSI.

1. 40% of the runtime went to reading and parsing the original qualities from the BAM.
Given that it is only used as a quality cutoff, we can just use the standard (recalibrated)
quals and save this pain.

2. Get rid of the nested maps; there are a known number of possible transitions so use an array.

3. Stop doing String addition in the Transition class; array lookups are faster.

4. Cache the upper-cased string version of the reference; don't create a string for every
base in every read.

### DIFF
--- a/src/java/picard/analysis/artifacts/ArtifactCounter.java
+++ b/src/java/picard/analysis/artifacts/ArtifactCounter.java
@@ -21,10 +21,7 @@ class ArtifactCounter {
     private final String sampleAlias;
     private final String library;
 
-    private final Set<String> fullContexts;
-    private final Map<String, String> leadingContextMap;
-    private final Map<String, String> trailingContextMap;
-    private final Map<String, String> zeroContextMap;
+    private final Map<String, RefContext> contextMap = new HashMap<>();
 
     private final ContextAccumulator fullContextAccumulator;
     private final ContextAccumulator halfContextAccumulator;
@@ -35,38 +32,51 @@ class ArtifactCounter {
     private final List<BaitBiasSummaryMetrics> baitBiasSummaryMetricsList;
     private final List<BaitBiasDetailMetrics> baitBiasDetailMetricsList;
 
+    private final Set<String> leadingContexts = new HashSet<>();
+    private final Set<String> trailingContexts = new HashSet<>();
+
+    // tuple to keep track of the different types of sub-contexts from a given reference context
+    protected final class RefContext {
+        final String ref, leading, trailing, zero;
+
+        public RefContext(final String ref, final String leading, final String trailing, final String zero) {
+            this.ref = ref;
+            this.leading = leading;
+            this.trailing = trailing;
+            this.zero = zero;
+        }
+    }
+
     public ArtifactCounter(final String sampleAlias, final String library, final int contextSize, final boolean expectedTandemReads) {
         this.sampleAlias = sampleAlias;
         this.library = library;
 
         // define the contexts
-        this.fullContexts = new HashSet<String>();
+        final HashSet<String> fullContexts = new HashSet<>();
         for (final byte[] kmer : SequenceUtil.generateAllKmers(2 * contextSize + 1)) {
-            this.fullContexts.add(StringUtil.bytesToString(kmer));
+            fullContexts.add(StringUtil.bytesToString(kmer));
         }
+
+        final Set<String> zeroContexts = new HashSet<>();
 
         // the half contexts specify either leading or trailing bases. the zero context is just the center.
         // NB: we use N to represent a wildcard base, rather than an ambiguous base. It's assumed that all of the input
         // contexts are unambiguous, and that any actual N's in the data have been dealt with elsewhere.
         final String padding = StringUtil.repeatCharNTimes('N', contextSize);
-        this.leadingContextMap = new HashMap<String, String>();
-        this.trailingContextMap = new HashMap<String, String>();
-        this.zeroContextMap = new HashMap<String, String>();
-        for (final String context : this.fullContexts) {
-            final String leading = context.substring(0, contextSize);
-            final String trailing = context.substring(contextSize + 1, context.length());
-            final char center = context.charAt(contextSize);
-            this.leadingContextMap.put(context, leading + center + padding);
-            this.trailingContextMap.put(context, padding + center + trailing);
-            this.zeroContextMap.put(context, padding + center + padding);
+        for (final String context : fullContexts) {
+            final char centralBase = context.charAt(contextSize);
+            final String leading = context.substring(0, contextSize) + centralBase + padding;
+            final String trailing = padding + centralBase + context.substring(contextSize + 1, context.length());
+            final String zero = padding + centralBase + padding;
+            contextMap.put(context, new RefContext(context, leading, trailing, zero));
+
+            leadingContexts.add(leading);
+            trailingContexts.add(trailing);
+            zeroContexts.add(zero);
         }
 
-        // set up the accumulators
-        final Set<String> halfContexts = new HashSet<String>();
-        halfContexts.addAll(leadingContextMap.values());
-        halfContexts.addAll(trailingContextMap.values());
-        final Set<String> zeroContexts = new HashSet<String>();
-        zeroContexts.addAll(zeroContextMap.values());
+        final Set<String> halfContexts = new HashSet<>(leadingContexts);
+        halfContexts.addAll(trailingContexts);
 
         this.fullContextAccumulator = new ContextAccumulator(fullContexts, expectedTandemReads);
         this.halfContextAccumulator = new ContextAccumulator(halfContexts, expectedTandemReads);
@@ -83,11 +93,12 @@ class ArtifactCounter {
      * Add a record to all the accumulators.
      */
     public void countRecord(final String refContext, final char calledBase, final SAMRecord rec) {
-        if (this.fullContexts.contains(refContext)) {
-            this.fullContextAccumulator.countRecord(refContext, calledBase, rec);
-            this.halfContextAccumulator.countRecord(this.leadingContextMap.get(refContext), calledBase, rec);
-            this.halfContextAccumulator.countRecord(this.trailingContextMap.get(refContext), calledBase, rec);
-            this.zeroContextAccumulator.countRecord(this.zeroContextMap.get(refContext), calledBase, rec);
+        if (this.contextMap.containsKey(refContext)) {
+            final RefContext contexts = contextMap.get(refContext);
+            this.fullContextAccumulator.countRecord(contexts.ref, calledBase, rec);
+            this.halfContextAccumulator.countRecord(contexts.leading, calledBase, rec);
+            this.halfContextAccumulator.countRecord(contexts.trailing, calledBase, rec);
+            this.zeroContextAccumulator.countRecord(contexts.zero, calledBase, rec);
         }
     }
 
@@ -147,8 +158,8 @@ class ArtifactCounter {
                 if (!metrics.preAdapterMetrics.CONTEXT.equals(metrics.baitBiasMetrics.CONTEXT)) {
                     throw new PicardException("Input detail metrics are not matched up properly - contexts differ.");
                 }
-                final boolean isLeading = this.leadingContextMap.containsValue(metrics.preAdapterMetrics.CONTEXT);
-                final boolean isTrailing = this.trailingContextMap.containsValue(metrics.preAdapterMetrics.CONTEXT);
+                final boolean isLeading = leadingContexts.contains(metrics.preAdapterMetrics.CONTEXT);
+                final boolean isTrailing = trailingContexts.contains(metrics.preAdapterMetrics.CONTEXT);
                 // if the original contextSize is 0, there's no difference between leading and trailing, so add it to both
                 if (isLeading) leadingMetricsForTransition.add(metrics);
                 if (isTrailing) trailingMetricsForTransition.add(metrics);
--- a/src/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
+++ b/src/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
@@ -108,6 +108,9 @@ public class CollectSequencingArtifactMetrics extends SinglePassSamProgram {
     private DbSnpBitSetUtil dbSnpMask;
     private SamRecordFilter recordFilter;
 
+    private String currentRefString = null;
+    private int currentRefIndex = -1;
+
     private final Set<String> samples = new HashSet<String>();
     private final Set<String> libraries = new HashSet<String>();
     private final Map<String, ArtifactCounter> artifactCounters = new HashMap<String, ArtifactCounter>();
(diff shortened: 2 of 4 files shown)


eitanbanks:
@ktibbett @yfarjoun This is the best I could do for now.  We go from consistently hitting 80 sec per 1M reads in the previous version to 15 sec now (so ~5 hours over the whole genome instead of 28 hours).

Yossi, I'd like you to really look this over and make sure I haven't changed the behavior at all.  It will likely mean having to sit side by side to review, so feel free to schedule something on my calendar (and bring coffee!).

src/java/picard/analysis/artifacts/ArtifactCounter.java
```diff
@@ -35,38 +32,50 @@
     private final List<BaitBiasSummaryMetrics> baitBiasSummaryMetricsList;
     private final List<BaitBiasDetailMetrics> baitBiasDetailMetricsList;
 
+    private final Set<String> leadingContexts = new HashSet<>();
+    private final Set<String> trailingContexts = new HashSet<>();
+
+    protected final class RefContext {
```
yfarjoun: comment: tuple to...

src/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
```diff
@@ -86,9 +86,6 @@
             "i.e. they're expected to face the same direction.")
     public boolean TANDEM_READS = false;
 
-    @Option(doc = "When available, use original quality scores for filtering.")
```
yfarjoun: instead of deleting default false and add comment in doc explaining the increase in run-time.

src/java/picard/analysis/artifacts/Transition.java
```diff
@@ -51,4 +46,46 @@ public Transition matchingRef() {
...
+        Arrays.fill(baseIndexMap, -1);
+        baseIndexMap['A'] = Base.A.ordinal();
+        baseIndexMap['C'] = Base.C.ordinal();
+        baseIndexMap['G'] = Base.G.ordinal();
+        baseIndexMap['T'] = Base.T.ordinal();
+    }
+
+    static private final Transition[][] transitionIndexMap = new Transition[4][4];
+    static {
+        transitionIndexMap[0][0] = AtoA;
```
yfarjoun: might as well use the ordinals here too...

src/java/picard/analysis/artifacts/Transition.java
```diff
@@ -51,4 +46,46 @@ public Transition matchingRef() {
...
+    static protected final int[] baseIndexMap = new int[256];
+    static {
+        Arrays.fill(baseIndexMap, -1);
+        baseIndexMap['A'] = Base.A.ordinal();
+        baseIndexMap['C'] = Base.C.ordinal();
+        baseIndexMap['G'] = Base.G.ordinal();
+        baseIndexMap['T'] = Base.T.ordinal();
+    }
+
+    static private final Transition[][] transitionIndexMap = new Transition[4][4];
```
yfarjoun: Base.values.size() rather than 4?

src/java/picard/analysis/artifacts/Transition.java
yfarjoun: could for-loop this, using the valueOf trick...would be nicer I think

src/java/picard/analysis/artifacts/Transition.java
```diff
@@ -51,4 +46,46 @@ public Transition matchingRef() {
...
+
+        private Base(final char base) {
+            this.base = (byte)base;
+        }
+    }
+
+    static protected final int[] baseIndexMap = new int[256];
+    static {
+        Arrays.fill(baseIndexMap, -1);
+        baseIndexMap['A'] = Base.A.ordinal();
```
yfarjoun: loop instead, each enum knows its base, so use it.

src/java/picard/analysis/artifacts/Transition.java
```diff
@@ -51,4 +46,46 @@ public Transition matchingRef() {
...
+        T ('T');
+
+        public byte base;
+
+        private Base(final char base) {
+            this.base = (byte)base;
+        }
+    }
+
+    static protected final int[] baseIndexMap = new int[256];
```
yfarjoun: explain that only 4 of the values will be used but that you want to be "allow" an attempted access by any byte/char value.

yfarjoun:
@mattsooknah I'm reviewing this, but feel free to take a look at this as well if you like.

src/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
```diff
@@ -196,6 +196,12 @@ protected void acceptRead(final SAMRecord rec, final ReferenceSequence ref) {
             throw new PicardException("Record contains library that is missing from header: " + library);
         }
 
+        // get the ArtifactCounter for this library
+        final ArtifactCounter counter = artifactCounters.get(library);
+        final byte[] readBases = rec.getReadBases();
+        final byte[] readQuals = rec.getBaseQualities();
```
yfarjoun: allow for OQ if the user wants.

src/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
```diff
@@ -218,26 +232,27 @@ protected void acceptRead(final SAMRecord rec, final ReferenceSequence ref) {
...
-
                 // count the base!
-                artifactCounters.get(library).countRecord(context, readBase, rec);
+                counter.countRecord(context, readBase, rec);
             }
         }
     }
 
+    private String getRefContext(final ReferenceSequence ref, final int contextStartIndex, final int contextFullLength) {
+        // cache the upper-cased string version of this reference so we don't need to create a string for every base in every read
```
yfarjoun: nice.

src/java/picard/analysis/artifacts/ContextAccumulator.java
```diff
@@ -20,27 +19,22 @@
 
     // are the PE reads expected to face the same direction?
     private final boolean expectedTandemReads;
-    private final Map<Transition, Map<String, AlignmentAccumulator>> artifactMap;
+    private final Map<String, AlignmentAccumulator[]> artifactMap;
```
yfarjoun: add comment that this maps from library to the accumulators.

eitanbanks: it's a mapping from contexts, not library

src/java/picard/analysis/artifacts/Transition.java
eitanbanks: valueOf doesn't work for enums with multiple values (that's why Matt had a String addition thing in the previous version).  This is the cleanest it's going to get.

yfarjoun:
That's awesome! I'm looking at the code now.

eitanbanks:
Yossi, just pushed a commit to address your first round of comments.  Will continue tomorrow.

src/java/picard/analysis/artifacts/ContextAccumulator.java
yfarjoun: except that it doesn't...my mistake...make a note stating what it maps to accumulators...

src/java/picard/analysis/artifacts/ContextAccumulator.java
```diff
@@ -48,9 +42,12 @@ public void countRecord(final String refContext, final char calledBase, final SA
      */
     public ListMap<Transition, DetailPair> calculateMetrics(final String sampleAlias, final String library) {
         final ListMap<Transition, DetailPair> detailMetricsMap = new ListMap<>();
-        for (final Transition altTransition : Transition.altValues()) {
-            final Transition refTransition = altTransition.matchingRef();
-            for (final String context : new TreeSet<>(this.artifactMap.get(altTransition).keySet())) {
+        for (final String context : new TreeSet<>(artifactMap.keySet())) {
+
+            final char refBase = context.charAt(context.length() / 2);
```
yfarjoun: why are you not using the old function with the sanity check? this cannot be too time consuming...

src/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
```diff
@@ -196,6 +199,12 @@ protected void acceptRead(final SAMRecord rec, final ReferenceSequence ref) {
             throw new PicardException("Record contains library that is missing from header: " + library);
         }
 
+        // get the ArtifactCounter for this library
+        final ArtifactCounter counter = artifactCounters.get(library);
+        final byte[] readBases = rec.getReadBases();
+        final byte[] readQuals = (USE_OQ && rec.getOriginalBaseQualities() != null) ? rec.getOriginalBaseQualities() : rec.getBaseQualities();
```
tfenne: FYI `rec.getOriginalBaseQualities()` does *not* cache, and so does the attribute lookup and decode from text on each call.  You probably want to replace this with something like:
```java
final byte[] readQuals;
if (USE_OQ) {
    byte[] tmp = rec.getOriginalBaseQualities();
    readQuals = tmp == null ? rec.getBaseQualities() : tmp;
else {
    readQuals = rec.getOriginalBaseQualities()
}
```

Might even be nice to have something like `SAMRecord.getOriginalQualsOrQuals()` which encapsulates that logic?

src/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
```diff
@@ -203,6 +212,14 @@ protected void acceptRead(final SAMRecord rec, final ReferenceSequence ref) {
                 final int readPos = block.getReadStart() + offset;
                 final int refPos = block.getReferenceStart() + offset;
 
+                // skip low BQ sites
+                final byte qual = readQuals[readPos - 1];
+                if (qual < MINIMUM_QUALITY_SCORE) continue;
+
+                // skip N bases in read
+                final char readBase = (char)readBases[readPos - 1];  // SAMRecord guarantees that all bases are already upper-cased
```
tfenne: Unless you're toUpper'ing the whole read earlier, `BAMRecord` guarantees all bases are upper case, but `SAMRecord` doesn't.

tfenne:
@eitanbanks @yfarjoun - Can I suggest reverting the default back to using OQ and seeing how the performance is?  I looked back at the original code and see that `SAMRecord.getOriginalQualities` was being called _twice per base in the read_!  Since it doesn't cache, no wonder it sucked up 40% of the runtime.  Since @eitanbanks moved it to being twice per read (soon once, hopefully), that would be ~40%/300 = 0.13% of the runtime now.

If that's the case I think it would be better to leave OQ being the default for two reasons:
1) In the case of really bad artifacts it's possible BQSR will deflate base quality enough that CollectSequencingArtifactMetrics will skip many of the bases and under-count the artifact.
2) I think it's preferable to leave defaults so that program behaviour doesn't change unexpectedly unless there's a really compelling reason.

tfenne:
Oh - one last thought... what do you think about adding some kind of mode that causes the program to exit once it has achieved sufficient power, instead of marching through the whole file?  E.g. you could specify a artifact rate you wanted to be able to detect (e.g. Q40), and a confidence (e.g. 0.999) and just stop once all the contexts have a 0.999 CI that doesn't straddle Q40.  For WGS data you could probably bail somewhere on chr1!

eitanbanks:
@tfenne is right, so I'll go back to making it the default again!  @yfarjoun, you can close that Palantir ticket now since it's no longer necessary.

Tim, your other thought is a good one, but I'll leave that for Yossi to implement in a different PR.  Because it will affect the results, he'll likely want lots of extra testing.

src/java/picard/analysis/artifacts/ContextAccumulator.java
eitanbanks: Actually, the modulo operator is pretty time consuming.  However, it was bad when it was being calculated many times per base per read; but now that it's only at the end it won't be that bad.  That being said, do you really want to put the check here and fail out after processing the entire genome?  Why not put the check when the artifactMap is created?  In fact, that's what I'll do.

eitanbanks: (and I'll change it to use the low bit instead of modulo)

eitanbanks:
Okay, just pushed a commit to address the 2nd round of comments from Yossi and Tim.

src/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
```diff
@@ -86,7 +86,7 @@
             "i.e. they're expected to face the same direction.")
     public boolean TANDEM_READS = false;
 
-    @Option(doc = "When available, use original quality scores for filtering.")
+    @Option(doc = "When available, use original quality scores for filtering.  Note that this tool is likely to run significantly slower with this argument enabled")
```
yfarjoun: correct comment as this is not true anymore.

src/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
```diff
@@ -196,6 +199,18 @@ protected void acceptRead(final SAMRecord rec, final ReferenceSequence ref) {
...
+        // set up some constants that don't change in the loop below
+        final int contextFullLength = 2 * CONTEXT_SIZE + 1;
+        final ArtifactCounter counter = artifactCounters.get(library);
+        final byte[] readBases = rec.getReadBases();
+        final byte[] readQuals;
+        if (USE_OQ) {
+            final byte[] tmp = rec.getOriginalBaseQualities();
+            readQuals = tmp == null ? rec.getBaseQualities() : tmp;
+        } else {
+            readQuals = rec.getOriginalBaseQualities();
```
yfarjoun: typo

yfarjoun: ( this was actually @tfenne typo...that you copied...) should be rec.getBaseQualities() here.

src/java/picard/analysis/artifacts/Transition.java
```diff
@@ -51,4 +46,46 @@ public Transition matchingRef() {
...
+    static {
+        Arrays.fill(baseIndexMap, -1);
+        for (final Base b : Base.values()) {
+            baseIndexMap[b.base] = b.ordinal();
+        }
+    }
+
+    static private final Transition[][] transitionIndexMap = new Transition[Base.values().length][Base.values().length];
+    static {
+        transitionIndexMap[Base.A.ordinal()][Base.A.ordinal()] = AtoA;
```
yfarjoun: I still think that a double for-loop here would be clearer and cleaner, even though it uses 16 * 2 string additions.

src/java/picard/analysis/artifacts/ContextAccumulator.java
yfarjoun: ok. sounds good.

yfarjoun: this part of the code doesn't happen that much though...it's not like the accumulators, right?

src/java/picard/analysis/artifacts/Transition.java
tfenne: It wouldn't even have to use string additions right?
```java
for (final Base b1 : Base.values()) {
    for (final Base b2 : Base.values()) {
        transitionIndexMap[b1.ordinal()][b2.ordinal()] = Transition.transitionOf(b1.base, b2.base);
    }
}
```

yfarjoun: that would be magical if it could happen!

yfarjoun: sorry, that might seem more snarky than I intended....
transitionOf uses the array we are trying to build here...

tfenne: I missed that!  Though having an enum access shared state in the enclosing class seems kinda nasty to me, which is probably why I didn't expect it!

yfarjoun:
OK. I'm done. :+1:

src/java/picard/analysis/artifacts/Transition.java
eitanbanks: @yfarjoun I tried it but you can't use valueOf in the static initializer...

yfarjoun: hmmm. I'm getting no errors (compile or runtime) with  
```java
static {
        for (final Base b1 : Base.values()) {
            for (final Base b2 : Base.values()) {
                transitionIndexMap[b1.ordinal()][b2.ordinal()] = Transition.valueOf(b1.toString() + "to" + b2.toString());
            }
        }
    }
```

eitanbanks:
Maybe not having the toString() is what got me.  I added it, committed, and rebased.
Next steps?

src/java/picard/analysis/artifacts/Transition.java
yfarjoun: I'll send you over hipchat a file to tie out with. OK?

eitanbanks:
@yfarjoun gave me a test file to run on and the results are exactly the same when run with the old version as with the new except for the header lines (different 'OUTPUT' and 'Started on' values).  Well, and the runtime was very different.  :)
I'm going to merge now.

yfarjoun:
:+1:

PR #369, opened 2015-12-30
Added note about why our PCT_EXC_UNPAIRED numbers are higher than non-RQC

eitanbanks:
No code changes, just a comment for posterity.  I assume I can just merge w/out a review.

### DIFF
--- a/src/java/picard/analysis/CollectWgsMetricsFromQuerySorted.java
+++ b/src/java/picard/analysis/CollectWgsMetricsFromQuerySorted.java
@@ -158,6 +158,8 @@ public class CollectWgsMetricsFromQuerySorted extends CommandLineProgram {
         metrics.metrics.TOTAL_BASES += totalReadBases;
         if (isPaired) metrics.metrics.TOTAL_READ_PAIRS++;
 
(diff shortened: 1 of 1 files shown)


src/java/picard/analysis/CollectWgsMetricsFromQuerySorted.java
```diff
@@ -158,6 +158,8 @@ private void calculateMetricsForRead(final QuerySortedReadPairIteratorUtil.ReadP
         metrics.metrics.TOTAL_BASES += totalReadBases;
         if (isPaired) metrics.metrics.TOTAL_READ_PAIRS++;
 
+        // Note that CollectWgsMetrics does NOT count paired reads that are both unmapped in the PCT_EXC_UNPAIRED, but we do so here
+        //   because this tool isn't a locus iterator and we need to ensure that our passing base numbers are accurate in the end
```
yfarjoun: The point of this note is lost, I think. can you specify more clearly what metric you are trying to make accurate and what metric will be different?

eitanbanks: It's in the comment: PCT_EXC_UNPAIRED.

PR #372, opened 2016-01-05
Formatted the normalized coverage quintile results

kbergin:


yfarjoun:
that point here is that the metric in the end should not produce too many digits, can you create a test that shows this happens?

Issue #380, opened 2016-01-08
Have fingerprint metrics suffix be command-line options


eitanbanks:
We need to allow the possibility for the metrics suffix from CheckFingerprint to be different.
See this PR for more information: https://github.com/broadinstitute/picard/pull/379/files#r49205323

yfarjoun:
perhaps have the suffixes as optional parameters with the static variables
as the default values?

nh13:
@yfarjoun to keep consistent with other tools, I think the full file paths should be specified.  I think having different behavior across tools or different input mechanisms confuses users.  @vdauwera what do you think?

nh13:
I think preserving the default behavior is valuable (since likely it is used in production) so I implemented both mechanisms.

vdauwera:
There's definitely some potential for confusion here -- I'd recommend including a worked-out example that specifically details what the file names will be depending on the options used.

nh13:
@vdauwera are you talking about the discussion here, or the pull request I just submitted.  If the former, can you comment on the latter?  I think the PR brings this tool in line with other tools in Picard.

vdauwera:
@nh13 To be honest I was confused by both, and I'm not sure I'm clear on the behavior in your PR. 

Generally speaking the behavior of Picard metrics tools re: output file naming conventions tends to be confusing for new users. Obviously consistency helps because once you get it for one, you got it for the rest. But getting it for that first one is still a hump. We recently (finally!) completed our first round of doc improvements (touching ~55 tools) which included adding examples for the file naming for some tools, but I think we could do more on that front. 

Going forward it would be great to see explicit examples added systematically whenever the behavior is not simply "you get the name that you give" (i.e. if you specify "metrics.txt" and out comes "metrics.txt", no need to spell it out) for increased transparency and great justice.

nh13:
@vdauwera I definitely see the improved documentation (ex. `CollectGcBiasMetrics`), kudos.  I would definitely support the addition of similar documentation to `CheckFingerprint` when it is out of alpha, though frankly, I don't believe I am best suited to write that documentation and it seems like a burden to be honest for such a small change to an alpha tool that fixes something that perhaps could have been addressed in the initial import pull request review.  

For now, I suggest the incremental improvement of making the functionality in `CheckFingerprint` consistent with tools like `CollectGcBiasMetrics` and pushing comprehensive documentation for this tool to be a future improvement (perhaps by someone on your team :+1:?).  Does that sound reasonable?

[...]

PR #381, opened 2016-01-09
Detail and summary metrics for CheckFingerprint can be specified

nh13:
individually.  The old behavior is still the default.

@eitanbanks @yfarjoun 
Address #380.

### DIFF
--- a/src/java/picard/fingerprint/CheckFingerprint.java
+++ b/src/java/picard/fingerprint/CheckFingerprint.java
@@ -24,9 +24,11 @@
 
 package picard.fingerprint;
 
-import htsjdk.samtools.SAMFileReader;
 import htsjdk.samtools.SAMReadGroupRecord;
+import htsjdk.samtools.SamReader;
+import htsjdk.samtools.SamReaderFactory;
 import htsjdk.samtools.metrics.MetricsFile;
+import htsjdk.samtools.util.CloserUtil;
 import htsjdk.samtools.util.IOUtil;
 import htsjdk.samtools.util.Log;
 import htsjdk.samtools.util.SequenceUtil;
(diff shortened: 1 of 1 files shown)


src/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -64,9 +66,14 @@
     @Option(shortName=StandardOptionDefinitions.INPUT_SHORT_NAME, doc = "Input SAM or BAM file.")
     public File INPUT;
 
-    @Option(shortName=StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "Base of output files to write.")
+    @Option(shortName=StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "If SUMMARY_OUTPUT is null, the base of output files to write, " +
+            "otherwise the full path to the detailed metrics.")
     public String OUTPUT;
 
+    @Option(shortName = "S", doc = "The text file to which to write summary metrics.  If this is null, then OUTPUT is the base name of " +
+            "the output files and the default suffixes will be used.", optional = true)
```
yfarjoun: can you point out what the suffixes are in the doc so that the user will not have to read the code?

src/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -64,9 +66,14 @@
     @Option(shortName=StandardOptionDefinitions.INPUT_SHORT_NAME, doc = "Input SAM or BAM file.")
     public File INPUT;
 
-    @Option(shortName=StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "Base of output files to write.")
+    @Option(shortName=StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "If SUMMARY_OUTPUT is null, the base of output files to write, " +
+            "otherwise the full path to the detailed metrics.")
```
yfarjoun: I think that the term "full path" is confusing here as some-one might give a relative path and it will be used as such.

yfarjoun:
minor quibbles regarding the doc, otherwise :+1:

tfenne:
@nh13 I'm actually not very keen on this change, for a couple of reasons:

1. I think that when tools produce multiple, related, outputs it's more convenient to be able to specify an output prefix and have the tool produce multiple files that share the prefix.  I think if anything we should be moving other tools in that direction.
2. If you *really* want to support both behaviours, I think it would be better to have three options, `OUTPUT`, `SUMMARY_METRICS` and `DETAIL_METRICS`, with `OUTPUT` being mutex with the other two.  The interpretation of `OUTPUT` conditional on whether or not `SUMMARY_OUTPUT` is null or not is going to cause confusion.

nh13:
@tfenne I am happy to take direction on where we are moving, so I have implemented option (2), since for at least some Picard users, they need a way to specify the file extension. 

For posterity, I think that all Picard metric files should have ".txt" file extension and then this would be moot.

nh13:
@yfarjoun rebased onto master.  I also fixed the missing word in the usage, and verified that mutex for those options work as expected.

yfarjoun:
:+1:

PR #385, opened 2016-01-11
moving metrics tallying to a public static method, so it can be used by

nh13:
others (it is really useful).

@tfenne

### DIFF
--- a/src/java/picard/illumina/ExtractIlluminaBarcodes.java
+++ b/src/java/picard/illumina/ExtractIlluminaBarcodes.java
@@ -235,6 +235,26 @@ public class ExtractIlluminaBarcodes extends CommandLineProgram {
         }
 
         // Finish metrics tallying.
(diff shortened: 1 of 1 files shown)


jacarey:
Would it make sense to pull BarcodeMetrics out to its own class so others can use that as well without the having to reference ExtractIlluminaBarcodes? It seems like if these are useful to others (not ExtractIlluminaBarcodes?) we should pull them up to the as to the top level.

nh13:
@jacarey that makes a lot of sense, however, I want to make the simplest change to not cause disruption to others and make progress on my own work.  Your suggestion may break the Broad pipeline, which relies on class names, but you could confirm.   In light of these, I think this should be fine to merge.

jacarey:
Agreed. I think this is fine for merge.

nh13:
@jacarey I can't merge in this repo, would you do the honors?

PR #387, opened 2016-01-11
Change fingerprint output file instantiation and check if is writeable as opposed to readable

jsotobroad:
Change fingerprint output file instantiation and check if is writeable as opposed to readable


ktibbett:
@nh13 , please take note; I am pushing this through without waiting on your input (which I would ordinarily do) since it's breaking things on our end, but let me know if you have concerns.

@jsotobroad: :+1:

nh13:
Your call. My mistake.

PR #388, opened 2016-01-12
now relying on getSamReadNameFromFastqHeader in HTSJDK's SequenceUtil…

nh13:
… for converting a read name in a FASTQ to a read name in a SAM/BAM file.

Relies on https://github.com/samtools/htsjdk/pull/431

[Review on Reviewable](https://reviewable.io/reviews/broadinstitute/picard/388)


yfarjoun:
Hi @nh13,

I know that you mostly moved this code from Picard to hts-jdk, and are
trying to preserve the behavior.

To that effect, could you add your new tests to picard in a clean branch as
well? that way it will be clear that the read-names do not change when
switching to your implementation in htsjdk.

nh13:
@yfarjoun I think the tests in `FastqToSamTest` cover the cases, specifically `testPairedBad` and `okPairedFiles`, and the testing of trailing numbers.

yfarjoun:
👍 thanks.

PR #389, opened 2016-01-12
First pass at adding optical duplicate flagging/removal into MarkDupes

tfenne:
First pass at some changes to MarkDuplicates to allow optical (co-localized?) duplicates to be annotated in the output file and/or excluded from the output file.  So far support is only included in MarkDuplicates, though with a little help from @nh13 or @bradtaylor I'd be happy to extend support to MarkDuplicatesWithMateCigar also.

@nh13, @yfarjoun : Could you take a first look at this please and let me know what you think of the direction I've taken?  I'm not ready (yet) for a detailed review, just looking to get some extra eyes on this before I sink more time into writing tests and shining it up.  Thanks!

### DIFF
--- a/src/java/picard/sam/markduplicates/MarkDuplicates.java
+++ b/src/java/picard/sam/markduplicates/MarkDuplicates.java
@@ -63,29 +63,50 @@ import java.util.*;
 )
 public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
     static final String USAGE_SUMMARY = "Identifies duplicate reads.  ";
-    static final String USAGE_DETAILS = "This tool locates and tags duplicate reads (both PCR and optical) in a BAM or SAM file, where " +
-            "duplicate reads are defined as originating from the same original fragment of DNA.  The PCR duplicates are identified as " +
-            "having identical start coordinates for both reads in a mate pair. Optical duplicates are identified using the " +
-            "READ_NAME_REGEX and the OPTICAL_DUPLICATE_PIXEL_DISTANCE options. For more information on duplicate reads, see the " +
-            "GATK user documentation guide.<br /><br />" +
-            "The tool's main output is a new SAM or BAM file in which duplicates have been identified (in the SAM flags field). " +
-            "In addition, it also outputs a metrics file containing the numbers of READ_PAIRS_EXAMINED, UNMAPPED_READS, UNPAIRED_READS, " +
-            "UNPAIRED_READ DUPLICATES, READ_PAIR_DUPLICATES, and READ_PAIR_OPTICAL_DUPLICATES. <br /> " +
-            "<h4>Usage example:</h4>" +
-            "<pre>" +
-            "java -jar picard.jar MarkDuplicates \\<br />" +
-            "      I=input.bam \\<br />" +
-            "      O=marked_duplicates.bam \\<br />" +
-            "      M=marked_dup_metrics.txt " +
-            "</pre>" +
-            "<hr />";
+    static final String USAGE_DETAILS =
+            "This tool locates and tags duplicate reads (both PCR and optical/sequencing-driven) in a BAM or SAM file, where\n" +
+                    "duplicate reads are defined as originating from the same original fragment of DNA. Duplicates are identified as read\n" +
+                    "pairs having identical 5' positions (coordinate and strand) for both reads in a mate pair (and optinally, matching\n" +
+                    "unique molecular identifier reads; see BARCODE_TAG option). Optical, or more broadly Sequencing, duplicates are\n" +
+                    "duplicates that appear clustered together spatially during sequencing and can arise from optical/imagine-processing\n" +
+                    "artifacts or from bio-chemical processes during clonal amplification and sequencing; they are identified using the\n" +
+                    "READ_NAME_REGEX and the OPTICAL_DUPLICATE_PIXEL_DISTANCE options.\n" +
+                    "\n" +
+                    "The tool's main output is a new SAM or BAM file in which duplicates have been identified in the SAM flags field, or\n" +
+                    "optionally removed (see REMOVE_DUPLICATE and REMOVE_SEQUENCING_DUPLICATES), and optionally marked with a duplicate type\n" +
+                    "in the 'DT' optional attribute. In addition, it also outputs a metrics file containing the numbers of\n" +
+                    "READ_PAIRS_EXAMINED, UNMAPPED_READS, UNPAIRED_READS, UNPAIRED_READ DUPLICATES, READ_PAIR_DUPLICATES, and\n" +
+                    "READ_PAIR_OPTICAL_DUPLICATES.\n" +
+                    "\n" +
+                    "Usage example: java -jar picard.jar MarkDuplicates I=input.bam \\\n" +
+                    "                 O=marked_duplicates.bam M=marked_dup_metrics.txt\n";
+
+    /** Enum used to control how duplicates are flagged in the DT optional tag on each read. */
+    public enum DuplicateTaggingPolicy { DontTag, OpticalOnly, All }
+
+    /** The optional attribute in SAM/BAM files used to store the duplicate type. */
+    public static final String DUPLICATE_TYPE_TAG = "DT";
+    /** The duplicate type tag value for duplicate type: library. */
+    public static final String DUPLICATE_TYPE_LIBRARY = "LB";
+    /** The duplicate type tag value for duplicate type: sequencing (optical & pad-hopping, or "co-localized"). */
+    public static final String DUPLICATE_TYPE_SEQUENCING = "SQ";
+
+    /** Enum for the possible values that a duplicate read can be tagged with in the DT attribute. */
+    public enum DuplicateType {
+        LIBRARY(DUPLICATE_TYPE_LIBRARY),
+        SEQUENCING(DUPLICATE_TYPE_SEQUENCING);
+
+        private final String code;
+        DuplicateType(final String code) { this.code = code; }
+        public String code() { return this.code; }
+    }
+
     private final Log log = Log.getInstance(MarkDuplicates.class);
 
     /**
      * If more than this many sequences in SAM file, don't spill to disk because there will not
      * be enough file handles.
      */
-
     @Option(shortName = "MAX_SEQS",
             doc = "This option is obsolete. ReadEnds will always be spilled to disk.")
     public int MAX_SEQUENCES_FOR_DISK_READ_ENDS_MAP = 50000;
@@ -109,9 +130,18 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
     @Option(doc = "Read two barcode SAM tag (ex. BX for 10X Genomics)", optional = true)
     public String READ_TWO_BARCODE_TAG = null;
 
+    @Option(doc = "If true remove 'optical' duplicates and other duplicates that appear to have arisen from the " +
+            "sequencing process instead of the library preparation process, even if REMOVE_DUPLICATES is false. " +
+            "If REMOVE_DUPLICATES is true, all duplicates are removed and this option is ignored.")
+    public boolean REMOVE_SEQUENCING_DUPLICATES = false;
+
+    @Option(doc= "Determines how duplicate types are recorded in the DT optional attribute.")
+    public DuplicateTaggingPolicy TAGGING_POLICY = DuplicateTaggingPolicy.DontTag;
+
     private SortingCollection<ReadEndsForMarkDuplicates> pairSort;
     private SortingCollection<ReadEndsForMarkDuplicates> fragSort;
     private SortingLongCollection duplicateIndexes;
+    private SortingLongCollection opticalDuplicateIndexes;
     private int numDuplicateIndices = 0;
 
     protected LibraryIdGenerator libraryIdGenerator = null; // this is initialized in buildSortedReadEndLists
@@ -154,7 +184,7 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
         log.info("Reading input file and constructing read end information.");
         buildSortedReadEndLists(useBarcodes);
         reportMemoryStats("After buildSortedReadEndLists");
-        generateDuplicateIndexes(useBarcodes);
+        generateDuplicateIndexes(useBarcodes, this.REMOVE_SEQUENCING_DUPLICATES || this.TAGGING_POLICY != DuplicateTaggingPolicy.DontTag);
         reportMemoryStats("After generateDuplicateIndexes");
         log.info("Marking " + this.numDuplicateIndices + " records as duplicates.");
 
@@ -181,6 +211,7 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
         // Now copy over the file while marking all the necessary indexes as duplicates
         long recordInFileIndex = 0;
         long nextDuplicateIndex = (this.duplicateIndexes.hasNext() ? this.duplicateIndexes.next() : -1);
+        long nextOpticalDuplicateIndex = this.opticalDuplicateIndexes != null && this.opticalDuplicateIndexes.hasNext() ? this.opticalDuplicateIndexes.next() : -1;
 
         final ProgressLogger progress = new ProgressLogger(log, (int) 1e7, "Written");
         final CloseableIterator<SAMRecord> iterator = headerAndIterator.iterator;
@@ -226,15 +257,28 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
                     rec.setDuplicateReadFlag(false);
                 }
             }
-            recordInFileIndex++;
 
-            if (!this.REMOVE_DUPLICATES || !rec.getDuplicateReadFlag()) {
-                if (PROGRAM_RECORD_ID != null) {
-                    rec.setAttribute(SAMTag.PG.name(), chainedPgIds.get(rec.getStringAttribute(SAMTag.PG.name())));
+            // Manage the flagging of optical/sequencing duplicates
+            final boolean isOpticalDuplicate = (recordInFileIndex == nextOpticalDuplicateIndex);
+            if (isOpticalDuplicate) nextOpticalDuplicateIndex = this.opticalDuplicateIndexes.hasNext() ? this.opticalDuplicateIndexes.next() : -1;
+            rec.setAttribute(DUPLICATE_TYPE_TAG, null);
+
+            if (this.TAGGING_POLICY != DuplicateTaggingPolicy.DontTag && rec.getDuplicateReadFlag()) {
+                if (isOpticalDuplicate) {
+                    rec.setAttribute(DUPLICATE_TYPE_TAG, DuplicateType.SEQUENCING.code());
+                } else if (this.TAGGING_POLICY == DuplicateTaggingPolicy.All) {
+                    rec.setAttribute(DUPLICATE_TYPE_TAG, DuplicateType.LIBRARY.code());
                 }
-                out.addAlignment(rec);
-                progress.record(rec);
             }
+
+            // Output the record if desired and bump the record index
+            recordInFileIndex++;
+            if (this.REMOVE_DUPLICATES            && rec.getDuplicateReadFlag()) continue;
+            if (this.REMOVE_SEQUENCING_DUPLICATES && isOpticalDuplicate)         continue;
+
+            if (PROGRAM_RECORD_ID != null)  rec.setAttribute(SAMTag.PG.name(), chainedPgIds.get(rec.getStringAttribute(SAMTag.PG.name())));
+            out.addAlignment(rec);
+            progress.record(rec);
         }
 
         // remember to close the inputs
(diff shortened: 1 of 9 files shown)


tfenne:
@yfarjoun I've updated this to:
1. Use the `DT` optional tag
2. Provide options for no-tagging, tag only optical duplicate, tag all duplicates
3. Add more to the documentation

Since I know you're interested in this kind of thing, would you be up for a review?  There's no huge rush to merge, but I'd like to get any major problems/changes identified and fixed quickly.

yfarjoun:
I'd be happy to take a look.

sooheelee:
When are you planning to make this public? I've been working on an [updated  MarkDuplicates tutorial](http://gatkforums.broadinstitute.org/gatk/discussion/6747/how-to-flag-duplicates-with-markduplicates-or-markduplicateswithmatecigar#latest) that includes MarkDuplicatesWithMateCigar.

tfenne:
@sooheelee It's in public now, but I assume you mean when am I planning on merging it?  I'm happy with it in it's current form, so really it will depend on @yfarjoun's availability to review, and what he has to say.

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -109,9 +109,19 @@
     @Option(doc = "Read two barcode SAM tag (ex. BX for 10X Genomics)", optional = true)
     public String READ_TWO_BARCODE_TAG = null;
 
+    @Option(doc = "If true remove optical duplicates from the output even if REMOVE_DUPLICATES is false.")
+    public boolean REMOVE_OPTICAL_DUPLICATES = false;
+
+    @Option(doc = "If set write a tag with this name and OPTICAL_DUPLICATE_TAG_VALUE on to each read that is an optical duplicate.", optional=true)
+    public String OPTICAL_DUPLICATE_TAG_NAME = null;
+
+    @Option(doc = "If true do not write duplicates to the output file instead of writing them with appropriate flags set.")
```
eitanbanks: I don't think this description is correct.

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -109,9 +109,19 @@
     @Option(doc = "Read two barcode SAM tag (ex. BX for 10X Genomics)", optional = true)
     public String READ_TWO_BARCODE_TAG = null;
 
+    @Option(doc = "If true remove optical duplicates from the output even if REMOVE_DUPLICATES is false.")
+    public boolean REMOVE_OPTICAL_DUPLICATES = false;
```
eitanbanks: Might want to mention that if REMOVE_DUPLICATES is set to true then it overrides this being set to false?

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -72,9 +72,12 @@ public OpticalDuplicateFinder(final String readNameRegex, final int opticalDupli
      * unmodified a copy of the list should be passed to this method.
      *
      * @param list a list of reads that are determined to be duplicates of one another
+     * @param keeper a single PhysicalLocation that is the one being kept as non-duplicate, and thus should never be
+     *               annotated as an optical duplicate. May in some cases be null, or a PhysicalLocation not
+     *               contained within the list!
```
eitanbanks: Could you please explain why this is necessary?  Is it possible to have multiple optical duplicate sets within a given duplicate set?

eitanbanks:
Review complete.

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -109,9 +109,19 @@
...
+    @Option(doc = "If set write a tag with this name and OPTICAL_DUPLICATE_TAG_VALUE on to each read that is an optical duplicate.", optional=true)
+    public String OPTICAL_DUPLICATE_TAG_NAME = null;
+
+    @Option(doc = "If true do not write duplicates to the output file instead of writing them with appropriate flags set.")
+    public String OPTICAL_DUPLICATE_TAG_VALUE = "+";
+
     private SortingCollection<ReadEndsForMarkDuplicates> pairSort;
     private SortingCollection<ReadEndsForMarkDuplicates> fragSort;
     private SortingLongCollection duplicateIndexes;
+    private SortingLongCollection opticalDuplicateIndices;
```
yfarjoun: even if both of the spellings of the plural of index are correct, I suggest we use one.

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -109,9 +109,19 @@
     @Option(doc = "Read two barcode SAM tag (ex. BX for 10X Genomics)", optional = true)
     public String READ_TWO_BARCODE_TAG = null;
 
+    @Option(doc = "If true remove optical duplicates from the output even if REMOVE_DUPLICATES is false.")
+    public boolean REMOVE_OPTICAL_DUPLICATES = false;
+
+    @Option(doc = "If set write a tag with this name and OPTICAL_DUPLICATE_TAG_VALUE on to each read that is an optical duplicate.", optional=true)
+    public String OPTICAL_DUPLICATE_TAG_NAME = null;
```
yfarjoun: according to https://github.com/samtools/hts-specs/issues/121 this should be hard-coded to DT, a new tag to be added to SAMUtils.

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -109,9 +109,19 @@
...
+
+    @Option(doc = "If true do not write duplicates to the output file instead of writing them with appropriate flags set.")
+    public String OPTICAL_DUPLICATE_TAG_VALUE = "+";
```
yfarjoun: should this default to "optical" rather than '+'? should this even be an option?

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -226,15 +237,22 @@ protected int doWork() {
...
+            final boolean isOpticalDuplicate = recordInFileIndex == nextOpticalDuplicateIndex;
+            if (isOpticalDuplicate) {
+                if (this.OPTICAL_DUPLICATE_TAG_NAME != null) rec.setAttribute(this.OPTICAL_DUPLICATE_TAG_NAME, this.OPTICAL_DUPLICATE_TAG_VALUE);
+                nextOpticalDuplicateIndex = this.opticalDuplicateIndices.hasNext() ? this.opticalDuplicateIndices.next() : -1;
             }
+
+            // Output the record if desired and bump the record index
+            recordInFileIndex++;
+            if (this.REMOVE_DUPLICATES && rec.getDuplicateReadFlag()) continue;
+            if (this.REMOVE_OPTICAL_DUPLICATES && isOpticalDuplicate) continue;
```
yfarjoun: if you move the index bump before the if(isBoolean), you will be able to put this continue inside the if statement. it's slightly simpler since it will put all the optical stuff in the same place.

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -460,10 +478,13 @@ private ReadEndsForMarkDuplicates buildReadEnds(final SAMFileHeader header, fina
      *
      * @return an array with an ordered list of indexes into the source file
      */
-    private void generateDuplicateIndexes(final boolean useBarcodes) {
+    private void generateDuplicateIndexes(final boolean useBarcodes, final boolean indexOpticalDuplicates) {
         // Keep this number from getting too large even if there is a huge heap.
-        final int maxInMemory = (int) Math.min((Runtime.getRuntime().maxMemory() * 0.25) / SortingLongCollection.SIZEOF,
-                (double) (Integer.MAX_VALUE - 5));
+        int maxInMemory = (int) Math.min((Runtime.getRuntime().maxMemory() * 0.25) / SortingLongCollection.SIZEOF, (double) (Integer.MAX_VALUE - 5));
```
yfarjoun: I think it will be clearer if maxInmemory remains final and you use maxinMemory/2 as the input for the opticalDupIndex. BTW, can you comment on why you need half the amount?

yfarjoun: Oh. sorry. I get it. you're using half the available memory for the regular indices and half for the opticals...

nm. ignore previous comment about final. just add a comment that explains this. thanks.

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -460,10 +478,13 @@ private ReadEndsForMarkDuplicates buildReadEnds(final SAMFileHeader header, fina
...
+        if (indexOpticalDuplicates) {
+            maxInMemory /= 2;
```
yfarjoun: add a log explaining that keeping track of optical dups will mean more IO.

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
yfarjoun: yes. imagine that you have two PCR copies, A and B and each has optical dups A' and B'. any of these could be keeper, so you want to make sure you do not mark that one as an optical dup.

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -96,13 +99,14 @@ public int compare(final PhysicalLocation lhs, final PhysicalLocation rhs) {
...
-                if (opticalDuplicateFlags[j]) continue;
+                if ((rhs == keeper && opticalDuplicateFlags[i]) || opticalDuplicateFlags[j]) continue;
                 if (lhs.getReadGroup() != rhs.getReadGroup()) continue outer;
                 if (lhs.getTile() != rhs.getTile()) continue outer;
                 if (rhs.getX() > lhs.getX() + this.opticalDuplicatePixelDistance) continue outer;
 
                 if (Math.abs(lhs.getY() - rhs.getY()) <= this.opticalDuplicatePixelDistance) {
-                    opticalDuplicateFlags[j] = true;
+                    if (rhs == keeper) opticalDuplicateFlags[i] = true;
+                    else               opticalDuplicateFlags[j] = true;
```
yfarjoun: the logic here seems right. but it's a bit messy....would be good to have another pair of eyes on it and also specific tests.

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -72,9 +72,12 @@ public OpticalDuplicateFinder(final String readNameRegex, final int opticalDupli
      * unmodified a copy of the list should be passed to this method.
      *
      * @param list a list of reads that are determined to be duplicates of one another
+     * @param keeper a single PhysicalLocation that is the one being kept as non-duplicate, and thus should never be
```
nh13: I keep thinking of sports analogies for `keeper`, how about `representative` instead like in htsjdk's [DuplicateSet](https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/samtools/DuplicateSet.java)?

yfarjoun:
@tfenne I think it's a good direction. I think it would be great value having these tags in the reads. 

With the newest sequencers 'short' doesn't cut it anymore for physical position, we need a full int to avoid overflow. I've run into this problem when writing the PositionBasedDownsampling, which is why I implemented PhysicalPositionInt. Do you think this would be a good time to switch everything over to ints or do you think that the memory requirements would be too onerous? an alternative would be to allow for overflow and take it into consideration when calculating the pixel distance (overflow is blatantly ignored currently) what do you think?

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -96,13 +99,14 @@ public int compare(final PhysicalLocation lhs, final PhysicalLocation rhs) {
             for (int j = i + 1; j < length; ++j) {
                 final PhysicalLocation rhs = list.get(j);
 
-                if (opticalDuplicateFlags[j]) continue;
+                if ((rhs == keeper && opticalDuplicateFlags[i]) || opticalDuplicateFlags[j]) continue;
```
nh13: `(rhs == keeper && opticalDuplicateFlags[i])` can actually be moved into the outer loop.

nh13: Not true since `opticalDuplicateFlags[i]` may be set below.  Wow this is tough to follow.

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -96,13 +99,14 @@ public int compare(final PhysicalLocation lhs, final PhysicalLocation rhs) {
...
                 if (lhs.getReadGroup() != rhs.getReadGroup()) continue outer;
                 if (lhs.getTile() != rhs.getTile()) continue outer;
                 if (rhs.getX() > lhs.getX() + this.opticalDuplicatePixelDistance) continue outer;
 
                 if (Math.abs(lhs.getY() - rhs.getY()) <= this.opticalDuplicatePixelDistance) {
-                    opticalDuplicateFlags[j] = true;
+                    if (rhs == keeper) opticalDuplicateFlags[i] = true;
```
nh13: can you break here since now `(rhs == keeper && opticalDuplicateFlags[i])` will always be true, which causes the continue statement on line `102` to always be invoked?

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
eitanbanks: Got it.  I know you guys have said this below, but I think there needs to be a lot of comments in the code explaining what is happening below.

tfenne: It's because, to do the optical duplicate detection, you need to see all the locations, including the read pair that is being kept as non-duplicate.  I.e. if you have two read pairs with same alignment positions, you need both to determine if one is an optical duplicate of the other, but you should only mark one of the with an optical duplicate tag.  And ideally *not* the one that isn't being marked as a duplicate.

src/java/picard/sam/markduplicates/MarkDuplicates.java
tfenne: I didn't even notice that I did that.  Sorry.

tfenne:
Hey @eitanbanks and @yfarjoun, I'm really sorry but I just realized that I messed up.  I made some substantial changes go this before I re-pinged for review, and I was sure I had pushed those changes up, but now I see that my push failed and I didn't see it at the time.

I'm going to go through and apply any changes you guys suggested that still make sense, then rebase and push again.  I'll try and highlight the differences when I do that.

src/java/picard/sam/markduplicates/MarkDuplicates.java
tfenne: Agreed.  This was changed but not pushed.  Sorry.

src/java/picard/sam/markduplicates/MarkDuplicates.java
tfenne: Ditto.

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
tfenne: I don't like `representative` because the name alone is not sufficient to convey that this _must_ be the records that is considered non-duplicate amongst the set.  How about `nonDuplicate` or `retained` instead?

tfenne:
@yfarjoun Regarding shorts vs. ints, it's probably time we just pushed everything over to ints, or (shudder) perhaps even floats?  That said, I don't really want to take that on as part of this PR.

The only other option I can think of, just to throw it out there, would be to _parse_ the locations out as ints, and then repeatedly divide by 10 and round until they fit in a short.  I say this because shorts should be enough to store a location within a (square) gigapixel image (2^15 * 2^15), and I think this is only an issue because of Illumina multiplying up the pixel coordinates to do sub-pixel stuff.  If we only need pixel resolution, we could push them back down?  Just a thought.

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
tfenne: @yfarjoun  is right that this is convoluted now.  I'll add a few tests, and then try what you suggest.

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -226,15 +254,28 @@ protected int doWork() {
                     rec.setDuplicateReadFlag(false);
                 }
             }
-            recordInFileIndex++;
 
-            if (!this.REMOVE_DUPLICATES || !rec.getDuplicateReadFlag()) {
-                if (PROGRAM_RECORD_ID != null) {
-                    rec.setAttribute(SAMTag.PG.name(), chainedPgIds.get(rec.getStringAttribute(SAMTag.PG.name())));
+            // Manage the flagging of optical/sequencing duplicates
```
tfenne: @yfarjoun, @eitanbanks : this section is probably the biggest change (along with the change in CLP args), as it now will optionally tag no duplicates, optical only, or all duplicates with a tag.

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -67,45 +67,36 @@ public OpticalDuplicateFinder(final String readNameRegex, final int opticalDupli
...
-                if (retval == 0) retval = lhs.getTile() - rhs.getTile();
-                if (retval == 0) retval = lhs.getX() - rhs.getX();
-                if (retval == 0) retval = lhs.getY() - rhs.getY();
-                return retval;
-            }
-        });
-
-        outer:
-        for (int i = 0; i < length; ++i) {
+        for (int i=0; i<length; ++i) {
```
tfenne: @yfarjoun , @eitanbanks , @nh13 : this method is now *much* simpler, though likely less efficient for large piles of duplicates.  None of us (me included) caught the fact that the method returns an array of flags that is index-synced to the list of PhysicalLocations - but only the post-sorted version.  That seems really nasty and error-prone, so I removed the sort and simplified the code.

tfenne:
Ready for round 2.  Sorry again about having screwed up the push before the review.

tfenne:
@yfarjoun Any chance you can take another look soon?

yfarjoun:
sure. thanks for the ping.

yfarjoun:
back to you @tfenne. Please add the test-case that I highlighted in the comments (or a similar one)

tfenne:
@yfarjoun Good catch on the clique/keeper test case. I've added that exact test case and fixed the code to make it work.  I think the resulting code is, largely, clearer.  Hit your other fixes too.  Do you want to take a quick look at the revamped `findOpticalDuplicates()` method or are you happy with all the tests passing (including the new one)?

tfenne:
Looks like the latest change broke a bunch of other MarkDuplicates tests, I'm investigating now.

tfenne:
Ok, one more try - this most recent push should have all the tests passing again.

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -65,47 +65,76 @@ public OpticalDuplicateFinder(final String readNameRegex, final int opticalDupli
...
+        if (keeper != null && keeper.hasLocation()) {
+            for (final PhysicalLocation loc : list) {
+                if (loc == keeper) return keeper;
+            }
+        }
+        return null;
+    }
+
+    /** Simple method to test whether two physical locations are close enough to each other to be deemed optical dupes. */
+    private boolean closeEnough(final PhysicalLocation lhs, final PhysicalLocation rhs, final int distance) {
```
yfarjoun: static

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -65,47 +65,76 @@ public OpticalDuplicateFinder(final String readNameRegex, final int opticalDupli
...
+                    opticalDuplicateFlags[index] = true;
                 }
             }
         }
+        
         return opticalDuplicateFlags;
     }
+
+    /** Returns the keeper if it is contained within the list and has location information, otherwise null. */
+    private PhysicalLocation keeperOrNull(final List<? extends PhysicalLocation> list, final PhysicalLocation keeper) {
```
yfarjoun: static

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -65,47 +65,76 @@ public OpticalDuplicateFinder(final String readNameRegex, final int opticalDupli
...
-                if (opticalDuplicateFlags[j]) continue;
-                if (lhs.getReadGroup() != rhs.getReadGroup()) continue outer;
-                if (lhs.getTile() != rhs.getTile()) continue outer;
-                if (rhs.getX() > lhs.getX() + this.opticalDuplicatePixelDistance) continue outer;
-
-                if (Math.abs(lhs.getY() - rhs.getY()) <= this.opticalDuplicatePixelDistance) {
-                    opticalDuplicateFlags[j] = true;
+                if (closeEnough(lhs, rhs, distance)) {
+                    // At this point we want to mark either lhs or rhs as duplicate. Generally we want to mark
+                    // rhs, unless rhs is already marked.  The only way rhs can already be marked is if it's a
```
yfarjoun: s/is if/as if/ in comment.

yfarjoun: I agree with the code, but not with the comment.

In the case of no keeper and x positions 10,20,30 with distance 25, by the time i=1, j=2 both flags will be set to true, so it's not the "only way". I'd say "The only way rhs can already be marked is if it's a duplicate of a "previous" read (perhaps the keeper) so in that case it's safe to mark the lhs, and so we do."

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -65,47 +65,76 @@ public OpticalDuplicateFinder(final String readNameRegex, final int opticalDupli
...
                 }
             }
         }
+        
         return opticalDuplicateFlags;
     }
+
+    /** Returns the keeper if it is contained within the list and has location information, otherwise null. */
+    private PhysicalLocation keeperOrNull(final List<? extends PhysicalLocation> list, final PhysicalLocation keeper) {
+        if (keeper != null && keeper.hasLocation()) {
```
yfarjoun: please test this (add a test where keeper isn't part of the list, and isn't null)

yfarjoun:
ok. only minor comments this time round. some comment changes and another test. what are the next steps here? squash and merge?

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
tfenne: I agree that this could be static, but I see no advantage to making it static - it's only referenced from a non-static context.

tfenne:
Nulls were being tested already.  I added two more tests:
1. Just randomly shuffles a list a hundred times and randomly selects the keeper also, and then ensures that it always gets the right number of optical duplicates back.
2. Tests with a keeper that's not in the list

And also fixed that comment that was out of date with the latest changes.  I've squashed - I think this is ready to merge.

tfenne:
@yfarjoun All tests are passing - want to do the honours?

yfarjoun:
One more thing (from me) could you extract the string literals to constants (both the sam tag and the enum strings)?

tfenne:
@yfarjoun Those strings have been constant-ized.  Merging now.

PR #390, opened 2016-01-13
target to generate code coverage report using jacoco

akiezun:
added a report-coverage target. It makes a report of coverage using jacoco.

to test it run
```
ant report-coverage
```

the report is generated in the report/ directory. Open the report/index.html file to start navigating


tfenne:
@akiezun That new jar should probably live in `lib/ant` since it's an ant plugin and not needed past built-time.

akiezun:
@tfenne no problem

build.xml
```diff
@@ -39,7 +39,7 @@
     <property name="reports" value="dist/test"/>
     <property name="test.output" value="dist/test"/>
     <property name="htsjdk_lib_dir" value="${dist}/htsjdk_lib_dir"/>
-    <property name="htsjdk_git_url" value="git@github.com:samtools/htsjdk.git"/>
+    <property name="htsjdk_git_url" value="https://github.com/samtools/htsjdk.git"/>
```
akiezun: oops, this is not needed. I needed it only for my system. Will delete this change

akiezun:
this PR should be seen as the first step towards setting up coverage in coveralls. Not sure how to that though.

ktibbett:
:+1: if @tfenne  is ok with the update.

akiezun:
btw, I addressed @tfenne 's comment

ktibbett:
@akiezun : :+1: .  Merge at will.

PR #391, opened 2016-01-13
Updated notes about differences between this tool and CollectWgsMetrics

eitanbanks:


ktibbett:
:+1:  When we have some idea of how much a of a difference this makes, it might be nice to add that.

vdauwera:
Might also be nice to add that information to the public-facing doc.

PR #393, opened 2016-01-14
Small updates to the RQC metrics collector.

eitanbanks:
1. Ignore non-PF reads.
2. Add TOTAL_READS_ALIGNED to the metrics.

### DIFF
--- a/src/java/picard/analysis/CollectWgsMetricsFromQuerySorted.java
+++ b/src/java/picard/analysis/CollectWgsMetricsFromQuerySorted.java
@@ -82,17 +82,27 @@ public class CollectWgsMetricsFromQuerySorted extends CommandLineProgram {
         public FILTERING_STRINGENCY TYPE;
 
         /** The total number of bases, before any filters are applied. */
-        public long TOTAL_BASES = 0;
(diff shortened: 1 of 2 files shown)


ktibbett:
@eitanbanks  (recognizing fully that it is a little late in the game for this kind of comment), at least when adding new metrics can we try to use the same name we use for a metric in other places?  In this case, I think TOTAL_READS_ALIGNED should be PF_READS_ALIGNED.

ktibbett:
Also, please consider whether other metrics names should be changed since they are now "PF" rather than "TOTAL."

eitanbanks:
@ktibbett I've gone ahead and renamed all of the metrics that started with TOTAL_ to PF_ as requested.
Tests passing.  Okay to merge?

yfarjoun:
don't you need a flyway script for this?

ktibbett:
@eitanbanks , please wait until we make the database changes internally to merge.

eitanbanks:
Roger that.
@yfarjoun flyways only needed for the uploader tools.

eitanbanks:
Got verbal okay to merge from Kathleen.

PR #394, opened 2016-01-14
Hard cap the depth of the het-Sensitivity calculation at 1000. 

yfarjoun:
This should not affect the result (as one should have very high _theoretical_ sensitivity at that depth) but it will affect performance (as in, it will improve it).

- White-space changes

@kbergin could you take a look?


kbergin:
@yfarjoun tests are failing because of some r / cran dependencies? Not sure what that is

kbergin:
Travis is failing for other reasons; when it's fixed, this PR looks good. Thank you for the help figuring out the problem :+1:

kbergin:
Rebased onto master to address Travis failures after PR #395 fixed it

kbergin:
Merging!

PR #396, opened 2016-01-14
Added the ability to produce insert size metrics including duplicates.

tfenne:
This is helpful when working with deeply over-sequenced libraries as in that case the post-dedupe distribution is skewed towards the under-represented fragments.


tfenne:
I also fixed a long-time typo/case problem with Histogram_WIDTH

yfarjoun:
looks like the default behavior is unchanged, except for the name of the parameter. I'm assuming that we do not use that parameter in production, but it should be made clear in the "release notes"...so please make sure it is clear in the commit message.

tfenne:
@yfarjoun The reason that has been that way for so long is that Picard's argument names are case-insensitive.  So you can still specify `Histogram_WIDTH` or `HISTOGRAM_WIDTH` or any other mix of case.  It just no longer hurts my eyes in the code.

yfarjoun:
thanks for the clarification.

ktibbett:
@yfarjoun , ok with this?

yfarjoun:
:+1:

PR #399, opened 2016-01-17
Added a new tool for counting the non-N bases in a reference sequence.

eitanbanks:
Can optionally take a set of intervals to restrict the counting to.
Ties out with the numbers we are seeing in production for whole genome territory.


src/java/picard/reference/NonNFastaSize.java
```diff
@@ -0,0 +1,103 @@
+package picard.reference;
```
nh13: license

src/java/picard/reference/NonNFastaSize.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * A tool to count the number of non-N bases in a fasta file
+ */
+@CommandLineProgramProperties(
+        usage = NonNFastaSize.USAGE_SUMMARY + NonNFastaSize.USAGE_DETAILS,
```
nh13: need spacing between the summary and details usage.

src/java/picard/reference/NonNFastaSize.java
```diff
@@ -0,0 +1,103 @@
...
+            "Note that it requires that the fasta file have associated index (.fai) and dictionary (.dict) files.<br />" +
+            "<h4>Usage example:</h4>" +
+            "<pre>" +
+            "java -jar picard.jar NonNFastaSize \\<br />" +
+            "      I=input_sequence.fasta \\<br />" +
+            "      O=count.txt" +
+            "</pre>" +
+            "<hr />"
+            ;
+    @Option(shortName= StandardOptionDefinitions.INPUT_SHORT_NAME, doc="The input FASTA file.")
```
nh13: we have `var= value`, `var=value`, and below `var = value`.  Should we make them all the same?

src/java/picard/reference/NonNFastaSize.java
```diff
@@ -0,0 +1,103 @@
...
+    public File INPUT;
+
+    @Option(shortName= StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc="The output file in which to record the count.")
+    public File OUTPUT;
+
+    @Option(shortName = "INTERVALS", doc = "An interval list file that contains the locations of the positions to assess.  If not provided, the entire reference will be used", optional = true)
+    public File INTERVALS = null;
```
nh13: null is not needed.

src/java/picard/reference/NonNFastaSize.java
```diff
@@ -0,0 +1,103 @@
...
+            "      I=input_sequence.fasta \\<br />" +
+            "      O=count.txt" +
+            "</pre>" +
+            "<hr />"
+            ;
+    @Option(shortName= StandardOptionDefinitions.INPUT_SHORT_NAME, doc="The input FASTA file.")
+    public File INPUT;
+
+    @Option(shortName= StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc="The output file in which to record the count.")
+    public File OUTPUT;
```
nh13: why not just print to stdout and let folks pipe it?  Just a thought.

src/java/picard/reference/NonNFastaSize.java
```diff
@@ -0,0 +1,103 @@
...
+        for (final SAMSequenceRecord rec : ref.getSequenceDictionary().getSequences()) {
+            // pull out the contig and set up the bases
+            final ReferenceSequence sequence = ref.getSequence(rec.getSequenceName());
+            final byte[] bases = sequence.getBases();
+            StringUtil.toUpperCase(bases);
+
+            for (int i = 0; i < bases.length; i++) {
+                // only investigate this position if it's within our mask
+                if (referenceSequenceMask.get(sequence.getContigIndex(), i+1)) {
+                    nonNbases += bases[i] == 'N' ? 0 : 1;
```
nh13: use `SequenceUtil.N` instead of `N` (perhaps overkill).  What about IUPAC bases, does that matter for your purposes?  If so, you could use `!SequenceUtil.isValidBase`.

src/tests/java/picard/reference/NonNFastaSizeTest.java
```diff
@@ -0,0 +1,88 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2013 The Broad Institute
```
nh13: 2013!

src/tests/java/picard/reference/NonNFastaSizeTest.java
```diff
@@ -0,0 +1,88 @@
...
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.lang.Exception;
+import java.lang.String;
+
+/**
+ * @author ebanks
+ */
+
```
nh13: extra nl

src/tests/java/picard/reference/NonNFastaSizeTest.java
```diff
@@ -0,0 +1,88 @@
...
+public class NonNFastaSizeTest {
+
+	private static final String REFERENCE = "testdata/picard/reference/test.fasta";
+
+	@Test
+	public void noIntervals() throws IOException {
+        final File input = new File(REFERENCE);
+        final File outfile   = File.createTempFile("nonNcount", ".txt");
```
nh13: why the spaces before the `=`

nh13:
@eitanbanks: this tool is so specialized as I see it (why not give a set of bases for it to count).  Could you provide a motivation for this tool?

In bash I would have just done something like `cat hs38DH.fa | grep -v "^>" | tr -cd "ACGTacgt" | wc -c` (does not count IUPAC bases).  If I wanted to restrict to regions, I would have used `samtools faidx hs38DH.fa chr1:1-12345 | grep -v "^>" | tr -cd "ACGTacgt" | wc -c` and summed for each interval.

src/java/picard/reference/NonNFastaSize.java
eitanbanks: I got dinged in a previous PR (not by you) for not having the null called out explicitly!
Would prefer to keep it here therefore.

src/java/picard/reference/NonNFastaSize.java
eitanbanks: Is that preferred?  I was trying to be consistent with other picard tools (which all seem to have INPUTs and OUTPUTs).

src/java/picard/reference/NonNFastaSize.java
eitanbanks: Ambiguous IUPAC bases should not be counted the same as Ns.

eitanbanks:
Sure!  The motivation here is that when we move to the hg38 reference (soon!), we don't want to count all of the ALT and HLA contigs as part of the genome territory against which to assess coverage; they account for 10s of millions of bases that shouldn't be part of the coverage calculation.  There have been several times that we've wanted to know exactly how many non-N bases are present in some subset of a reference genome, and this tool represents a super simple way to achieve that answer.  While your bash-foo is exemplary, ultimately it's more error prone than running a simple picard command-line tool (which is why I was asked to put it here).

nh13:
@eitanbanks :+1: from me

eitanbanks:
Okay, squashing and merging.  Thanks!

tfenne:
@eitanbanks I realize this is already merged, but given your stated motivation you might consider using the `ref.fa.alt` file if present, to (optionally) automatically restrict to the core assembly without having to manually (i.e. error-prone-ly) create in interval list.

PR #400, opened 2016-01-18
When using the non-default scoring strategy, the scoring strategy is not

nh13:
applied to the first duplicate set considered in
SimpleMarkDuplicatesWithMateCigar.

@eitanbanks this relates to the change in htsjdk: https://github.com/samtools/htsjdk/pull/440


eitanbanks:
Yup, looks good.  Merging.

PR #401, opened 2016-01-19
Revert "coverage report using jacoco"

yfarjoun:
This reverts commit


yfarjoun:
@akiezun this is not compiling locally (while it is on travis...) happy to have this in when it works.

ktibbett:
:+1:

akiezun:
I'll have a look at this. Weird that travis works and local build fails.

PR #402, opened 2016-01-19
adding an explicit call to ant to travis.yml

lbergelson:
adding a call to 'ant' before 'ant test' since ant with no targets does a different build than 'ant test'
this should catch problems like the jacoco build issue

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,5 @@ before_install:
   - sed -i -e 's_git@github.com:_https://github.com/_' build.xml
 script:
   - ant -lib lib/ant clone-htsjdk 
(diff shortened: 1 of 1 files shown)


lbergelson:
DON'T merge this as is.  I've added a re-revert of the jacoco stuff to test that travis will fail now.

lbergelson:
Once that fails I'll remove that commit and we can merge the updated travis file.

lbergelson:
Ok.  Travis was doing something funny with the pull request commit.

@yfarjoun Good to merge?

yfarjoun:
:+1:

PR #403, opened 2016-01-19
Added read-end quality trimming to SamToFastq.

tfenne:


tfenne:
@ktibbett Care to take a look at this one?  Default behaviour is unchanged, but there's now an option to trim the ends of reads using the same quality trimming as you would get with `bwa aln -q`.

ktibbett:
:+1:

tfenne:
Thanks @ktibbett !

PR #404, opened 2016-01-20
test coverage report by jacoco. 

akiezun:
Take 2 (fixing failing build from https://github.com/broadinstitute/picard/pull/390) 
The build is fixed by excluding jacocoant.jar from picard.jar

to see the report it run
```
ant test-coverage-report
```

the report is generated in the report/ directory. Open the report/index.html file to start navigating

BTW the current coverage is 66% and package-wise looks like this:
image


build.xml
```diff
@@ -263,7 +289,9 @@
        <mkdir dir="${dist}"/>
        <mkdir dir="${dist.tmp}"/>
        <unjar dest="${dist.tmp}">
-           <fileset dir="${lib}" />
+           <fileset dir="${lib}">
+               <exclude name="**/jacocoant.jar"/> <!-- must exclude this jar from packing into picard - otherwise build fails -->
```
yfarjoun: I'd change the comment. The truth is that we do not want jacoco packed together with picard...right?

yfarjoun:
does it take long to run the report? Would it be difficult to set up a badge with the current coverage?

akiezun:
@yfarjoun it takes almost no time but I'm not sure how to do coveralls upload or a badge (because Picard uses Ant for the build system and nothing supports Ant anymore)

yfarjoun:
I see. OK.

yfarjoun:
@akiezun please change the comment and then :+1:

PR #405, opened 2016-01-20
Collect number of chimeric reads

eitanbanks:


src/java/picard/analysis/CollectWgsMetricsFromQuerySorted.java
```diff
@@ -101,6 +104,12 @@
          */
         public long PF_NOISE_READS = 0;
 
+        /**
+         * The number of PF reads that map outside of a maximum insert size (100kb) or that have
+         * the two ends mapping to different chromosomes.
+         */
+        public double PF_CHIMERA_PAIRS;
```
tfenne: `PF_CHIMERIC_PAIRS`?

src/java/picard/analysis/CollectWgsMetricsFromQuerySorted.java
```diff
@@ -270,6 +280,21 @@ private boolean isNoiseRead(final SAMRecord record) {
     }
 
     /**
+     * @param readPair the read pair
+     * @param minimumMappingQuality the minimum mapping quality
+     * @return true if the read pair is considered chimeric, false otherwise
+     */
+    private boolean isChimericReadPair(final QuerySortedReadPairIteratorUtil.ReadPair readPair, final int minimumMappingQuality) {
```
tfenne: Since @nh13 just did something very similar in CollectAlignmentSummaryMetrics it would probably be good to extract this into a re-usable spot so that there's a reasonable fixed definition of Chimeras somewhere.  I could imagine:
```java
public class ChimeraUtil {
    public static int DEFAULT_INSERT_SIZE_LIMIT = 100000;
    public static int DEFAULT_MAPQ_LIMIT = 20;
    public Set<PairOrientation> DEFAULT_EXPECTED_ORIENTATIONS = EnumSet.of(PairOrientation.FR);

    public void isChimeric(final SAMRecord rec);
    public void isChimeric(final SAMRecord r1, final SAMRecord r2);
```
}

eitanbanks:
Good point.  Changed it.

ktibbett:
Can you please change all instances of "adaptor" to "adapter" to match the spelling we have used for this elsewhere in the code base?

eitanbanks:
@tfenne @ktibbett I've refactored things a bit and this PR is much larger than originally planned, so I squashed the commits and am requesting a new review from scratch if that's okay.  Does either of you want to do that?

tfenne:
I’m happy to take a pass through it, but probably won’t have time until tomorrow.

src/java/picard/analysis/AdapterUtil.java
```diff
@@ -0,0 +1,105 @@
...
+    private static final int MAX_ADAPTER_ERRORS = 1;
+
+    public static List<String> DEFAULT_ADAPTER_SEQUENCE = CollectionUtil.makeList(
+            IlluminaUtil.IlluminaAdapterPair.SINGLE_END.get5PrimeAdapter(),
+            IlluminaUtil.IlluminaAdapterPair.SINGLE_END.get3PrimeAdapter(),
+            IlluminaUtil.IlluminaAdapterPair.PAIRED_END.get5PrimeAdapter(),
+            IlluminaUtil.IlluminaAdapterPair.PAIRED_END.get3PrimeAdapter(),
+            IlluminaUtil.IlluminaAdapterPair.INDEXED.get5PrimeAdapter(),
+            IlluminaUtil.IlluminaAdapterPair.INDEXED.get3PrimeAdapter()
+    );
```
tfenne: Perhaps it's time we added `DUAL_INDEXED` to this list?

src/java/picard/analysis/AdapterUtil.java
```diff
@@ -0,0 +1,105 @@
...
+import htsjdk.samtools.util.CollectionUtil;
+import htsjdk.samtools.util.SequenceUtil;
+import htsjdk.samtools.util.StringUtil;
+import picard.util.IlluminaUtil;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class AdapterUtil {
```
tfenne: Could you add a little javadoc to explain what this does and how it's different from `ClippingUtility` that looks vaguely similar?  I.e. that it's trying to match the starts of reads to any part of the adapter, vs. finding the start of the adapter anywhere in the read.

src/java/picard/analysis/ChimeraUtil.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015 The Broad Institute
```
tfenne: 2016

src/java/picard/analysis/AdapterUtil.java
```diff
@@ -0,0 +1,105 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015 The Broad Institute
```
tfenne: 2016

src/java/picard/analysis/ChimeraUtil.java
```diff
@@ -0,0 +1,70 @@
...
+
+import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.SamPairUtil;
+import htsjdk.samtools.SamPairUtil.PairOrientation;
+
+import java.util.EnumSet;
+import java.util.Set;
+
+public class ChimeraUtil {
+
```
tfenne: Channeling Nils - extra newline here.

src/java/picard/analysis/ChimeraUtil.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Checks whether the given read is part of a chimeric pair.
+     *
+     * @param rec                    the read
+     * @param maxInsertSize          max insert size to be considered non-chimeric
+     * @param expectedOrientations   set of orientations that are not chimeric; must not ne null
+     * @return true if this record is part of a chimeric read pair, false otherwise
+     */
+    public static boolean isChimeric(final SAMRecord rec, final int maxInsertSize, final Set<PairOrientation> expectedOrientations) {
+        return (Math.abs(rec.getInferredInsertSize()) > maxInsertSize ||
```
tfenne: I think this method should either:
1. start with `if (rec.getReadUnmappedFlag() || !rec.getReadPairedFlag() || rec.getMateUnmappedFlag()) return false`
2. or make it explicit in the javaodoc that it's only legitimate to call this method on a read that is paired with both it and the mate mapped

src/java/picard/analysis/ChimeraUtil.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Checks whether the given read is part of a chimeric pair.
+     *
+     * @param r1                     first read of the pair
+     * @param r2                     second read of the pair
+     * @param maxInsertSize          max insert size to be considered non-chimeric
+     * @param expectedOrientations   set of orientations that are not chimeric
+     * @return true if this pair is chimeric, false otherwise
+     */
+    public static boolean isChimeric(final SAMRecord r1, final SAMRecord r2, final int maxInsertSize, final Set<PairOrientation> expectedOrientations) {
```
tfenne: Similarly, I think this should either test that the two reads are mapped, or explicitly state that the read must be mapped.

src/java/picard/analysis/CollectAlignmentSummaryMetrics.java
```diff
@@ -98,20 +97,13 @@
     private static final Log log = Log.getInstance(CollectAlignmentSummaryMetrics.class);
 
     @Option(doc="Paired-end reads above this insert size will be considered chimeric along with inter-chromosomal pairs.")
-    public int MAX_INSERT_SIZE = 100000;
+    public int MAX_INSERT_SIZE = ChimeraUtil.DEFAULT_INSERT_SIZE_LIMIT;
 
     @Option(doc="Paired-end reads that do not have this expected orientation will be considered chimeric.")
-    public PairOrientation EXPECTED_PAIR_ORIENTATION = PairOrientation.FR;
+    public Set<PairOrientation> EXPECTED_PAIR_ORIENTATIONS = ChimeraUtil.DEFAULT_EXPECTED_ORIENTATIONS;
```
tfenne: To be safe, this should probably copy-construct the set of orientations.  I believe EnumSet is mutable, and that CommandLineProgram will mutate it!

src/java/picard/analysis/AdapterUtil.java
tfenne: Looping back here, now that I see how this is used.  Why not make this class stateful instead of static?  I.e. turn `prepareAdapterSequences` into a constructor, hide all the `byte[]` and `byte[][]` stuff inside the class, and have the test method become just `public boolean isAdapterSequence(final byte[] read)`.

src/java/picard/analysis/CollectWgsMetricsFromQuerySorted.java
```diff
@@ -195,10 +212,21 @@ private void calculateMetricsForRead(final QuerySortedReadPairIteratorUtil.ReadP
         // now compute metrics...
         metrics.metrics.PF_BASES += totalReadBases;
         if (isNoiseRead(pairToAnalyze.read1)) metrics.metrics.PF_NOISE_READS++;
-        if (!pairToAnalyze.read1.getReadUnmappedFlag()) metrics.metrics.PF_READS_ALIGNED++;
+
+        if (pairToAnalyze.read1.getReadUnmappedFlag()) {
```
tfenne: Could be re-written as the following if you prefer:
```java
if (!pairToAnalyze.read1.getReadUnmappedFlag()) metrics.metrics.PF_READS_ALIGNED++;
else if (isAdapterRead(pairToAnalyze.read1)) metrics.metrics.PF_ADAPTER_READS++;
```

tfenne:
Looks good @eitanbanks - a bunch of suggestions from me, but nothing critical.

src/java/picard/analysis/AdapterUtil.java
eitanbanks: Just spoke to @ktibbett.  I'm not going to add that into this PR because it could change downstream results, but am adding a TODO in the code and will file an issue in github.

eitanbanks:
Reviews addressed.  Thanks, @tfenne.
Squashing and merging.

PR #407, opened 2016-01-20
runCollectTargetedMetricsTest was not writing to the correct output

nh13:
file.  Also fixed up the expected values for the test, and commented
about what it is testing.


nh13:
@tfenne

tfenne:
Thanks @nh13

nh13:
Relies on https://github.com/samtools/htsjdk/pull/444.

PR #410, opened 2016-01-21
Updating the content in the Picard overview.

rcalixte:
Update content to reflect latest changes.


vdauwera:
We're updating all the pages on the user-facing doc website, and for that page we're just going to point people to the Github readme, which has been reorganized to be more readable and comprehensive. It will be easier having only one source for building and other source code-related instructions. Otherwise things eventually drift out of sync.

PR #414, opened 2016-01-25
Adds WMC (Warn on missing contig) option to LiftoverVcf

fleharty:
@yfarjoun This is a new branch (from mf_liftoverbug)
I'd like to kill the other pull request, and use this one instead.

Unfortunately I've lost your previous PR comments, but I have addressed the following:

"break this up into a test and a dataprovider" in LiftoverVcfTest.java

"long name should be descriptive: WARN_ON_MISSING_CONTIG.
as opposed to what? what happens if this is false?" 

"I don't think this should be an input option. make it a protected (for testing) static value."

"use
} else {  
(all in the same line)"

"here too } else {"

"reference -> target reference"

"What about the same protection for LiftOverIntervalList? there you'll see the same problem:
the add() in line 99 will throw an IllegalArgumentException if the contig isn't in the dictionary."

I will address this in a separate pull request

### DIFF
--- a/src/java/picard/vcf/LiftoverVcf.java
+++ b/src/java/picard/vcf/LiftoverVcf.java
@@ -85,6 +85,13 @@ public class LiftoverVcf extends CommandLineProgram {
                     "accompanying sequence dictionary (.dict file).")
     public File REFERENCE_SEQUENCE = Defaults.REFERENCE_FASTA;
 
+    // Option on whether or not to provide a warning, or error message and exit if a missing contig is encountered
+    @Option(shortName = "WMC", doc = "Warn on missing contig.", optional = true)
+    public boolean WARN_ON_MISSING_CONTIG = false;
+
+    // When a contig used in the chain is not in the reference, exit with this value instead of 0.
+    protected static int EXIT_CODE_WHEN_CONTIG_NOT_IN_REFERENCE = 1;
+
     /** Filter name to use when a target cannot be lifted over. */
     public static final String FILTER_CANNOT_LIFTOVER_INDEL = "ReverseComplementedIndel";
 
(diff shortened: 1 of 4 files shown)


src/java/picard/vcf/LiftoverVcf.java
```diff
@@ -174,7 +181,19 @@ public static void main(final String[] args) {
                 rejects.add(new VariantContextBuilder(ctx).filter(reason).make());
                 failedLiftover++;
             }
-            else {
+            else if (!refSeqs.containsValue(target.getContig())) {
```
yfarjoun: } else if ... {

src/java/picard/vcf/LiftoverVcf.java
```diff
@@ -174,7 +181,19 @@ public static void main(final String[] args) {
...
+                final String reason = FILTER_NO_TARGET;
```
yfarjoun: remove this temporary string.

src/java/picard/vcf/LiftoverVcf.java
```diff
@@ -85,6 +85,13 @@
                     "accompanying sequence dictionary (.dict file).")
     public File REFERENCE_SEQUENCE = Defaults.REFERENCE_FASTA;
 
+    // Option on whether or not to provide a warning, or error message and exit if a missing contig is encountered
+    @Option(shortName = "WMC", doc = "Warn on missing contig.", optional = true)
+    public boolean WARN_ON_MISSING_CONGIG = false;
```
yfarjoun: Typo in CONGIG

yfarjoun:
a few comments, then :+1:

PR #417, opened 2016-01-27
Added an option for per-base coverage output...

tfenne:
...for both CollectHsMetrics and CollectTargetedPcrMetrics.

The output file is fairly minimalistic (just chrom,pos,coverage tab-separated) and when gzipped is approximately 10X larger than the per-target output, so I'd generally recommend it only for small targeted panels, not exomes or similar.

### DIFF
--- a/src/java/picard/analysis/directed/TargetMetricsCollector.java
+++ b/src/java/picard/analysis/directed/TargetMetricsCollector.java
@@ -36,6 +36,7 @@ import htsjdk.samtools.reference.ReferenceSequenceFile;
 import htsjdk.samtools.util.CollectionUtil;
 import htsjdk.samtools.util.CoordMath;
 import htsjdk.samtools.util.FormatUtil;
+import htsjdk.samtools.util.IOUtil;
 import htsjdk.samtools.util.QualityUtil;
 import htsjdk.samtools.util.Histogram;
 import htsjdk.samtools.util.Interval;
@@ -83,8 +84,8 @@ public abstract class TargetMetricsCollector<METRIC_TYPE extends MultilevelMetri
     public static final int NEAR_PROBE_DISTANCE_DEFAULT = 250;
     private int nearProbeDistance = NEAR_PROBE_DISTANCE_DEFAULT;
 
-    //If perTargetCoverage != null then coverage is computed for each specified target and output to this file
-    private final File perTargetCoverage;
+    private final File perTargetCoverage;  // If not null, per-target coverage summaries are written to this file
+    private final File perBaseCoverage;    // If not null, per-base(!) coverage summaries are written to this file
 
     //The name of the set of probes used
     private final String probeSetName;
@@ -240,6 +241,7 @@ public abstract class TargetMetricsCollector<METRIC_TYPE extends MultilevelMetri
                                   final List<SAMReadGroupRecord> samRgRecords,
                                   final ReferenceSequenceFile refFile,
                                   final File perTargetCoverage,
+                                  final File perBaseCoverage,
                                   final IntervalList targetIntervals,
                                   final IntervalList probeIntervals,
                                   final String probeSetName,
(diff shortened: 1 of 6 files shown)


tfenne:
@yfarjoun Following up on our chat the other day - this is a good example of how complicated some of the metrics collection code has gotten.  One fairly simple change (about 10-15 lines of real code), requires altering 6 different files to keep everything in sync :disappointed:

tfenne:
@ktibbett Do you want someone on your team to review/bless this, or would an @nh13 review be ok with you?  The "real" change is pretty minimal, and default behaviour is unchanged, but it does touch a bunch of classes.

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -653,6 +663,33 @@ public void finish() {
...
+                final int[] cov = entry.getValue().getDepths();
+                for (int i = 0; i < cov.length; ++i) {
+                    out.print(chrom);
+                    out.print('\t');
+                    out.print(firstBase + i);
+                    out.print('\t');
+                    out.print(interval.getName());
+                    out.print('\t');
+                    out.print(cov[i]);
+                    out.println();
```
nh13: why not:
```
out.print(String.format("%s\t%d\t%s\t%d\n", chrom, firstBase+i, interval.getName(), cov[i]));
```
?

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -653,6 +663,33 @@ public void finish() {
...
+                }
+            }
+
```
nh13: newline is not needed

tfenne: I tned to leave a newline after closing a block, to provide some visual separation.

src/java/picard/analysis/directed/TargetMetricsCollector.java
tfenne: I have a confession to make - I've never loved `printf()` and it's descendants.  It's just not my go-to way of building up strings.

nh13: That's fine, `printf` can wait for you to come around.

ktibbett:
@tfenne , thanks for the heads up (also, completely with you on printf!).  Just because of number of classes touched, I'd like to ask @geoffjentry to have a look as well....

geoffjentry:
soooo mannnnyyyyyy semicolons! :)

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -333,9 +337,8 @@ public TargetMetricsCollector(final Set<MetricAccumulationLevel> accumulationLev
     @Override
     protected PerUnitMetricCollector<METRIC_TYPE, Integer, SAMRecord> makeAllReadCollector() {
         final PerUnitTargetMetricCollector collector = (PerUnitTargetMetricCollector) makeChildCollector(null, null, null);
-        if (perTargetCoverage != null) {
-            collector.setPerTargetOutput(perTargetCoverage);
-        }
+        if (perTargetCoverage != null) collector.setPerTargetOutput(perTargetCoverage);
+        if (perBaseCoverage   != null) collector.setPerBaseOutput(perBaseCoverage);
```
geoffjentry: Thar be some funky spacing up in here

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -392,6 +396,11 @@ public void setPerTargetOutput(final File perTargetOutput) {
             this.perTargetOutput = perTargetOutput;
         }
 
+        /** If set, the metrics collector will output per target coverage information to this file. */
```
geoffjentry: I realize the phrasing of this is because the other one is the same but why are these "if set" and the other setter is "sets"? Just seems weird :)

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -653,6 +663,33 @@ public void finish() {
             return coverageDistribution;
         }
 
+        /** Emits a file with per base coverage if an output file has been set. */
+        private void emitPerBaseCoverageIfRequested() {
+            if (this.perBaseOutput == null) return;
```
geoffjentry: This is where Option would be nice (despite @ktibbett admitting early disdain of it!) because you could just foreach on the value. Alas.

I'm not a huge fan of the "call a function which maybe actually does something" type of function (now watch me have a bunch of them in my git history) vs. "call this function if you want it to do something", but I don't care strongly enough to hold this up.

src/java/picard/analysis/directed/TargetMetricsCollector.java
geoffjentry: While I've taken up the torch of the patented "Nils Newline Nitpick" I agree w/ @tfenne on this one

geoffjentry:
:+1:

src/java/picard/analysis/directed/TargetMetricsCollector.java
tfenne: I rephrased both to:
`        /** Sets the (optional) File to write per-[base/target] coverage information to. If null (the default), no file is produced. */`

src/java/picard/analysis/directed/TargetMetricsCollector.java
tfenne: I totally agree, my main goal here was to keep the `finish()` method from getting much longer or more convoluted, hence the `ifRequested()` in the name to indicate that it might not do anything.  I sorely miss scala when working in Java these days, and `java.util.Optional` isn't really well enough integrated/idiomatic enough to be a good replacement to scala's `Option`.

tfenne:
Thanks @geoffjentry - I've adjusted those comments and squashed.  I'll merge once the tests go green again.

src/java/picard/analysis/directed/TargetMetricsCollector.java
geoffjentry: Yeah it's not a big deal, in fine with with ifRequested but since I'm supposed to be airing every petty grievance there ya go :)

PR #425, opened 2016-01-28
Added yet more tests for Yossi

eitanbanks:


eitanbanks:
Just verbal ok from Yossi

PR #427, opened 2016-01-29
Fixed IlluminaBaseCallsToFastq's dealing with UMI's

yfarjoun:
- fixed IlluminaBaseCallsToFastq's dealing with UMI's. The problem was that the encoding for the sorting collection wasn't done right, but tests didn't have to spill-to-disk and so this wasn't exposed.
- changed tests so that spill-to-disk happens (now tests fail without code changes)
- java8 sugar

### DIFF
--- a/src/java/picard/illumina/IlluminaBasecallsConverter.java
+++ b/src/java/picard/illumina/IlluminaBasecallsConverter.java
@@ -453,8 +453,8 @@ public class IlluminaBasecallsConverter<CLUSTER_OUTPUT_RECORD> {
 
         private synchronized SortingCollection<CLUSTER_OUTPUT_RECORD> newSortingCollection() {
             final int maxRecordsInRam =
-                    maxReadsInRamPerTile /
-                            barcodeRecordWriterMap.size();
+                    Math.max(1, maxReadsInRamPerTile /
+                            barcodeRecordWriterMap.size());
             return SortingCollection.newInstance(
                     outputRecordClass,
                     codecPrototype.clone(),
--- a/src/java/picard/illumina/IlluminaBasecallsToFastq.java
+++ b/src/java/picard/illumina/IlluminaBasecallsToFastq.java
@@ -115,7 +115,7 @@ public class IlluminaBasecallsToFastq extends CommandLineProgram {
     public File MULTIPLEX_PARAMS;
 
     @Option(doc = "Which adapters to look for in the read.")
-    public List<IlluminaUtil.IlluminaAdapterPair> ADAPTERS_TO_CHECK = new ArrayList<IlluminaUtil.IlluminaAdapterPair>(
+    public List<IlluminaUtil.IlluminaAdapterPair> ADAPTERS_TO_CHECK = new ArrayList<>(
             Arrays.asList(IlluminaUtil.IlluminaAdapterPair.INDEXED,
                     IlluminaUtil.IlluminaAdapterPair.DUAL_INDEXED,
                     IlluminaUtil.IlluminaAdapterPair.NEXTERA_V2,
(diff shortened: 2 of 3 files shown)


src/java/picard/illumina/IlluminaBasecallsConverter.java
```diff
@@ -453,8 +453,8 @@ public synchronized void addRecord(final String barcode, final CLUSTER_OUTPUT_RE
 
         private synchronized SortingCollection<CLUSTER_OUTPUT_RECORD> newSortingCollection() {
             final int maxRecordsInRam =
-                    maxReadsInRamPerTile /
-                            barcodeRecordWriterMap.size();
+                    Math.max(1, maxReadsInRamPerTile /
```
eitanbanks: I know you're just trying to protect yourself from 0 here, but 1 is a very weird number to assign to maxRecordsInRam.  Could it be e.g. 1000?

src/java/picard/illumina/IlluminaBasecallsToFastq.java
```diff
@@ -238,8 +233,8 @@ private void initialize() {
             demultiplex = true;
         }
         final int readsPerCluster = readStructure.templates.length() + readStructure.sampleBarcodes.length();
-        basecallsConverter = new IlluminaBasecallsConverter<FastqRecordsForCluster>(BASECALLS_DIR, BARCODES_DIR, LANE, readStructure,
-                sampleBarcodeFastqWriterMap, demultiplex, MAX_READS_IN_RAM_PER_TILE / readsPerCluster, TMP_DIR, NUM_PROCESSORS,
+        basecallsConverter = new IlluminaBasecallsConverter<>(BASECALLS_DIR, BARCODES_DIR, LANE, readStructure,
+                sampleBarcodeFastqWriterMap, demultiplex, Math.max(1, MAX_READS_IN_RAM_PER_TILE / readsPerCluster), TMP_DIR, NUM_PROCESSORS,
```
eitanbanks: Same comment here

src/java/picard/illumina/IlluminaBasecallsToFastq.java
```diff
@@ -487,7 +482,7 @@ public void encode(final FastqRecordsForCluster val) {
             if (numSampleBarcodes != val.sampleBarcodeRecords.length) throw new IllegalStateException();
             encodeArray(val.templateRecords);
             encodeArray(val.sampleBarcodeRecords);
- //           encodeArray(val.molecularBarcodeRecords);
+            encodeArray(val.molecularBarcodeRecords);
```
eitanbanks: Could you please confirm that you did mean to uncomment this?  Because IlluminaBasecallsToFastq is a core tool, I want to make sure this was on purpose.

eitanbanks:
A few questions, but otherwise good

src/java/picard/illumina/IlluminaBasecallsToFastq.java
yfarjoun: I have no idea why it was uncommented to begin with! this was the reason the tool was failing in production. The tests didn't discover it because they were not spilling to disk and so encode wasn't called...I've changed that first (made the tests fail) and then fixed this.

src/java/picard/illumina/IlluminaBasecallsConverter.java
yfarjoun: I want it to be small so that the tests spill to disk, it was previously unprotected and becomes 0 due to integer division...this bumps that 0 up to 1.

eitanbanks: But is 1 a good number to have in production code?

yfarjoun: Not ideal, but if someone is setting the number of reads in memory to be that low, then spillage to disk will ensure. I'm giving enough rope to make the program run (slowly perhaps) without barfing (what would happen if 0 was allowed). This isn't the problem that lead me here in the first place. the actual fix is the un-commented line.

PR #428, opened 2016-01-31
Protect against massive pileups so that we do not OOM even with several Gb of RAM

eitanbanks:
The signature of these pileups are several million copies of the same exact read which, when recorded in the locus iterator, spike our memory consumption.
What we lose by throwing these reads out is that the PCT_EXC_CAPPED will not be completely accurate.  However, I happen to know (from work on Rapid QC) that our lab at least does not look at that number at all. 
What we gain is the ability to always run CollectWgsMetrics and CollectRawWgsMetrics in under 2Gb of RAM, which greatly simplifies pipelines and also keeps compute costs way down.

Depends on: https://github.com/samtools/htsjdk/pull/456

### DIFF
--- a/src/java/picard/analysis/CollectWgsMetrics.java
+++ b/src/java/picard/analysis/CollectWgsMetrics.java
@@ -92,8 +92,11 @@ public class CollectWgsMetrics extends CommandLineProgram {
     @Option(shortName = "Q", doc = "Minimum base quality for a base to contribute coverage.", overridable = true)
     public int MINIMUM_BASE_QUALITY = 20;
 
-    @Option(shortName = "CAP", doc = "Treat bases with coverage exceeding this value as if they had coverage at this value.", overridable = true)
+    @Option(shortName = "CAP", doc = "Treat positions with coverage exceeding this value as if they had coverage at this value (but calculate the difference for PCT_EXC_CAPPED).", overridable = true)
     public int COVERAGE_CAP = 250;
+
+    @Option(doc="At positions with coverage exceeding this value, completely ignore reads that accumulate beyond this value (so that they will not be considered for PCT_EXC_CAPPED).  Used to keep memory consumption in check, but could create bias if set too low", overridable = true)
+    public int LOCUS_ACCUMULATION_CAP = 100000;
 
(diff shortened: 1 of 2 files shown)


src/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -92,9 +92,12 @@
     @Option(shortName = "Q", doc = "Minimum base quality for a base to contribute coverage.", overridable = true)
     public int MINIMUM_BASE_QUALITY = 20;
 
-    @Option(shortName = "CAP", doc = "Treat bases with coverage exceeding this value as if they had coverage at this value.", overridable = true)
+    @Option(shortName = "CAP", doc = "Treat positions with coverage exceeding this value as if they had coverage at this value (but calculate the difference for PCT_EXC_CAPPED).", overridable = true)
     public int COVERAGE_CAP = 250;
 
+    @Option(shortName="LOCUS_ACCUMULATION_CAP", doc="At positions with coverage exceeding this value, completely ignore reads that accumulate beyond this value (so that they will not be considered for PCT_EXC_CAPPED).  Used to keep memory consumption in check, but could create bias if set too low")
+    public int LOCUS_ACCUMULATION_CAP = 100000;
```
yfarjoun: check that LOCUS_ACCUMULATION_CAP is bigger than COVERAGE_CAP. I woudl actualyl recommend that it be much bigger, say 5x times so that the exclusion actually does something.

yfarjoun:
could you change the default value of LOCUS_ACCUMULATION_CAP in the RAW WGS metrics CLP to 200,000?

src/java/picard/analysis/CollectWgsMetrics.java
eitanbanks: It doesn't have to be bigger.  If you don't care about PCT_EXC_CAPPED then it's totally fine to have them equal.
If LOCUS_ACCUMULATION_CAP is smaller than COVERAGE_CAP then I will just raise the LOCUS_ACCUMULATION_CAP and warn the user.

eitanbanks:
Comments addressed.
Will merge once tests pass.

PR #434, opened 2016-02-12
modified usage doc for OPTICAL_DUPLICATE_PIXEL_DISTANCE

yfarjoun:
@tfenne would you agree with this change to the argument doc? I felt that what was there before was not accurate enough.


tfenne:
That seems reasonable to me.

PR #435, opened 2016-02-12
Changed htsjdk clone path

gotolino:
Changed htsjdk clone path to https instead of git for vagrant usage


ktibbett:
I think we're not willing to make this change but would consider a PR that made it possible to do this optionally...  TH\hanks.

PR #437, opened 2016-02-12
Add basic HTML parsing so CLI help is readable

vdauwera:
Fixes the problem we introduced when we started using HTML formatting in tool documentation.

### DIFF
--- a/src/java/picard/cmdline/CommandLineParser.java
+++ b/src/java/picard/cmdline/CommandLineParser.java
@@ -23,6 +23,7 @@
  */
 package picard.cmdline;
 
+import com.google.common.base.CharMatcher;
 import htsjdk.samtools.util.CloserUtil;
 import htsjdk.samtools.util.CollectionUtil.MultiMap;
 import htsjdk.samtools.util.StringUtil;
@@ -230,9 +231,12 @@ public class CommandLineParser {
         } else {
             usagePreamble += defaultUsagePreambleWithPositionalArguments;
         }
+
         if (null != this.programVersion && 0 < this.programVersion.length()) {
             usagePreamble += "Version: " + getVersion() + "\n";
         }
+        checkForNonASCII(usagePreamble, "preamble");
+
         return usagePreamble;
     }
 
@@ -300,7 +304,7 @@ public class CommandLineParser {
      */
     public void usage(final PrintStream stream, final boolean printCommon) {
         if (prefix.isEmpty()) {
-            stream.print(getStandardUsagePreamble(callerOptions.getClass()) + getUsagePreamble());
+            stream.print(convertFromHtml(getStandardUsagePreamble(callerOptions.getClass()) + getUsagePreamble()));
             stream.println("\nVersion: " + getVersion());
             stream.println("\n\nOptions:\n");
 
@@ -338,6 +342,26 @@ public class CommandLineParser {
         for (final CommandLineParser childClp : childClps) {
             childClp.usage(stream, printCommon);
         }
+    }
+
+    private void checkForNonASCII(String documentationText, String location) {
+        if (!CharMatcher.ASCII.matchesAllOf(documentationText)) {
+            throw new AssertionError("Non-ASCII character used in documentation of "+callerOptions.getClass().getSimpleName()+" ("+location+"). Only ASCII characters are allowed.");
+        }
+    }
+
+    private String convertFromHtml(final String textToConvert) {
+
+        final String regex1 = "<br />|<p>|</p>|<table>|</table >|<a href='(.*)'>(.*)</a>|<h4>|<pre>|</pre>|<ul>|</ul>|</li>|</tr>|<hr />";
+        final String regex2 = "<li>";
+        final String regex3 = "</th>";
+        final String regex4 = "</h4>|<tbody>|</tbody>|<tr>|<th>|<.*?>";
+
+        return textToConvert
+                .replaceAll(regex1, "\n")
+                .replaceAll(regex2, " - ")
+                .replaceAll(regex3, "\t")
+                .replaceAll(regex4, "");
     }
 
     private Collection<CommandLineParser> getChildParsersForHelp() {
@@ -398,6 +422,9 @@ public class CommandLineParser {
      */
     private void htmlPrintOptionTableRows(final PrintStream stream, final boolean printCommon) {
         for (final OptionDefinition optionDefinition : optionDefinitions) {
+
+            checkForNonASCII(optionDefinition.doc, optionDefinition.name);
+
             if (!optionDefinition.isCommon || printCommon) {
                 printHtmlOptionUsage(stream, optionDefinition);
             }


ktibbett:
:+1: when tests pass.

src/java/picard/cmdline/CommandLineParser.java
```diff
@@ -340,6 +340,16 @@ public void usage(final PrintStream stream, final boolean printCommon) {
         }
     }
 
+    private String convertFromHtml(final String textToConvert) {
```
yfarjoun: I noticed that there are some-times non-ascii characters embedded into the docs, could you look for these at this point and fail if this happens so that we do not have to rely on reviewers to catch this?

vdauwera:
@yfarjoun Please have a look at the output you get in the CLI help with this branch and let me know if there's anything that you think still looks bad. I think I covered all the HTML tags we use.

vdauwera:
Oh hah y'all already jumped on this, wonderful :)

src/java/picard/cmdline/CommandLineParser.java
vdauwera: I'll see what I can do. When you mean "fail if this happens" you mean fail the build? That would have to happen somewhere else, I think.

src/java/picard/cmdline/CommandLineParser.java
```diff
@@ -340,6 +340,16 @@ public void usage(final PrintStream stream, final boolean printCommon) {
...
+
+        String regex1 = "<br />|<p>|</p>|<table>|</table >|<a href='(.*)'>(.*)</a>|<h4>|<pre>|</pre>|<ul>|</ul>|</li>|</tr>|<hr />";
+        String regex2 = "</h4>|<tbody>|</tbody>|<tr>|<th>";
```
yfarjoun: looking at how (for example) CollectAlignmentSummaryMetrics is set-up, seems that </h4> should also translate to a \n, no?

yfarjoun: what about `<hr />`

yfarjoun: oops I see it now. in the regexp.

src/java/picard/cmdline/CommandLineParser.java
vdauwera: OK, after a quick review I think I can set up a check for non-ascii in the code that builds the html docs, which will fail at build time if non-ascii is detected so whoever added the character won't be able to pass tests.

src/java/picard/cmdline/CommandLineParser.java
yfarjoun: given that the </li> will make a newline, do we need </ul> to make another newline?

yfarjoun: For the case that an html tag was missed should we replace all remaining "<.*?>" with "" ?

yfarjoun: can this be tested somehow? perhaps add an html doc (that uses all the tags you are looking for and some that you are not) and the expected output and make sure that they are the same?

src/java/picard/cmdline/CommandLineParser.java
vdauwera: Added the check. It looks for non-ascii in both the preamble and in the argument definitions (not in names though). If non-ascii is found, throws an AssertionError (seemed appropriate, happy to change the type if someone has a better idea) that reads either e.g.

    java.lang.AssertionError: Non-ASCII character used in documentation of CollectInsertSizeMetrics (DEVIATIONS). Only ASCII characters are allowed.

for an argument definition or

    java.lang.AssertionError: Non-ASCII character used in documentation of CollectInsertSizeMetrics (preamble). Only ASCII characters are allowed.

for a tool doc preamble found in violation of the rule.

To test, add ascii characters into your favorite tool doc and try to build.

src/java/picard/cmdline/CommandLineParser.java
```diff
@@ -340,6 +344,22 @@ public void usage(final PrintStream stream, final boolean printCommon) {
...
+            throw new AssertionError("Non-ASCII character used in documentation of "+callerOptions.getClass().getSimpleName()+" ("+location+"). Only ASCII characters are allowed.");
+        }
+    }
+
+    private String convertFromHtml(final String textToConvert) {
+
+        String regex1 = "<br />|<p>|</p>|<table>|</table >|<a href='(.*)'>(.*)</a>|<h4>|<pre>|</pre>|<ul>|</ul>|</li>|</tr>|<hr />";
+        String regex2 = "</h4>|<tbody>|</tbody>|<tr>|<th>";
+        String regex3 = "<li>";
+        String regex4 = "</th>";
```
nh13: these should be `final`

src/java/picard/cmdline/CommandLineParser.java
vdauwera: > looking at how (for example) CollectAlignmentSummaryMetrics is set-up, seems that should also translate to a \n, no?

Which element are you referring to? All of those on that line or just one?

src/java/picard/cmdline/CommandLineParser.java
```diff
@@ -340,6 +344,22 @@ public void usage(final PrintStream stream, final boolean printCommon) {
...
+
```
nh13: Could you search for any unmatched html tags, in case someone adds one not supported here to the docs?   `regex=(<[^<]+>)`  when the others don't match.

src/java/picard/cmdline/CommandLineParser.java
vdauwera: > given that the </li> will make a newline, do we need </ul> to make another newline?

I wanted to make sure we get a blank line after the list.

src/java/picard/cmdline/CommandLineParser.java
```diff
@@ -340,6 +344,22 @@ public void usage(final PrintStream stream, final boolean printCommon) {
...
+        return textToConvert.replaceAll(regex1, "\n").replaceAll(regex2, "").replaceAll(regex3, " - ").replaceAll(regex4, "\t");
+    }
```
nh13: why not just make an array/list of regexes, and then iterate through them all and `replaceAll`?  It seems easier if we want to add more later.

src/java/picard/cmdline/CommandLineParser.java
vdauwera: > For the case that an html tag was missed should we replace all remaining "<.*?>" with "" ?

Sure -- but I'll have to make sure it's running non-greedy then.

src/java/picard/cmdline/CommandLineParser.java
```diff
@@ -398,6 +418,9 @@ public void htmlPrintOptions(final PrintStream stream, final boolean printCommon
      */
     private void htmlPrintOptionTableRows(final PrintStream stream, final boolean printCommon) {
         for (final OptionDefinition optionDefinition : optionDefinitions) {
+
+            checkForNonASCII(optionDefinition.doc, optionDefinition.name);
```
nh13: what about in the command line tool's description?  Does this cover that case?  It would be nice to have any printed text be checked for non-ASCII characters and html codes.

nh13:
@vdauwera your call, but just made a few comments in case @yfarjoun is enjoying AGBT too much.

src/java/picard/cmdline/CommandLineParser.java
vdauwera: > can this be tested somehow? perhaps add an html doc (that uses all the tags you are looking for and some that you are not) and the expected output and make sure that they are the same?

If I had infinite time I would say yes, but in a time-constrained world I think it's overkill. I'm happy eyeballing a few docs and set the threshold at "do they look reasonable/readable?". If people find problems they'll be easy to fix quickly now that we have this set up.

src/java/picard/cmdline/CommandLineParser.java
vdauwera: Yep Yossi already requested that.

src/java/picard/cmdline/CommandLineParser.java
vdauwera: "preamble" covers both `usage` and `usageShort`. Is there any other description text?

src/java/picard/cmdline/CommandLineParser.java
vdauwera: It was simpler this way when I started out with one regex for testing, then built on that. Now I'm too lazy to change it (and life is short).

src/java/picard/cmdline/CommandLineParser.java
vdauwera: Note to self: the ? as proposed makes it non-greedy. Duh.

src/java/picard/cmdline/CommandLineParser.java
vdauwera: done

vdauwera:
@yfarjoun @nh13 Addressed your comments (including adding a catch-all pattern to clean up residue); except I'm taking a pass on setting up a formal test and on rewriting the regexing as a loop because it's Friday afternoon and my kid is being unbearably cute. 

I eyeballed a bunch of tools' docs and everything looked good enough for a first pass. If anything looks funky just let me know and we can iterate on that quickly. I think right now the priority should be to get this out asap to relieve the pain of those who use the CLI help. 

Unless there are any further/major concerns I'll merge when tests pass, as blessed by @ktibbett earlier.

yfarjoun:
OK. but in that case, please add a ticket to circle back and

- add a test for the results
- add a test for non-ascii character
- rewriting the regexp using a stream/fold or forloop.

:+1:

vdauwera:
Squashed, will merge when tests pass. 

Follow up ticket in #439

yfarjoun:
do you know why one of the tests isn't running?

vdauwera:
Nope. In another thread Jeff mentioned Travis might be acting up, I think.
On Fri, Feb 12, 2016 at 8:11 PM Yossi Farjoun <notifications@github.com>
wrote:

> do you know why one of the tests isn't running?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/broadinstitute/picard/pull/437#issuecomment-183550323>
> .
>
-- 
Geraldine A. Van der Auwera, PhD
Manager of Outreach and Communications
Data Sciences and Data Engineering
Broad Institute

vdauwera:
Continued in Yossi's branch.

PR #441, opened 2016-02-13
Yf try more formatting help

yfarjoun:
@vdauwera took your pr #437  and added tests and the functionality we discussed.

care to look and comment?

### DIFF
--- a/src/java/picard/cmdline/CommandLineParser.java
+++ b/src/java/picard/cmdline/CommandLineParser.java
@@ -23,9 +23,11 @@
  */
 package picard.cmdline;
 
+import com.google.common.base.CharMatcher;
 import htsjdk.samtools.util.CloserUtil;
 import htsjdk.samtools.util.CollectionUtil.MultiMap;
 import htsjdk.samtools.util.StringUtil;
+import picard.PicardException;
 
 import java.io.BufferedReader;
 import java.io.File;
@@ -48,8 +50,8 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-
-import picard.PicardException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * Annotation-driven utility for parsing command-line arguments, checking for errors, and producing usage message.
@@ -177,19 +179,19 @@ public class CommandLineParser {
     private int maxPositionalArguments;
 
     // List of all the data members with @Option annotation
-    private final List<OptionDefinition> optionDefinitions = new ArrayList<OptionDefinition>();
+    private final List<OptionDefinition> optionDefinitions = new ArrayList<>();
 
     // Maps long name, and short name, if present, to an option definition that is
     // also in the optionDefinitions list.
-    private final Map<String, OptionDefinition> optionMap = new HashMap<String, OptionDefinition>();
+    private final Map<String, OptionDefinition> optionMap = new HashMap<>();
 
     // Maps child options prefix to CommandLineParser for the child object.
     // Key: option prefix.
-    private final Map<String, CommandLineParser> childOptionsMap = new LinkedHashMap<String, CommandLineParser>();
+    private final Map<String, CommandLineParser> childOptionsMap = new LinkedHashMap<>();
 
     // Holds the command-line arguments for a child option parser.
     // Key: option prefix.  Value: List of arguments for child corresponding to that prefix (with prefix stripped).
-    private final MultiMap<String, ChildOptionArg> childOptionArguments = new MultiMap<String, ChildOptionArg>();
+    private final MultiMap<String, ChildOptionArg> childOptionArguments = new MultiMap<>();
 
     // For printing error messages when parsing command line.
     private PrintStream messageStream;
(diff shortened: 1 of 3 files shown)


src/java/picard/cmdline/CommandLineParser.java
```diff
@@ -299,8 +304,11 @@ public String getVersion() {
      * @param stream Where to write the usage message.
      */
     public void usage(final PrintStream stream, final boolean printCommon) {
+
         if (prefix.isEmpty()) {
-            stream.print(getStandardUsagePreamble(callerOptions.getClass()) + getUsagePreamble());
+            final String prefix = htmlUnescape(convertFromHtml(getStandardUsagePreamble(callerOptions.getClass()) + getUsagePreamble()));
+            checkForNonASCII(prefix, "prefix");
```
vdauwera: "prefix" is a weird way to refer to this part of the text. Prefix is generally understood as a short string prepended to some other string, like a name or identifier. If you really don't like "preamble" (which to be honest I'm not crazy about either, I was just using the existing nomenclature) then why not use something more descriptive like "tool description". Though for the variable name I would stick with "preamble" since that's how it's referred to in the rest of the code, for readability and consistency.

src/java/picard/cmdline/CommandLineParser.java
```diff
@@ -363,7 +395,7 @@ public void htmlUsage(final PrintStream stream, final String programName, final
         stream.println("<a id=\"" + programName + "\"/>");
         stream.println("<h3>" + programName + "</h3>");
         stream.println("<section>");
-        stream.println("<p>" + getUsagePreamble() + "</p>");
+        stream.println("<p>" + htmlEscape(getUsagePreamble()) + "</p>");
```
vdauwera: Nice, I should have thought of that.

src/java/picard/cmdline/CommandLineParser.java
```diff
@@ -405,14 +440,23 @@ private void htmlPrintOptionTableRows(final PrintStream stream, final boolean pr
         for (final CommandLineParser childParser : getChildParsersForHelp()) {
             childParser.htmlPrintOptionTableRows(stream, false);
         }
+    }
+
+    private static final Map<String, String> htmlToText = new LinkedHashMap<String, String>(){{
+        put("<","&lt;");
```
vdauwera: This kills a whole bunch of tags; for example see in the source for MarkDuplicatesWithMateCigar:

````
<p>Identifies duplicate reads, accounting for mate CIGAR.  This tool locates and tags duplicate reads (both PCR and optical) in a BAM or SAM file, where duplicate reads are defined as originating from the same original fragment of DNA, taking into account the CIGAR string of read mates. &lt;br />&lt;br />It is intended as an improvement upon the original MarkDuplicates algorithm, from which it differs in several ways, includingdifferences in how it breaks ties. It may be the most effective duplicate marking program available, as it handles all cases including clipped and gapped alignments and locates duplicate molecules using mate cigar information. However, please note that it is not yet used in the Broad's production pipeline, so use it at your own risk. &lt;br />&lt;br />Note also that this tool will not work with alignments that have large gaps or deletions, such as those from RNA-seq data.  This is due to the need to buffer small genomic windows to ensure integrity of the duplicate marking, while large skips (ex. skipping introns) in the alignment records would force making that window very large, thus exhausting memory. &lt;br />&lt;h4>Usage example:&lt;/h4>&lt;pre>java -jar picard.jar MarkDuplicatesWithMateCigar \&lt;br />      I=input.bam \&lt;br />      O=mark_dups_w_mate_cig.bam \&lt;br />      M=mark_dups_w_mate_cig_metrics.txt&lt;/pre>&lt;hr /></p>
````

Comment (vdauwera) [inline src/java/picard/cmdline/CommandLineParser.java]: This function should not be invoked within the htmlUsage processing

File: src/java/picard/cmdline/CommandLineParser.java (hunk 2 above)
Comment (vdauwera) [inline src/java/picard/cmdline/CommandLineParser.java]: Actually I meant I should have thought of the unescape the unescape for regular usage. Why would you want to strip out html elements in the html usage section???

**Comment by vdauwera** [comment]
Some good additions but use of the htmlEscape function to process the htmlUsage is problematic -- in the sense that it kills most of the html formatting. See comments (with screenshots) for details. FYI to check the effect of your changes on the html, see the html files in dist/html/program_usage.

**Comment by yfarjoun** [comment]
oops. good catch. I'll fix.

File: src/java/picard/cmdline/CommandLineParser.java [hunk 4]
```diff
@@ -363,7 +395,7 @@ public void htmlUsage(final PrintStream stream, final String programName, final
         stream.println("<a id=\"" + programName + "\"/>");
```
vdauwera: While you're at it (I know it's not your code, but boyscout rules apply) -- this is evil. It causes many browsers to color all the text after this in link blue until the next closing `</a>` tag.

This line should be deleted and the next line should be replaced by         

        stream.println("<h3 id=\"" + programName + "\">" + programName + "</h3>"); 

So we can still get the convenient anchor effect without the formatting wrongness. 

I made the corresponding change in the autogenerated pages on my fork's gh-pages branch, if you want to test that anchoring still works.

yfarjoun:
according to discussion here: http://stackoverflow.com/questions/30535268/regex-to-strip-out-greater-than-and-less-than-characters-from-html-string-ig
  
one shouldn't look for `<` and `>` in html using regex. one needs a full on parser. So I'm taking out the escapeHtml function and we will require that the usage be written html-escaped

yfarjoun:
want to take another look @vdauwera ?

vdauwera:
Makes sense. 

Yep, will have a look tonight.

vdauwera:
Looks good to me. Status says there are some conflicts; I'll see if I can resolve them. If yes then I'll squash and then I think we'll be ready to merge.

vdauwera:
Rebased and squashed; conflicts are gone; waiting for tests to pass.

vdauwera:
@yfarjoun we're good to go. Anyone else need to opine on this before we merge?

yfarjoun:
nope.

PR #447, opened 2016-02-16
Formatting tweaks to user-facing docs

vdauwera:
Includes edits to Tool usage pages that need to be replicated from within the auto-docs system to be permanent.
Also makes explain-flags page look nicer.


vdauwera:
Note to self: don't delete the gh-pages branch after merging!

PR #453, opened 2016-02-21
squid:HiddenFieldCheck - Local variables should not shadow class fields

m-ezzat:
This pull request is focused on resolving occurrences of Sonar squid:HiddenFieldCheck - Local variables should not shadow class fields

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:HiddenFieldCheck

M-Ezzat

### DIFF
--- a/src/java/picard/analysis/directed/TargetMetricsCollector.java
+++ b/src/java/picard/analysis/directed/TargetMetricsCollector.java
@@ -674,15 +674,15 @@ public abstract class TargetMetricsCollector<METRIC_TYPE extends MultilevelMetri
                 final String chrom = interval.getContig();
                 final int firstBase = interval.getStart();
 
-                final int[] cov = entry.getValue().getDepths();
-                for (int i = 0; i < cov.length; ++i) {
+                final int[] covLocal = entry.getValue().getDepths();
+                for (int i = 0; i < covLocal.length; ++i) {
                     out.print(chrom);
                     out.print('\t');
                     out.print(firstBase + i);
                     out.print('\t');
                     out.print(interval.getName());
                     out.print('\t');
-                    out.print(cov[i]);
+                    out.print(covLocal[i]);
                     out.println();
                 }
             }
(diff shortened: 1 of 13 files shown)


yfarjoun:
This looks fine. Thanks. :+1:

tfenne:
@yfarjoun FWIW I don't think this should have been merged.  IMHO this makes it more likely that bugs will be introduced during maintenance, not less.  Simply appending the word `Local` to a bunch of variable names so that they don't shadow instance properties does not seem like an improvement.

yfarjoun:
Thanks for chiming in, @tfenne!

My thinking was that when reviewing small pieces of code (as one often does
during a review) could be difficult to know whether a variable is local or
not. This change at least let's the reviewer know that there's something
going on and that there are two similarly types variables with a similar
name.

What is the advantage, in your view of having the same variable name for a
variable and a property? How does this help maintenance?

tfenne:
I'm not necessarily arguing that shadowing is _good_, just that it's not always _bad_.  But what I think is bad is mechanically altering code like this to meet some arbitrary definition of "good".

Calling things `covLocal`, `prefixLocal`, etc. does the following for me:
1) Make me wonder where `covRemote` or `covNonLocal` is and what it might be, even though it's often not at all relevant to the method in question.  I.e. it misleads me into thinking there is more going on than there is.
2) Makes it more likely that I or someone else will accidentally reference the instance members instead in future, because `cov` is more natural than `covLocal`, and now if I type `cov` that's both valid and the wrong thing.

If someone really cares enough to fix these things then they should be fixed with thought and care, not mechanically, perhaps by finding more meaningful names for either the instance variables or the local variables.

akiezun:
Can we have a process, a'la Apache projects, by which the reviewer asks committers for opinions and code only gets merged if it does **not** get a -1 from anyone? See here http://www.apache.org/foundation/voting.html#votes-on-code-modification
For example, we could say that all PRs are decided by lazy consensus 2 days after the reviewer puts his/her initial +1. 

It would have helped here because @tfenne would have voted -1 which would have blocked the merge.

Just an idea.

yfarjoun:
I see your points @tfenne, and I agree that there should be more time given
for :-1:

I'll revert this.

PR #455, opened 2016-02-22
squid:S1197 - Array designators [] should be on the type, not the var…

m-ezzat:
This pull request is focused on resolving occurrences of Sonar squid:S1197 - Array designators "[]" should be on the type, not the variable

You can find more information about the issue here: 
https://dev.eclipse.org/sonar/coding_rules#q=squid:S1197

M-Ezzat

### DIFF
--- a/src/java/picard/analysis/directed/TargetMetricsCollector.java
+++ b/src/java/picard/analysis/directed/TargetMetricsCollector.java
@@ -604,8 +604,8 @@ public abstract class TargetMetricsCollector<METRIC_TYPE extends MultilevelMetri
             double totalCoverage = 0;
 
             // The "how many bases at at-least X" calculations.
-            final int targetBasesDepth[] = {0, 1, 2, 10, 20, 30, 40, 50, 100}; // NB: this should be in ascending order
-            final int targetBases[] = new int[targetBasesDepth.length]; // counts for how many target bases are at at least X coverage, where X corresponds to the value at the same offset in targetBasesDepth
+            final int[] targetBasesDepth = {0, 1, 2, 10, 20, 30, 40, 50, 100}; // NB: this should be in ascending order
+            final int[] targetBases = new int[targetBasesDepth.length]; // counts for how many target bases are at at least X coverage, where X corresponds to the value at the same offset in targetBasesDepth
 
             // consider all bases in calculating the mean, median etc.
             for (final Coverage c : this.coverageByTarget.values()) {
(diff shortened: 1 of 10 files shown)


yfarjoun:
hmmmm. This seems to be aesthetic. the rule could have been defined as the opposite:

"Array designators should always be located on the _variable_ for better code readability. Otherwise, developers must look both at the type and the variable name to know whether or not a variable is an array."...

yfarjoun:
personally :-1:

lbergelson:
This is a completely aesthetic decision.  I agree though that the `[]` should go with the type though, since the brackets are part of the type.  It's a bizarre that they're even allowed on the variable name.

akiezun:
java style guide clearly wants brackets on the type: https://google.github.io/styleguide/javaguide.html#s4.8.3.2-array-declarations

i agree they should not have been allowed on the variable name. @yfarjoun do you agree that **all** declarations should use the same style?  If yes, what style do you think predominates (by orders of magnitude more examples)?

yfarjoun:
a cursory glance indicates indeed many more appear on the type, and in
return types of functions it isn't even possible (I think) to put on the
"variable" name...so I agree that this is the reasonable choice :+0:

akiezun:
thanks @yfarjoun. I'll merge in 1 day unless someone objects with a good reason

akiezun:
@yfarjoun BTW, FYI it is possible to put the array-ness on the 'variable name' in return types, using the following horrible style (warning, what follows may be disturbing to some people, reader discretion is advised):
```
    int foo()[]{
        return null;
    }
```

And it gets worse: you can put some dimensions on the type and some more on the variable - this function returns a 2D array
```
    int[] foo()[]{
        return null;
    }
```

you can do it on variables too - this is 3D array:
```
 int[] foo[][];
```

akiezun:
@lbergelson do you know why tests are have not run here on travis?

pgrosu:
@akiezun it possibly because @m-ezzat might not be a member with write access to [picard](https://github.com/broadinstitute/picard/) based on the following:

https://docs.travis-ci.com/user/pull-requests

The specific line I'm referring to is the following:

> For the protection of secure data, Travis CI makes it available only on pull requests coming from the same repository. These are considered trustworthy, as only members with write access to the repository can send them.

~p

lbergelson:
@akiezun @pgrosu I'm not sure why they didn't run on travis.  In general travis runs tests on pull requests from external users but doesn't provide encrypted variables to them.  Other pull requests from the same user have built on travis.  This seems like a bug.  I suspect that somehow travis didn't hear the notification that github sent out and never did a build because of that.

@m-ezzat Could you make a tiny change somewhere and push it to force travis to run?  You can then force push back to the original request.

m-ezzat:
@lbergelson Ok, I've triggered another build

lbergelson:
@m-ezzat Thanks!  Sorry about that.  I don't think there is any other way to force the tests to run if they didn't get kicked off initially.

m-ezzat:
@lbergelson no worries :)

PR #457, opened 2016-02-22
restore correct usage of 'recordsRead' local variable

jabbarish:
Restore correct usage of 'recordsRead' local variable in EstimateLibraryComplexity

It seems that increment of the variable was accidentally deleted during the following refactoring:
git show ee98947e72d81ec7539217ef76334888b24ced46 -- **/EstimateLibraryComplexity.java

This pull request restores it.

This local variable influences on the groups omitting during the final phase of ELC.

### DIFF
--- a/src/java/picard/sam/markduplicates/EstimateLibraryComplexity.java
+++ b/src/java/picard/sam/markduplicates/EstimateLibraryComplexity.java
@@ -406,7 +406,7 @@ public class EstimateLibraryComplexity extends AbstractOpticalDuplicateFinderCom
         log.info("Will store " + MAX_RECORDS_IN_RAM + " read pairs in memory before sorting.");
 
         final List<SAMReadGroupRecord> readGroups = new ArrayList<SAMReadGroupRecord>();
-        final int recordsRead = 0;
+        int recordsRead = 0;
         final SortingCollection<PairedReadSequence> sorter;
         final boolean useBarcodes = (null != BARCODE_TAG || null != READ_ONE_BARCODE_TAG || null != READ_TWO_BARCODE_TAG);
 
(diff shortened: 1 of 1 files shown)


tfenne:
Sorry about that @jabbarish !  Looking at your change, wouldn't it also work to just eliminate `recordsRead` entirely and replace any usages with `progress.getCount()`?

jabbarish:
@tfenne ,  thank you for the quick reply!
I am afraid progress.getCount() would cause problems in the future changes: we are going to propose a series of pull requests to introduce internal parallelizm. Usage of progress.getCount() in multi-threaded environment would require either changes in it, or syncing access to it via locks.  On the other hand, it would be easy to change local variable to lock-free Atomic, or find other ways to count processed records. Thus, in order to encapsulate threadsafe-related activity in the EstimateLibraryComplexity as much as possible, I would propose not to use progress.getCount()

pgrosu:
Probably the easiest is to just add the `synchronized` keyword to the methods.

jabbarish:
@pgrosu ,  synchronized  will affect (slow down) all metrics which use the ProgressLogger class.

pgrosu:
@jabbarish ok, then how about [log4j](http://logging.apache.org/log4j/2.x/)?

pgrosu:
Another alternative is to compartmentalize it via something like [Jetty](http://www.eclipse.org/jetty/), to keep the application and log service as separate as possible.

jabbarish:
@pgrosu,   I think it is not the case to use such major things.  It is just a counter.

@tfenne, I have looked at the code in AbstractProgressLogger and realized that synchronized is already used there in record() methods. Thus, we can use progress.getCount(), indeed.   The only thing I would propose to change in AbstractProgressLogger is to make 'long processed' field volatile, to provide atomic read of it's value it in getCount().   I will prepare another pull request.

jabbarish:
I've just created the new pull request: https://github.com/broadinstitute/picard/pull/458
Thanks to all for the comments!

PR #460, opened 2016-02-23
Adding a category and plot to WgsMetrics

nh13:
@yfarjoun this pull request is not complete, as it requires updating and adding tests, as well as one final pass.  What I wanted to do prior to putting any more effort into this change, is to get a :+1: that if I fix current tests and add some more tests, and respond to all the comments, that this will get merged.  If not, I can move the version of this tool elsewhere.  My reasoning is that this commit changes `WgsMetrics`, which may impact folks who rely on this class having a certain format, so it could be a painful change for some.  Nonetheless, I think the plot that is produced is useful, and the QC metrics (for non-zero coverage regions) are useful for folks doing low-coverage sequencing for technology development.

If you agree that this PR will be included in Picard, I will go back and do the rest of the work and then ask for a review.  Thanks for taking a look.

### DIFF
--- a/src/java/picard/analysis/CollectWgsMetrics.java
+++ b/src/java/picard/analysis/CollectWgsMetrics.java
@@ -23,6 +23,7 @@
  */
 package picard.analysis;
 
+import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SamReader;
 import htsjdk.samtools.SamReaderFactory;
 import htsjdk.samtools.filter.SamRecordFilter;
@@ -31,12 +32,17 @@ import htsjdk.samtools.metrics.MetricBase;
 import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.reference.ReferenceSequence;
 import htsjdk.samtools.reference.ReferenceSequenceFileWalker;
-import htsjdk.samtools.util.*;
+import htsjdk.samtools.util.Histogram;
+import htsjdk.samtools.util.IOUtil;
+import htsjdk.samtools.util.Log;
+import htsjdk.samtools.util.ProgressLogger;
+import htsjdk.samtools.util.QualityUtil;
+import htsjdk.samtools.util.SamLocusIterator;
 import picard.cmdline.CommandLineProgram;
 import picard.cmdline.CommandLineProgramProperties;
 import picard.cmdline.Option;
-import picard.cmdline.programgroups.Metrics;
 import picard.cmdline.StandardOptionDefinitions;
+import picard.cmdline.programgroups.Metrics;
 import picard.filter.CountingDuplicateFilter;
 import picard.filter.CountingFilter;
 import picard.filter.CountingMapQFilter;
@@ -47,7 +53,6 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
-import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 /**
@@ -74,7 +79,7 @@ public class CollectWgsMetrics extends CommandLineProgram {
             "</pre>" +
             "Please see " +
             "<a href='https://broadinstitute.github.io/picard/picard-metric-definitions.html#CollectWgsMetrics.WgsMetrics'>" +
-            "the WgsMetrics documentation</a>for detailed explanations of the output metrics." +
+            "the WgsMetrics documentation</a> for detailed explanations of the output metrics." +
             "<hr />";
 
     @Option(shortName = StandardOptionDefinitions.INPUT_SHORT_NAME, doc = "Input SAM or BAM file.")
@@ -110,11 +115,14 @@ public class CollectWgsMetrics extends CommandLineProgram {
     @Option(doc="Sample Size used for Theoretical Het Sensitivity sampling. Default is 10000.", optional = true)
     public int SAMPLE_SIZE=10000;
 
+    private SAMFileHeader header = null;
+
     private final Log log = Log.getInstance(CollectWgsMetrics.class);
     private static final double LOG_ODDS_THRESHOLD = 3.0;
 
     /** Metrics for evaluating the performance of whole genome sequencing experiments. */
     public static class WgsMetrics extends MetricBase {
+
         /** The number of non-N bases in the genome reference over which coverage will be evaluated. */
         public long GENOME_TERRITORY;
         /** The mean coverage in bases of the genome territory, after all filters are applied. */
@@ -198,8 +206,9 @@ public class CollectWgsMetrics extends CommandLineProgram {
         final ReferenceSequenceFileWalker refWalker = new ReferenceSequenceFileWalker(REFERENCE_SEQUENCE);
         final SamReader in = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(INPUT);
         final SamLocusIterator iterator = getLocusIterator(in);
+        this.header = in.getFileHeader();
 
-        final List<SamRecordFilter> filters = new ArrayList<SamRecordFilter>();
+        final List<SamRecordFilter> filters = new ArrayList<>();
         final CountingFilter dupeFilter = new CountingDuplicateFilter();
         final CountingFilter mapqFilter = new CountingMapQFilter(MINIMUM_MAPPING_QUALITY);
         final CountingPairedFilter pairFilter = new CountingPairedFilter();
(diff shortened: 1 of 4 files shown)


yfarjoun:
If I understand correctly, the new version will emit two lines instead of 1, with one line as before, and another line that only collects information about "non-zero" bases. Also there is a coverage plot that is added that will show both coverage plots (all and non-zero). 

This seems targeted at users that have low-coverage files so that the zero-coverage sites overwhelm the rest and it could be difficult to see the coverage over the non-zero sites. For "deep" samples this should be a non-issue since the zero-covered sites should be a small blip in the plot. 

Do you find that you need both modes at the same time? would it be easier/less disruptive to have a flag switch between the two different modes of operation? I think that the second line is that useful in general (also because truncated Poisson's are harder to analyze than regular Poisson's and so this might be more confusing than helping....I think. 

Other than the removal of the i=0 bin in the histogram (and the resulting difference to the metrics),  is there differences between the two lines?

nh13:
In an automated pipeline having to figure out which mode to run is not always the easiest and running it twice is inefficient. Also some of the metrics across both categories are important in both low and high coverage situations. The two histograms are both useful, even the truncated one. I can see differences in good versus bad experiments (good and bad coverage uniformity). Your intuition is correct in most cases above, it is just that I have found the two categories and histograms useful at various times. I have used a modified version of this tool for over 150 experiments to give you a sense.

yfarjoun:
I'm still not clear about the histograms. is the truncated one the same as
the non-truncated except for the coverage=0 bin?

yfarjoun:
Since you have used this already in many experiments, could you give me an
example of how you would use the new results and show that the old results
were insufficient for this purpose? I'm still struggling to understand what
new information it contains. Sorry for being so thick.

nh13:
> is the truncated one the same as the non-truncated except for the coverage=0 bin?
Your exactly right.  

A few things:
1. For the plotting, having the mean coverage calculated without zero-coverage regions is important for the curve "non-zero coverage regions".  I _could_ probably calculate this from the whole genome generated histogram.  The denominator (# of non-zero sites) would be the the sum of values excluding the zero-coverage bin.  The numerator would be the sum of `(coverage * bin_value)` for all but the zero-coverage bin.  This would eliminate the need for a second histogram.
2. For the metrics, the various metrics require a second row, since we compute means and medians ignoring zero coverage sites.
3. To visually assess the coverage uniformity on a whole genome, we can plot the fraction (or percentage) of bases covered versus the fold mean coverage (normalized coverage).  Ideally, we want want to see a large fraction of bases covered at small fold (fraction) of the mean coverage.  A whiteboard would be useful here.  If we are assessing various prep conditions, we may not be able to afford to sequence each condition to 30x coverage, so we sequence it to 0.1x, 1x, or something less than 30x.  The former plot becomes not useful, as the denominator is too large: we just don't sequence enough to get the y-axis (fraction of bases) above zero.  But we want to know how is coverage uniformity on the bases we _do_ cover?  This is where the second plot, and set of metrics, that consider only the covered bases as the target region are useful.  Hopefully I am making sense, and if not, I'll grab you in-person.

yfarjoun:
I think I understand what you mean, but if so, I disagree with your claim
:-) yeah. let's meet and talk about this in person.

nh13:
@yfarjoun in the meantime, I am going to move these changes to a sub-class.

nh13:
@yfarjoun this is ready for review.  If you have scientific concerns, lets discuss in person, as this is may be a new tool to Picard, but its utility has been shown to the users who want this in a public place.

src/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -73,8 +78,8 @@
             "      R=reference_sequence.fasta " +
             "</pre>" +
             "Please see " +
-            "<a href='https://broadinstitute.github.io/picard/picard-metric-definitions.html#CollectWgsMetrics.WgsMetrics'>" +
-            "the WgsMetrics documentation</a>for detailed explanations of the output metrics." +
+            "<a href='https://broadinstitute.github.io/picard/picard-metric-definitions.html#CollectWgsMetrics.WgsMetricsWithNonZeroCoverage'>" +
```
yfarjoun: why reference this metric here? wrong file?

src/java/picard/analysis/CollectWgsMetricsWithNonZeroCoverage.java
```diff
@@ -0,0 +1,171 @@
...
+        usage = CollectWgsMetricsWithNonZeroCoverage.USAGE_SUMMARY + CollectWgsMetricsWithNonZeroCoverage.USAGE_DETAILS,
+        usageShort = CollectWgsMetricsWithNonZeroCoverage.USAGE_SUMMARY,
+        programGroup = Metrics.class
+)
+public class CollectWgsMetricsWithNonZeroCoverage extends CollectWgsMetrics {
+
+    static final String USAGE_SUMMARY = "Collect metrics about coverage and performance of whole genome sequencing (WGS) experiments.";
+    static final String USAGE_DETAILS = "This tool collects metrics about the percentages of reads that pass base- and mapping- quality " +
+            "filters as well as coverage (read-depth) levels. Both minimum base- and mapping-quality values as well as the maximum " +
+            "read depths (coverage cap) are user defined.  This extends CollectWgsMetrics by including metrics related to only sites" +
```
yfarjoun: to only -> only to

src/java/picard/analysis/CollectRawWgsMetrics.java
```diff
@@ -68,8 +68,8 @@
             "</pre>" +
             "<hr />" +
             "Please see " +
-            "<a href='https://broadinstitute.github.io/picard/picard-metric-definitions.html#CollectWgsMetrics.WgsMetrics'>" +
-            "the WgsMetrics documentation</a> for detailed explanations of the output metrics." +
+            "<a href='https://broadinstitute.github.io/picard/picard-metric-definitions.html#CollectWgsMetrics.WgsMetricsWithNonZeroCoverage'>" +
```
yfarjoun: this shouldn't have been changed either, right?

src/java/picard/analysis/CollectWgsMetricsWithNonZeroCoverage.java
```diff
@@ -0,0 +1,171 @@
...
+import picard.filter.CountingPairedFilter;
+import picard.util.RExecutor;
+
+import java.io.File;
+import java.util.List;
+
+@CommandLineProgramProperties(
+        usage = CollectWgsMetricsWithNonZeroCoverage.USAGE_SUMMARY + CollectWgsMetricsWithNonZeroCoverage.USAGE_DETAILS,
+        usageShort = CollectWgsMetricsWithNonZeroCoverage.USAGE_SUMMARY,
+        programGroup = Metrics.class
```
yfarjoun: Could this be in the alpha group for now? I hope that we will be able to discuss and shrink the growing collection of Collect*WGSMetrics back into one program in the near future....I'd rather not collect API calls we need to support.

yfarjoun:
minor comments. :+1: thanks.

src/java/picard/analysis/CollectRawWgsMetrics.java
nh13: It is a new metrics class with a new field, so why not?

nh13: Good catch: let's blame intellij.

src/java/picard/analysis/CollectWgsMetrics.java
nh13: Good catch: let's blame intellij.

src/java/picard/analysis/CollectWgsMetricsWithNonZeroCoverage.java
nh13: Done

src/java/picard/analysis/CollectRawWgsMetrics.java
yfarjoun: so confused...this is CollectRawWgsMetrics.java, why are you sending the reader to WgsMetricsWithNonZeroCoverage?

nh13: oh wow, you're right, it shouldn't be.  fixing now

yfarjoun:
:+1:

nh13:
@yfarjoun rebased onto master and squashed: ready to merge after tests pass.

src/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -73,8 +78,8 @@
             "      R=reference_sequence.fasta " +
             "</pre>" +
             "Please see " +
-            "<a href='https://broadinstitute.github.io/picard/picard-metric-definitions.html#CollectWgsMetrics.WgsMetrics'>" +
-            "the WgsMetrics documentation</a>for detailed explanations of the output metrics." +
+            "<a href='https://broadinstitute.github.io/picard/picard-metric-definitions.html#CollectWgsMetrics.CollectWgsMetrics'>" +
```
yfarjoun: Still a broken link. This should end in CollectWgsMetrics.WgsMetrics as it needs to point to the metric not the collector. also please revert the text in the `<a>`. 

while you're here, could you please add a space after the `</a>`?

src/java/picard/analysis/CollectWgsMetricsWithNonZeroCoverage.java
```diff
@@ -0,0 +1,171 @@
...
+            "<pre>"  +
+            "java -jar picard.jar CollectWgsMetricsWithNonZeroCoverage \\<br /> " +
+            "      I=input.bam \\<br /> "+
+            "      O=collect_wgs_metrics.txt \\<br /> " +
+            "      CHART=collect_wgs_metrics.pdf  \\<br /> " +
+            "      R=reference_sequence.fasta " +
+            "</pre>" +
+            "Please see " +
+            "<a href='https://broadinstitute.github.io/picard/picard-metric-definitions.html#CollectWgsMetricsWithNonZeroCoverage.WgsMetricsWithNonZeroCoverage'>" +
+            "the WgsMetricsWithNonZeroCoverage documentation</a>for detailed explanations of the output metrics." +
```
yfarjoun: add a space here too after `</a>`

src/java/picard/analysis/CollectWgsMetricsWithNonZeroCoverage.java
```diff
@@ -0,0 +1,171 @@
...
+import picard.filter.CountingPairedFilter;
+import picard.util.RExecutor;
+
+import java.io.File;
+import java.util.List;
+
+@CommandLineProgramProperties(
+        usage = CollectWgsMetricsWithNonZeroCoverage.USAGE_SUMMARY + CollectWgsMetricsWithNonZeroCoverage.USAGE_DETAILS,
+        usageShort = CollectWgsMetricsWithNonZeroCoverage.USAGE_SUMMARY,
+        programGroup = Alpha.class
```
yfarjoun: thanks.

yfarjoun:
sorry, @nh13 two more comments.

nh13:
@yfarjoun thanks for checking those links.  I used vim this time to make the changes and will blame IntelliJ for any past errors.  After the tests pass, it is ready to merge.

PR #461, opened 2016-02-23
squid:S2864 - entrySet() should be iterated when both the key and val…

m-ezzat:
This pull request is focused on resolving occurrences of Sonar squid:S2864 - "entrySet()" should be iterated when both the key and value are needed

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S2864

M-Ezzat


tfenne:
Sorry to be so negative, but I'm :-1: on this one too.  While iterating the entry set is more efficient, it's my opinion that:
1) in non-performance sensitive sections of code it's often clearer to iterate the keys and pull the values out inside the loop
2) the names given to the variables in this PR appear to be mechanically generated to satisfy the arbitrary rule, and do not reflect the semantics of the data, and thus make things less clear

yfarjoun:
I want to strengthen your fingertips @tfenne. Commits should be to make the
codebase better, not to blindly adhere to a coding standard.

akiezun:
closing then, right?

PR #464, opened 2016-02-27
adding a test-case for the chimeric read that simply has an SA tag bu…

### LINKED ISSUE #426: Add a test to CollectAlignmentSummaryMetrics for reads that have the SA tag
This is the only type of chimeric read type that we don't currently test

yfarjoun:
…t is otherwise not chimeric

resolves #426 

@eitanbanks please review.


eitanbanks:
Looks good

PR #465, opened 2016-02-27
squid:S2275 - Printf-style format strings should not lead to unexpected

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule squid:S2275 - Printf-style format strings should not lead to unexpected behavior at runtime

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S2275

M-Ezzat

### DIFF
--- a/src/java/picard/cmdline/PicardCommandLine.java
+++ b/src/java/picard/cmdline/PicardCommandLine.java
@@ -228,7 +228,7 @@ public class PicardCommandLine {
 
             if (!commandListOnly) {
                 builder.append(KWHT + "--------------------------------------------------------------------------------------\n" + KNRM);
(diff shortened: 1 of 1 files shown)


yfarjoun:
Title and comment of the PR are misleading, as there's no "unexpected" results in the output.

The only thing you did here was to change the '\n' in the format string in CommandlineProgram (the base class only) to '%n' which (i had to look it up) mean that the output will be "platform independent".

What about all the '\n' in the usage strings? Every commandline program defines USAGE string that are emitted during runtime and have lots of newline characters. It seems pointless to change just a few of the \n's in the output. 

Anyone see the utility of this?

yfarjoun:
:-1: due to lack of interest.

PR #466, opened 2016-02-28
trying to fix the intellij setting so that it points to the 1.8 java sdk

yfarjoun:
@ktibbett, we talked about this a while back. Can you take a look?

### DIFF
--- a/Picard-public.iml
+++ b/Picard-public.iml
@@ -5,7 +5,7 @@
       <configuration sdkName="" />
     </facet>
   </component>
-  <component name="NewModuleRootManager" inherit-compiler-output="false">
+  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_1_8" inherit-compiler-output="false">
     <output url="file://$MODULE_DIR$/intellij.classes" />
     <output-test url="file://$MODULE_DIR$/intellij.testclasses" />
     <exclude-output />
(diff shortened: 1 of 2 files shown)


lbergelson:
@yfarjoun We'd be better off deleting the project files than trying to keep them up to date I think.  Everyone in the world has their own unique IDE settings and committing project files means everyone has to fight with the committed ones.

yfarjoun:
hmmm. the point of having this is to help people have a good initial setup
of the project that "works" or very nearly works. I suspect that it is
helpful for folks who are just starting with picard. I know that setting up
GATK (which doesn't have intelliJ settings) is a royal pain if you do not
have someone telling you what to do...so I'm inclined to leaving these
around.

I'm happy to fix them every now and again when they break.

tfenne:
I think the ideal situation is to, eventually, move over to a build system that provides better IDE integration.  My experience with projects that use either _gradle_ or _sbt_ has been that importing into IntelliJ does 99% of the configuration I need and want.  And then the build file(s) live in git, but the IntelliJ files don't because they can be auto-generated easily.

akiezun:
+1 for gradle

tfenne:
Just to be contrarian - the more I use sbt, the more I like it :smile:

yfarjoun:
I have absolutely no problem with an alternative system that makes it easy
to start developing. until then, I think that the intelliJ files should be
included.

cwhelan:
Another option is to provide an build target that generates IntelliJ files.
For example in Gradle the idea plugin can do this:

https://docs.gradle.org/current/userguide/idea_plugin.html

Customizing the task in the build file lets you fix any things that IDEA's
imports don't get 100% correct, if you keep it up to date. I worked on a
project before where this seemed to work well.

ktibbett:
:+1: as the interim solution

Issue #467, opened 2016-02-28
Running CollectSequencingArtifactMetrics with CollectMultipleMetrics and a FILE_EXTENSION does not produce the correct file name


nh13:
It looks like the `OUTPUT` is not being set correctly here:
https://github.com/broadinstitute/picard/blob/master/src/java/picard/analysis/CollectMultipleMetrics.java#L277
@davidangb it looks like perhaps it was an addition you made.

yfarjoun:
Are you seeing a problem with the output or just seeing it from the code?
CollectSequencingArtifactMetrics is supposed to take OUTPUT and generate 4
different output files with different extensions....


On Sun, Feb 28, 2016 at 1:00 PM, Nils Homer <notifications@github.com>
wrote:
>
> It looks like the OUTPUT is not being set correctly here:
>
https://github.com/broadinstitute/picard/blob/master/src/java/picard/analysis/CollectMultipleMetrics.java#L277

nh13:
In the output.

nh13:
@yfarjoun I am also seeing that `SUMMARY_OUTPUT` for `CollectGcBiasMetrics` is not updated as well: https://github.com/broadinstitute/picard/blob/master/src/java/picard/analysis/CollectMultipleMetrics.java#L226

I have to work around both issues in my use of `CollectMultipleMetrics`.

yfarjoun:
Thanks @nh13! I'll add test for that and fix!

yfarjoun:
oh, you're fixing it! thanks!

yfarjoun:
this is fixed, right @nh13 ?

nh13:
It is, closing.

PR #468, opened 2016-02-29
CollectMultipleMetric and CollectSequencingArtifactMetrics use

nh13:
FILE_EXTENSION properly.

Addresses https://github.com/broadinstitute/picard/issues/467

@yfarjoun


yfarjoun:
:+1:  thanks! I'll add a test to it and merge, if that's OK with you!

yfarjoun:
testing this is annoyingly hard....because the existing test files require extra arguments to be passed to the PROGRAM and there's no mechanism to do that....merging as is. thanks!

PR #470, opened 2016-03-01
Moves CollectWgsMetricsFromSampledSites functionality and tests to Co…

meganshand:
…llectWgsMetrics

### DIFF
--- a/src/java/picard/analysis/CollectWgsMetrics.java
+++ b/src/java/picard/analysis/CollectWgsMetrics.java
@@ -34,6 +34,7 @@ import htsjdk.samtools.reference.ReferenceSequence;
 import htsjdk.samtools.reference.ReferenceSequenceFileWalker;
 import htsjdk.samtools.util.Histogram;
 import htsjdk.samtools.util.IOUtil;
+import htsjdk.samtools.util.IntervalList;
 import htsjdk.samtools.util.Log;
 import htsjdk.samtools.util.ProgressLogger;
 import htsjdk.samtools.util.QualityUtil;
@@ -115,6 +116,13 @@ public class CollectWgsMetrics extends CommandLineProgram {
     @Option(doc="Sample Size used for Theoretical Het Sensitivity sampling. Default is 10000.", optional = true)
     public int SAMPLE_SIZE=10000;
 
+    @Option(shortName = "INTERVALS", doc = "An interval list file that contains the positions to restrict the assessment. Please note that " +
+            "all bases of reads that overlap these intervals will be considered, even if some of those bases extend beyond the boundaries of " +
+            "the interval. The ideal use case for this argument is to use it to restrict the calculation to a subset of (whole) contigs. To " +
+            "restrict the calculation just to individual positions without overlap, please see CollectWgsMetricsFromSampledSites.",
+            optional = true, overridable = true)
+    public File INTERVALS = null;
+
     private SAMFileHeader header = null;
 
     private final Log log = Log.getInstance(CollectWgsMetrics.class);
@@ -194,6 +202,9 @@ public class CollectWgsMetrics extends CommandLineProgram {
         IOUtil.assertFileIsReadable(INPUT);
         IOUtil.assertFileIsWritable(OUTPUT);
         IOUtil.assertFileIsReadable(REFERENCE_SEQUENCE);
+        if (INTERVALS != null) {
+            IOUtil.assertFileIsReadable(INTERVALS);
+        }
 
         // it doesn't make sense for the locus accumulation cap to be lower than the coverage cap
         if (LOCUS_ACCUMULATION_CAP < COVERAGE_CAP) {
@@ -265,12 +276,16 @@ public class CollectWgsMetrics extends CommandLineProgram {
         return new WgsMetrics();
     }
 
+    /**
+     * If INTERVALS is specified, this will count bases beyond the interval list when the read overlaps the intervals and extends beyond the
+     * edge. Ideally INTERVALS should only include regions that have hard edges without reads that could extend beyond the boundary (such as a whole contig).
+     */
     protected long getBasesExcludedBy(final CountingFilter filter) {
         return filter.getFilteredBases();
     }
 
     protected SamLocusIterator getLocusIterator(final SamReader in) {
-        return new SamLocusIterator(in);
+        return (INTERVALS != null) ? new SamLocusIterator(in, IntervalList.fromFile(INTERVALS)) : new SamLocusIterator(in);
     }
 
     protected WgsMetricsCollector getCollector(final int coverageCap) {
--- a/src/java/picard/analysis/CollectWgsMetricsFromSampledSites.java
+++ b/src/java/picard/analysis/CollectWgsMetricsFromSampledSites.java
@@ -46,10 +46,11 @@ import java.io.File;
                 "whole genome sequencing experiments, but only at a set of sampled positions.  " +
                 "It is important that the sampled positions be chosen so that they are spread out " +
                 "at least further than a read's length apart; otherwise, you run the risk of double-counting " +
-                "reads in the metrics.",
+                "reads in the metrics. If contig-sized intervals are needed, use INTERVALS argument in CollectWgsMetrics.",
         usageShort = "Writes whole genome sequencing-related metrics for a SAM or BAM file",
         programGroup = Metrics.class
 )
+
 public class CollectWgsMetricsFromSampledSites extends CollectWgsMetrics {
 
     @Option(shortName = "INTERVALS", doc = "An interval list file that contains the locations of the positions to assess.", optional = false)
--- a/src/tests/java/picard/analysis/CollectWgsMetricsFromSampledSitesTest.java
+++ b/src/tests/java/picard/analysis/CollectWgsMetricsFromSampledSitesTest.java
@@ -27,10 +27,15 @@ import htsjdk.samtools.metrics.MetricsFile;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 import picard.cmdline.CommandLineProgramTest;
+import picard.cmdline.PicardCommandLine;
 
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 /**
  * Tests CollectWgsMetricsFromSampledSites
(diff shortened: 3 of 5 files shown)


meganshand:
This consolidates CollectWgsMetrics and CollectWgsMetricsFromSampledSites. @yfarjoun Could you please take a look?

nh13:
@yfarjoun this could be a painful rebase for me at #460.  How would you like them ordered.  My PR will be ready shortly.

src/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -110,6 +110,9 @@
     @Option(doc="Sample Size used for Theoretical Het Sensitivity sampling. Default is 10000.", optional = true)
     public int SAMPLE_SIZE=10000;
 
+    @Option(shortName = "INTERVALS", doc = "An interval list file that contains the locations of the positions to assess. It is important that the sampled positions be chosen so that they are spread out at least further than a read's length apart; otherwise, you run the risk of double-counting reads in the metrics.", optional = true)
```
eitanbanks: Instead of "sampled positions" say "intervals"

src/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -350,10 +353,17 @@ protected WgsMetrics generateWgsMetrics() {
     }
 
     protected long getBasesExcludedBy(final CountingFilter filter) {
+        if(INTERVALS != null){
+            return filter.getFilteredRecords();
+        }
         return filter.getFilteredBases();
```
eitanbanks: It's very important to copy over the comment from CollectWgsMetricsFromSampledSites because the behavior is different.  You may even want to mention this in the docs for the INTERVALS argument too.

eitanbanks: Also, this whole method could be shortened to a single line:
return (INTERVALS != null) ? filter.getFilteredRecords() : filter.getFilteredBases();

Oh, and another thing: protocol in this codebase is to put spaces after the 'if' and before the braces.  Otherwise, you risk incurring the wrath of @nh13.

src/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -350,10 +353,17 @@ protected WgsMetrics generateWgsMetrics() {
...
     }
 
     protected SamLocusIterator getLocusIterator(final SamReader in) {
+        if(INTERVALS != null){
+            IOUtil.assertFileIsReadable(INTERVALS);
```
eitanbanks: Same note here about the spacing

eitanbanks:
Review complete on my end.  @yfarjoun may still want to take a look.

src/java/picard/analysis/CollectWgsMetrics.java
nh13: exactly @eitanbanks

src/java/picard/analysis/CollectWgsMetrics.java
nh13: Could you add this check to the start of the `doWork` method? I think the pattern is to call the `IOUtil` asserts as soon as possible directly in the `doWork` the method.

meganshand:
I made the changes from @eitanbanks and @nh13. Could someone please take another look?

eitanbanks:
This looks really good to me.  @yfarjoun do you want to take a quick look too given that it's a commonly used tool?

nh13:
It would be great if #460 could be merged first, and this rebased on top of that, since this is a much smaller change.

yfarjoun:
Yes. let's do that. I'll review #460 later today. thanks for the
understanding @nh13

yfarjoun:
need to remove line from xml as well.

yfarjoun:
since it's deprecated, I'd rather folks not find the tool.

src/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -349,12 +358,18 @@ protected WgsMetrics generateWgsMetrics() {
...
 
+    /**
+     * By design, if an INTERVALS file is provided, we want to count just those bases at the positions we care about,
+     * not across the entire read. Therefore, we call filter.getFilteredRecords() so that only the bases in the pileup
+     * at a given position are included in the calculations (with filter.getFilteredBases() we would be including other
+     * bases in the read too).
+     */
     protected long getBasesExcludedBy(final CountingFilter filter) {
-        return filter.getFilteredBases();
+        return (INTERVALS != null) ? filter.getFilteredRecords() : filter.getFilteredBases();
```
yfarjoun: this only works when "sampled sites" are a single base here and there. This needs to work when INTERVALS is complete contigs...and I think it doesn't. @eitanbanks could you help @meganshand figure out how to make it work...this might be a little more effort than we first thought.

src/tests/java/picard/analysis/CollectWgsMetricsTest.java
```diff
@@ -101,6 +101,66 @@ public void testMetricsFromWGS(final File input, final File outfile, final Strin
...
+            Assert.assertEquals(metrics.PCT_EXC_MAPQ, 0.272727); // 3 of 11
+            Assert.assertEquals(metrics.PCT_EXC_DUPE, 0.181818); // 2 of 11
+            Assert.assertEquals(metrics.PCT_EXC_UNPAIRED, 0.090909); // 1 of 9
+            Assert.assertEquals(metrics.PCT_EXC_BASEQ, 0.090909); // 1 of 9
+            Assert.assertEquals(metrics.HET_SNP_SENSITIVITY, 0.34655, .02);
+        }
+    }
+
+    @Test
+    public void testContiguousIntervals() throws IOException {
```
yfarjoun: could you add a test that has two close loci and one far away in the INTERVAL? I think the code will currently fail on that test.

src/java/picard/analysis/CollectWgsMetrics.java
eitanbanks: Yeah, that's hard.  That's why it was called "FromSampledSites".
@yfarjoun what if we check up front and fail if any of the intervals are greater than 1 base?

yfarjoun: but they will be! we want to use this to filter on 1-22,x,y,mt...right?

eitanbanks: Oh, duh.  Great point.  You are totally right.
Can we whiteboard this later today?

nh13:
@yfarjoun my suggestion was to merge this PR _after_ #460.  My apologies if that was not clear.  Does that still work for you?

src/java/picard/analysis/CollectWgsMetrics.java
nh13: Sorry if I am missing something, but aren't you using a `SamLocusIterator` that returns only records overlapping the intervals, and so you could just return `getFilteredBases` in either case?  What base would be counted that shouldn't?

yfarjoun:
@nh13 my position is unchanged. We will merge #460  first and then #470. Please do not fret. :ok: ?

src/java/picard/analysis/CollectWgsMetrics.java
yfarjoun: The difficulty is that we want this to work when the INTERVALS are isolated loci or large intervals...

meganshand:
After talking to @eitanbanks, we decided to keep CollectWGSMetricsFromSampledSites and also add an argument to CollectWGSMetrics that will take interval lists with larger intervals such as a whole contig. As long as the LARGE_INTERVALS don't have reads that can overhang an edge it should work. For smaller intervals with overhanging reads CollectWGSMetricsFromSampledSites should still be used. Could you please take another look @yfarjoun?

src/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -110,6 +110,12 @@
     @Option(doc="Sample Size used for Theoretical Het Sensitivity sampling. Default is 10000.", optional = true)
     public int SAMPLE_SIZE=10000;
 
+    @Option(shortName = "LARGE_INTERVALS", doc = "An interval list file that contains contig-sized locations of the positions to assess. It is important that " +
+            "the intervals be chosen so that they won't have overlapping reads that extend beyond the boundaries of the interval. For smaller sites use " +
+            "CollectWgsMetricsFromSampledSites.",
+            optional = true)
```
eitanbanks: I think I would prefer to call this just INTERVALS and change the doc text to something like this:

An interval list file that contains the positions to restrict the assessment.  Please note that all bases of reads that overlap these intervals will be considered, even if some of those bases extend beyond the boundaries of the interval.   The ideal use case for this argument is to use it to restrict the calculation to a subset of (whole) contigs.  To restrict the calculation just to individual positions without overlap, please see CollectWgsMetricsFromSampledSites.

What do you guys think?

meganshand: I think if we want this argument to be INTERVALS, then the argument in CollectWgsMetricsFromSampledSites will need to be something other than INTERVALS. I thought it was better to not change the FromSampledSites version in case there were scripts out there already using it.

eitanbanks: I defer to @yfarjoun

yfarjoun: I think it's fine to use the same argument as long as the doc is clear.

eitanbanks: Note that one is optional and the other is required.

meganshand:
I changed the doc to the one suggested by @eitanbanks because I thought it was clearer and I changed the argument name for both tools to INTERVALS. Should be ready for another look, @yfarjoun?

yfarjoun:
@h13's PR has been merged.

src/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -110,6 +110,13 @@
     @Option(doc="Sample Size used for Theoretical Het Sensitivity sampling. Default is 10000.", optional = true)
     public int SAMPLE_SIZE=10000;
 
+    @Option(shortName = "INTERVALS", doc = "An interval list file that contains the positions to restrict the assessment. Please note that " +
+            "all bases of reads that overlap these intervals will be considered, even if some of those bases extend beyond the boundaries of " +
+            "the interval. The ideal use case for this argument is to use it to restrict the calculation to a subset of (whole) contigs. To " +
+            "restrict the calculation just to individual positions without overlap, please see CollectWgsMetricsFromSampledSites.",
+            optional = true, overridable = true)
```
yfarjoun: why overridable = true?

src/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -349,12 +359,16 @@ protected WgsMetrics generateWgsMetrics() {
         return new WgsMetrics();
     }
 
+    /**
+     * If INTERVALS is specified, this will count bases beyond the interval list when the read overlaps the intervals and extends beyond the
+     * edge. INTERVALS should only include regions that have hard edges without reads that could extend beyond the boundary (such as a whole contig).
```
yfarjoun: I would add the word "Ideally" before "INTERVALS"

src/java/picard/analysis/CollectWgsMetrics.java
meganshand: So that it can be overridden by the INTERVALS argument to CollectWgsMetricsFromSampledSites. Without it if both tools have an INTERVALS argument and you get a command line parser error when you run CollectWgsMetricsFromSampledSites.

src/java/picard/analysis/CollectWgsMetricsFromSampledSites.java
```diff
@@ -46,10 +46,11 @@
                 "whole genome sequencing experiments, but only at a set of sampled positions.  " +
                 "It is important that the sampled positions be chosen so that they are spread out " +
                 "at least further than a read's length apart; otherwise, you run the risk of double-counting " +
-                "reads in the metrics.",
+                "reads in the metrics. If a contig-sized interval is needed, use INTERVALS argument in CollectWgsMetrics.",
```
yfarjoun: "**if co**ntig-sized interval**s are** needed"

src/tests/java/picard/analysis/CollectWgsMetricsFromSampledSitesTest.java
```diff
@@ -101,4 +101,36 @@ public void testContiguousIntervals() throws IOException {
...
+                "SAMPLE_SIZE=" + sampleSize
+        };
+        Assert.assertEquals(runPicardCommandLine(args), 0);
+
+        final MetricsFile<CollectWgsMetrics.WgsMetrics, Comparable<?>> output = new MetricsFile<CollectWgsMetrics.WgsMetrics, Comparable<?>>();
+        output.read(new FileReader(outfile));
+
+        for (final CollectWgsMetrics.WgsMetrics metrics : output.getMetrics()) {
+            Assert.assertEquals(metrics.GENOME_TERRITORY, 404);
+            Assert.assertNotEquals(metrics.PCT_EXC_MAPQ, 0.271403);
```
yfarjoun: This is too easy a test to pass. could you subset the "forMetrics.sam" file (temporarily) to only contain contig chr7, run (manually) the normal WGSMetrics and record the results from there? then do an assert*Equals* rather than a *not*equals...

yfarjoun: oh, nm. this is testing CollectWgsMetricsFromSampledSites...

yfarjoun: I'm concerned that it isn't clear where these numbers are coming from. An alternative is to run CollectWGSMetrics here as well and compare the metrics directly, rather than "remembering" the results from the other test.

yfarjoun:
- minor changes to text.
- a request for improving the test. 
- you need to rebase, you have conflicts.

src/java/picard/analysis/CollectWgsMetrics.java
yfarjoun: got it. thanks!

meganshand:
@yfarjoun I changed the comments and fixed the test as you requested. It's only testing if CollectWgsMetrics returns a different result than CollectWgsMetricsFromSampledSites if you use the exact same inputs (same INTERVALS list).

meganshand:
@yfarjoun I added the comment and rebased. Could you take one more quick look to make sure the rebase went smoothly?

src/tests/java/picard/analysis/CollectWgsMetricsFromSampledSitesTest.java
```diff
@@ -101,4 +106,53 @@ public void testContiguousIntervals() throws IOException {
...
+        args.put("OUTPUT", "OUTPUT=" + collectWgsOutfile.getAbsolutePath());
+
+        CollectWgsMetrics collectWgsMetrics = new CollectWgsMetrics();
+        collectWgsMetrics.instanceMain(args.values().toArray(new String[]{}));
+
+        final MetricsFile<CollectWgsMetrics.WgsMetrics, Comparable<?>> collectWgsMetricsOutput = new MetricsFile<CollectWgsMetrics.WgsMetrics, Comparable<?>>();
+        collectWgsMetricsOutput.read(new FileReader(collectWgsOutfile));
+
+        for (final CollectWgsMetrics.WgsMetrics metrics : output.getMetrics()) {
+            Assert.assertEquals(metrics.GENOME_TERRITORY, 404);
```
yfarjoun: extract this "404" to a variable up near the definition of the file. (same in the test below)

yfarjoun: ah, nevermind...too petty :-)

yfarjoun:
:+1:

PR #474, opened 2016-03-08
Adding logging to opticalDuplicateFinder when duplicate set is large

### LINKED ISSUE #472: MarkDuplicates dies a horrible death when large duplicates sets exists.
The code for finding optical duplication is implemented as a n^2 code. We have seen a case when n is .5e6 and this breaks the code. While the data is dubious, it would be good to get an output from the program. I think that the "findOpticalDuplicates" code should be implemented using a kdtree rather than a brute force approach. 

Any thoughts?

pinging @nh13 @tfenne @davidbenjamin

yfarjoun:
Resolves #472 by slowing *down* and providing "safe points" to the code (I guess)...

This seems to make the problem go away..

@nh13 @tfenne It seems that you are the authors of this code, would one of you be able to take a look? I didn't change anything functional, just added a few logging and progress statements when the duplicate set are large. These cases should be _very_ rare (duplicate sets greater than 1000)

### DIFF
--- a/src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
+++ b/src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2014 The Broad Institute
+ * Copyright (c) 2014-2016 The Broad Institute
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -25,11 +25,10 @@
 package picard.sam.markduplicates.util;
 
 import htsjdk.samtools.util.Log;
+import htsjdk.samtools.util.ProgressLogger;
 import picard.sam.util.PhysicalLocation;
 import picard.sam.util.ReadNameParser;
 
-import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
 
 /**
@@ -43,14 +42,28 @@ public class OpticalDuplicateFinder extends ReadNameParser {
     public int opticalDuplicatePixelDistance;
 
     public static final int DEFAULT_OPTICAL_DUPLICATE_DISTANCE = 100;
+    public static final int DEFAULT_BIG_DUPLICATE_SET_SIZE = 1000;
 
     /**
-     * Uses the default duplicate distance {@value DEFAULT_OPTICAL_DUPLICATE_DISTANCE} and the default read name regex
+     * Sets the size of a set that is big enough to log progress about.
+     * Defaults to {@value OpticalDuplicateFinder#DEFAULT_BIG_DUPLICATE_SET_SIZE}
+     *
+     * @param bigDuplicateSetSize the size of a set that is big enough to log progress about
+     */
+    public void setBigDuplicateSetSize(final int bigDuplicateSetSize) {
+        this.bigDuplicateSetSize = bigDuplicateSetSize;
+    }
+
+    private int bigDuplicateSetSize;
+
+    /**
+     * Uses the default duplicate distance {@value OpticalDuplicateFinder#DEFAULT_OPTICAL_DUPLICATE_DISTANCE} and the default read name regex
      * {@link ReadNameParser#DEFAULT_READ_NAME_REGEX}.
      */
     public OpticalDuplicateFinder() {
         super();
         this.opticalDuplicatePixelDistance = DEFAULT_OPTICAL_DUPLICATE_DISTANCE;
+        this.bigDuplicateSetSize = DEFAULT_BIG_DUPLICATE_SET_SIZE;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2014 The Broad Institute
```
nh13: `2014-2016`?

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -87,20 +86,41 @@ public OpticalDuplicateFinder(final String readNameRegex, final int opticalDupli
 
         final PhysicalLocation actualKeeper = keeperOrNull(list, keeper);
 
+        final Log log = Log.getInstance(OpticalDuplicateFinder.class);
+        ProgressLogger progressLogger = new ProgressLogger(log, 10000, "examined", "ReadEnds to keeper");
```
nh13: could be final?

nh13: Also, I think it should only be created if the length is too large.

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -87,20 +86,41 @@ public OpticalDuplicateFinder(final String readNameRegex, final int opticalDupli
...
+        if(length > 1000) {
```
nh13: Can this be a parameter to this method?  And then have another version sets this to `1000`.  I definitely see this being adjusted by folks operating on duplicate sets.

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -87,20 +86,41 @@ public OpticalDuplicateFinder(final String readNameRegex, final int opticalDupli
...
-            for (int i=0; i<length; ++i) {
+            for (int i = 0; i < length; ++i) {
                 final PhysicalLocation other = list.get(i);
                 opticalDuplicateFlags[i] = closeEnough(actualKeeper, other, distance);
+
+                // The main point of adding this log and if statement (also below) is a workaround a bug in the JVM
+                // which causes a deep exception (https://github.com/broadinstitute/picard/issues/472)
+                // it seems that this is related to https://bugs.openjdk.java.net/browse/JDK-8033717 which
+                // was closed due to non-reproducibility. We came across a bam file that evoked this error
+                // every time we tried to mark duplicates it. It had a duplicate-set of size 500,000, and
```
nh13: `time we tried to mark duplicates it. It had a duplicate-set` -> `time we tried to duplicate mark it, it had a duplicate-set`.

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
nh13: cannot be final.

src/tests/java/picard/sam/markduplicates/util/OpticalDuplicateFinderTest.java
```diff
@@ -1,13 +1,10 @@
 package picard.sam.markduplicates.util;
 
-import htsjdk.samtools.util.CollectionUtil;
 import htsjdk.samtools.util.Log;
-import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 import org.testng.Assert;
 import picard.sam.util.PhysicalLocation;
 import picard.sam.util.PhysicalLocationInt;
-import picard.sam.util.PhysicalLocationShort;
```
nh13: brutal

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
yfarjoun: thanks for the comments @nh13. to be clear, you want the 1000 value to be a parameter to the method? Would it be OK if I make it a private member of the class with a setter and default 1000?

src/tests/java/picard/sam/markduplicates/util/OpticalDuplicateFinderTest.java
yfarjoun: This was all that remained from a big change where I tested the new version with a KDTree...but then I realized that the logging itself solved the problem...while the kdtree is much faster for the crazy edge case that I'm solving here, for most cases it is not...so I decided against using KDtree for faster analysis of an edge case that happened once in 5 years!

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
nh13: Your call, I just want to be able to set it somehow in some code I have.

src/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
yfarjoun: done.

yfarjoun:
Thanks for the review @nh13. When you have a moment, I've responded to your comments.

nh13:
Looks good to me, @tfenne ?

tfenne:
Seems reasonable to me.  I'm sad that it's necessary, but whatcha gonna do?

yfarjoun:
@tfenne I hope you are not concerned too much about performance, but rather about aesthetics...

Given that by default there is no logging on duplicate sets of size < 1000, the only performance hit is a `if` statement against a final false, which probably gets optimized away in the loop. I do not expect any performance hit.  

There's a post I saw in the ORACLE website (http://bugs.java.com/bugdatabase/view_bug.do?bug_id=7116986) that seems to claim that this bug has been fixed in version 9. I'm happy to get back to this when we move there if I am reminded...the problem is that I need to create a synthetic test case that fails without this change, and I've not yet managed to do that...Thank you both for your help.

tfenne:
Yup, totally aesthetic @yfarjoun, which is why, since it solves a real problem, I make no objection :smile:

PR #477, opened 2016-03-09
Fixed LiftoverVcf to contig name checking.

tfenne:
Fixed LiftoverVcf to check target contig names against the _names_ in the target genome instead of the `byte[]`s of sequences.


tfenne:
@yfarjoun, @fleharty can you take a look please?

yfarjoun:
Argh, what a silly bug. Thanks @tfenne :+1:

PR #479, opened 2016-03-10
Changes the List of ProgramInterfaces to a Set 

yfarjoun:
so that the output will not have duplication of programs if the user neglects to include a null as the first element. Also clarify documentation wrt INTERVALS and DB_SNP which are (currently) only use by one PROGRAM.

### DIFF
--- a/src/java/picard/analysis/CollectMultipleMetrics.java
+++ b/src/java/picard/analysis/CollectMultipleMetrics.java
@@ -326,27 +326,29 @@ public class CollectMultipleMetrics extends CommandLineProgram {
     private final Set<MetricAccumulationLevel> accumLevelDefault = CollectionUtil.makeSet(MetricAccumulationLevel.ALL_READS);
 
     @Option(shortName="LEVEL", doc="The level(s) at which to accumulate metrics.")
-    public Set<MetricAccumulationLevel> METRIC_ACCUMULATION_LEVEL = new HashSet<MetricAccumulationLevel>(accumLevelDefault);
+    public Set<MetricAccumulationLevel> METRIC_ACCUMULATION_LEVEL = new HashSet<>(accumLevelDefault);
 
     @Option(shortName = "EXT", doc="Append the given file extension to all metric file names (ex. OUTPUT.insert_size_metrics.EXT). None if null", optional=true)
     public String FILE_EXTENSION = null;
 
-    @Option(doc = "List of metrics programs to apply during the pass through the SAM file.")
-    public List<Program> PROGRAM = CollectionUtil.makeList(Program.CollectAlignmentSummaryMetrics, Program.CollectBaseDistributionByCycle,
-            Program.CollectInsertSizeMetrics, Program.MeanQualityByCycle, Program.QualityScoreDistribution);
+    @Option(doc = "Set of metrics programs to apply during the pass through the SAM file.")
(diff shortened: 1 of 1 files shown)


yfarjoun:
@bradtaylor could you take a look?

yfarjoun:
bump.

src/java/picard/analysis/CollectMultipleMetrics.java
```diff
@@ -325,27 +325,29 @@ public SinglePassSamProgram makeInstance(final String outbase, final File input,
...
-    public Set<MetricAccumulationLevel> METRIC_ACCUMULATION_LEVEL = new HashSet<MetricAccumulationLevel>(accumLevelDefault);
+    public Set<MetricAccumulationLevel> METRIC_ACCUMULATION_LEVEL = new HashSet<>(accumLevelDefault);
 
     @Option(shortName = "EXT", doc="Append the given file extension to all metric file names (ex. OUTPUT.insert_size_metrics.EXT). None if null", optional=true)
     public String FILE_EXTENSION = null;
 
-    @Option(doc = "List of metrics programs to apply during the pass through the SAM file.")
-    public List<Program> PROGRAM = CollectionUtil.makeList(Program.CollectAlignmentSummaryMetrics, Program.CollectBaseDistributionByCycle,
+    @Option(doc = "Set of metrics programs to apply during the pass through the SAM file.")
+    public Set<Program> PROGRAM = CollectionUtil.makeSet(Program.CollectAlignmentSummaryMetrics, Program.CollectBaseDistributionByCycle,
```
tfenne: Maybe use a `LinkedHashSet` here so that the insertion order is maintained, making the re-constructed command line deterministic?

src/java/picard/analysis/CollectMultipleMetrics.java
```diff
@@ -368,7 +370,7 @@ public static void main(final String[] args) {
      * Use this method when invoking CollectMultipleMetrics programmatically to run programs other than the ones
      * available via enum.  This must be called before doWork().
      */
-    public void setProgramsToRun(final List<ProgramInterface> programsToRun) {
+    public void setProgramsToRun(final Set<ProgramInterface> programsToRun) {
```
tfenne: It would be nicer to change this to `Collection<ProgramInterface>` since that would be backwards compatible, and then implement as `this.programToRun.clear(); this.programsToRun.add(programsToRun)` so that a) it works and b) the caller can't silently modify the collection be retaining a reference.

yfarjoun: nice idea. I'll do that. thanks.

src/java/picard/analysis/CollectMultipleMetrics.java
yfarjoun: sure. You mean the default command-line, yes? not sure what here is **re**constructed.

tfenne: When `CommandLineProgram` emits the command line (both to stderr, and/or in any metrics file headers) it **re**constructs it from the `@Option` instance members.

yfarjoun: I see. I understand now. Will do.

bradtaylor:
👍 from me once Tim's comments are resolved.

yfarjoun:
did the things...

yfarjoun:
I had to rebase since changes in htsjdk were breaking, but I kept the commits separate so that you can still see my changes due to the review.

tfenne:
Thanks @yfarjoun.  This looks good to me.

yfarjoun:
waiting on tests.

PR #480, opened 2016-03-10
Issue with InsertSizeMetricsCollector and CollectInsertSizeMetricsTest  [Please review, DON'T MERGE]

SHuang-Broad:
There are currently several issues in the master branch that causes Picard to generate wrong results on Histogram bin width values when using `CollectInsertSizeMetrics` tool, and since the unit test in `CollectInsertSizeMetricsTest` is not catching it, this subtle bug slipped through the watching eyes.

Specifically, this PR addressed the following issues:
1. Added several `MetricAccumulationLevel` values so unit tests now could catch errors (master branch currently only test `MetricAccumulationLevel::ALL_READS`).
2. Corrected several expected values in the unit tests that were from running production Picard on the test bam file (see git diff for which values are changed).
3. The proposed correction on line 152 in `InsertSizeMetricsCollector` now generates "correct" bin width values for library "Solexa-41734", whose bin width, before the proposed fix, were calculated to be (5,5,5,5,5,0,0,0,0,0), which is apparently problematic, because this sequence should be a (not necessarily strict) monotonically increasing sequence.

@yfarjoun proposed this minimum-modification correction that will correct the mistake and leave other values unchanged. Alternative proposals on how to fix is to follow.

Thanks.

PS: some code is commented out for quick fix (R script for generating histogram plot). Python-related code change is mysteriously made by Intellij to work, which I don't quite understand yet. But these don't interfere with the issue raised.

### DIFF
--- a/src/main/java/picard/analysis/directed/InsertSizeMetricsCollector.java
+++ b/src/main/java/picard/analysis/directed/InsertSizeMetricsCollector.java
@@ -149,7 +149,7 @@ public class InsertSizeMetricsCollector extends MultiLevelCollector<InsertSizeMe
                         double low = median;
                         double high = median;
 
-                        while (low >= histogram.getMin() || high <= histogram.getMax()) {
+                        while (low >= histogram.getMin()-1 || high <= histogram.getMax()+1) {
                             final Histogram.Bin<Integer> lowBin = histogram.get((int) low);
                             if (lowBin != null) covered += lowBin.getValue();
 
--- a/src/test/java/picard/analysis/CollectInsertSizeMetricsTest.java
+++ b/src/test/java/picard/analysis/CollectInsertSizeMetricsTest.java
@@ -54,7 +54,10 @@ public class CollectInsertSizeMetricsTest extends CommandLineProgramTest {
         final String[] args = new String[] {
                 "INPUT="  + input.getAbsolutePath(),
                 "OUTPUT=" + outfile.getAbsolutePath(),
-                "HISTOGRAM_FILE=" + pdf.getAbsolutePath()
+                "HISTOGRAM_FILE=" + pdf.getAbsolutePath(),
+                "LEVEL=SAMPLE",
+                "LEVEL=LIBRARY",
+                "LEVEL=READ_GROUP"
         };
         Assert.assertEquals(runPicardCommandLine(args), 0);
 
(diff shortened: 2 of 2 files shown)


yfarjoun:
@SHuang-Broad The removal of the Rcode is confusing as is the inclusion of the .ipr and .iml files. could you please undo those changes (we refrain from including iml and ipr changes since intelliJ is very specific about them and insignificant changes will otherwise get submitted.)

That commit (which does the above) should be squashed into the first commit so that the proposal can be evaluated more cleanly (if you're not sure how to do that reordering, let me know I'm happy to assist). I'm glad to see that the proposed change didn't have an effect on the other tests, only on the edge case with 2 points. 

Regarding the change itself, this highlights that in the (rare) case where the median is a half-integer, the algorithm as currently implemented is slightly biased towards slightly low values since the casting will **subtract** 0.5 from both `low` and `high`. My personal view is that `low` and `high` should be initialized (in lines 149-150) to be the floor and the ceil of the median (respectively) this should cause minimal disruption but make the calculation unbiased.

Happy to hear comments.

yfarjoun:
@SHuang-Broad I think that the deafening silence is an indication that this is such an edge case that there are no strong opinions one way or another. I would advise to rebase and squash the two commits and then merge this PR, without further modification.

:+1:

SHuang-Broad:
Sorry @yfarjoun , I wasn't paying enough attention for this particular thread.
I recall that when we discussed about this issue, you have another proposed fix, which actually affects output, and isn't pushed here yet. Now that it seems not so many people care enough about this edge case, would it useful to push the other fix and merge that?

yfarjoun:
@SHuang-Broad , I think that the changes here are good enough. could you respond to the review, and rebase-squash?

SHuang-Broad:
@yfarjoun , just rebase-squashed with the minimal fix.

yfarjoun:
@SHuang-Broad could you fix the conflicts and merge this PR?

SHuang-Broad:
Yes, will do it today. @yfarjoun

SHuang-Broad:
@yfarjoun , just resolved the conflict. Would you please review (again)? Thanks!

yfarjoun:
👍 looks good! thanks. please fix your commit message and title and then merge.

cmnbroad:
Just noticed it looks like this nicely fixes https://github.com/broadinstitute/picard/issues/581, so that should be closed when this is merged.

PR #483, opened 2016-03-15
Add a method that exposes logging in TheoreticalSensitivity.

nh13:
@yfarjoun the motivation comes from wanting to call this method many times and not have so many logging messages.

Also, @davidbenjamin, could you possibly add a comment for the class explaining the motivation and reasoning behind this method?  Or provide a link to a paper/reference that does?  It would be great for folks who want more information about the motivation, math, and what it is trying to estimate.

### DIFF
--- a/src/java/picard/analysis/TheoreticalSensitivity.java
+++ b/src/java/picard/analysis/TheoreticalSensitivity.java
@@ -37,27 +37,41 @@ import java.util.List;
  * Created by David Benjamin on 5/13/15.
  */
 public class TheoreticalSensitivity {
-    /**
-     * @param the probability of depth n is depthDistribution[n] for n = 0, 1. . . N - 1
-     * @param the probability of quality q is qualityDistribution[q] for q = 0, 1. . . Q
-     * @param sample size is the number of random sums of quality scores for each m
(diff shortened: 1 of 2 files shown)


src/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -156,7 +170,7 @@ public int draw() {
         }
 
         //get samples of sums of 0, 1, 2,. . .  N - 1 draws
-        public List<ArrayList<Integer>> sampleCumulativeSums(final int maxNumberOfSummands, final int sampleSize) {
+        public List<ArrayList<Integer>> sampleCumulativeSums(final int maxNumberOfSummands, final int sampleSize, final boolean withLogging) {
```
yfarjoun: need to change the test as well...as it is currently failing.

yfarjoun:
except for the failing test, this looks fine. I'm looking into finding a good home for the documentation (and to have it referenced in the code), in the meantime it can found here www.broadinstitute.org/~farjoun/white_papers/theoretical-het-sensitivity.pdf

nh13:
@yfarjoun all set.

yfarjoun:
thanks! 

:+1:

PR #484, opened 2016-03-17
Fixes to usage examples on gh-pages docs

andrewquitadamo:
I fixed the background for two usage examples (MarkDuplicates and CollectMultipleMetrics) so they match the other usage examples.
I've also standardized the indentation of the usage examples and the capitalization (Usage example vs Usage Example).


vdauwera:
Hi @andrewquitadamo,

The content you've corrected is actually pulled into the webpages from the source code every time a new version is released. So the right place to make these corrections is in each tool's source code file, not here in the html pages, which will be overwritten by the next release. We can take these changes in so they'll look nice in the meantime of course.

vdauwera:
If you'd still like us to take these in we'll need you to squash them into a single commit, by the way.

yfarjoun:
I disagree. There's no point in taking in a change to auto-generated files.

:-1:

andrewquitadamo:
Oops. 

If you would like I can make these fixes to the source files. I suppose a new pull request would be better for that? Let me know the best way to proceed.

yfarjoun:
Sure, a new PR would be best. Thank-you!

ktibbett:
@andrewquitadamo , ok to close this PR?

PR #486, opened 2016-03-18
moved initialization of variable (bigDuplicateSetSize out of one cons…

yfarjoun:
…tructor, as there are two)

This was causing VERY chatty logs when using the other constructor...sorry for the oversight @ktibbett


ktibbett:
:+1:

PR #487, opened 2016-03-21
Outputs arguments with redundant short names and removes INTERVALS short…

meganshand:
…Name

Currently, if the short name of an argument is set to be the same as the long name, and if that argument is overridable, the argument is not included in the output of the command line arguments. This changes CommandLineParser to output arguments with matching short names that are overridable (as well as those that are not overridable which are already currently being output) and removes the redundant short name for the INTERVALS argument in CollectWgsMetrics.


yfarjoun:
Given that this changes the basics of the command-line parser,
could you verify that other programs that use "optional" parameters do not have their help output change?

meganshand:
I looked at other "overridable" parameters and I can confirm that their help output doesn't change.

yfarjoun:
:+1:

yfarjoun:
rebase/squash please...

PR #489, opened 2016-03-24
Store source chr/position on liftover

bbimber:
LiftoverVCF will lift variants to an alternate genome, given a chain file.  This patch will add attributes to each record to store the source chromosome/position for the lifted variant.  It can be a nice improvement when trying to work w/ the lifted VCF.

### DIFF
--- a/src/java/picard/vcf/LiftoverVcf.java
+++ b/src/java/picard/vcf/LiftoverVcf.java
@@ -21,6 +21,8 @@ import htsjdk.variant.variantcontext.writer.VariantContextWriterBuilder;
 import htsjdk.variant.vcf.VCFFileReader;
 import htsjdk.variant.vcf.VCFFilterHeaderLine;
 import htsjdk.variant.vcf.VCFHeader;
+import htsjdk.variant.vcf.VCFHeaderLineType;
+import htsjdk.variant.vcf.VCFInfoHeaderLine;
 import htsjdk.variant.vcf.VCFRecordCodec;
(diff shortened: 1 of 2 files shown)


tfenne:
I appreciate the utility of this, but I think it should probably be optional - in some cases I would definitely want this, but in others I would rather not clutter up my output VCF and don't want to have to post-process it to remove extraneous fields.

yfarjoun:
Agreed. thanks @tfenne. @bbimber feel free to add this as an optional, not-default behavior.

bbimber:
that's reasonable - just committed it.  does this follow your standard for command line args?  i didnt add a shortName, since I didnt find one that seemed to fit all that well.  happy to add "WOP" (WRITE_ORIGINAL_POSITION), or change it to a better name if anyone has one.

yfarjoun:
Could you please add a small test?

yfarjoun:
No need to worry about the shortname.

bbimber:
test added

yfarjoun:
please do not checkin your changes to iml and ipr. otherwise I think this is fine. :+1:

please squash and rebase your commits.

bbimber:
sorry about the project files - trying to do too many things at once.

changes squashed, but i would like to propose one minor addition.  this adds LIFTOVER_MIN_MATCH as a command line option (retaining the current default of 1.0, so behavior unchanged).  if you have concerns or want this split into a separate PR let me know.

yfarjoun:
Thanks @bbimber that is fine.

PR #490, opened 2016-03-24
Return 0 when lhs and rhs are identical in EstimateLibraryComplexity.PairedReadComparator

jabbarish:
EstimateLibraryComplexity.PairedReadComparator returns System.identityHashCode(lhs) - System.identityHashCode(rhs) when lh and rhs are equal.

This leads to slightly different results in different sequential launches of the metric. It makes difficult to run  regression tests.

We propose to replace the returned expression with zero to provide the results' consistency.

### DIFF
--- a/src/java/picard/sam/markduplicates/EstimateLibraryComplexity.java
+++ b/src/java/picard/sam/markduplicates/EstimateLibraryComplexity.java
@@ -339,8 +339,9 @@ public class EstimateLibraryComplexity extends AbstractOpticalDuplicateFinderCom
 
     /**
      * Comparator that orders read pairs on the first N bases of both reads.
+     * There is no tie-breaking, so any sort is stable, not total.
      */
-    class PairedReadComparator implements Comparator<PairedReadSequence> {
+    private class PairedReadComparator implements Comparator<PairedReadSequence> {
         final int BASES = EstimateLibraryComplexity.this.MIN_IDENTICAL_BASES;
 
         public int compare(final PairedReadSequence lhs, final PairedReadSequence rhs) {
@@ -356,7 +357,7 @@ public class EstimateLibraryComplexity extends AbstractOpticalDuplicateFinderCom
                 if (retval != 0) return retval;
             }
 
-            return System.identityHashCode(lhs) - System.identityHashCode(rhs);
+            return 0;
         }
     }
 


yfarjoun:
I think it would be better to check for equality upfront and return 0 only in that case.

jabbarish:
Yossi, could you clarify you point of view?  Here is the code excerpt:

```java
    class PairedReadComparator implements Comparator<PairedReadSequence> {
        final int BASES = EstimateLibraryComplexity.this.MIN_IDENTICAL_BASES;

        public int compare(final PairedReadSequence lhs, final PairedReadSequence rhs) {
            // First compare the first N bases of the first read
            for (int i = 0; i < BASES; ++i) {
                final int retval = lhs.read1[i] - rhs.read1[i];
                if (retval != 0) return retval;
            }

            // Then compare the first N bases of the second read
            for (int i = 0; i < BASES; ++i) {
                final int retval = lhs.read2[i] - rhs.read2[i];
                if (retval != 0) return retval;
            }

            return System.identityHashCode(lhs) - System.identityHashCode(rhs);
        }
    }
```
The two for cycles compare lhs and rhs respectively.  If they are equal, the last return statement applies, which returns the expression based on lhs & rhs identityHashCodes. We propose to return just 0 to indicate that the sequences are identical.

nh13:
How about using `SAMUtils.getCanonicalRecordName`?

tfenne:
That comparator is effectively private within `EstimateLibraryComplexity` and it's design is really only to order reads by the first `n` bases, after that the ordering is arbitrary - and since it's a comparator it only has to guarantee `if (a < b) then (b > a)`.  So I think the proposed change is fine, but that the comparator should be made `private` and a comment added that there is no tie-breaking, so any sort is stable, not total.

jabbarish:
@tfenne done what you proposed
@yfarjoun please look and say if further changes needed

yfarjoun:
fine by me :+1:

yfarjoun:
I defer to @tfenne on this.

jabbarish:
@tfenne could you please have a look at it?..

tfenne:
Works for me.  Please squash and I will then merge.

jabbarish:
@tfenne done

PR #495, opened 2016-03-31
Expose SAM tags for molecular indexes in IlluminaBasecallsToSam.

nh13:
Add options to IlluminaBasecallsToSam to set the SAM tag for the
molecular index bases and base qualities.  Also added an option to allow
each molecular index to be stored in its own tag.


nh13:
@yfarjoun when you are back from vacation, could you take a look at this one.  Or @eitanbanks if he remains in the wild, would you mind taking a look?

src/java/picard/illumina/ClusterDataToSamConverter.java
```diff
@@ -65,6 +67,10 @@
     private final int outputRecordsPerCluster;
     private final ReadNameEncoder readNameEncoder;
 
+    private String molecularIndexTag          = "RX";
```
yfarjoun: perhaps add a todo to change this to reference SAMTagsUtil when I finally get around to changing the samspec: https://github.com/samtools/hts-specs/pull/119

src/java/picard/illumina/ClusterDataToSamConverter.java
```diff
@@ -134,12 +169,29 @@ private SAMRecord createSamRecord(final ReadData readData, final String readName
             sam.setAttribute(SAMTag.BC.name(), unmatchedBarcode);
         }
 
-        if (molecularIndex != null) {
+        if (!molecularIndexes.isEmpty()) {
             //TODO: add RX and QX to the list of SAMTags and change this. initial discussion
```
yfarjoun: the todos here become very out of place..coudl you move them to where I commented above instead?

src/java/picard/illumina/ClusterDataToSamConverter.java
```diff
@@ -101,10 +107,39 @@ public ClusterDataToSamConverter(final String runBarcode,
     }
 
     /**
+     *  Sets the SAM tag to use to store the molecular index bases.  If multiple molecular indexes exist, it will concatenate them
+     * and store them in this tag.
+     */
+    public ClusterDataToSamConverter withMolecularIndexTag(final String molecularIndexTag) {
```
yfarjoun: I'm intregued by the use of the word "with" in the name of this method, when the verb is clearly "set"...why not the set* pattern we have in many other factories?

src/java/picard/illumina/ClusterDataToSamConverter.java
```diff
@@ -101,10 +107,39 @@ public ClusterDataToSamConverter(final String runBarcode,
...
+        return this;
+    }
+
+    /**
+     * Sets the SAM tags to use to store the bases each molecular index.  This will only be used if there are more than one molecular
+     * index. If fewer tags are given than molecular indexes found, then the remaining molecular indexes will be concatenated and stored
+     * in the last tag.  If more tags are provided than molecular indexes found, the additional tags will not be used.
+     */
+    public ClusterDataToSamConverter withTagPerMolecularIndex(final List<String> tagPerMolecularIndex) {
+        if (tagPerMolecularIndex != null) this.tagPerMolecularIndex = tagPerMolecularIndex;
```
yfarjoun: Should this throw a IllegalArgumentException rather than silently ignore a null being passed in?

src/java/picard/illumina/ClusterDataToSamConverter.java
```diff
@@ -134,12 +169,29 @@ private SAMRecord createSamRecord(final ReadData readData, final String readName
...
             //TODO: add RX and QX to the list of SAMTags and change this. initial discussion
             //TODO: here: https://github.com/broadinstitute/picard/issues/287
             //TODO: HTS-spec issue: https://github.com/samtools/hts-specs/issues/109
-            sam.setAttribute("RX", molecularIndex);
-            sam.setAttribute("QX", molecularIndexQ);
+            sam.setAttribute(this.molecularIndexTag, String.join("", molecularIndexes));
+            sam.setAttribute(this.molecularIndexQualityTag, String.join("", molecularIndexQualities));
+            if (!this.tagPerMolecularIndex.isEmpty() && molecularIndexes.size() > 1) {
+                if (tagPerMolecularIndex.size() < molecularIndexes.size()) {
+                    // fewer tags than indexes, so store the remaining in the last tag
```
yfarjoun: This is a strange outcome that should result in a warning or an exception...

src/java/picard/illumina/ClusterDataToSamConverter.java
nh13: done

src/java/picard/illumina/ClusterDataToSamConverter.java
nh13: done

src/java/picard/illumina/ClusterDataToSamConverter.java
nh13: I like "with*" since it reads more naturally:

```
new ClusterDataToSamConverter(...).withMolecularIndexTag(...).withMolecularIndexQualityTag(...).withTagPerMolecularIndex(...)
```

versus

```
new ClusterDataToSamConverter(...).setMolecularIndexTag(...).setMolecularIndexQualityTag(...).setTagPerMolecularIndex(...)
```

Also, I don't see a lot of factories in Picard, three to be exact, so perhaps `with*` is reasonable?  As this is a style question, perhaps others @ktibbett or @tfenne could weigh in?

src/java/picard/illumina/ClusterDataToSamConverter.java
nh13: Happy either way, I will do you ask.

src/java/picard/illumina/ClusterDataToSamConverter.java
nh13: This is explained both in the definition of the `withTagPerMolecularIndex` method and in the `IlluminaBasecallsToSam`'s `TAG_PER_MOLECULAR_INDEX` option. 

What I *really* want is a read structure that can describe which molecular indexes should go into which tags (with concatenation), and possible ordering in concatenation.  For example `8M{RX}50T50T4M{ZA}4M{RX}` but that is for a different day. 

 Nonetheless, the convention of concatenating the molecular indexes when fewer tags are available is up for debate here.  I am in favor of this behavior.  Do others have thoughts?  @ktibbett @tfenne?

I will at minimum put a warning in place.

nh13:
@yfarjoun thanks for the comments.  I'd like to get feedback from @ktibbett or @tfenne about two comments you made:
1. `with*` versus `set*`.
2. concatenation of the remaining molecular indexes when fewer tags than molecular indexes are given. 

I will push updated the branch once those two issues are resolved so we don't lose the comments :)

yfarjoun:
FYI: I'm not entirely done with the review, yet...just took a break for
dinner. :-)

eitanbanks:
(FYI I'm back online too and can help review if needed.  But I don't know nearly as much about molecular barcodes as Yossi...)

src/java/picard/illumina/ClusterDataToSamConverter.java
```diff
@@ -163,33 +215,29 @@ private SAMRecord createSamRecord(final ReadData readData, final String readName
...
 
-        final String joinedMolecularIndex ;
-        final String joinedMolecularIndexQ ;
+        final List<String> molecularIndexes;
+        final List<String> molecularIndexQualities;
         if (hasMolecularBarcode) {
-            final StringBuilder joinedMolecularIndexQBuilder = new StringBuilder();
-            final byte[][] molecularIndex = new byte[molecularBarcodeIndices.length][];
-            final byte[][] molecularIndexQ = new byte[molecularBarcodeIndices.length][];
+            molecularIndexes         = new ArrayList<>();
```
yfarjoun: may I suggest better alignment here?

src/java/picard/illumina/ClusterDataToSamConverter.java
```diff
@@ -163,33 +215,29 @@ private SAMRecord createSamRecord(final ReadData readData, final String readName
...
-                joinedMolecularIndexQBuilder.append(SAMUtils.phredToFastq(molecularIndexQ[i]));
+                molecularIndexes.add(new String(cluster.getRead(molecularBarcodeIndices[i]).getBases()).replace('.', 'N'));
+                molecularIndexQualities.add(SAMUtils.phredToFastq(cluster.getRead(molecularBarcodeIndices[i]).getQualities()));
             }
-            joinedMolecularIndex  = IlluminaUtil.byteArrayToString(molecularIndex, "").replace('.', 'N');
-            joinedMolecularIndexQ = joinedMolecularIndexQBuilder.toString();
         } else {
-            joinedMolecularIndex = null;
-            joinedMolecularIndexQ = null;
+            molecularIndexes         = Collections.emptyList();
```
yfarjoun: alignment

yfarjoun:
one over-all question: I'm wondering why you didn't do the same for the qualities...are you not interested in the qualities? if you're going to go through the trouble of parsing the QX field to get the qualities of each barcode, you could have done the same for the barcodes themselves...I'm not asking that you do this...just wondering why you didn't..

src/java/picard/illumina/ClusterDataToSamConverter.java
yfarjoun: I think that a null is an Illegal argument..so I'd vote for an exception.

src/java/picard/illumina/ClusterDataToSamConverter.java
```diff
@@ -101,10 +107,39 @@ public ClusterDataToSamConverter(final String runBarcode,
     }
 
     /**
+     *  Sets the SAM tag to use to store the molecular index bases.  If multiple molecular indexes exist, it will concatenate them
+     * and store them in this tag.
+     */
+    public ClusterDataToSamConverter withMolecularIndexTag(final String molecularIndexTag) {
+        this.molecularIndexTag = molecularIndexTag;
```
yfarjoun: what about null checking here?

this brings up a point: should we allow for a null `molecularIndexTag` indicating that you do not want the result stored concatenated in a tag but only in the  `tagPerMolecularIndex` ?

src/java/picard/illumina/ClusterDataToSamConverter.java
```diff
@@ -101,10 +107,39 @@ public ClusterDataToSamConverter(final String runBarcode,
...
+        return this;
+    }
+
+    /**
+     * Sets the SAM tag to use to store the molecular index base qualities.  If multiple molecular indexes exist, it will concatenate them
+     * and store them in this tag.
+     */
+    public ClusterDataToSamConverter withMolecularIndexQualityTag(final String molecularIndexQualityTag) {
```
yfarjoun: null check.

src/java/picard/illumina/ClusterDataToSamConverter.java
```diff
@@ -134,12 +169,29 @@ private SAMRecord createSamRecord(final ReadData readData, final String readName
...
-        if (molecularIndex != null) {
+        if (!molecularIndexes.isEmpty()) {
             //TODO: add RX and QX to the list of SAMTags and change this. initial discussion
             //TODO: here: https://github.com/broadinstitute/picard/issues/287
             //TODO: HTS-spec issue: https://github.com/samtools/hts-specs/issues/109
-            sam.setAttribute("RX", molecularIndex);
-            sam.setAttribute("QX", molecularIndexQ);
+            sam.setAttribute(this.molecularIndexTag, String.join("", molecularIndexes));
+            sam.setAttribute(this.molecularIndexQualityTag, String.join("", molecularIndexQualities));
+            if (!this.tagPerMolecularIndex.isEmpty() && molecularIndexes.size() > 1) {
```
yfarjoun: why the protection against `molecularIndexes.size() == 1` ?

src/java/picard/illumina/ClusterDataToSamConverter.java
```diff
@@ -134,12 +169,29 @@ private SAMRecord createSamRecord(final ReadData readData, final String readName
...
+                    for (int i = 0; i < tagPerMolecularIndex.size()-1; i++) {
+                        sam.setAttribute(tagPerMolecularIndex.get(i), molecularIndexes.get(i));
+                    }
+                    final String lastTag = tagPerMolecularIndex.get(tagPerMolecularIndex.size()-1);
+                    final String lastValue = String.join("", molecularIndexes.subList(tagPerMolecularIndex.size()-1, molecularIndexes.size()));
+                    sam.setAttribute(lastTag, lastValue);
+                }
+                else {
+                    // at least as many tags as indexes
+                    for (int i = 0; i < tagPerMolecularIndex.size(); i++) {
```
yfarjoun: should be `i < molecularIndexes.size()`, no?

src/tests/java/picard/illumina/IlluminaBasecallsToSamTest.java
```diff
@@ -101,6 +101,27 @@ public void testNonBarcodedWithMoleclarIndex() throws Exception {
         IOUtil.assertFilesEqual(outputBam, new File(TEST_DATA_DIR, "nonBarcodedWithMolecularIndex8M.sam"));
     }
 
+    // This *should not* store any molecular indexes in ZA or ZB since there is only one molecular index
+    @Test
```
yfarjoun: please add a test with 3 tags and 2 barcodes

src/java/picard/illumina/ClusterDataToSamConverter.java
yfarjoun: (this is what the request for the extra test should test)

src/java/picard/illumina/IlluminaBasecallsToSam.java
```diff
@@ -206,6 +206,16 @@
             "BAMs for only a subset of the barcodes in a lane.", shortName="INGORE_UNEXPECTED")
     public boolean IGNORE_UNEXPECTED_BARCODES = false;
 
+    @Option(doc="The tag to use to store any molecular indexes.  If more than one molecular index is found, they will be concatenated and stored here.")
```
yfarjoun: optional?

src/java/picard/illumina/IlluminaBasecallsToSam.java
```diff
@@ -206,6 +206,16 @@
...
+    public String MOLECULAR_INDEX_TAG = "RX";
+
+    @Option(doc="The tag to use to store any molecular index base qualities.  If more than one molecular index is found, their qualities will be concatenated and stored here.")
```
yfarjoun: optional?

src/java/picard/illumina/IlluminaBasecallsToSam.java
```diff
@@ -206,6 +206,16 @@
...
+    public String MOLECULAR_INDEX_BASE_QUALITY_TAG = "QX";
+
+    @Option(doc="The list of tags to store each molecular index.  If fewer tags are provided than molecular indexes, the remaining molecular indexes " +
```
yfarjoun: optional?

src/java/picard/illumina/IlluminaBasecallsToSam.java
```diff
@@ -206,6 +206,16 @@
...
+            "will be concatenated and stored in the last tag.  If more tags are provided than molecular indexes found, the remaining tags will be ignored.")
+    public List<String> TAG_PER_MOLECULAR_INDEX;
```
yfarjoun: default empty list?

src/tests/java/picard/illumina/IlluminaBasecallsToSamTest.java
```diff
@@ -117,7 +138,52 @@ public void testNonBarcodedWithDualMoleclarIndex() throws Exception {
                 "LIBRARY_NAME=Hello, World"
         });
         IOUtil.assertFilesEqual(outputBam, new File(TEST_DATA_DIR, "nonBarcodedWithMolecularIndex4M4M.sam"));
+    }
+
+    // This *should* store any molecular indexes in ZA and ZB since there is more than one molecular indexes
```
yfarjoun: remove "any" from comment.

yfarjoun: is->are in comment

src/tests/java/picard/illumina/IlluminaBasecallsToSamTest.java
```diff
@@ -117,7 +138,52 @@ public void testNonBarcodedWithDualMoleclarIndex() throws Exception {
...
+                "SAMPLE_ALIAS=HiDad",
+                "LIBRARY_NAME=Hello, World",
+                "TAG_PER_MOLECULAR_INDEX=ZA",
+                "TAG_PER_MOLECULAR_INDEX=ZB"
+
+        });
+        IOUtil.assertFilesEqual(outputBam, new File(TEST_DATA_DIR, "nonBarcodedWithTagPerMolecularIndex4M4M.sam"));
+    }
 
+    // This *should* store any molecular indexes in ZA, ZB, and ZC since there is more than one molecular indexes.  There are
```
yfarjoun: any, is

src/java/picard/illumina/ClusterDataToSamConverter.java
nh13: How about I check for a null or empty string?

src/java/picard/illumina/ClusterDataToSamConverter.java
nh13: I suppose I don't really need it, but the thinking was this would `act` the same as `molecularIndexTag`.  I'll remove it.

src/java/picard/illumina/ClusterDataToSamConverter.java
nh13: gotcha

src/java/picard/illumina/ClusterDataToSamConverter.java
nh13: you may

src/java/picard/illumina/ClusterDataToSamConverter.java
nh13: mapping </joke>

src/java/picard/illumina/IlluminaBasecallsToSam.java
nh13: Sure, so someone could give `null` to not populate the tag.

src/java/picard/illumina/IlluminaBasecallsToSam.java
nh13: Same as above, thanks!

src/java/picard/illumina/IlluminaBasecallsToSam.java
nh13: Not needed since `minElements` defaults to zero.

nh13: But I will add optional and empty list since I am rusty with the clp parsing conventions in picard.

src/java/picard/illumina/IlluminaBasecallsToSam.java
nh13: will do, I am not sure if it is necessary.

src/tests/java/picard/illumina/IlluminaBasecallsToSamTest.java
nh13: good idea, easy to do, done

nh13:
@yfarjoun I have addressed your comments and squashed (sorry).  I think the last thing that needs broader agreement is if we should concatenated additional molecular indexes in the last provided tag, which you thought was odd, and I thought was creative.

yfarjoun:
The problem with the concatenation is that is feels like a user error. I mean the user specified in the same invocation of the tool a different number of M operators in the ReadStructure than per-barcode-tags. do you see a use for this, or is it just a way around a possible mistake on the user's part? if the latter, I would prefer a inputValidation error upfront.

src/java/picard/illumina/ClusterDataToSamConverter.java
yfarjoun: Actually, now that you mention empty strings....

null -> error.
empty string -> no tag. 

would be my preference.

yfarjoun:
thinking about this more, I do not see any use-case in which the number of tags should be different than the number of M-operators. do you? If you agree, I'd say that the validation should enforce the same number, and the code will be simpler.

src/java/picard/illumina/ClusterDataToSamConverter.java
```diff
@@ -101,10 +115,40 @@ public ClusterDataToSamConverter(final String runBarcode,
...
+    }
+
+    /**
+     * Sets the SAM tags to use to store the bases each molecular index.  This will only be used if there are more than one molecular
+     * index. If fewer tags are given than molecular indexes found, then the remaining molecular indexes will be concatenated and stored
+     * in the last tag.  If more tags are provided than molecular indexes found, the additional tags will not be used.
+     */
+    public ClusterDataToSamConverter withTagPerMolecularIndex(final List<String> tagPerMolecularIndex) {
+        if (tagPerMolecularIndex == null) throw new IllegalArgumentException("Null given for tagPerMolecularIndex");
+        this.tagPerMolecularIndex = tagPerMolecularIndex;
```
yfarjoun: according to sam-spec: "TAG is a two-character string that matches /[A-Za-z][A-Za-z0-9]/".. the code should probably enforce this in order to generate valid SAM. (x3)

yfarjoun:
- with vs. set :+1:
- #tags vs. #M operators, unless you have a usecase (not a contrived one) where this is important, I'd advocate for a validation that enforces equality
- sam tags should be of the form set out in the sam-spec (please validate)

Thank-you for this PR.

src/java/picard/illumina/ClusterDataToSamConverter.java
nh13: Oh how I wish java had `nonEmpty`.  Will do.

src/java/picard/illumina/ClusterDataToSamConverter.java
nh13: `setAttribute` will do this for you :+1:

nh13:
@yfarjoun addressed all your comments.  I validated equal # of tags as molecular indices in both `IllluminaBasecallsToSam` and `ClusterDataToSamConverter`.  I didn't validate the sam tag spec, since that is handled elsewhere (in `setAttribute`).  If you :+1: I will rebase and squash.

yfarjoun:
thanks. :+1:  after minor comments.

src/java/picard/illumina/ClusterDataToSamConverter.java
yfarjoun: ok.

nh13:
@yfarjoun rebase and squashed.  Click the merge when tests pass?

yfarjoun:
Thanks!

PR #497, opened 2016-04-04
Moved calculation of values for %selected back to before filtering to unique reads.

tfenne:
Previously %selected was calculated over all reads, and in the changes introduced in or around e6d54be it was moved to be calculated on only unique reads.  For most experiments the values calculated on all reads and unique reads only are very similar.  However, for experiments at extremely deep coverage, and thus with very high on-target duplication, if the rate of duplicate off-target is sufficiently different the %selected values calculated on all reads vs. just unique reads can diverge quite dramatically.

This change restores the old behaviour of calculating on all reads, which I think is more correct in that it is a more accurate estimate of the fraction of molecules/sequencing reads that were _selected_ for, vs. those that are off-target.

### DIFF
--- a/src/java/picard/analysis/directed/TargetMetricsCollector.java
+++ b/src/java/picard/analysis/directed/TargetMetricsCollector.java
@@ -476,16 +476,9 @@ public abstract class TargetMetricsCollector<METRIC_TYPE extends MultilevelMetri
                 }
             }
 
-            ///////////////////////////////////////////////////////////////////
-            // Duplicate reads can be totally ignored beyond this point
-            ///////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


tfenne:
@yfarjoun @ktibbett This is a relatively small code change, but probably requires some thought/review from your end before proceeding.  Could you take a look please?

yfarjoun:
:+1: from me. I expect some wiggle in the results, but it's for the better.

tfenne:
@ktibbett I'm not going to merge this until I hear from you too :smile:

tfenne:
Thanks @yfarjoun

ktibbett:
:+1:

PR #500, opened 2016-04-07
Makes SortSam honor the (common) input parameters 

yfarjoun:
CREATE_INDEX and CREATE_MD5.

No change in default behavior expected.


tfenne:
The change looks good to me, but I'm a little confused as to why it's necessary.  I would have thought that `CommandLineProgram:176-180` would handle this automatically?

So a curious :+1: from me.

yfarjoun:
sorry. I must have been smoking something. I was annoyed that an index wasn't created when the SortOrder was queryname...

PR #504, opened 2016-04-10
Fix typo in log.warn in EstimateLibraryComplexity

jabbarish:
Fix log.warn when omitting group because of group size exceeds expected mean number of read pairs.
It outputs prs.read1 twice instead of prs.read1 and prs.read2.


yfarjoun:
this seems very reasonable. thanks for the fix 👍

PR #509, opened 2016-04-13
Deprecating tools that we no longer want to support

eitanbanks:
@ktibbett: deprecating these classes as requested.
I'll also add an issue here to remove them in a few months.

### DIFF
--- a/src/tests/java/picard/analysis/CollectWgsMetricsFromQuerySortedTest.java
+++ b/src/tests/java/picard/analysis/CollectWgsMetricsFromQuerySortedTest.java
@@ -16,7 +16,7 @@ import java.util.List;
  *
  * @author Eric Banks
  */
-
+@Deprecated
 public class CollectWgsMetricsFromQuerySortedTest extends CommandLineProgramTest {
 
     private static final File TEST_DATA_DIR = new File("testdata/picard/sam");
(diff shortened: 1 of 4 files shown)


ktibbett:
:+1:

src/tests/java/picard/analysis/CollectWgsMetricsFromQuerySortedTest.java
```diff
@@ -16,7 +16,7 @@
  *
  * @author Eric Banks
  */
-
+@Deprecated
```
yfarjoun: please comment what class to use instead.

src/tests/java/picard/analysis/CollectWgsMetricsFromSampledSitesTest.java
```diff
@@ -40,6 +40,7 @@
 /**
  * Tests CollectWgsMetricsFromSampledSites
  */
+@Deprecated
```
yfarjoun: same.

src/tests/java/picard/analysis/CollectWgsMetricsFromQuerySortedTest.java
yfarjoun: sorry, I mean on the main class, not the test.

eitanbanks:
Added a note about which tool to use instead.  Unless I hear any objections, I will merge this later on today.

PR #511, opened 2016-04-14
Making more classes and methods public in Genotype Concordance.

nh13:
Also fixed a NPE when run on a whole genome (no intervals).


nh13:
@eitanbanks exposed a few more methods.  For context, this should allow other tools to easily compute the same metrics as `GenotypeConcordance`, or compute the concordance state of individual variants.

nh13:
One more commit coming right up

nh13:
Should be all set, I wont touch this, and will make another PR if I find anything else that needs to be exposed.

eitanbanks:
Yes, looks good.  Need me to merge it for you?

nh13:
Yes please. I don't have merge privileges.

PR #514, opened 2016-04-14
handle divide by 0 case

vivster7:


nh13:
@vivster7 which metric was zero?  `GENOME_TERRITORY`? `PF_BASES`?  most likely `PF_ORIENTED_PAIRS`?   I would argue that it is not needed for the former two.

Since this class in `Deprecated` in master, is this really needed?  So many other tools would at first glance need something like this too (they don't).

Finally, since the numerator and denominator are typically not `Doubles`, then why not just check if the denominator is zero? (`insertSizeSum` could be a `long`)
```
if (metrics.metrics.PF_ORIENTED_PAIRS > 0) metrics.metrics.MEAN_INSERT_SIZE = metrics.insertSizeSum / (double) metrics.metrics.PF_ORIENTED_PAIRS;
```
This would be consistent with other metric collectors (see `AlignmentSummaryMetricsCollector`).

vivster7:
Whoops, sorry was supposed to close this PR because as you mentioned this tool is deprecated.

Both ```PF_BASES``` and  ```PF_ORIENTED_PAIRS``` are zero when this metric collector is run on a BAM with no reads

PR #515, opened 2016-04-15
addressing review comments for yf_add_CLP_FixNmAndUqTags 

yfarjoun:
I merged yf_add_CLP_FixNmAndUqTags too quickly and failed to incorporate some reviewer comments.

@nh13:  Sorry about that. I'm addressing your comments here.
I didn't change the issue regarding the sort order because I think that if the input file is valid then the output should be valid...please correct me if I'm missing something. I agree that if the input has "coordinate" in the header, but is not actually sorted, then it will fail...but that's true for other programs too, no?


yfarjoun:
@eitanbanks quick review?

eitanbanks:
Looks good!  +1

Issue #160, opened 2015-02-13
CompareSAMs sometimes reports a file as being different from itself


mattsooknah:
Suppose you have a coordinate-sorted file ```foo.sam``` with multiple entries having the same read name and start coordinate (for whatever reason). Calling ```CompareSAMs foo.sam foo.sam``` will tell you that ```foo.sam``` differs from itself!

This is because the code currently assumes that we only have one entry for each read name at a given coordinate. Moreover, it traverses the "left" and "right" files asymmetrically, so that duplicate entries will get respected in one file but not the other.

Issue #284, opened 2015-09-11
CompareSAMs reports file different from itself


akiezun:
picard 1.138
[ac12.sam | uploaded via ZenHub](https://files.zenhub.io/55f24bf2c633a8e0736339bc)

commandline:
`java -jar ~/bin/picard-tools-1.138/picard.jar CompareSAMs VALIDATION_STRINGENCY=SILENT ac12.sam ac12.sam`

result:
```
Executing as akiezun@wm494-874 on Mac OS X 10.9.5 x86_64; Java HotSpot(TM) 64-Bit Server VM 1.8.0_25-b17; Picard version: 1.138(aa51703435dc6a423013e74e56b0b68405facd79_1439324166) JdkDeflater
Match	1
Differ	0
Unmapped_both	0
Unmapped_left	0
Unmapped_right	0
Missing_left	1
Missing_right	0
SAM files differ.
```

This is unexpected - files should be equal to themselves.

mattsooknah:
Is this potentially related to https://github.com/broadinstitute/picard/issues/160 ?

yfarjoun:
looks like the same issue from here. why was #160 closed? lack of interest?

mattsooknah:
I remember fixing this in the Hellbender port of CompareSAMs (see https://github.com/broadinstitute/hellbender/pull/18) but I guess I forgot to copy the fix back into Picard. So it seems that ticket was closed prematurely.

Although https://github.com/broadinstitute/hellbender/pull/898 suggests that it's still broken in the Hellbender port as well ... @akiezun is that true?

PR #516, opened 2016-04-16
Yf fix compare sam

yfarjoun:
Fixed bug in CompareSams that occurs when two read mates have the same genomic start position. 

Previously this would show up as "missing left" since the key was read-name and position and the first read would be overwritten by the second. Now I also include "first" or "second" in the key and this solves the problem. 

- added a test that previously failed.


yfarjoun:
resolves #160  and #284  was fixed in GATK here: https://github.com/broadinstitute/gatk/pull/183 with a similar solution

yfarjoun:
tests still failing. will fix and reopen

yfarjoun:
@mattsooknah would you like to review this? You seem to be the last person to have thought about this program...though your fix was in gatk4.

src/java/picard/sam/CompareSAMs.java
```diff
@@ -501,6 +501,11 @@ private void reportDifference(Object o1, Object o2, final String label) {
         reportDifference(o1.toString(), o2.toString(), label);
     }
 
+    private String getKeyForRecord(final SAMRecord record) {
+        boolean t = record.getReadPairedFlag() && record.getSecondOfPairFlag();
```
eitanbanks: Should be final.  And it would be nice to get a more informative variable name.  isSecondOfPair?

src/java/picard/sam/CompareSAMs.java
```diff
@@ -501,6 +501,11 @@ private void reportDifference(Object o1, Object o2, final String label) {
...
+        return record.getReadName() + "-" + (t ? "first" : "second");
```
eitanbanks: Wait, shouldn't it be the opposite?  t is true if record.getSecondOfPairFlag() is true.

eitanbanks:
Review done!  +1

PR #517, opened 2016-04-19
Fix an error with too many open files for Multi-Tile BCLs.

nh13:
Please do not merge yet, as the real-world test case is in progress, but reviews are welcome (@jacarey).

### Description

When `PerTileCycleParser` seeks to a tile, it would call `close()` on the
underlying `cycleFilesParser`, which is fine if the parser parses only a
single tile.  When seeking to a new tile with `MultiTileBclParser`, we
want to re-use the underlying `cycleFileParser`, so the `close()` call on the
underlying iterator was commented out.  This leads to an error with "too
many open files", even when setting `ulimit` to 800K files.

The solution is to only call close on the underlying iterator if the
base class' (`PerTileCycleParser`) `close()` method was called, not when the
`seekToTile()` method is called.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing
- [ ] Failing real-world test case now succeeds.

### DIFF
--- a/src/java/picard/illumina/parser/MultiTileBclParser.java
+++ b/src/java/picard/illumina/parser/MultiTileBclParser.java
@@ -66,6 +66,12 @@ public class MultiTileBclParser extends BclParser {
     }
 
     @Override
+    protected CycleFilesParser<BclData> makeCycleFileParser(final List<File> files, final CycleFilesParser<BclData> cycleFilesParser) {
+        // NB: do not close the underlying reader like the parent does just yet.
+        return makeCycleFileParser(files);
+    }
+
+    @Override
     protected CycleFilesParser<BclData> makeCycleFileParser(final List<File> files) {
         if (cycleFileParser == null) {
             cycleFileParser = new MultiTileBclDataCycleFileParser(files, currentTile);
@@ -97,7 +103,7 @@ public class MultiTileBclParser extends BclParser {
 
         @Override
         public void close() {
-            //underlyingIterator.close();
+            underlyingIterator.close();
         }
 
         @Override
--- a/src/java/picard/illumina/parser/PerTileCycleParser.java
+++ b/src/java/picard/illumina/parser/PerTileCycleParser.java
@@ -86,11 +86,17 @@ abstract class PerTileCycleParser<ILLUMINA_DATA extends IlluminaData> implements
     }
 
     /**
-     * For a given cycle, return a CycleFilesParser.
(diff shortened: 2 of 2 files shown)


nh13:
I should mention this affects NextSeq users primarily, as `bcl.bgzf` files are mainly produced by NextSeq instruments (see [this link](http://support.illumina.com/content/dam/illumina-support/documents/documentation/software_documentation/bcl2fastq/bcl2fastq2-v216-Software-Guide-15051736-E.pdf)).

jacarey:
:+1:  I'll see if i can track down some nextseq data that we can use for testing.

nh13:
@jacarey that would be wonderful if you can verify the bug and fix on your side.

jcarey:
You have an incorrect email address on your list.  Please remove jcarey100@gmail.com <mailto:jcarey100@gmail.com>

jacarey:
I wasn't able to reproduce the bug with the NextSeq data as I suspect it has to do with the number of tiles in the run. I can see, however, that this PR has the desired result on the number of open files after tile reading..

Before:
Open files: 97 after 0 reads
Open files: 97 after 10000000 reads
...
Open files: 97 after 120000000 reads
Open files before: 37 Open files after: 96

After: 
Open files: 97 after 0 reads
Open files: 97 after 10000000 reads
...
Open files: 97 after 120000000 reads
Open files before: 37 Open files after: 38

nh13:
@jacarey thanks for the verification.  The run on the original failing dataset is going on now.

jacarey:
@nh13 Please let me know if if passes and we will merge this in.

nh13:
@jacarey will do, I will probably know by the end of the day.

jacarey:
@nh13 I found a relatively small data set that exhibited this behavior when I kept my open file handles low. I hooked it up to the profiler and saw that this change vastly reduces the number of open files as the run continues and gives us a nice flat open file telemetry.

I'm happy enough to merge this PR with your :+1:

nh13:
👍

nh13:
@jacarey the last tile in each lane is missing from the final BAM in this implementation.  Can you spot anything obvious that would create this behavior?

jacarey:
@nh13 Nothing obvious jumps out at me. I still have my test case set up.. I'll give it a look

jacarey:
@nh13 I'm not seeing the same thing. The last tile in my data set is indeed output to the final fastq.

nh13:
👍 confirmed in my hands this is good to go.  Merge at will.

jacarey:
@nh13  Thanks!

PR #523, opened 2016-04-25
Use string builder instead of format for memory reduction and perform…

jacarey:
…ance gain.


kbergin:
If the tests pass, I don't see anything wrong. Didn't see any differences from the old version to the new version regarding order of variables or delimiting characters. 👍

PR #525, opened 2016-04-27
New option: ALLOW_MISSING_FIELDS_IN_HEADER in LiftoverVcf.

nh13:
I am trying to lift over [NIST's NA12878 reference vcf](http://ftp-trace.ncbi.nlm.nih.gov/giab/ftp/release/NA12878_HG001/NISTv2.19/NISTIntegratedCalls_14datasets_131103_allcall_UGHapMerge_HetHomVarPASS_VQSRv2.19_2mindatasets_5minYesNoRatio_all_nouncert_excludesimplerep_excludesegdups_excludedecoy_excludeRepSeqSTRs_noCNVs.vcf.gz) to b38.  Unfortunately it has some info and format fields in the VCF records which are not found (described) in the VCF header.  I would like the tool not to fail when such records are encountered.

Relies on https://github.com/samtools/htsjdk/pull/590

### DIFF
--- a/src/java/picard/vcf/LiftoverVcf.java
+++ b/src/java/picard/vcf/LiftoverVcf.java
@@ -98,6 +98,9 @@ public class LiftoverVcf extends CommandLineProgram {
     @Option(doc = "The minimum percent match required for a variant to be lifted.", optional = true)
     public double LIFTOVER_MIN_MATCH = 1.0;
 
+    @Option(doc = "Allow INFO and FORMAT in the records that are not found in the header", optional = true)
+    public boolean ALLOW_MISSING_FIELDS_IN_HEADER = false;
+
     // When a contig used in the chain is not in the reference, exit with this value instead of 0.
     protected static int EXIT_CODE_WHEN_CONTIG_NOT_IN_REFERENCE = 1;
 
@@ -168,10 +171,13 @@ public class LiftoverVcf extends CommandLineProgram {
             for (final VCFInfoHeaderLine line : ATTRS) outHeader.addMetaDataLine(line);
         }
         final VariantContextWriter out = new VariantContextWriterBuilder().setOption(Options.INDEX_ON_THE_FLY)
+                .modifyOption(Options.ALLOW_MISSING_FIELDS_IN_HEADER, ALLOW_MISSING_FIELDS_IN_HEADER)
                 .setOutputFile(OUTPUT).setReferenceDictionary(walker.getSequenceDictionary()).build();
         out.writeHeader(outHeader);
 
-        final VariantContextWriter rejects = new VariantContextWriterBuilder().setOutputFile(REJECT).unsetOption(Options.INDEX_ON_THE_FLY).build();
+        final VariantContextWriter rejects = new VariantContextWriterBuilder().setOutputFile(REJECT).unsetOption(Options.INDEX_ON_THE_FLY)
+                .modifyOption(Options.ALLOW_MISSING_FIELDS_IN_HEADER, ALLOW_MISSING_FIELDS_IN_HEADER)
+                .build();
         final VCFHeader rejectHeader = new VCFHeader(in.getFileHeader());
         for (final VCFFilterHeaderLine line : FILTERS) rejectHeader.addMetaDataLine(line);
         if (WRITE_ORIGINAL_POSITION) {
@@ -188,7 +194,7 @@ public class LiftoverVcf extends CommandLineProgram {
         log.info("Lifting variants over and sorting.");
 
         final SortingCollection<VariantContext> sorter = SortingCollection.newInstance(VariantContext.class,
-                new VCFRecordCodec(outHeader, VALIDATION_STRINGENCY != ValidationStringency.STRICT),
+                new VCFRecordCodec(outHeader, ALLOW_MISSING_FIELDS_IN_HEADER || VALIDATION_STRINGENCY != ValidationStringency.STRICT),
                 outHeader.getVCFRecordComparator(),
                 MAX_RECORDS_IN_RAM,
                 TMP_DIR);


nh13:
Relies on https://github.com/samtools/htsjdk/pull/594.  The second commit makes `LiftoverVcf` a little more memory efficient by storing only the bases for reference sequences that are needed for the current contig.  I found that this allows me to run `LiftoverVcf` on my laptop versus not.

tfenne:
@nh13 I like the change to allow for missing INFO/FORMAT field definitions in the header since I've also run into that when lifting over VCFs from 3rd parties.

However, I think the reference management stuff is far too complicated for the benefit it offers.  A few thoughts:
1) I've never personally had a problem running this on my laptop.  I given it enough memory to load the whole genome (4G+) and it works just fine.
2) I _think_ this either introduces an assumption that the input file is sorted, or becomes very inefficient with reference loading/unloading when the file is not sorted
3) If each chromosome maps to somewhere >> 1 other chromosome and << all other chromosomes, we'll see a lot of re-loading of reference sequences
4) It's a lot of extra code to manage loading/unloading of reference sequences

If you _really_ want to make the memory use lower I would suggest implementing a very simple `CachingReferenceSequenceFile` on top of the basic (non-indexed) `ReferenceSequenceFile` interface that wraps another `ReferenceSequenceFile` and holds onto the returned entries in a `Map<String,WeakReference<ReferenceSequence>>` so that when under memory pressure cached references can be discarded (and re-loaded when needed).

vdauwera:
Will the vcf output by liftover contain header lines for those fields? If not, would it be possible to add them? Otherwise we are just propagating badness.

yfarjoun:
I'm not sure that LiftOverVcf should be in the business of fixing
headerlines. If we want a tool that fixed headerlines, we should write
that...

nh13:
@tfenne 
1) I used 8G of RAM and it failed at 45 minutes on the NIST NA12878 reference.  With these changes it completed in about 5 minutes.  Can you try it on your laptop? 
```
java -Xmx8G -jar /Users/nhomer/git/picard/dist/picard.jar \
    LiftoverVcf \
    I=NISTIntegratedCalls_14datasets_131103_allcall_UGHapMerge_HetHomVarPASS_VQSRv2.19_2mindatasets_5minYesNoRatio_all_nouncert_excludesimplerep_excludesegdups_excludedecoy_excludeRepSeqSTRs_noCNVs.vcf.gz \
    O=NISTIntegratedCalls_14datasets_131103_allcall_UGHapMerge_HetHomVarPASS_VQSRv2.19_2mindatasets_5minYesNoRatio_all_nouncert_excludesimplerep_excludesegdups_excludedecoy_excludeRepSeqSTRs_noCNVs.b38.vcf.gz \
    C=b37ToHg38.over.chain \  
    REJECT=NISTIntegratedCalls_14datasets_131103_allcall_UGHapMerge_HetHomVarPASS_VQSRv2.19_2mindatasets_5minYesNoRatio_all_nouncert_excludesimplerep_excludesegdups_excludedecoy_excludeRepSeqSTRs_noCNVs.b38.reject.vcf.gz \
    ALLOW_MISSING_FIELDS_IN_HEADER=true \
    R=hs38DH.fa
```
2) It does introduce that assumption.  An alternative is that it uses this method if the input is sorted.
3) Yup, but it really isn't that slow given we can move right to the correct sequences in the file.
4) It's 25 lines of extra code that are really not that complicated, so I think the only issue is the performance degradation if the input is not sorted.  I addressed that in (2) but I think the VCF specification states the positions must be sorted numerically within each contig, but there is no guarantee about contig order.  Anyhow, I just don't buy the "it's a lot of extra code argument"; could you provide better justification.

@vdauwera @yfarjoun I would love to have two tools in Picard:
1) Fixes up a VCF header.
2) Validate VCF similar to `ValidateSam`.

src/java/picard/vcf/LiftoverVcf.java
```diff
@@ -281,6 +311,7 @@ public static void main(final String[] args) {
                 else {
                     sorter.add(builder.make());
                 }
+                reverseComplementAlleleMap.clear();
```
nh13: @tfenne I am wondering if this line may be fixing my heap problems instead of the new reference management code?

nh13: Nope, just tested it, memory goes up to 8G and it looks like GC starts to take over.

vdauwera:
Agree that a standalone utility would be useful -- though it would be convenient to also have the ability to fix this sort of thing on the fly. But if I had to choose I would go for the standalone. 

GATK has a ValidateVariants tool, it's not perfect but it comes in handy.

nh13:
@vdauwera GATK has a different license that is incompatible for some.  Any chance we could port `ValidateVariants` over to Picard?

vdauwera:
ValidateVariants is a utility tool in the public module so it's under MIT license. Feel free to use directly or port (it could probably use some improvements).

https://github.com/broadgsa/gatk/blob/master/public/gatk-tools-public/src/main/java/org/broadinstitute/gatk/tools/walkers/variantutils/ValidateVariants.java

nh13:
@vdauwera 👍  thank-you for the clarification.  It always is great to have a tool and example test cases to start from.

tfenne:
@nh13 That's definitely weird.  I've lifted over Exac, dbSNP and some smaller files with genotypes, all fine in 8GB for sure, and I think less.  It's got to be something related to the large number of records with genotypes I think because both dbSNP and Exac are sizeable VCFs.

nh13:
It would be great if someone could reproduce my failure using 8g of memory as I would think that a lot of folks would use the NIST dataset, and since I only see a b37 release, lift over will be necessary. I would be happy to revert if it is just me but otherwise I would advocate for the change.

tfenne:
@nh13 I disagree.  As the original author of this tool I think this change does make it significantly more complicated.  Moreover, since it runs fine on other VCFs with loading in the whole genome, it seems clear to me that the real problem is likely a memory leak elsewhere, and that by making this change to how the reference is worked with you've freed up somewhere < 3GB of memory, and that's enough to lift over the file you care about in 8GB.  But other files may still require yet more memory.  I.e. this relieves the pressure on memory for the specific set of file and heap size you're working with, but is not a general solution to the problem!

Can you run the liftover in a profiler, or use `jvisualvm` to monitor the heap usage and try to identify where it's all coming from?

tfenne:
One other thought @nh13 - `LiftoverVcf` does sort the results, and uses a `SortingCollection` with max records in RAM set by the `MAX_RECORDS_IN_RAM` parameter that defaults to 500,000.  It could be just holding 500,000 of the records in memory is the problem?  the GIAB VCF does have a lot of stuff in the INFO field.  

Without your other change to support the f'd up headers it's hard to test this myself.  Maybe you could try with `MAX_RECORDS_IN_RAM=100000`?

yfarjoun:
There is a ValidateVariants in GATK...it's in the public part, so you may
even use it without a license (AFAIK, but IINAL)

nh13:
@tfenne moved it over to the first commit for this PR.  I will try the `MAX_RECORDS_IN_RAM`, which is a good suggestion.

nh13:
@tfenne `MAX_RECORDS_IN_RAM` set to a low value worked.  I am going to defer diving into the `vcf` to find the memory leak as well as improving the memory needed for the reference when the input is sorted since the first commit was the original intent.

tfenne:
👍 from me when the tests pass.

nh13:
@tfenne or @yfarjoun: care to hit that merge button?

PR #528, opened 2016-04-28
Adding a new tool ReplaceVcfHeader.

nh13:
@tfenne @yfarjoun testing the appetite for this tool.  I will add tests if folks 👍.

### DIFF
--- a/src/java/picard/vcf/ReplaceVcfHeader.java
+++ b/src/java/picard/vcf/ReplaceVcfHeader.java
@@ -1,0 +1,115 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Nils Homer
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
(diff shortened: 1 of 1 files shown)


vdauwera:
This is a great start but I'd love to see an option to automatically generate a new header (instead of replacing with a hand written one) where missing header lines have been added on the fly. Possibly even with an option to provide a description string for missing fields on the command line, to populate the lines of the added field definitions. Does that make sense? I can provide a worked out example if it helps.

nh13:
@vdauwera that is a lot more effort than this tool (see the analogous `ReplaceSamHeader` tool). It is also difficult to know the various fields to insert for the info/format fields (type, number, description, source, version).   I agree a tool that could do this automatically more useful to the user, but this better than nothing, and the desire for the former should not remove the appetite for the latter.

vdauwera:
That's fair -- just putting the idea out there (as the French lottery slogan goes, 100% of winners tried their luck). But I agree that if this is as far as you want to go for now, it is still very worthwhile and we should take it in.

nh13:
@yfarjoun not sure if I would have this tool separate from `FixVcfHeader` in https://github.com/broadinstitute/picard/pull/530.  I could keep both tools but remove the replace header functionality from `FixVcfHeader` or close this PR.  Either way is fine by me.

yfarjoun:
without this PR, you will not be able to replace a header as is, since
FixVcfHeader will fix the new header...I'd advocate leaving both.

yfarjoun:
Thinking about this more, I think that a single tools with optional HEADER and boolean FIX_HEADER and boolean REQUIRE_SAME_SAMPLES would be ideal. 

that single tools together with some test cases would get a 💯  from me. 😄

nh13:
Closing this in favor of https://github.com/broadinstitute/picard/pull/530.

PR #531, opened 2016-05-03
Allow RevertSam to output by read group

dshiga:
### Summary of changes

RevertSam's default behavior is unchanged, but you can now optionally output by read group. When run with OUTPUT_BY_READGROUP=true, multiple output bams will be created, each containing the reads for a single read group.

### Usage details

When outputting by read group, the user must provide either OUTPUT -- which must be a directory where the output files will be written (with default names of the form [read_group_id].bam) -- or OUTPUT_MAP, which is a two-column tab-separated file providing the output paths for each read group id.

### Implementation

When outputting by read group, we make a collection of writers -- one per read group -- and send each read to the appropriate writer. When not outputting by read group we send all reads to a single writer, as always.

When sorting and outputting by read group, we use a collection of sorters. When not outputting by read group, we use a single sorter, as always.


nh13:
👎 

It's fewer than [50 lines of code in Python](https://gist.github.com/seandavi/2014542) to split into one file per read group, and then you can run `RevertSam` on each, what am I missing?

yfarjoun:
@nh13 
while I don't know the details behind the need for this functionality (@ktibbett ?) I can think of a few comments to your remark:

1. The fact that it "can be done" outside of picard isn't reason enough to exclude it from picard (IMHO). 
2. This solution solves in a single pass what yours does in two, so in that regard it's better.

nh13:
1. The purpose of the python script example was to show that it can be done much more simply, and that the current PR is too complex (how many new classes?).
2. That requirement was never mentioned anywhere.  With all the sorting collections it is effectively buffering to disk with large input, so is it really a single pass?

Nonetheless, I think this one needs clear motivation and careful review as you can see my first instinct is to question the need for so much new code.

src/java/picard/sam/revertsam/OutputMap.java
```diff
@@ -0,0 +1,82 @@
+package picard.sam.revertsam;
```
nh13: license for all non-test sources.

src/java/picard/sam/revertsam/OutputMap.java
```diff
@@ -0,0 +1,82 @@
...
+import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Represents the mapping from each read group id to the corresponding output file
+ * that RevertSam is expected to write to.
+ */
+public class OutputMap {
```
nh13: Why is this class public by the constructors package protected?  Shouldn't the class be package protected?

src/java/picard/sam/revertsam/OutputMap.java
```diff
@@ -0,0 +1,82 @@
...
+    final Map<String, File> idToOutput = new HashMap<String, File>();
+
+    OutputMap(final File outputMapFile) {
```
nh13: Why not simplify and have two methods that return a `Map<String,File>` that stores for each read group id the corresponding paths?

src/java/picard/sam/revertsam/OutputMap.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private static OutputMapEntry createEntry(final TabbedTextFileWithHeaderParser.Row row, final File outputMapFile) {
+        final String id = row.getField("READ_GROUP_ID");
+        final String output = row.getField("OUTPUT");
+        validateRow(id, output, outputMapFile);
+        File outputPath = new File(output);
+        return new OutputMapEntry(id, outputPath);
+    }
+    
+    private static void validateRow(final String id, final String output, final File outputMapFile) {
```
nh13: this is not used elsewhere, why not inline it?

src/java/picard/sam/revertsam/OutputMap.java
```diff
@@ -0,0 +1,82 @@
...
+        if (!"READ_GROUP_ID".equals(columnLabels.get(0))) {
+            return false;
+        }
+        if (!"OUTPUT".equals(columnLabels.get(1))) {
+            return false;
+        }
+        return true;
+    }
+
+    private static OutputMapEntry createEntry(final TabbedTextFileWithHeaderParser.Row row, final File outputMapFile) {
```
nh13: why is this a separate function, inline it?

src/java/picard/sam/revertsam/OutputMap.java
```diff
@@ -0,0 +1,82 @@
...
+        final String id = row.getField("READ_GROUP_ID");
+        final String output = row.getField("OUTPUT");
+        validateRow(id, output, outputMapFile);
+        File outputPath = new File(output);
```
nh13: `final`?

src/java/picard/sam/revertsam/OutputMap.java
```diff
@@ -0,0 +1,82 @@
...
+        return new OutputMapEntry(id, outputPath);
+    }
+    
+    private static void validateRow(final String id, final String output, final File outputMapFile) {
+        if (id == null || output == null) {
+            throw new PicardException("Missing READ_GROUP_ID or OUTPUT in OUTPUT_MAP " + outputMapFile);
+        }
+    }
+
+    private static class OutputMapEntry {
```
nh13: Use a `Map<String,File>`?

src/java/picard/sam/revertsam/PerReadGroupSorter.java
```diff
@@ -0,0 +1,52 @@
+package picard.sam.revertsam;
```
nh13: I'm not complaining about the missing license.

src/java/picard/sam/revertsam/PerReadGroupSorter.java
```diff
@@ -0,0 +1,52 @@
...
+        return iterators;
+    }
+    
+    @Override
+    public void add(final SAMRecord rec) {
+        final SortingCollection<SAMRecord> sorter = getSorter(rec.getReadGroup());
+        sorter.add(rec);
+    }
+
+    private SortingCollection<SAMRecord> getSorter(final SAMReadGroupRecord readGroup) {
```
nh13: again, what's the need for added methods used only once elsewhere?

src/java/picard/sam/revertsam/PerReadGroupSorter.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private final Map<String, SortingCollection<SAMRecord>> sorterMap;
+    private final List<PeekableIterator<SAMRecord>> iterators;
+    
+    PerReadGroupSorter(final Map<String, SortingCollection<SAMRecord>> sorterMap) {
+        this.sorterMap = sorterMap;
+        this.iterators = createIterators(sorterMap.values());
+    }
+
+    private static List<PeekableIterator<SAMRecord>> createIterators(final Collection<SortingCollection<SAMRecord>> sorters) {
```
nh13: put this in the constructor?

src/java/picard/sam/revertsam/RevertSam.java
```diff
@@ -148,75 +144,51 @@ public static void main(final String[] args) {
...
+        final SAMFileHeader singleOutHeader = RevertSamUtil.createSingleOutHeader(SORT_ORDER, inHeader, REMOVE_ALIGNMENT_INFORMATION);
+        if (SAMPLE_ALIAS != null) overrideSample(singleOutHeader, SAMPLE_ALIAS);
+        if (LIBRARY_NAME != null) overrideLibrary(singleOutHeader, LIBRARY_NAME);
+        final RevertSamWriter out = RevertSamUtil.makeSingleOrPerReadGroupWriter(inHeader, singleOutHeader, presorted, OUTPUT_BY_READGROUP, readGroups, OUTPUT);
 
         ////////////////////////////////////////////////////////////////////////////
         // Build a sorting collection to use if we are sanitizing
         ////////////////////////////////////////////////////////////////////////////
-        final SortingCollection<SAMRecord> sorter;
+        RevertSamSorter sorter = null;
```
nh13: Instead of a whole new set of classes, why not just two maps:
```
final Map<String, PeekableIterator<SAMRecord> iterators;
final Map<String, SortingCollection<SAMRecord> sorters;
```
In the case of one output, just have them map to the same iterator or sorter (you'll need to unique them later), or just condition on `OUTPUT_BY_READGROUP` laster.  There are probably other ways, but are three new classes to store two simple maps necessary?

src/java/picard/sam/revertsam/RevertSam.java
```diff
@@ -240,37 +212,97 @@ protected int doWork() {
...
+            rec.setAlignmentStart(SAMRecord.NO_ALIGNMENT_START);
+            rec.setCigarString(SAMRecord.NO_ALIGNMENT_CIGAR);
+            rec.setMappingQuality(SAMRecord.NO_MAPPING_QUALITY);
+
+            if (!rec.getReadUnmappedFlag()) {
+                rec.setInferredInsertSize(0);
+                rec.setNotPrimaryAlignmentFlag(false);
+                rec.setProperPairFlag(false);
+                rec.setReadUnmappedFlag(true);
+
```
nh13: newline

src/java/picard/sam/revertsam/RevertSam.java
```diff
@@ -240,37 +212,97 @@ protected int doWork() {
...
+            }
+
+            // Then remove any mate flags and info related to alignment
+            if (rec.getReadPairedFlag()) {
```
nh13: I would prefer that *all* mate information is removed, even if the `getReadPairedFlag` is not set.  We are trying to sanitize/revert the SAM, perhaps from a bad or inconsistent state.

src/java/picard/sam/revertsam/RevertSam.java
```diff
@@ -240,37 +212,97 @@ protected int doWork() {
...
+            }
+
+            // And then remove any tags that are calculated from the alignment
+            for (final String tag : ATTRIBUTE_TO_CLEAR) {
+                rec.setAttribute(tag, null);
             }
+        }
+    }
 
+    private long[] sanitize(
```
nh13: You don't use the `RevertSamSorter` here, so I am more convinced `RevertSamSorter` can be simplified.

src/java/picard/sam/revertsam/RevertSam.java
```diff
@@ -359,56 +379,19 @@ public boolean filterOut(final SAMRecord first, final SAMRecord second) {
...
-     */
-    public void revertSamRecord(final SAMRecord rec) {
-        if (RESTORE_ORIGINAL_QUALITIES) {
-            final byte[] oq = rec.getOriginalBaseQualities();
-            if (oq != null) {
-                rec.setBaseQualities(oq);
-                rec.setOriginalBaseQualities(null);
-            }
-        }
+    private String getDefaultExtension(final String input) {
```
nh13: `/dev/stdin`? Urls?

src/java/picard/sam/revertsam/RevertSam.java
```diff
@@ -359,56 +379,19 @@ public boolean filterOut(final SAMRecord first, final SAMRecord second) {
...
+        return input.endsWith(".sam") ? ".sam" : ".bam";
+    }
 
-        if (REMOVE_DUPLICATE_INFORMATION) {
-            rec.setDuplicateReadFlag(false);
+    private void overrideSample(final SAMFileHeader header, final String sampleAlias) {
```
nh13: `overwriteSample`

src/java/picard/sam/revertsam/RevertSam.java
```diff
@@ -359,56 +379,19 @@ public boolean filterOut(final SAMRecord first, final SAMRecord second) {
...
-                rec.setMateNegativeStrandFlag(false);
-                rec.setMateReferenceIndex(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX);
-                rec.setMateUnmappedFlag(true);
-            }
-
-            // And then remove any tags that are calculated from the alignment
-            for (final String tag : ATTRIBUTE_TO_CLEAR) {
-                rec.setAttribute(tag, null);
-            }
+    private void overrideLibrary(final SAMFileHeader header, final String libraryName) {
```
nh13: `overwriteLibrary`?

src/java/picard/sam/revertsam/RevertSamSorter.java
```diff
@@ -0,0 +1,16 @@
...
+
+import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.util.PeekableIterator;
+
+import java.util.List;
+
+/**
+ * Allows callers to call the sorter in the same way, regardless of whether it's a SingleSorter or a PerReadGroupSorter.
+ */
+public interface RevertSamSorter {
```
nh13: make package protected if this stays around.

src/java/picard/sam/revertsam/RevertSamUtil.java
```diff
@@ -0,0 +1,188 @@
...
+import java.io.File;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Functions for creating writers, headers, and other objects needed by RevertSam.
+ */
+public class RevertSamUtil {
```
nh13: what's the benefit of shoving these methods into a generic `Util` class that is so tied to `RevertSam` that it's in the name?

src/java/picard/sam/revertsam/RevertSamUtil.java
```diff
@@ -0,0 +1,188 @@
...
+        for (final SAMReadGroupRecord samReadGroup : inHeader.getReadGroups()) {
+            final SAMFileHeader samHeader = createOutHeaderForReadGroup(samReadGroup, inHeader, sortOrder, removeAlignmentInformation);
+            final String id = samReadGroup.getId();
+            final ReadGroup readGroup = new ReadGroup(id, samHeader, outputMap.get(id));
+            readGroupList.add(readGroup);
+        }
+        return readGroupList;
+    }
+
+    private static SAMFileHeader createOutHeaderForReadGroup(
```
nh13: I am just not convinced that this makes it more readable to have a private static method that is used only once.

src/java/picard/sam/revertsam/RevertSamUtil.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.Map;
+
+/**
+ * Functions for creating writers, headers, and other objects needed by RevertSam.
+ */
+public class RevertSamUtil {
+
+    private final static Log log = Log.getInstance(RevertSamUtil.class);
+
+    public static List<ReadGroup> makeReadGroupList(
```
nh13: If this was a public class and method, it should be documented.

src/java/picard/sam/revertsam/RevertSamUtil.java
```diff
@@ -0,0 +1,188 @@
...
+        outHeader.addReadGroup(samReadGroup);
+        outHeader.setSortOrder(sortOrder);
+        if (!removeAlignmentInformation) {
+            outHeader.setSequenceDictionary(inHeader.getSequenceDictionary());
+            outHeader.setProgramRecords(inHeader.getProgramRecords());
+        }
+        return outHeader;
+    }
+
+    public static RevertSamWriter makeSingleOrPerReadGroupWriter(
```
nh13: This seems like a lot of extra code and effort to hide if we are writing to one output file or one per read group.  What does this buy you?  It's not like `RevertSam` would be too much more complex if you had `Map<String, SAMFileWriter> writers`.

src/java/picard/sam/revertsam/RevertSamUtil.java
```diff
@@ -0,0 +1,188 @@
...
+
+        final Map<String, SAMFileWriter> writerMap = new HashMap<String, SAMFileWriter>();
+        for (final ReadGroup readGroup : readGroups) {
+            final SAMFileWriter writer = factory.makeSAMOrBAMWriter(readGroup.getSamHeader(), presorted, readGroup.getOutput());
+            writerMap.put(readGroup.getId(), writer);
+        }
+        return writerMap;
+    }
+
+    public static SAMFileHeader createSingleOutHeader(
```
nh13: It was confusing that this is called `createSingleOutHeader` yet its output is eventually used only by `makeSingleOrPerReadGroupWriter` and only when creating `innerWriter`.  Seems hard to follow.  Why not just put it in `createWriterMap`?

src/java/picard/sam/revertsam/RevertSamUtil.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+        outHeader.setSortOrder(sortOrder);
+        if (!removeAlignmentInformation) {
+            outHeader.setSequenceDictionary(inHeader.getSequenceDictionary());
+            outHeader.setProgramRecords(inHeader.getProgramRecords());
+        }
+        return outHeader;
+    }
+
+    public static Map<SAMReadGroupRecord, FastqQualityFormat> createReadGroupFormatMap(
```
nh13: Again, I question why this is in a Util class.

src/java/picard/sam/revertsam/RevertSamUtil.java
```diff
@@ -0,0 +1,188 @@
...
+            log.info("Detected quality format for " + r.getReadGroupId() + ": " + readGroupToFormat.get(r));
+        }
+        if (readGroupToFormat.values().contains(FastqQualityFormat.Solexa)) {
+            throw new PicardException("No quality score encoding conversion implemented for " + FastqQualityFormat.Solexa);
+        }
+
+        return readGroupToFormat;
+    }
+
+    public static OutputMap createOutputMap(
```
nh13: can be removed after getting rid of the `OutputMap` class

src/java/picard/sam/revertsam/RevertSamUtil.java
```diff
@@ -0,0 +1,188 @@
...
+        } else {
+            final SortingCollection<SAMRecord> singleSorter = SortingCollection.newInstance(SAMRecord.class, new BAMRecordCodec(singleOutHeader), new SAMRecordQueryNameComparator(), maxRecordsInRam);
+            return new SingleSorter(singleSorter);
+        }
+    }
+
+    private static Map<String, SortingCollection<SAMRecord>> createSorterMap(
+            final List<ReadGroup> readGroups,
+            final int maxRecordsInRam) {
+
```
nh13: newline

src/java/picard/sam/revertsam/SingleSamFileWriter.java
```diff
@@ -0,0 +1,27 @@
...
+    @Override
+    public void addAlignment(SAMRecord alignment) {
+        writer.addAlignment(alignment);
+    }
+
+    @Override
+    public void close() {
+        writer.close();
+    }
+
```
nh13: nl

src/java/picard/sam/revertsam/SingleSamFileWriter.java
```diff
@@ -0,0 +1,27 @@
+package picard.sam.revertsam;
```
nh13: licenses again

src/java/picard/sam/revertsam/ValidationUtil.java
```diff
@@ -0,0 +1,124 @@
...
+import picard.util.TabbedTextFileWithHeaderParser;
+
+import java.io.File;
+import java.nio.file.Files;
+import java.util.List;
+
+/**
+ * Helper functions for validating RevertSam parameters. 
+ */
+public class ValidationUtil {
```
nh13: inner class in `RevertSam`?  Just have the methods in `RevertSam`?  package protected class?

src/java/picard/util/TabbedTextFileWithHeaderParser.java
```diff
@@ -145,6 +149,8 @@ public boolean hasColumn(final String columnLabel) {
     public Set<String> columnLabels() {
         return columnLabelIndices.keySet();
     }
+    
+    public List<String> columnLabelsList() { return Collections.unmodifiableList(columnLabels); }
```
nh13: no docs, and can you make a list from `columnLabelIndices.keySet()` so you don't need a new member?

nh13:
@dshiga @yfarjoun this is certainly a lot of effort, and I definitely see why you would want this tool, since regularly one receives (mapped) BAM files that need to be reverted or sanitized to create "unmapped" BAMs as input to analysis pipeline.  Nonetheless, I think the code could be simplified with some more thought.  I suggested a way that may not be the best way but is a start to the discussion.

nh13:
Also, I omitted marking all missing licenses, the many places where classes or methods scope should be changed and documentation added, as well the many `private static` methods that are used only once.

yfarjoun:
Thank-you very much @nh13 for the through review and suggestions!

src/java/picard/sam/revertsam/PerReadGroupSorter.java
geoffjentry: Some would say that methods ideally should only be a few lines long and anything longer should be broken down. I have my beers with the guy but IMO this is one that uncle bob got right

geoffjentry: And by beers I mean beefs. Clearly one can see which word I type more frequently on my iphone

nh13: I think in this case a simple lookup as a private method is too much.

geoffjentry: Now that I'm looking at code and not just commenting on comments I agree with you here.

geoffjentry:
I generally agree with @nh13 that this could be simplified down a fair bit. Nils biased this witness a bit too much so I found myself agreeing w/ him (egads!) vs. being able to come up with another alternative. Nils is smarter than me anyways.

dshiga:
Thanks @nh13, lots of good food for thought here.

I have moved things in the direction you suggested. There are no longer any new classes outside of RevertSam. Many of the methods that were in separate classes before are now just private methods of RevertSam. I have inlined all the methods you suggested inlining.

I had broken things out into independent classes and methods because I personally find that approach easier to read and understand. But based on feedback from you and others, I'm clearly in the minority in seeing it that way, at least as far as this PR is concerned. Since it seemed to be making it harder to read for most people, which was the exact opposite of my intent, I've reversed course.

My guess is that it was particularly unwelcome to have to jump around to a lot of different files to see all the code. So all the new code is in RevertSam.

I did find that some of the new methods naturally go together, and have grouped them into three new inner classes: RevertSamWriter, RevertSamSorter, and ValidationUtil. My hope is that grouping methods together in this way will make it easier for developers to navigate around in RevertSam and to safely ignore code that's not relevant to a particular task. If it hinders clarity at all, that would be good to know.

The refactoring has also had these happy effects:

- No need for adding licenses now that there are no new non-test files
- No new public classes or methods, so no need to add docs for them

There are a few other comments I will address individually.

src/java/picard/sam/revertsam/RevertSam.java
dshiga: This is not new code. It just got shifted to a different spot within RevertSam. I could remove the conditional, but it would mean changing behavior that some users may have come to expect. Should I add a parameter to determine this behavior instead? Does it belong in a different PR? I'm not sure how strict this project tends to be about maintaining backwards compatibility.

src/java/picard/sam/revertsam/RevertSam.java
dshiga: I don't understand the question. Are you asking for getDefaultExtension to be able to handle input from stdin and urls? I have inlined this method now so maybe this is a non issue.

src/java/picard/sam/revertsam/RevertSamUtil.java
dshiga: Having simplified by removing classes, etc., hopefully this is clearer now. singleOutHeader is used for creating the writer and the sorter when OUTPUT_BY_READGROUP=false.

src/java/picard/sam/revertsam/RevertSam.java
nh13: Some software *incorrectly* has mate information even when the read paired flag is not set, and it would be great if `RevertSam` removed that mate information.  If the goal of this tool is to have a clean and valid BAM with no mapping information in the output, I don't see what harm removing all this information regardless of whether or not the `getReadPairedFlag` is set.  @yfarjoun would you confirm my reasoning?

src/java/picard/sam/revertsam/RevertSam.java
nh13: If I want to output by read group, but the input comes from either a URL or `/dev/stdin`, then calling `getDefaultExtension` on `INPUT` will return `.bam`.  Also, this doesn't support `.cram`.  I would add documentation that says the output is BAM by default unless the input ends with `.sam` or .cram`.

src/java/picard/sam/revertsam/RevertSam.java
yfarjoun: Yes. I agree. Mate information is mapping information and should be removed
during revert regardless of how the bam got to its sad state.

src/java/picard/sam/revertsam/RevertSam.java
nh13: Just to me more explicit, I think defaulting to output `.bam` is reasonable but `.cram` isn't supported and some docs are needed.

dshiga: Okay - add documentation in the doc string for the OUTPUT param, or in the usage details documentation? Or did you mean a code comment where we define this default?

nh13: I think in the usage, since you will want two usage examples anyhow, one for the outputting to one file, and one when outputting per-read-group. @vdauwera does having two examples make sense?

dshiga:
Okay, I just pushed some more changes, which should address the last two outstanding comments.

- Added documentation about the output format in the usage details string
- Removed the conditional so that we will remove mate information for all reads regardless of whether its read paired flag is set

Does anyone have any further feedback, or can I get a thumbs up?

nh13:
@dshiga do you mind if I do one more review as likely your changes were significant?  If you could give me this weekend I can get to it by EOD Monday.

dshiga:
@nh13 that would be great, thanks!

src/java/picard/sam/RevertSam.java
```diff
@@ -78,18 +82,43 @@
             "information, which can be used to produce an unmapped BAM (uBAM) from a previously aligned BAM. It is also capable of " +
             "restoring the original quality scores of a BAM file that has already undergone base quality score recalibration (BQSR) if the" +
             "original qualities were retained." +
-            "<h4>Usage example:</h4>" +
+            "<h4>Usage example 1:</h4>" +
```
nh13: Perhaps instead of `...example 1` state `example single output`.

src/java/picard/sam/RevertSam.java
```diff
@@ -78,18 +82,43 @@
...
             "<pre>" +
             "java -jar picard.jar RevertSam \\<br />" +
             "     I=input.bam \\<br />" +
             "     O=reverted.bam" +
             "</pre>" +
+            "Output format is BAM by default, or SAM if the input path ends with '.sam'." +
+            "<h4>Usage example 2:</h4>" +
```
nh13: Perhaps instead of `...example 2` state `example output per read group`.

src/java/picard/sam/RevertSam.java
```diff
@@ -148,76 +177,57 @@ public static void main(final String[] args) {
...
+        final SAMFileHeader singleOutHeader = createOutHeader(inHeader, SORT_ORDER, REMOVE_ALIGNMENT_INFORMATION);
+        for (final SAMReadGroupRecord readGroup : inHeader.getReadGroups()) {
+            singleOutHeader.addReadGroup(readGroup);
         }
 
-        final SAMFileWriter out = new SAMFileWriterFactory().makeSAMOrBAMWriter(outHeader, presorted, OUTPUT);
+        final Map<String, File> outputMap;
+        final Map<String, SAMFileHeader> headerMap;
+        if (OUTPUT_BY_READGROUP) {
+            String defaultExtension = INPUT.toString().endsWith(".sam") ? ".sam" : ".bam";
```
nh13: add support for writing a cram with `.cram`.

src/java/picard/sam/RevertSam.java
```diff
@@ -78,18 +82,43 @@
...
             "</pre>" +
+            "Output format is BAM by default, or SAM if the input path ends with '.sam'." +
+            "<h4>Usage example 2:</h4>" +
+            "<pre>" +
+            "java -jar picard.jar RevertSam \\<br />" +
+            "     I=input.bam \\<br />" +
+            "     OUTPUT_BY_READGROUP=true \\<br />" +
+            "     OUTPUT_MAP=reverted_bam_paths.tsv" +
+            "</pre>" +
+            "Will output a BAM/SAM file per read group. By default, all outputs will be in BAM format. " +
```
nh13: document `.cram` support.

nh13:
@ktibbett I would also like a 👍 that this PR couldn't be solved by running `SplitSamByLibrary` then the previous version of `RevertSam` on each output BAM (which could be parallelized).

src/java/picard/sam/RevertSam.java
```diff
@@ -148,76 +177,57 @@ public static void main(final String[] args) {
      */
     @Override
     protected String[] customCommandLineValidation() {
-        if (SANITIZE && SORT_ORDER != SortOrder.queryname) {
-            return new String[]{"SORT_ORDER must be queryname when sanitization is enabled with SANITIZE=true."};
-        }
+        List<String> errors = new ArrayList<String>();
```
nh13: final?

src/java/picard/sam/RevertSam.java
```diff
@@ -148,76 +177,57 @@ public static void main(final String[] args) {
...
-        }
-        outHeader.setSortOrder(SORT_ORDER);
-        if (!REMOVE_ALIGNMENT_INFORMATION) {
-            outHeader.setSequenceDictionary(inHeader.getSequenceDictionary());
-            outHeader.setProgramRecords(inHeader.getProgramRecords());
+        final boolean presorted = isPresorted(inHeader, SORT_ORDER, sanitizing);
+        if (SAMPLE_ALIAS != null) overwriteSample(inHeader.getReadGroups(), SAMPLE_ALIAS);
+        if (LIBRARY_NAME != null) overwriteLibrary(inHeader.getReadGroups(), LIBRARY_NAME);
+        final SAMFileHeader singleOutHeader = createOutHeader(inHeader, SORT_ORDER, REMOVE_ALIGNMENT_INFORMATION);
+        for (final SAMReadGroupRecord readGroup : inHeader.getReadGroups()) {
```
nh13: covert to a stream?

src/java/picard/sam/RevertSam.java
nh13: `final`

src/java/picard/sam/RevertSam.java
```diff
@@ -148,76 +177,57 @@ public static void main(final String[] args) {
...
-        final SAMFileWriter out = new SAMFileWriterFactory().makeSAMOrBAMWriter(outHeader, presorted, OUTPUT);
+        final Map<String, File> outputMap;
+        final Map<String, SAMFileHeader> headerMap;
+        if (OUTPUT_BY_READGROUP) {
+            String defaultExtension = INPUT.toString().endsWith(".sam") ? ".sam" : ".bam";
+            outputMap = createOutputMap(OUTPUT_MAP, OUTPUT, defaultExtension, inHeader.getReadGroups());
+            ValidationUtil.assertAllReadGroupsMapped(outputMap, inHeader.getReadGroups());
+            headerMap = createHeaderMap(inHeader, SORT_ORDER, REMOVE_ALIGNMENT_INFORMATION);
+        } else {
+            outputMap = null;
```
nh13: Do we want nulls anymore or should we use `Optional`?

src/java/picard/sam/RevertSam.java
```diff
@@ -240,37 +250,96 @@ protected int doWork() {
...
+                readGroupToFormat = createReadGroupFormatMap(inHeader, REFERENCE_SEQUENCE, VALIDATION_STRINGENCY, INPUT, RESTORE_ORIGINAL_QUALITIES);
+            } catch (PicardException e) {
+                log.error(e.getMessage());
+                return -1;
+            }
 
-            long total = 0, discarded = 0;
-            final PeekableIterator<SAMRecord> iterator = new PeekableIterator<SAMRecord>(sorter.iterator());
-            final Map<SAMReadGroupRecord, FastqQualityFormat> readGroupToFormat = new HashMap<SAMReadGroupRecord, FastqQualityFormat>();
+            long[] sanitizeResults = sanitize(readGroupToFormat, sorter, out);
```
nh13: `final`.  IntelliJ should highlight these as yellow if you are using it.

geoffjentry: I don't know if this is the case for @dshiga but I learned recently that for people using the dark theme it's kinda hard to spot those warnings, the yellow shade they use there is subtle.

src/java/picard/sam/RevertSam.java
```diff
@@ -240,37 +250,96 @@ protected int doWork() {
...
+            final NumberFormat fmt = new DecimalFormat("0.000%");
+            log.info("Discarded " + discarded + " out of " + total + " (" + fmt.format(discardRate) + ") reads in order to sanitize output.");
 
-                    public boolean filterOut(final SAMRecord first, final SAMRecord second) {
-                        throw new UnsupportedOperationException();
-                    }
-                };
-                readGroupToFormat.put(rg, QualityEncodingDetector.detect(QualityEncodingDetector.DEFAULT_MAX_RECORDS_TO_ITERATE, new FilteringIterator(reader.iterator(), filter), RESTORE_ORIGINAL_QUALITIES));
-                CloserUtil.close(reader);
+            if (discarded / (double) total > MAX_DISCARD_FRACTION) {
```
nh13: use `discardRate`?

src/java/picard/sam/RevertSam.java
```diff
@@ -240,37 +250,96 @@ protected int doWork() {
...
+                rec.setReadNegativeStrandFlag(false);
             }
 
+            // Remove all alignment based information about the read itself
+            rec.setReferenceIndex(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX);
+            rec.setAlignmentStart(SAMRecord.NO_ALIGNMENT_START);
+            rec.setCigarString(SAMRecord.NO_ALIGNMENT_CIGAR);
+            rec.setMappingQuality(SAMRecord.NO_MAPPING_QUALITY);
+
+            if (!rec.getReadUnmappedFlag()) {
```
nh13: Just like clearing all the mate information, I think that these should all be set regardless of the `getReadUnmappedFlag` flag.

src/java/picard/sam/RevertSam.java
```diff
@@ -240,37 +250,96 @@ protected int doWork() {
...
 
-            final ProgressLogger sanitizerProgress = new ProgressLogger(log, 1000000, "Sanitized");
+            // Then remove any mate flags and info related to alignment
+            rec.setMateAlignmentStart(SAMRecord.NO_ALIGNMENT_START);
+            rec.setMateNegativeStrandFlag(false);
+            rec.setMateReferenceIndex(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX);
+            rec.setMateUnmappedFlag(true);
 
+            // And then remove any tags that are calculated from the alignment
+            for (final String tag : ATTRIBUTE_TO_CLEAR) {
```
nh13: `forEach`?

src/java/picard/sam/RevertSam.java
```diff
@@ -240,37 +250,96 @@ protected int doWork() {
...
+                rec.setAttribute(tag, null);
+            }
+        }
+    }
+
+    private long[] sanitize(
+            final Map<SAMReadGroupRecord, FastqQualityFormat> readGroupToFormat,
```
nh13: formatting

src/java/picard/sam/RevertSam.java
```diff
@@ -359,56 +416,344 @@ public boolean filterOut(final SAMRecord first, final SAMRecord second) {
         return out;
     }
 
+    private void overwriteSample(final List<SAMReadGroupRecord> readGroups, final String sampleAlias) {
+        for (final SAMReadGroupRecord rg : readGroups) {
```
nh13: `forEach`?

src/java/picard/sam/RevertSam.java
```diff
@@ -359,56 +416,344 @@ public boolean filterOut(final SAMRecord first, final SAMRecord second) {
...
+            rg.setSample(sampleAlias);
+        }
+    }
+
+    private void overwriteLibrary(final List<SAMReadGroupRecord> readGroups, final String libraryName) {
+        for (final SAMReadGroupRecord rg : readGroups) {
```
nh13: `forEach`?

src/java/picard/sam/RevertSam.java
```diff
@@ -359,56 +416,344 @@ public boolean filterOut(final SAMRecord first, final SAMRecord second) {
...
+            rg.setLibrary(libraryName);
+        }
+    }
+
+    Map<String, File> createOutputMap(
```
nh13: could this be `static`?

src/java/picard/sam/RevertSam.java
```diff
@@ -359,56 +416,344 @@ public boolean filterOut(final SAMRecord first, final SAMRecord second) {
...
+    private Map<String, File> createOutputMapFromFile(final File outputMapFile) {
+        final Map<String, File> outputMap = new HashMap<String, File>();
+        final TabbedTextFileWithHeaderParser parser = new TabbedTextFileWithHeaderParser(outputMapFile);
+        for (final TabbedTextFileWithHeaderParser.Row row : parser) {
+            final String id = row.getField("READ_GROUP_ID");
+            final String output = row.getField("OUTPUT");
+            final File outputPath = new File(output);
+            outputMap.put(id, outputPath);
+        }
+        parser.close();
```
nh13: `CloserUtil.close`

nh13:
@dshiga this looks a lot better, could you fix up some of the formatting issues and then 👍 .

src/java/picard/sam/RevertSam.java
nh13: I use the dark theme.

nh13: Also, the top of the source will show yellow too.

src/java/picard/sam/revertsam/RevertSam.java
vdauwera: I think in this case it makes sense, yes. In general we should be wary of including multiple examples just to show how a particular argument is invoked (eg how we went overboard with SelectVariants in GATK) but these are likely to be equally important use cases. So +1 from me.

src/java/picard/sam/RevertSam.java
vdauwera: I agree, specifying in the title of the example what distinguishes/characterizes it is very helpful.

ktibbett:
@nh13 , this was at my request, so I do 👍 .  This output-by-readgroup functionality is analogous to what we do elsewhere, and (though I do take your point about SplitSamByLibrary, whose existence I had forgotten) this implementation means we read the file once rather than twice.

nh13:
I'd like to point out that due to sorting you may read the contents of the file more than once, but you are correct, the original file is only read once.  👍

src/java/picard/sam/RevertSam.java
dshiga: I'll change to use forEach here - don't think there's any need to use a stream.

src/java/picard/sam/RevertSam.java
dshiga: In general I like getting away from nulls, however I don't think Optional is helpful in this case. We should only ever try to access outputMap and headerMap if OUTPUT_BY_READGROUP is true. If we ever try to access them when OUTPUT_BY_READGROUP is false then there is a bug in the code somewhere, and it's appropriate to blow up in that case. I'd like to avoid masking the bug with a more obscure stack trace related to Optional, or silently succeeding but producing garbage.

src/java/picard/sam/RevertSam.java
dshiga: I had those "could be final" warnings turned off. No longer.

dshiga:
@nh13 I've addressed all your comments now. I will merge at the end of the day unless you shout.

yfarjoun:
I think that protocol should be that we wait for an explicit response
rather than assume silent acquiescence. If @nh13 takes a very long time to
respond, we can discuss how to move forward...but I doubt that would happen.

dshiga:
Okay, I'll wait -- I'm not trying to sneak this in or anything! His last comment asked for some formatting changes, then gave the 👍. In my experience that means go ahead and merge after you've made those changes. Is the convention different here?

nh13:
@dshiga 
zgpk3

dshiga:
Thanks @nh13, merging away!

yfarjoun:
@dshiga You are correct, I was slightly oversensitive here, and you would have been 💯 to merge your formatting changes. This PR was big and chatty that I wanted to make sure everything in on the up-and-up!

Thanks to all who helped here!!

PR #532, opened 2016-05-04
fixed number of bases in output (missing a + sign)

yfarjoun:
In the log message, in the end of the run, when scattering to multiple files, the total number of bases was actually the number of bases in the last intervalList.

There is no test for this since we rarely (if ever?) test the log output values. The CLP produced correct output, just incorrect log-message.


nh13:
👍

PR #535, opened 2016-05-11
New EstimateLibraryComplexity algorithm proposal

ZLyanov:
We propose to use new duplicates search algorithm:
Divide each PairRead into subsequences of a length of k (calculated based on MAX_DIFF_RATE) and for each subsequence and its position within the read, calculate a hash and store in a collection. We assume that similar PairReads produce similar hash sets, thus making it possible to filter out PairReads that are certainly not duplicates according to the metric’s logic. Thus we form the groups of PairReads that are candidates for duplicates. For these groups we perform pair-to-pair comparison of PairReads, which takes less time due to reduced number of comparisons.

Find below the graphs of the performance of the two algorithms depending on group size, the graph with zoom for group size from 0 to 500 and the histogram of distribution groups by size.

Test file: ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/phase3/data/NA12878/exome_alignment/NA12878.mapped.ILLUMINA.bwa.CEU.exome.20121211.bam

image

image

image

Performance tests:

<table>
  <tbody>
    <tr>
      <th>Algorithm</th>
      <th>Full time</th>
      <th>First part of metric</th>
      <th colspan="1">Second part of metric</th>
    </tr>
    <tr>
      <td>Identical bases (standard)</td>
      <td>20.36 minutes</td>
      <td>15.51 minutes</td>
      <td colspan="1">4.86 minutes</td>
    </tr>
    <tr>
      <td>Hash based (new)</td>
      <td>18.03 minutes</td>
      <td>15.41 minutes</td>
      <td colspan="1">2.61 minutes</td>
    </tr>
  </tbody>
</table>

### Tests for algorithms correctness checking.

In these tests we created instances of ElcHashBasedDuplicatesFinder and ElcIdenticalBasesDuplicatesFinder, gave each of them different data sets and compared results they returned to known correct answers.
First test fills the histogram based on given data.
Second test searches for duplicates and after that fills the histogram based on the data it calculated.
 Below, there are all the test cases we tested:
### Test with filling histogram
- Empty list of dups and one keeper PairedReadSequence,  that mean increment first bin
- 10 dupes + keeper PairedReadSequence,  that mean increment 11 bin
- 10 dupes +  keeper PairedReadSequence, that mean increment 11 bin and isOpticalDuplicates is true, then increment opticalHisto 11 bin 10 times 
### Test with search duplicates
- Empty list of dups,  that mean increment first bin
- 10 dupes,  that mean increment 10 bin
- 10 pair read sequences but only 9 dupes, then increment 9 bin
- 10 dupes, that mean increment 10 bin and isOpticalDuplicates is true, then increment opticalHisto 10 bin 9 times
- 10 pair read sequences but only 9 dupes, then increment 9 bin and isOpticalDuplicates is true, then increment opticalHisto 9 bin 8 times


ktibbett:
@tfenne , would you be willing to review and comment on this PR?

yfarjoun:
Just to be sure that I understand the purpose of this PR: This is a replacement/enhancement of EstimateLIbraryComplexity, not MarkDuplicates. ELC currently looks at the ends of the read pairs first, and within a set of reads that has the same bases at the ends, compares the bases in the read  a certain forgiveness for differences  considers the similar one duplicates, the library complexity is calculated from the histogram of duplicate-set-sizes. 

The proposed algorithm will do the same thing in the first part (namely sort the read-pairs according to the ends) but then when calculating the number of different bases in reads it will skip "chunks" that have the same hash rather than plot through the bases and verifying that they are, in-fact, equal. The number of hashes used is the number of bases/number of allowed errors + 1. Finally, this modification is only employed when the number of read in a set is greater than 100, presumably since the O(N^2) comparison algorithm is dominated by the larger sets of reads that have similar starts but then at some-point diverge. This algorithm "jumps" directly to the point of divergence and computer the differences there.

Issues:
- The obvious possible "problem" is due to hash collision, but I can't see how this will cause big differences in the library complexity. 
- The tests currently in place are quite sparse, and this PR does nothing to alleviate that. given that the implementation is dependent on several parameters, and read-lengths, it would be good to include other tests that have reads in various lengths to assure that there is no read-dependent problem. 
- Also I'm not sure that the test included contains sets of size > 100 so I'm not sure that any of this new code is actually tested in our test suite.

"Mathematically", the PR seems fine, there are issues with formatting, naming conventions, private/public access but I wanted to look at the underlying algorithm first.

yfarjoun:
Finally, it would be good to know what the difference in runtime is for the bam overall. I see the plot, but that hides the fact that most "groups" are actually very small, so you're showing us the conversion from n^2 to n  (which is great!) but it isn't clear what the overall effect is.

jabbarish:
@yfarjoun thank you for the review, we will come back with answers soon.

jabbarish:
@yfarjoun we did what you asked for, please check.

Regarding hashes: the hash is calculated for the strings with equals length on the alphabet of 5 symbols. The probability of duplicate hashes is near 0. Keeping in mind the fact that ELC  doesn't give exact nubmers at all and skips large groups, this potential coincidence is not significant.

jabbarish:
The following two files not related to the topic were touched :
src/java/picard/analysis/CollectJumpingLibraryMetrics.java  - simply removed ^M from the ends of lines (see commit 786ec68adea6b547249dfffb3a9a87bef0c2eb71)
src/java/picard/sam/markduplicates/util/LibraryIdGenerator.java  - replaced tab with spaces, formatted

src/java/picard/sam/markduplicates/ElcDuplicatesFinder.java
```diff
@@ -0,0 +1,78 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009 The Broad Institute
```
yfarjoun: 2016

src/java/picard/sam/markduplicates/ElcDuplicatesFinder.java
```diff
@@ -0,0 +1,78 @@
...
+            }
+        } else {
+            duplicationHisto.increment(1);
+        }
+    }
+
+    protected int minLength(byte[] read1, byte[] read2) {
+        return Math.min(Math.min(read1.length, read2.length), maxReadLength);
+    }
+
```
yfarjoun: extra empty line

src/java/picard/sam/markduplicates/ElcDuplicatesFinderResolver.java
```diff
@@ -0,0 +1,78 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009 The Broad Institute
```
yfarjoun: 2016

src/java/picard/sam/markduplicates/ElcDuplicatesFinderResolver.java
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Resolve algorithm is used for the current PairedReadSequence group
+ */
+class ElcDuplicatesFinderResolver {
+
+    /**
+     * This parameter determines the choice of the algorithm: if the group size > SUPREMUM_LIBRARY_SIZE, the modified
+     * algorithm applies
+     */
+    private static final int SUPREMUM_LIBRARY_SIZE = 100;
```
yfarjoun: MAXIMAL?

src/java/picard/sam/markduplicates/EstimateLibraryComplexity.java
```diff
@@ -180,14 +182,25 @@
      * Little class to hold the sequence of a pair of reads and tile location information.
      */
     static class PairedReadSequence extends PhysicalLocationShort {
+
+        static final  int NUMBER_BASES_IN_READ = 150;
```
yfarjoun: "EXPECTED_BASES_IN_READ"? please add a note to make clear if the code will function with a different (larger or smaller) read length and what will be the effect.

src/java/picard/sam/markduplicates/EstimateLibraryComplexity.java
```diff
@@ -180,14 +182,25 @@
...
+
         short readGroup = -1;
         boolean qualityOk = true;
         byte[] read1;
         byte[] read2;
         short libraryId;
 
+        // Hashes corresponding to read1 and read2
+        int[] hashes1;
+		int[] hashes2;
```
yfarjoun: alignment

src/java/picard/sam/markduplicates/EstimateLibraryComplexity.java
```diff
@@ -201,6 +214,31 @@ public static int getSizeInBytes() {
...
+			for (int i = 0; i < hashNum; ++i) {
+				int st = skippedBases + i * hashLength;
+				int end = st + hashLength;
+
+                // use custom hash to avoid using System.arraycopy
+                hashValues[i] = 1;
+                for (int j = st; j < end; ++j) {
+                    hashValues[i] = 31 * hashValues[i] + read[j];
+                }
+			}
```
yfarjoun: alignment

src/tests/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinderTest.java
```diff
@@ -0,0 +1,158 @@
...
+                {
+                        duplicatesFinder,
+                        new Histogram<>(),
+                        new Histogram<>(),
+                        generateSeqs(10, false),
+                        10,
+                        0,
+                        0
+                },
+                // 10 prs but only 9 dupes, that mean increment 9 bin
```
yfarjoun: what's prs?

yfarjoun: oh, paired-reads...please expand.

src/tests/java/picard/sam/markduplicates/ElcIdenticalBasesDuplicatesFinderTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+    protected ArrayList<PairedReadSequence> generateSeqs(int seqsSize, boolean isOpticalDuplicates) {
+        ArrayList<PairedReadSequence> seq = new ArrayList<>(seqsSize);
+        for (int i = 0; i < seqsSize; i++) {
+            seq.add(generatePrs(isOpticalDuplicates));
+        }
+        return seq;
+    }
+
+    protected ArrayList<PairedReadSequence> generateSeqsWithNotDup(int seqsSize, boolean isOpticalDuplicates) {
```
yfarjoun: not->no

jabbarish:
@yfarjoun Hi!  We have done what you asked for. Check please.

yfarjoun:
@ZLyanov, now that the timing results show such a small improvement (10% overall). I'm wondering whether the increase in complexity is worth it. 

@tfenne, @ktibbett @nh13 could you please chime in as well? The code itself gets my 👍 but it's more complex than the previous "compare the bases to find duplicates" model..... not sure how to proceed here.

SilinPavel:
@yfarjoun if we talk about small files like exomes it would be about 10% of overall work time, but it's still a half of algorithm's work time. And if we look at work time for whole genome files we can see that the difference between running time is about 60 minutes or 30%, and it's still 50% of algorithm's time. So the bigger the groups size, the greater the performance growth is.
We attached full logs files. 

open file for test:  ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/phase3/data/NA12878/high_coverage_alignment/NA12878.mapped.ILLUMINA.bwa.CEU.high_coverage_pcr_free.20130906.bam

hash based alg: picard.sam.markduplicates.EstimateLibraryComplexity done. Elapsed time: 160.86 minutes.
standard alg: picard.sam.markduplicates.EstimateLibraryComplexity done. Elapsed time: 220.70 minutes.

[log_standard_alg.txt](https://github.com/broadinstitute/picard/files/583100/log_standard_alg.txt)
[log_hash_based_alg.txt](https://github.com/broadinstitute/picard/files/583101/log_hash_based_alg.txt)

yfarjoun:
OK. I'm convinced that it's worth it since for large files the increase is significant.

Thanks!

ZLyanov:
Great! Thanks!

yfarjoun:
Thank you!!

PR #536, opened 2016-05-16
new CLP: FindMendelianViolations

yfarjoun:
- moved FindMendelianViolations from the private repository  …
- added a test
- cleaned up the code a-la-java8
- now with mergeable metrics
- now testing with haploid violations
- found a bug with the original implementation (regarding DP with haploids)
- extended MergableMetricBase so that @MergeByAssertEquals allows merging if field is null (for initialization)

@tfenne, I didn't realize that this was hiding inside a private repository. Would you like to review, or suggest someone to review (given that you seem to be the original author...)?


tfenne:
@yfarjoun I'll try to take a look over the next few days.  I think that if you read through and "freshened up" the code that counts as one review.  And it's been long enough I think me reviewing my own code is probably still a good review!

yfarjoun:
bump!

yfarjoun:
bump?

src/java/picard/vcf/FindMendelianViolations.java
```diff
@@ -0,0 +1,301 @@
...
+    public int THREAD_COUNT = 1;
+
+    @Option(doc = "If true then fields need to be delimited by a single tab. If false the delimiter is one or more whitespace characters." +
+            " Note that tab mode does not strictly follow the PED spec")
+    public boolean TAB_MODE = false;
+
+    private final Log LOG = Log.getInstance(FindMendelianViolations.class);
+    private final ProgressLogger progressLogger = new ProgressLogger(LOG, 10000, "variants analyzed");
+
+    private final Lazy<VCFHeader> inputHeader = new Lazy<>(new Lazy.LazyInitializer<VCFHeader>() {
```
tfenne: Would be nice to update `Lazy.LazyInitializer` to be a `@Functional` interface, or replace it with `Supplier<T>`, that way you could condense all of this to:
```java
... inputHeader = new Lazy<>(() -> try (final VCFFileReader in = new VCFFileReader(INPUT)) { return in.getFileHeader(); }
```

src/java/picard/vcf/FindMendelianViolations.java
```diff
@@ -0,0 +1,301 @@
...
+    });
+
+    private final Lazy<PedFile> pedFile = new Lazy<>(new Lazy.LazyInitializer<PedFile>() {
+        @Override
+        public PedFile make() {
+            return PedFile.fromFile(TRIOS, TAB_MODE);
+        }
+    });
+
+    private final List<Interval> parIntervals = new ArrayList<>();
```
tfenne: No `Lazy` here?

src/java/picard/vcf/FindMendelianViolations.java
```diff
@@ -0,0 +1,301 @@
...
+
+    // Stock main method.
+    public static void main(final String[] args) {
```
tfenne: These `main` methods haven't been needed since @nh13 implement the single CLP 😛

src/java/picard/vcf/FindMendelianViolations.java
```diff
@@ -0,0 +1,301 @@
...
+        return 0;
+    }
+
+    private void writeAllViolations(final MendelianViolationDetector.Result result) {
+        if (VCF_DIR != null) {
+            LOG.info(String.format("Writing family violation VCFs to %s...", VCF_DIR.getAbsolutePath()));
+            final VariantContextComparator vcComparator = new VariantContextComparator(inputHeader.get().getContigLines());
+            final Set<VCFHeaderLine> headerLines = new LinkedHashSet<>(inputHeader.get().getMetaDataInInputOrder());
+            headerLines.add(new VCFInfoHeaderLine("MV", 1, VCFHeaderLineType.String, "Type of mendelian violation."));
+            headerLines.add(new VCFInfoHeaderLine("AC_Orig", VCFHeaderLineCount.A, VCFHeaderLineType.Integer, "Original AC"));
```
tfenne: These attribute strings should probably be constant-ized somewhere.

src/java/picard/vcf/MendelianViolationDetector.java
```diff
@@ -0,0 +1,338 @@
...
+        logger.record(ctx.getContig(), ctx.getStart());
+
+        final String variantChrom = ctx.getContig();
+        final int variantPos = ctx.getStart();
+
+        // Skip anything a little too funky
+        if (ctx.isFiltered()) return;
+        if (!ctx.isVariant()) return;
+        if (!ctx.isSNP()) return;
+        if (!ctx.isBiallelic()) return;
```
tfenne: I wonder if with large callsets this shouldn't be handled differently?  Perhaps either allowing tri-allelic sites, or only filtering those where the first two alleles don't count for > some fraction (e.g. 99.5%)

src/java/picard/vcf/MendelianViolationDetector.java
```diff
@@ -0,0 +1,338 @@
...
+            if (FEMALE_CHROMS.contains(variantChrom)) {
+                if (trio.OFFSPRING_SEX == Sex.Female) {
+                    haploid = false;
+                } else if (isInPseudoAutosomalRegion(variantChrom, variantPos)) {
+                    haploid = false;
+                } else if (trio.OFFSPRING_SEX == Sex.Male) {
+                    haploid = true;
+                    haploidParentalGenotype = momGt;
+                } else {
+                    continue;
```
tfenne: Does this condition ever get evaluated?  I guess maybe if sex is unknown?  Maybe add a comment to that effect?

src/java/picard/vcf/MendelianViolationDetector.java
```diff
@@ -0,0 +1,338 @@
...
+                }
+            }
+
+            // the PAR on the male chromosome should be masked so that reads
```
tfenne: This is a safe assumption at Broad - but perhaps it should be documented in the usage for others?

src/java/picard/vcf/MendelianViolationDetector.java
```diff
@@ -0,0 +1,338 @@
...
+                } else if (kidGt.isHom() && ((momGt.isHom() && dadGt.isHet()) || (momGt.isHet() && dadGt.isHom()))) {
+                    trio.NUM_HOM_HET_HOM++;
+                    type = MendelianViolation.Hom_Het_Hom;
+                } else {
+                    trio.NUM_OTHER++;
+                    type = MendelianViolation.Other;
+                }
+            }
+
+            // Output a record into the family's violation VCF
```
tfenne: It's unfortunate that all of the violation VariantContexts are kept in memory until the end of the program run, and only then written to disk.  For large call sets with 1000s of trios this could become a problem.  I think this was introduced with the parallelization of the code, and I'm not sure what we can do about it.

src/java/picard/vcf/MendelianViolationDetector.java
```diff
@@ -0,0 +1,338 @@
...
+
+        return false;
+    }
+
+    /** Tests whether the alleles of the offspring are possible given the alleles of the parents. */
+    private boolean isMendelianViolation(final Genotype p1, final Genotype p2, final Genotype off) {
+        final Allele offAllele1 = off.getAllele(0);
+        final Allele offAllele2 = off.getAllele(1);
+
+        for (final Allele a1 : p1.getAlleles()) {
```
tfenne: I may have written this, but I find it a little confusing to look at now.  Perhaps it could just be:
```java
return !((p1.contains(offAllele1) && p2.contains(offAllele2)) || (p2.contains(offAllele1) && p1.contains(offAllele2)))
```

src/java/picard/vcf/MendelianViolationDetector.java
```diff
@@ -0,0 +1,338 @@
...
+            return metrics;
+        }
+
+        Map<String, Collection<VariantContext>> violations() {
+            return violations;
+        }
+
+        public static MendelianViolationDetector.Result merge(final Collection<MendelianViolationDetector.Result> results) {
+            final Collection<Collection<MendelianViolationMetrics>> metricCollections = new ArrayList<>();
+            final Collection<Map<String, Collection<VariantContext>>> violationCollections = new ArrayList<>();
```
tfenne: This makes my eyes water a bit!  How about defining a simple class:
```java
class ViolationsByFamily extends HashMap<String, Collection<VariantContext>>
```
to make it clearer what's going on?

src/tests/java/picard/vcf/FindMendelianViolationsTest.java
```diff
@@ -0,0 +1,148 @@
...
+ * Created by farjoun on 5/13/16.
+ */
+public class FindMendelianViolationsTest {
+
+    private static final File TEST_DATA_DIR = new File("testdata/picard/vcf");
+
+
+    @Test
+    public void testFindMedelianViolations() throws IOException {
+        final File vcfFile = new File(TEST_DATA_DIR, "NA12878.vcf");
```
tfenne: It's not just NA12878 though right?  Maybe call it NA12878_family.vcf?

src/tests/java/picard/vcf/FindMendelianViolationsTest.java
```diff
@@ -0,0 +1,148 @@
...
+        final File vcfIndexFile = new File(TEST_DATA_DIR, "NA12878.vcf.idx");
+        vcfIndexFile.deleteOnExit();
```
tfenne: Why are you deleting the index file (and possibly re-creating?)?  Also, is there no way to not require the index if the program is run single-threaded?

tfenne:
Looks good @yfarjoun .  A bunch of mostly minor comments from me.

src/java/picard/vcf/FindMendelianViolations.java
yfarjoun: no need...it's final without being lazy.

src/java/picard/vcf/FindMendelianViolations.java
yfarjoun: I'll make them public final static right here, until there's a better place. the only other place  that seems possible is HTSJDK.

src/java/picard/vcf/MendelianViolationDetector.java
yfarjoun: I could wait for the genotypes and check there...

same for "isSNP"?

yfarjoun: I think that ideally, it would. but it will take some effort to get the multiallelic code right and I'd rather have something that works for most call-sets than nothing...

src/java/picard/vcf/FindMendelianViolations.java
tfenne: True - but it's only final because you initialize it to an empty list, and then fill the list at the start of operations.

src/java/picard/vcf/MendelianViolationDetector.java
tfenne: I think that's what I would do - in the `for` loop for the trios, pull the three genotypes, and then ensure that across all three you have only two alleles and they are both `SNP` alleles.

yfarjoun: I'd like to avoid dealing with 1/2 genotypes for now, does that seems reasonable?

src/java/picard/vcf/MendelianViolationDetector.java
yfarjoun: actually, I'm not sure I understand this comment. As you know, I'm not the original author of this...do you know how the PAR is masked on the Y? I looked in the reference and didn't see it being masked.

src/java/picard/vcf/MendelianViolationDetector.java
tfenne: Totally - so only sites that contain ref and a single alt SNP allele in the family.

src/java/picard/vcf/MendelianViolationDetector.java
tfenne: Generally speaking you'd want to use a reference in an NGS pipeline that only contains one copy of the PARs, so as not to cause all reads from those regions to be at mapq and spread across X and Y.  It's common practice to leave the regions alone on X and mask them to Ns on Y.  The GRC page has a nice little table on it of the X and Y coordinates of the PARs here: http://www.ncbi.nlm.nih.gov/projects/genome/assembly/grc/human/

src/java/picard/vcf/FindMendelianViolations.java
yfarjoun: OK. I made it into a Lazy<Set> and put in an unmodifiableSet....

src/java/picard/vcf/MendelianViolationDetector.java
yfarjoun: "this comment" being the one in the code, not your comment in github.

src/java/picard/vcf/MendelianViolationDetector.java
yfarjoun: yes...not sure how to get around it...these could be written to disk and then merged...but that would be much work.

src/java/picard/vcf/MendelianViolationDetector.java
yfarjoun: That's what I **assumed** happens...but then I cracked open IGV and saw bases in both Y PARs in both GRC 37 and hg38...and so I'm confused....

src/java/picard/vcf/MendelianViolationDetector.java
yfarjoun: sure. though it has to extend DefaultingMap for some of the shinanigans.

src/tests/java/picard/vcf/FindMendelianViolationsTest.java
yfarjoun: Not as currently written. The code was written using the VariantProcessor framework, which doesn't seem to have an option of "just process the whole damn file!"....

src/java/picard/vcf/MendelianViolationDetector.java
tfenne: I don't know what IGV is using for it's reference...

yfarjoun: good point.

yfarjoun: yes, the IGV reference for hg19 isn't masked, but the broad's "production" one is...how odd.

src/java/picard/vcf/MendelianViolationDetector.java
yfarjoun: fixed.

src/tests/java/picard/vcf/FindMendelianViolationsTest.java
yfarjoun: Also, I kept changing the vcf during testing and so having the test framework build the index for me was helpful. No need during commit, but I'll keep around as commented code to help when adding more tests

yfarjoun:
@tfenne I responded to your comments. please re-review.

yfarjoun:
this needed a rebase. waiting for tests to pass. 

@tfenne I'm assuming that since I responded to your "mostly minor" comments, and that you had no more comments, that I"m good to go here. Will wait a few more days after this note, please respond (with a short note) if you want more time to review.

yfarjoun:
@vdauwera a new tool! please make sure it is documented to your satisfaction!

vdauwera:
Ah, the most important thing to do is to add the tool to the build.xml file, otherwise it won't get documented on the website... There's a couple other tools that are also not included; I'll make a PR to fix that and beef up the docs if needed. Thanks for the heads up in any case.

PR #538, opened 2016-05-17
Small changes to make HetSense deterministic

kbergin:
Tests for theoretical het sensitivity passed locally, having trouble with my picard-htsjdk set up to try testing all so opening a PR so it tests for me :)


kbergin:
@yfarjoun I believe this checks out!

yfarjoun:
I'll take that as a 👍

PR #540, opened 2016-05-18
MergeVcfs should not clear options be default.

nh13:
For example, VALIDATION_STRINGENCY=SILENT will be cleared!


nh13:
@tfenne or @yfarjoun can you take a look?

tfenne:
Seems reasonable to me.

PR #542, opened 2016-05-20
Added ivy dependency for htsjdk. To build htsjdk from source use the …

jacarey:
…build-from-src.xml build file.

### DIFF
--- a/build-from-src.xml
+++ b/build-from-src.xml
@@ -2,7 +2,7 @@
 <!--
   ~ The MIT License
   ~
-  ~ Copyright (c) 2009 The Broad Institute
+  ~ Copyright (c) 2016 The Broad Institute
   ~
   ~ Permission is hereby granted, free of charge, to any person obtaining a copy
   ~ of this software and associated documentation files (the "Software"), to deal
--- a/build-from-src.xml
+++ b/build-from-src.xml
@@ -23,7 +23,8 @@
   ~ THE SOFTWARE.
   -->
 
-<project name="picard-public" basedir="." default="all" xmlns:ivy="antlib:org.apache.ivy.ant">
+<project name="picard-public" basedir="." default="all">
+
     <property name="src" value="src/java"/>
     <property name="src.scripts" value="src/scripts"/>
     <property name="src.test" value="src/tests"/>
@@ -37,6 +38,10 @@
     <property name="scripts" value="src/scripts"/>
     <property name="reports" value="dist/test"/>
     <property name="test.output" value="dist/test"/>
+    <property name="htsjdk_lib_dir" value="${dist}/htsjdk_lib_dir"/>
+    <property name="htsjdk_git_url" value="git@github.com:samtools/htsjdk.git"/>
+    <!-- TODO: get this from the build.xml in htsjdk -->
+    <property name="htsjdk-classes" value="htsjdk/classes"/>
 
     <property name="javac.target" value="1.8"/>
     <property name="javac.debug" value="true"/>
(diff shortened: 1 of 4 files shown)


build-from-src.xml
```diff
@@ -0,0 +1,636 @@
+<?xml version="1.0"?>
+<!--
+  ~ The MIT License
+  ~
+  ~ Copyright (c) 2009 The Broad Institute
```
nh13: Year?

jacarey: copy/paste... I'll update it.

src/scripts/release_picard.sh
```diff
@@ -184,12 +184,12 @@ cd $TMPDIR
 # clone
 git clone $PICARDGITROOT picard
 cd picard
-ant clone-htsjdk
+ant -buildfile build-from-src.xml clone-htsjdk
```
droazen: I'm not sure it's a good idea to have Picard release with potentially a different version of htsjdk than is pulled down during a typical build and tested with in travis, etc. If Picard needs to release with a newer version of htsjdk than is in master, why not just do a PR to update `ivy.xml` before doing the release?

jacarey: I was just leaving current functionality in the release scripts (which pulls the latest tag), but I agree that changing the ivy xml is preferable. I will change this.

jacarey: Ah now I'm remembering why. This is for the github release.. which releases both picard and htsjdk simultaneously. So if we didn't grab the htsjdk master we would be releasing the previous version of htsjdk. Really these two releases should be decoupled even though they are released at the same time.

droazen: Is it technically possible to decouple htsjdk and picard github releases as things stand now, or would that be hard?

jacarey: It is possible. Just a little work. Alternately we could just ensure that the builds have been tested and released to OSS. And **then** do the release to github. I think for simplicity's sake we really should only be releasing to one place (OSS) and the github release script could just become a documentation update.

droazen: The idea of releasing only to OSS sounds good to me -- I imagine someone would still have to run a "git tag" command before each release, so the release workflow might just become: tag (and update version number in build.sbt?), release to OSS, and publish documentation via the script.

Is this simpler option do-able in this branch, or would you prefer to change the release process in a separate PR?

jacarey: I think I'd prefer a separate PR since I'm not sure if I will be able to get to it right away.

build.xml
```diff
@@ -80,28 +103,21 @@
     </target>
     <target name="set_excluded_test_groups" depends="set_excluded_test_groups_unix,set_excluded_test_groups_non_unix"/>
 
+	<target name="resolve" description="--> retrieve dependencies with ivy" depends="init-ivy">
```
droazen: indentation

build.xml
```diff
@@ -307,24 +257,19 @@
...
-                <include name="*.jar"/>
-            </fileset>
         </unjar>
 
         <jar destfile="${dist}/picard.jar" compress="no">
             <fileset dir="${classes}" includes="picard/**/*.*, META-INF/**/*"/>
             <fileset dir="${src.scripts}" includes="**/*.R"/>
-            <fileset dir="${htsjdk-classes}" includes ="${htsjdk}/*/**/*.*"/>
-            <fileset dir="${dist.tmp}" includes="**/*"/>
+			<fileset dir="${dist.tmp}" includes="**/*"/>
```
droazen: indentation

build.xml
```diff
@@ -523,10 +431,7 @@
             description="Default build target">
         <zip zipfile="${dist}/picard-tools-${picard-version}.zip">
             <zipfileset dir="${dist}" includes="*.jar" prefix="picard-tools-${picard-version}"/>
-            <zipfileset dir="${htsjdk_lib_dir}" includes="htsjdk-*.jar" prefix="picard-tools-${picard-version}"/>
             <zipfileset dir="${lib}" includes="snappy*.jar"/>
-            <!-- distribute libJniDeflater.so in same directory as jarfiles-->
-            <zipfileset dir="${htsjdk}/lib/jni" includes="*" prefix="picard-tools-${picard-version}"/>
```
droazen: Can you check whether `libJniDeflater.so` gets packaged in the Picard jar with the new build file? Currently htsjdk/Picard require that the path to the deflator be passed in as a system property, but shortly we'll be changing this so that it finds the deflator on the classpath and extracts and loads it at runtime.

ivy.xml
```diff
@@ -0,0 +1,6 @@
+<ivy-module version="2.0">
+    <info organisation="org.github.broadinstitute" module="picard"/>
+    <dependencies>
+        <dependency org="com.github.samtools" name="htsjdk" rev="2.3.0" />
```
droazen: Are there any non-htsjdk dependencies currently checked in as jars (eg., guava) that could be pulled down from maven as well?

src/scripts/release_picard.sh
droazen: Ok, sounds good -- I'll create a separate ticket to track the release process changes.

droazen:
Review complete, back to @jacarey, merge after addressing comments.

src/scripts/release_picard.sh
droazen: Created https://github.com/broadinstitute/picard/issues/545

yfarjoun:
do we need this or can we close it?

vdauwera:
Is this still a thing or was it an alternate/partial piece of the big reorg/migration that already happened?

lbergelson:
This is old and no longer relevant as far as I know.  Going to close.  @jacarey Reopen if you disagree.

PR #544, opened 2016-05-23
Output the read count per target.

nh13:
When mapping to short amplicons, we care about the # of reads that map to the amplicon.  This commit counts the # of reads that have at least one base mapping to a given interval.

### DIFF
--- a/src/java/picard/analysis/directed/TargetMetricsCollector.java
+++ b/src/java/picard/analysis/directed/TargetMetricsCollector.java
@@ -524,6 +524,7 @@ public abstract class TargetMetricsCollector<METRIC_TYPE extends MultilevelMetri
             } else rec = record;
 
             // Find the target overlaps
+            final Set<Interval> coveredTargets = new HashSet<>();
             for (final AlignmentBlock block : rec.getAlignmentBlocks()) {
                 final int length = block.getLength(), refStart = block.getReferenceStart(), readStart = block.getReadStart();
 
@@ -538,7 +539,6 @@ public abstract class TargetMetricsCollector<METRIC_TYPE extends MultilevelMetri
                         boolean isOnTarget = false;
                         for (final Interval target : targets) {
                             if (refPos >= target.getStart() && refPos <= target.getEnd()) {
-                                isOnTarget = true;
                                 ++this.metrics.ON_TARGET_BASES;
                                 if (mappedInPair) ++this.metrics.ON_TARGET_FROM_PAIR_BASES;
 
@@ -546,6 +546,11 @@ public abstract class TargetMetricsCollector<METRIC_TYPE extends MultilevelMetri
                                 final Coverage coverage = this.coverageByTarget.get(target);
                                 coverage.addBase(targetOffset);
                                 baseQHistogramArray[baseQualities[offset]]++;
+                                if (!coveredTargets.contains(target)) {
+                                    coverage.incrementReadCount();
+                                    coveredTargets.add(target);
+                                }
+                                isOnTarget = true;
                             }
                         }
 
@@ -702,7 +707,7 @@ public abstract class TargetMetricsCollector<METRIC_TYPE extends MultilevelMetri
                 try {
                     if (perTargetOutput != null) {
                         out = new PrintWriter(perTargetOutput);
-                        out.println("chrom\tstart\tend\tlength\tname\t%gc\tmean_coverage\tnormalized_coverage\tmin_normalized_coverage\tmax_normalized_coverage\tmin_coverage\tmax_coverage\tpct_0x");
+                        out.println("chrom\tstart\tend\tlength\tname\t%gc\tmean_coverage\tnormalized_coverage\tmin_normalized_coverage\tmax_normalized_coverage\tmin_coverage\tmax_coverage\tpct_0x\tread_count");
                     }
                     else {
                         out = null;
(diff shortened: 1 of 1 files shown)


src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -814,6 +817,11 @@ public void addBase(final int offset) {
             }
         }
 
+        /** Increments the # of reads mapping to this target. */
+        public void addReadCount() {
```
tfenne: `incrementReadCount()`?

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -538,14 +538,15 @@ public void acceptRecord(final SAMRecord record) {
...
-                                isOnTarget = true;
                                 ++this.metrics.ON_TARGET_BASES;
                                 if (mappedInPair) ++this.metrics.ON_TARGET_FROM_PAIR_BASES;
 
                                 final int targetOffset = refPos - target.getStart();
                                 final Coverage coverage = this.coverageByTarget.get(target);
                                 coverage.addBase(targetOffset);
                                 baseQHistogramArray[baseQualities[offset]]++;
+                                if (!isOnTarget) coverage.addReadCount();
+                                isOnTarget = true;
```
tfenne: Looking at the change here, this will _only_ increment the read count for the _first_ target that a read maps over, which seems a little odd.  E.g. if you have two targets close together, separated by 20 bases, both at ~100X with most reads straddling both you could end up with the second target having ~100X coverage but very low read count.

tfenne: Oh, and will also increment the same target multiple times if there's an indel, causing the read to have multiple alignment blocks overlapping the same target.

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: The reason was there was an `addBase` and so I wanted to make it easy to search for `add*`.  Updating to `incrementReadCount`.

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: Done.

nh13:
@tfenne updated for your review

nh13:
@tfenne bump.

tfenne:
This looks ok to me, but given how complicates this class is now, I think a second set of eyes on it would be good.  @yfarjoun or @ktibbett could one of you (or someone you choose) take a peek please?

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -538,21 +539,26 @@ public void acceptRecord(final SAMRecord record) {
...
+                                coveredTargets.add(target);
+                                isOnTarget = true;
                             }
                         }
 
                         if (!isOnTarget) this.metrics.PCT_EXC_OFF_TARGET++;
                     }
                 }
             }
+            for (final Interval target : coveredTargets) {
```
vivster7: this for loop can be replaced with 
`if (!coveredTargets.contains(target)) coverage.incrementReadCount();`
or
```
if (!coveredTargets.contains(target)) {
  coverage.incrementReadCount();
  coveredTargets.add(target);
}
```

If the reason we are keeping distinct targets in `coveredTargets` is to keep a read count then it makes sense for the variables to live together.

nh13: The main goal of this loop is to increment only once the read count per unique covered target.   `coveredTargets` collects the unique targets covered by this record, while `coverageByTarget` is the coverage per target across all targets.  I don't understand your point here.  Can you take a look again?

src/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -538,21 +539,26 @@ public void acceptRecord(final SAMRecord record) {
...
                             if (refPos >= target.getStart() && refPos <= target.getEnd()) {
-                                isOnTarget = true;
                                 ++this.metrics.ON_TARGET_BASES;
                                 if (mappedInPair) ++this.metrics.ON_TARGET_FROM_PAIR_BASES;
 
                                 final int targetOffset = refPos - target.getStart();
                                 final Coverage coverage = this.coverageByTarget.get(target);
                                 coverage.addBase(targetOffset);
                                 baseQHistogramArray[baseQualities[offset]]++;
+                                coveredTargets.add(target);
```
vivster7: if you add the line
```
if (!coveredTargets.contains(target)) coverage.incrementReadCount();
``` 
above `coveredTargets.add(target);` then I think you can get rid of the for loop below.

src/java/picard/analysis/directed/TargetMetricsCollector.java
vivster7: I've added a comment on line 549 (where the if-statement should be inserted). I think that might be bit clearer.

src/java/picard/analysis/directed/TargetMetricsCollector.java
nh13: Ah, this loop :)

nh13:
@vivster7 take a look at the last commit, and after a 👍 I will rebase, wait until tests pass, and merge (look who has those privileges now!).

vivster7:
👍 , looks great

PR #546, opened 2016-05-23
Added text to metrics definitions percent = fraction

dekling:
Added information to metricsdefinitions page to indicate that percent = fraction

### DIFF
--- a/_includes/picard-metric-definitions.html
+++ b/_includes/picard-metric-definitions.html
@@ -1,6 +1,6 @@
 <h2>Picard Metrics Definitions</h2>
 <section>
-<p> Click on a metric to see a description of its fields.</p>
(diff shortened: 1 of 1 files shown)


_includes/picard-metric-definitions.html
```diff
@@ -57,6 +57,9 @@
 <li><a href="#TargetedPcrMetrics">TargetedPcrMetrics</a>: Metrics class for targeted pcr runs such as TSCA runs</li>
 </ol>
 </section>
+<section>
+<p><b>Please note: Although metrics are indicated as percents, they are actually fractions.</b></p>
```
vdauwera: This should go at the top of the page; otherwise people might scroll down only until they find the tool they're interested in, click on it and miss the warning. It can even be on the same line as that first "Click on a metric to see a description of its fields." sentence.

vdauwera:
Just one comment on placement of the new text.

vdauwera:
OK, merging

PR #553, opened 2016-06-01
Printing the program list is not an error

dakl:
I propose that printing the program list is not an error and should exit with a 0.


yfarjoun:
👎  This seems to only matter in a pipeline setting anyway, in which case you probably want to know that you didn't give the ProgramName. current behavior seems right.

PR #554, opened 2016-06-02
First pass at picard gradle build file.

jacarey:


droazen:
@jacarey Rather than having a separate build file to build htsjdk from source, why not just add `mavenLocal()` as a repository in the main build file, and have people install local htsjdk snapshots whenever they need to use a locally-built version? This is only 1 extra command, and avoids the need for a separate parallel build system.

build.gradle
```diff
@@ -0,0 +1,354 @@
...
+    id 'signing'
+    id 'jacoco'
+    id 'com.palantir.git-version' version '0.5.1'
+    id 'com.github.johnrengelman.shadow' version '1.2.3'
+    id "com.github.kt3k.coveralls" version "2.6.3"
+    id 'org.ajoberstar.grgit' version "1.4.2"
+}
+
+repositories {
+    mavenCentral()
```
droazen: consider adding `mavenLocal()` as well to allow local htsjdk snapshots to be used (potential replacement for the build from source script below)

build.gradle
```diff
@@ -0,0 +1,354 @@
...
+
+    // Include the Broad artifactory for intel-gkl snapshots. This is temporary
+    // until intel-gkl is available on maven central.
+    //maven {
+    //    url "https://artifactory.broadinstitute.org/artifactory/libs-snapshot/"
+    //}
```
droazen: Should probably remove this commented-out reference to the artifactory, since `intel-gkl` is not currently a Picard dependency.

build.gradle
```diff
@@ -0,0 +1,354 @@
...
+    }
+}
+
+jacoco {
+    toolVersion = "0.7.5.201505241946"
+}
+
+dependencies {
+    compile 'com.google.guava:guava:15.0'
+    compile 'com.github.samtools:htsjdk:2.4.1'
```
droazen: woohoo! :)

build.gradle
```diff
@@ -0,0 +1,354 @@
...
+                      "picard.sam.markduplicates.MarkDuplicates", "picard.sam.markduplicates.MarkDuplicatesWithMateCigar", "picard.analysis.MeanQualityByCycle",
+                      "picard.sam.MergeBamAlignment", "picard.sam.MergeSamFiles", "picard.vcf.MergeVcfs", "picard.reference.NormalizeFasta", "picard.sam.PositionBasedDownsampleSam",
+                      "picard.reference.ExtractSequences", "picard.analysis.QualityScoreDistribution", "picard.vcf.RenameSampleInVcf", "picard.sam.ReorderSam",
+                      "picard.sam.ReplaceSamHeader", "picard.sam.RevertSam", "picard.sam.RevertOriginalBaseQualitiesAndAddMateCigar", "picard.sam.SamFormatConverter",
+                      "picard.sam.SamToFastq", "picard.util.ScatterIntervalsByNs", "picard.sam.SortSam", "picard.vcf.SortVcf", "picard.sam.SplitSamByLibrary",
+                      "picard.vcf.UpdateVcfSequenceDictionary", "picard.vcf.VcfFormatConverter", "picard.illumina.MarkIlluminaAdapters", "picard.vcf.SplitVcfs",
+                      "picard.sam.ValidateSamFile", "picard.sam.ViewSam", "picard.vcf.VcfToIntervalList"]
```
droazen: This list will have to be manually updated whenever a new Picard tool is added?

build.gradle
```diff
@@ -0,0 +1,354 @@
...
+    repositories {
+        mavenDeployer {
+            beforeDeployment { MavenDeployment deployment -> signing.signPom(deployment) }
+
+            repository(url: "https://oss.sonatype.org/service/local/staging/deploy/maven2/") {
+                authentication(userName: project.findProperty("sonatypeUsername"), password: project.findProperty("sonatypePassword"))
+            }
+
+            snapshotRepository(url: "https://artifactory.broadinstitute.org/artifactory/libs-snapshot-local/") {
+                authentication(userName: System.env.ARTIFACTORY_USERNAME, password: System.env.ARTIFACTORY_PASSWORD)
```
droazen: Have you done a test deployment to the artifactory to see if it works?

build.gradle
```diff
@@ -0,0 +1,354 @@
...
+task updateWebsite(dependsOn: 'copyStaticWebFiles') << {
+    def grgit = Grgit.open(project.projectDir)
+    grgit.branch.remove(names:['gh-pages'], force: true)
+    grgit.checkout(branch: 'gh-pages', startPoint: 'origin/gh-pages', createBranch: true)
+    //add javadoc files
+    grgit.add(patterns: ['javadoc'])
+
+    println("Adding website files for release: $version")
+
+    grgit.commit(message: "Adding website files for release: $version", all: true)
```
droazen: Is this going to be a complete replacement for the release script, or will that still be needed?

droazen:
@jacarey I've added some comments. @lbergelson Could you do a review pass on the new gradle build scripts for correctness?

buildWithHtsjdkSrc.gradle
```diff
@@ -0,0 +1,172 @@
+import javax.tools.ToolProvider
```
lbergelson: I'm not sure I understand why this is a separate build file.  It seems like it should be a different configuration in the main build file.

build.gradle
```diff
@@ -0,0 +1,354 @@
...
+    classifier = 'javadoc'
+    from 'build/docs/javadoc'
+}
+
+task sourcesJar(type: Jar) {
+    from sourceSets.main.allSource
+    classifier = 'sources'
+}
+
+task libJar(type: Jar) {
```
lbergelson: I'm not sure I understand the lib jar.  Is it a jar that includes all the libraries packaged in it?  That's usually done using a shadowJar task.

build.gradle
```diff
@@ -0,0 +1,354 @@
...
+        System.out.println("Uploading version $version")
+    }
+}
+
+task copyJavadoc(dependsOn: 'javadoc', type: Copy) {
+    from 'build/docs/javadoc'
+    into 'javadoc'
+}
+
+task updateGithubDocumentation(dependsOn: 'copyJavadoc') << {
```
lbergelson: cool, we should add this to the htsjdk build too in the future.

buildWithHtsjdkSrc.gradle
```diff
@@ -0,0 +1,172 @@
...
+        afterSuite { desc, result ->
+            if (!desc.parent) { // will match the outermost suite
+                println "Results: ${result.resultType} (${result.testCount} tests, ${result.successfulTestCount} successes, ${result.failedTestCount} failures, ${result.skippedTestCount} skipped)"
+            }
+        }
+    }
+}
+
+test {
+    description = "Runs the unit tests other than the SRA tests"
```
lbergelson: This description should probably be updated since I don't think there are any SRA tests.  We had separate `test` and `tasks.withType(Test)` configurations in htjsdk because we had multiple tasks of type test, but since there's only one here you could just roll the above configuration into this task.

build.gradle
```diff
@@ -0,0 +1,354 @@
...
+    outputs.upToDateWhen { false } // tests will always rerun
+    useTestNG()
+
+    // set heap size for the test JVM(s)
+    minHeapSize = "1G"
+    maxHeapSize = "2G"
+
+    jvmArgs '-Djava.awt.headless=true'  //this prevents awt from displaying a java icon while the tests are running
+
+    if (System.env.CI == "true") {  //if running under a CI output less into the logs
```
lbergelson: We added this to htsjdk/gatk because there was a problem with too many tests being too verbose.  It may not be necessary for picard.

build.gradle
```diff
@@ -0,0 +1,354 @@
...
+}
+// This is a hack to disable the java 8 default javadoc lint until we fix the html formatting
+if (JavaVersion.current().isJava8Compatible()) {
+    tasks.withType(Javadoc) {
+        options.addStringOption('Xdoclint:none', '-quiet')
+    }
+}
+
+test {
+    description = "Runs the unit tests other than the SRA tests"
```
lbergelson: We had separate `test` and `tasks.withType(Test)` configurations in htjsdk because we had multiple tasks of type test, but since there's only one here you could just roll the above configuration into this task.

lbergelson: Probably don't want to talk about SRA tests here either.

build.gradle
```diff
@@ -0,0 +1,354 @@
...
+    useTestNG {
+        if (OperatingSystem.current().isUnix()) {
+            excludeGroups "slow", "broken"
+        } else {
+            excludeGroups "slow", "broken", "unix"
+        }
+    }
+}
+
+task picardZip(type: Zip) {
```
lbergelson: using the `application` plugin will provide a distZip task that does essentially this, as well as some tasks like `installDist` and `run`.

build.gradle
```diff
@@ -0,0 +1,354 @@
...
+
+task documentInit(dependsOn: 'classes') {
+    commandLineHtmlDir.delete()
+    commandLineHtmlDir.mkdirs()
+}
+
+task documentAll
+
+task documentStandardOptions(dependsOn: 'documentInit', type: JavaExec) {
+    main = 'picard.cmdline.CreateHtmlDocForStandardOptions'
```
lbergelson: You should probably configure the `outputs` for custom tasks that produce files.  That way it will both understand when the task has run before, and properly configure `gradle clean` to remove them.

build.gradle
```diff
@@ -0,0 +1,354 @@
...
+        args mainClass
+        standardOutput = new FileOutputStream(new File(commandLineHtmlDir, mainClass.substring(mainClass.lastIndexOf(".") + 1) + ".html"))
+        doLast {
+            usageFile.append("{% include ${mainClass.substring(mainClass.lastIndexOf(".") + 1) + ".html"} %}")
+            usageFile.append(System.getProperty("line.separator"))
+            sidebarFile.append("<li><a href=\"command-line-overview.html#${mainClass.substring(mainClass.lastIndexOf(".") + 1)}\">${mainClass.substring(mainClass.lastIndexOf(".") + 1)}</a>")
+            sidebarFile.append(System.getProperty("line.separator"))
+        }
+    }
+    if (lastDocTask != null) currentDocTask.dependsOn lastDocTask
```
lbergelson: I'm assuming this ordering is important?   It seems a bit awkward the way it's done right now.  Maybe it would be cleaner to use `.eachwithindex` instead of the global properties.  Maybe not though.

.gitignore
```diff
@@ -13,4 +13,5 @@ htsjdk
 target
```
lbergelson: might want to remove the old ant ignores so that people delete them and they don't get confused

build.gradle
```diff
@@ -0,0 +1,354 @@
...
+task createMetricsDoc(dependsOn: ['documentInit', 'classes'], type: Javadoc){
+    source = sourceSets.main.allJava
+    classpath = sourceSets.main.runtimeClasspath
+    destinationDir = file('dist/html')
+    options.doclet = 'picard.util.MetricsDoclet'
+    options.docletpath = sourceSets.main.runtimeClasspath.asType(List)
+    documentAll.dependsOn "createMetricsDoc"
+}
+
+task javadocJar(type: Jar, dependsOn: javadoc) {
```
lbergelson: should this depend on `documentAll` instead of javadoc?  It might be best to configure the default javadoc instead of using a custom task here.  We can always revise later though if needed.

build.gradle
lbergelson: this might be less verbose using [org.ajoberstar.github-pages](https://github.com/ajoberstar/gradle-git) instead of the raw grgit commands

lbergelson:
@jacarey Back to you with some comments and questions.  Looks pretty good to me although I'd really like to be able to remove or greatly reduce that second gradle build file.  I think it can probably be represented as a separate build configuration with a different dependency set, but I may be missing something about it.

I like the inclusion of the github pages publishing in the gradle script, but I'm hoping we can maybe make it and the documentation somewhat less verbose.  Not a blocking issue, maybe something we can update in the future as we figure out cleaner ways to do it.

build.gradle
jacarey: Yeah.. sadly that's how it has always been. A list could be generated using reflection but I don't think anyone wanted to take that time to incorporate that into the build.

build.gradle
jacarey: I have not yet

build.gradle
jacarey: This just mirrors what the ant build was building for a picard-library.jar  If shadow jar can build the same thing then i'm all for it.

build.gradle
jacarey: Oh nice I'll look in to that

jacarey:
I'll try to address the concern about two build files. It shouldn't be difficult now that I've got both working. I'm likely not going to have as much time to put in to this in the following week if someone wants to pick things up from where it is now. Also I need to update all our internal build scripts at some point to support gradle.

build.gradle
jacarey: Does this ensure that maven local will be checked first? Also wouldn't you need to change the dependency to the snapshot id every time you rebuilt htsjdk?

build.gradle
jacarey: Good to know I'll look in to it.

build.gradle
droazen: It will be checked first if you list it first :) And we should be able to get around the need for a manual change to the snapshot dependency after every local htsjdk build / local maven install due to the way snapshot versioning works:

    There are many situations when you want to use the latest version of a particular dependency, or the latest in a range of versions. This can be a requirement during development, or you may be developing a library that is designed to work with a range of dependency versions. You can easily depend on these constantly changing dependencies by using a dynamic version. A dynamic version can be either a version range (e.g. 2.+) or it can be a placeholder for the latest version available (e.g. latest.integration).

    Alternatively, sometimes the module you request can change over time, even for the same version. An example of this type of changing module is a Maven SNAPSHOT module, which always points at the latest artifact published. In other words, a standard Maven snapshot is a module that never stands still so to speak, it is a “changing module”.

build.gradle
jacarey: it seems like it would work if I could remember my artifactory login

jacarey:
Need to update the htsjdk gradle build in order to support publishToLocalMaven.  PR https://github.com/samtools/htsjdk/pull/632

akiezun:
@jacarey @lbergelson where do we stand on this?

jacarey:
@akiezun I'm going to be working on finishing this up today.

build.gradle
```diff
@@ -0,0 +1,348 @@
...
+    }
+}
+
+jacoco {
+    toolVersion = "0.7.5.201505241946"
+}
+
+dependencies {
+    compile 'com.google.guava:guava:15.0'
+    compile ('com.github.samtools:htsjdk:2.4.1+')
```
droazen: Can we change this to depend on a specific htsjdk version @jacarey? (since one of our goals is to be able to release htsjdk without breaking the current picard).

jacarey: @droazen I'm a bit confused now. We want to support dynamic version so we can build with locally published htsjdk. However if we specify a specific version then we would need to update this every time we rebuilt a new local version as well.

droazen: The master branch should depend on a fixed version, so that releases of htsjdk don't break it. For local development, you can specify a SNAPSHOT version that doesn't need to be updated on every local build (it will use the latest locally-installed snapshot). See, eg., https://docs.gradle.org/current/userguide/dependency_management.html#sec:dependency_management_overview

droazen: What I'm specifically referring to is the section on "changing modules":

```
Alternatively, sometimes the module you request can change over time, even for the same version. An example of this type of changing module is a Maven SNAPSHOT module, which always points at the latest artifact published. In other words, a standard Maven snapshot is a module that never stands still so to speak, it is a “changing module”.

The main difference between a dynamic version and a changing module is that when you resolve a dynamic version, you'll get the real, static version as the module name. When you resolve a changing module, the artifacts are named using the version you requested, but the underlying artifacts may change over time.
```

We could depend on a fixed version in master (which lets us decouple the two projects), and tell people to switch to a snapshot dependency when doing local development in htsjdk and picard simultaneously.

jacarey: @droazen Ah OK. Thanks I'm on board now.

jacarey:
@droazen @lbergelson I think this is ready for another review pass. I removed the second build file and used gh-pages plugin for documentation push. Do we want to incorporate a flag to do an ivy publish instead of maven in this PR as per @tfenne 's comment?

build.gradle
```diff
@@ -0,0 +1,326 @@
...
+
+jacoco {
+    toolVersion = "0.7.5.201505241946"
+}
+
+dependencies {
+    compile 'com.google.guava:guava:15.0'
+    compile ('com.github.samtools:htsjdk:2.4.1')
+    //tools dependency for doclet requires sdk devel
+    compile(files(((URLClassLoader) ToolProvider.getSystemToolClassLoader()).getURLs()))
```
lbergelson: The doclet is only needed for building documentation at compile time?

lbergelson:
@jacarey I think it would be great to have ivy publishing support if that's a feature that people need.  I'm not certain how best to incorporate it.  We're currently using the old style maven plugin to do publishing and it is supposed to have some Ivy support.  I'm not certain how to use it though. 

```gradle
task installIvy(type:Upload) {
    configuration = configurations.archives
    uploadDescriptor = true
    repositories {
        ivy {
            url = "file:///Users/louisb/.ivy2/local"
        }

    }
}
```
puts a jar in the designated folder, but I don't know if ivy also requires other things analogous to the maven pom.xml

@tfenne Since you have ivy experience could you test a task like that and see if it does what you need?  

If it doesn't, we can switch to using the newer ivy-publish and maven-publish plugins, but that will require some changes to the existing maven publishing task.

droazen:
Recommend that we add ivy publish support in a future PR, in the interests of getting this branch in before it gets stale (since it touches so many files).

build.gradle
jacarey: I believe that is true. We only use it to generate the metrics documentation.

tfenne:
Dealing with ivy support in a subsequent round makes total sense to me.  @lbergelson I tried out your suggested snippet (with my own directory obviously) and while it did dump some stuff into a subdir of `.ivy2/local` I think it was insufficient and possibly also in the wrong structure.  A subsequent build that tried to reference it could not find it.

What I got was some jars directly in `com.github.broadinstitute/picard/2.4.1-24-g6147dc5-SNAPSHOT`.  When I publish from other projects I get a similar `org/product/version` structure, but underneath that are subdirs for `docs`, `ivys`, `jars`, `poms` and `srcs` which contain the actual JARs.

While I'm here - can I ask what the motivation is in including a hash in the snapshot name?  It makes it much more difficult to iterate through the `edit->publish->test in downstream project` cycle, as the version name changes every time you touch the source.

lbergelson:
@tfenne Ah, I suspected it might not be structured right.  I figured if it was that easy we could include it straight way, but it looks like we'll have to get to it in a later request.  

The reasoning for including the hash in the snapshots is that we treat them like numbered releases instead of mutable snapshots.  A new snapshot is published on every commit to master, and easily on demand.  That way it's easy to pick which exact version development branches can build off of in other repositories without having to have a proper release.  It also makes it so that it's safe for developers to publish their own snapshots without risk of collision.  I see how it makes it annoying if you actually want to treat them as development snapshots and rapidly iterate though them though.

Would exposing a `-Dversion` parameter in the build that lets you specify exactly what name you want be helpful?  That way the defaults could be the safely non-conflicting snapshots, but if you wanted to actually use them the way maven intended you could give them a specific set name.

yfarjoun:
can we create a latest.jar or a link to the latest?

lbergelson:
@yfarjoun Yeah, we could do that.  Gatk-protected makes a symlink to the latest jar as part of their build script to help make their Queue scripts simpler.  Or just make a copy of it called latest.jar.  No objections from me if someone would find that useful.

droazen:
@tfenne We will definitely provide a way to depend on the "latest locally-installed htsjdk snapshot" when doing local development without having to manually update the hash in the version every time, which would be incredibly annoying. That is something that snapshot versioning is specifically designed to enable:

```
Sometimes the module you request can change over time, even for the same version. An example of this
 type of changing module is a Maven SNAPSHOT module, which always points at the latest artifact published. In other
 words, a standard Maven snapshot is a module that never stands still so to speak, it is a “changing module”.
```

jacarey:
Upgraded to htsjdk 2.5.0

droazen:
@lbergelson Does @jacarey have the green light to merge this when ready?

lbergelson:
👍

jacarey:
I will be merging these changes as soon as I have the update for internal release scripts done and reviewed so they can also be merged directly afterwards. This will likely happen later this afternoon or early tomorrow (6/28)

PR #556, opened 2016-06-06
new algorithm for CollectWgsMetrics

ZLyanov:
This is a draft for the new CollectWGSMetrics algorithm. It will compile and work correctly only with HTSJDK branch zl_refactor_SamLocusIterator_for_wgs

In the attached pdf please find full description of proposed changes:

[CollectWgsMetricOptimization.pdf](https://github.com/broadinstitute/picard/files/301321/CollectWgsMetricOptimization.pdf)

The intension of this pull request is to provide access to the sources to help review pull request https://github.com/samtools/htsjdk/pull/637


yfarjoun:
this needs to be rebased I'm guessing?

ZLyanov:
@yfarjoun Yes, it is. We are working with it. Unfortunately, too many things changed in Picard, so we have to make some code corrections and tests before. Thanks!

yfarjoun:
Great to hear! sorry about the changes...hopefully it's all for the better!

Y

ZLyanov:
@yfarjoun rebase is done.

src/main/java/picard/analysis/CollectWgsMetricsWithNonZeroCoverage.java
```diff
@@ -240,4 +241,4 @@ public boolean areHistogramsEmpty() {
         }
     }
```
kbergin: I know this was already here, but extra new line

yfarjoun:
I'm working my way through this PR...FYI.

src/main/java/picard/analysis/WgsMetricsProcessorImpl.java
```diff
@@ -0,0 +1,126 @@
...
+import htsjdk.samtools.util.*;
+import picard.PicardException;
+import picard.filter.CountingFilter;
+import picard.filter.CountingPairedFilter;
+
+import java.util.Arrays;
+import java.util.stream.LongStream;
+
+/**
+ * Implementation of <code>WgsMetricsProcessor</code> that gets input data from a given iterator
```
yfarjoun: please use {@link} notation  instead of <code> </code>. Here and throughout.

src/main/java/picard/analysis/WgsMetricsProcessorImpl.java
```diff
@@ -0,0 +1,126 @@
...
+            }
+
+            progress.record(info.getSequenceName(), info.getPosition());
+            if (collector.isTimeToStop(++counter)) {
+                break;
+            }
+            collector.setCounter(counter);
+        }
+        // check that we added the same number of bases to the raw coverage histogram and the base quality histograms
+        final long sumBaseQ= Arrays.stream(collector.unfilteredBaseQHistogramArray).sum();
```
yfarjoun: space on both sides of `=`

yfarjoun: If I'm not mistaken, this code was excised from the main branch since it was slowing things down significantly. please do the same here.

src/test/java/picard/analysis/WgsMetricsProcessorImplTest.java
```diff
@@ -0,0 +1,83 @@
...
+        AbstractLocusIterator iterator = createReadEndsIterator(exampleSamOneRead);
+        CollectWgsMetrics collectWgsMetrics = new CollectWgsMetrics();
+        collectWgsMetrics.STOP_AFTER = 16;
+        AbstractWgsMetricsCollector collector =  new FastWgsMetricsCollector(collectWgsMetrics, 100, null);
+        WgsMetricsProcessorImpl wgsMetricsProcessor =
+                new WgsMetricsProcessorImpl(iterator, getReferenceSequenceFileWalker(), collector, progress);
+        wgsMetricsProcessor.processFile();
+        assertEquals(15, collector.counter);
+    }
+
```
yfarjoun: extra newline.

yfarjoun:
We would love to have this PR taken the last few steps and merged....it would be a real loss to miss out on the work we did in htsjdk....

yfarjoun:
I guess I wastn' clear that I'm handing this back to you @ZLyanov to respond to my review...

ZLyanov:
@yfarjoun Sorry for delay, we will try to speed up the work. We will come back soon.

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -118,6 +120,12 @@
     @Option(doc="Sample Size used for Theoretical Het Sensitivity sampling. Default is 10000.", optional = true)
     public int SAMPLE_SIZE=10000;
 
+    @Option(doc = "If true, fast algorithm is used.")
+    public boolean USE_FAST_ALGORITHM = false;
+
+    @Option(doc = "Average read length in the file. Default is 150.", optional = true)
```
mzueva: This interface was used during earlier stages of development. Code structure was changed and we don't need it anymore. We've removed it. Thank you for pointing it out.
If READ_LENGTH is much smaller than actual read length program will crash with ArrayIndexOutOfBoundException trying to increment value corresponding reference sequence index in Counter. We`ve added this possible cause of crashing into exception message.

src/test/java/picard/analysis/CollectWgsMetricsTest.java
```diff
@@ -225,8 +235,8 @@ public void testLargeIntervals() throws IOException {
         }
     }
 
-    @Test
-    public void testExclusions() throws IOException {
+    @Test(dataProvider = "wgsAlgorithm")
+    public void testExclusions(final String useFastAlgorithm) throws IOException {
```
mzueva: Fixed test. Now it use provided argument. Thank you for pointing it out

src/test/java/picard/analysis/CollectWgsMetricsTest.java
```diff
@@ -289,8 +299,8 @@ public void testExclusions() throws IOException {
         Assert.assertEquals((long) highQualityDepthHistogram.get(3).getValue(), 2*10);
     }
 
-    @Test
-    public void testPoorQualityBases() throws IOException {
+    @Test(dataProvider = "wgsAlgorithm")
+    public void testPoorQualityBases(final String useFastAlgorithm) throws IOException {
```
mzueva: Fixed test. Now it use provided argument. Thank you for pointing it out

src/test/java/picard/analysis/CollectWgsMetricsTestUtils.java
```diff
@@ -96,6 +118,97 @@ private static void createTestSAM(String testSamName) throws IOException {
...
+        ReferenceSequenceFile referenceSequenceFile = createReferenceSequenceFile(referenceString);
+        return new ReferenceSequenceFileWalker(referenceSequenceFile);
+    }
+
+    static ReferenceSequenceFileWalker getReferenceSequenceFileWalker(){
+        String referenceString = ">ref\nACCTACGTTCAATATTCTTC";
+        return getReferenceSequenceFileWalker(referenceString);
+    }
+
+    static ReferenceSequenceFile createReferenceSequenceFile(String referenceString) {
```
mzueva: InMemoryReferenceSequence in htsjdk semmes to have a package-private access level and as we can see is used only for testing. Should we rewrite it in picard tools?

src/main/java/picard/analysis/CounterManager.java
```diff
@@ -0,0 +1,189 @@
...
+     * Method added for use in unit tests
+     *
+     * @param offset from the reference sequence, that will be represented by index 0 of <code>Counter</code> arrays.
+     */
+    void setOffset(int offset) {
+        this.offset = offset;
+    }
+
+    /**
+     * Method checks that new locus position is not out of bounds of Counter arrays and there is enough
```
mzueva: We`ve included such test suite.

src/main/java/picard/analysis/CounterManager.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+    }
+
+    /**
+     * Rebases inner Counter arrays so that 0 index of array represents the new locus position
+     *
+     * @param locusPosition position in the reference sequence,
+     *                      that will be represented by index 0 of <code>Counter</code> arrays.
+     */
+    private void rebase(int locusPosition) {
```
mzueva: Assertion added

src/main/java/picard/analysis/CounterManager.java
```diff
@@ -0,0 +1,189 @@
...
+        for (Counter counter : arrays) {
+            int[] array = counter.array;
```
mzueva: done

src/main/java/picard/analysis/CounterManager.java
```diff
@@ -0,0 +1,189 @@
...
+        }
+        offset = locusPosition;
+    }
+
+    /**
+     * Clears all inner Counter arrays
+     */
+    public void clear() {
+        for (Counter counter : arrays) {
+            int[] array = counter.array;
```
mzueva: done

src/main/java/picard/analysis/CounterManager.java
```diff
@@ -0,0 +1,189 @@
...
+        offset = 0;
+    }
+
+    /**
+     * Creates a new Counter object and adds it to the list of managed Counters.
+     *
+     * @return <code>Counter</code>, that will be managed by current <code>CounterManager</code>
+     */
+    public Counter newCounter() {
+        Counter counter = new Counter(arrayLength);
```
mzueva: done

src/main/java/picard/analysis/CounterManager.java
```diff
@@ -0,0 +1,189 @@
...
+        arrays.add(counter);
+        return counter;
+    }
+
+    /**
+     * Class represents an integer array with methods to increment and get the values form it with respect
```
mzueva: done

src/main/java/picard/analysis/CounterManager.java
```diff
@@ -0,0 +1,189 @@
...
+     * to offset of outer <code>CounterManager</code>.
+     */
+    public class Counter {
+
+        /**
+         * Inner array to store accumulated values
+         */
+        private int[] array;
```
mzueva: done

src/main/java/picard/analysis/CounterManager.java
```diff
@@ -0,0 +1,189 @@
...
+        private Counter(int arrayLength) {
+            array = new int[arrayLength];
+        }
+
+        /**
+         * Increments value corresponding to reference sequence index
+         *
+         * @param index in the reference sequence
+         */
+        public void inc(int index) {
```
mzueva: done

src/main/java/picard/analysis/CounterManager.java
```diff
@@ -0,0 +1,189 @@
...
+         * @param index in the reference sequence
+         * @return value from the inner array, corresponding to input reference index
+         */
+        public int get(int index) {
+            checkIndex(index);
+            return array[index - offset];
+        }
+
+        private void checkIndex(int index) {
+            if ((index - offset) < 0 || (index - offset) >= array.length) {
```
mzueva: Yes, array access itself will throw identical exception but with not meaningful index. The real index used in array is [index – offset]. So, for getting more clear exception message we`ve perform this test here.

src/main/java/picard/analysis/WgsMetricsProcessorImpl.java
mzueva: This code in the master branch was moved out of the cycle, so that check is performed only once after the whole metric's run. We've made the same changes here.

src/main/java/picard/analysis/CounterManager.java
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * Rebases inner Counter arrays so that 0 index of array represents the new locus position
+     *
+     * @param locusPosition position in the reference sequence,
+     *                      that will be represented by index 0 of <code>Counter</code> arrays.
+     */
+    private void rebase(int locusPosition) {
+        for (Counter counter : arrays) {
+            int[] array = counter.array;
+            System.arraycopy(array, locusPosition - offset, array, 0, arrayLength - locusPosition + offset);
```
mzueva: done

src/test/java/picard/analysis/CollectWgsMetricsTestUtils.java
yfarjoun: hmmm. no. Perhaps it should be made public exactly for cases like this....out of scope. here, ofcourse.

src/test/java/picard/analysis/FastWgsMetricsCollectorTest.java
```diff
@@ -0,0 +1,215 @@
...
+    private ReferenceSequence ref;
+
+    @BeforeTest
+    public void setUp(){
+        String referenceString = ">chrM\nACCTACGTTCAATATTCTTCACCTACGTTCAATATTCTTCACCTACGTTCAATATTCTTCACCTACGTTCAATATTCTTCACCTACGTTCAATATTCTTC";
+        ref = new ReferenceSequence("chrM", 0, referenceString.getBytes());
+        sequence = new SAMSequenceRecord("chrM", 100);
+        record = new SAMRecord(new SAMFileHeader());
+        record.setReadName("test");
+        record.setBaseQualities(qualities);
```
yfarjoun: how come the length of baseQualities is shorter than the number of base?

yfarjoun:
Hi @zaal. Thanks for this. I looked at the outputs and I get some small differences between the old and new versions. They are not large, but they exists. I would like your help understanding where they are coming from.

I ran with the argument INCLUDE_BQ_HISTOGRAM=true, and in the base-quality histogram there's a difference (here's the result of running `diff` on the resulting metrics files, same jar, left file is USE_FAST_ALGORITHM=false and on the right it's true) 

[I added stars around the actual differences]
```
< 3	4258103	6975**1994**
< 4	3583608	14383**2006**
< 5	3177722	1802**48314**
< 6	2937996	5682**78048**
---
> 3	4258103	6975**4344**
> 4	3583608	14383**3597**
> 5	3177722	1802**50094**
> 6	2937996	5682**97247**
22c22
< 10	2778204	20090**53947**
---
> 10	2778204	20090**91942**
32c32
< 20	8652577	7192**277944**
---
> 20	8652577	7192**079839**
42c42
< 30	63535055	108490**394578**
---
> 30	63535055	108490**533842**
52c52
< 40	157384770	36682**4596**
---
> 40	157384770	36682**0522**
```

The differences are on the order of 1e-5...so a small test will not show this.

Initially I suspected the LOCUS_ACCUMULATION_CAP, but I didn't see the warning that the cap has been reached, and also the only difference is in the base-quality histogram and derived field (Het Sensitivity).

Any ideas?

Thanks!

yfarjoun:
@ZLyanov and @mzueva I have a small question in the PR about WGS metrics. thanks!

src/main/java/picard/analysis/FastWgsMetricsCollector.java
```diff
@@ -0,0 +1,195 @@
...
+                if (unfilteredDepthSize.get(index) < coverageCap) {
+                    unfilteredBaseQHistogramArray[quality]++;
+                    unfilteredDepthSize.increment(index);
+                }
+                if (quality < collectWgsMetrics.MINIMUM_BASE_QUALITY || SequenceUtil.isNoCall(bases[i + record.getOffset()])){
+                    basesExcludedByBaseq++;
+                } else {
+                    final int bsq = excludeByQuality(recordsAndOffsetsForName, index);
+                    if (recordsAndOffsetsForName.size() - bsq > 0) {
+                        basesExcludedByOverlap++;
```
yfarjoun: I"m confused about this. you call `excludeByQuality` and then increment `basesExcludedByOverlap`?

src/main/java/picard/analysis/FastWgsMetricsCollector.java
```diff
@@ -0,0 +1,195 @@
...
+            counterManager.clear();
+            previousSequenceIndex = info.getSequenceIndex();
+        }
+        counterManager.checkOutOfBounds(info.getPosition());
+    }
+
+    private int excludeByQuality(final Set<EdgingRecordAndOffset> setForName, int position) {
+        int bsq = 0;
+        for (EdgingRecordAndOffset recordAndOffset : setForName) {
+            if (position - recordAndOffset.getRefPos() >= recordAndOffset.getLength()
```
yfarjoun: first part of if statement isn't clear here. can you add a comment explaining it?

src/main/java/picard/analysis/FastWgsMetricsCollector.java
```diff
@@ -0,0 +1,195 @@
...
+                basesExcludedByBaseq++;
+            } else {
+                if (unfilteredDepthSize.get(index) < coverageCap) {
+                    unfilteredBaseQHistogramArray[quality]++;
+                    unfilteredDepthSize.increment(index);
+                }
+                if (quality < collectWgsMetrics.MINIMUM_BASE_QUALITY || SequenceUtil.isNoCall(bases[i + record.getOffset()])){
+                    basesExcludedByBaseq++;
+                } else {
+                    final int bsq = excludeByQuality(recordsAndOffsetsForName, index);
```
yfarjoun: perhaps a better variable name for bsq could help here.

mzueva:
@yfarjoun 
Please, take a look at this explanation for the result's difference, that we've prepared:
The depth histogram is exactly the same, the quality and some statistics values slightly differ because of different order of reads processing. Consider the following alignment:
 	1	2	3	4	5	6	7	8	9	10	11	12	13	14	15
reference	A	C	G	T	T	T	T	T	T	T	T	C	C	C	C
read1	A	C	G	T	-	-	-	-	-	T	T	C	C	C	C
read2	 	 	G	T	T	T	T	T	T	T	T	C	C	C	C
In current implementation for positions 10 - 15 read1 will be processed at first, for new implementation - read2 will be the first, so if we exceed the coverage cap for this region the qualities of the read2 will be added to the base quality histogram and read1 qualities will be omitted. 
Also results can be affected by the LOCUS_ACCUMULATION_CAP, used in SamLocusIterator, since new ReadEndsIterator doesn’t use this cap.
Bu these changes don't affect the overall result of metrics.

yfarjoun:
I see. Presumably this is because (in your example) there are 6 readEnds:

1. Read1-pos1-start
2. Read1-pos4-end
3. Read1-pos10-start
4. Read1-pos15-end
5. Read2-pos3-start
6. Read2-pos15-end

Then, at position 10 the list of read-starts is [Read2-pos3-start (left from before), Read1-pos10-start (just added)] which is why the order is not according to the order of the reads, and the first `COVERAGE_CAP` bases will be used for the metric. 

I don't think that it is a big issue, but it is somewhat confusing...as there will be a subsetting of alignment blocks rather an a subsetting of reads, am I right?

yfarjoun:
I'm not clear about your comment regarding "these changes don't affect the overall result of metrics." They will change some metrics if the relevant reads are differentially excluded due to the coverage cap. for example, the overlap exclusion could change (by a little bit) depending on which reads are excluded. Same for the baseQ exclusion. I agree that the differences should be small if the coverage cap is large enough.

yfarjoun:
Thanks for the explanation. It's too bad that the results to not match up exactly, but I'm not concerned with the effects of randomly choosing different reads (or parts of reads) to throw away when we hit the coverage cap.

Thanks for a very long process. When you next pass by, I'd like to take you out for some 🍻 

👍

yfarjoun:
bitmoji

ZLyanov:
Hurray! Many thanks to everybody involved in the process!  Special regards to Mariia and Yossi!

PR #559, opened 2016-06-10
Fix revert sam for single-end reads

ktibbett:
RevertSam was 'undoing' mate flags on single-end reads.  This resulted in invalid records that had the isMateUnmapped flag incorrectly set to true.  This change makes sure we un-set the mate info only if the read is paired.


nh13:
I would like to review as I disagree but I am on mobile and want to explain why clearly.

ktibbett:
@nh13 Will wait until I hear from you... no rush.

ktibbett:
Context: setting setMateUnmappedFlag to true on an unpaired read causes ValidateSamFile to fail when stringency = strict.

nh13:
@ktibbett that's what I wanted to say, which was that we should have `setMateUnmappedFlag(false)` for single end reads.  I have seen in the wild single-end reads that have mate information that I want `RevertSam` to clean up, so I would like to retain all the other `setMate*` calls for single-end reads too.

 Therefore, I would have:
```
rec.setMateAlignmentStart(SAMRecord.NO_ALIGNMENT_START);
rec.setMateNegativeStrandFlag(false);
rec.setMateReferenceIndex(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX);
rec.setMateUnmappedFlag(rec.getReadPairedFlag());
  ```

This probably needs a test in `RevertSamTest`, where we give it single-end reads with mate information and make sure it does the right thing (removes all mate information).

ktibbett:
@nh13 , good points, all.  Thanks!  Should have another PR for review by Monday.

ktibbett:
@nh13 , care to look over the changes?

nh13:
👍

PR #560, opened 2016-06-12
Tag reads in duplicate sets with the representative read and the duplicate set size

lhogstrom:
Optional flag added to MarkDuplicates. If true, each record that is part of a duplicate set is tagged with the representative read that is chosen by MarkDuplicates. This is accomplished with the ReadEndsForMarkDuplicatesTagRepresentativeRead class that retains read name information for each read pair examined. The RepresentativeReadName class stores the representative read name and set size for each duplicate set, spilling to disk if necessary. This functionality is tested by MarkDuplicatesTagRepresentativeReadTest that also executes the tests in AbstractMarkDuplicatesCommandLineProgramTest.

PR #561, opened 2016-06-12
Revert "Tag reads in duplicate sets with the representative read and the duplicate set size"

lhogstrom:
Reverts broadinstitute/picard#560


yfarjoun:
I checked and this indeed reverts all the changes, and the current master is identical to what it was before your PR. 👍

PR #562, opened 2016-06-13
Tag reads in duplicate sets with the representative read and the duplicate set size (v2)

lhogstrom:
This commit re-applies the changes to enable tagging of representative reads. The summary of this behavior is as follows: 

Optional flag added to MarkDuplicates. If true, each record that is part of a duplicate set is tagged with the representative read that is chosen by MarkDuplicates. This is accomplished with the ReadEndsForMarkDuplicatesTagRepresentativeRead class that retains read name information for each read pair examined. The RepresentativeReadName class stores the representative read name and set size for each duplicate set, spilling to disk if necessary. This functionality is tested by MarkDuplicatesTagRepresentativeReadTest that also executes the tests in AbstractMarkDuplicatesCommandLineProgramTest.

This reverts commit 934b217c96ba553910b90a8d3721f35ae7585364.

### DIFF
--- a/src/main/java/picard/sam/markduplicates/MarkDuplicates.java
+++ b/src/main/java/picard/sam/markduplicates/MarkDuplicates.java
@@ -37,16 +37,9 @@ import htsjdk.samtools.*;
 import htsjdk.samtools.util.CloseableIterator;
 import htsjdk.samtools.util.SortingCollection;
 import htsjdk.samtools.util.SortingLongCollection;
-import picard.sam.markduplicates.util.AbstractMarkDuplicatesCommandLineProgram;
-import picard.sam.markduplicates.util.DiskBasedReadEndsForMarkDuplicatesMap;
-import picard.sam.markduplicates.util.LibraryIdGenerator;
-import picard.sam.markduplicates.util.ReadEnds;
-import picard.sam.markduplicates.util.ReadEndsForMarkDuplicates;
-import picard.sam.markduplicates.util.ReadEndsForMarkDuplicatesCodec;
-import picard.sam.markduplicates.util.ReadEndsForMarkDuplicatesMap;
 import htsjdk.samtools.DuplicateScoringStrategy.ScoringStrategy;
-import picard.sam.markduplicates.util.ReadEndsForMarkDuplicatesWithBarcodes;
-import picard.sam.markduplicates.util.ReadEndsForMarkDuplicatesWithBarcodesCodec;
+import picard.sam.markduplicates.util.*;
+import picard.sam.util.RepresentativeReadIndexer;
 
 import java.io.*;
 import java.util.*;
@@ -125,6 +118,10 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
     public static final String DUPLICATE_TYPE_LIBRARY = "LB";
     /** The duplicate type tag value for duplicate type: sequencing (optical & pad-hopping, or "co-localized"). */
     public static final String DUPLICATE_TYPE_SEQUENCING = "SQ";
+    /** The attribute in the SAM/BAM file used to store which read was selected as representative out of a duplicate set */
+    public static final String DUPLICATE_SET_INDEX_TAG = "DI";
+    /** The attribute in the SAM/BAM file used to store the size of a duplicate set */
+    public static final String DUPLICATE_SET_SIZE_TAG = "DS";
 
     /** Enum for the possible values that a duplicate read can be tagged with in the DT attribute. */
     public enum DuplicateType {
@@ -165,6 +162,14 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
     @Option(doc = "Read two barcode SAM tag (ex. BX for 10X Genomics)", optional = true)
     public String READ_TWO_BARCODE_TAG = null;
 
+    @Option(doc = "If a read appears in a duplicate set, add two tags. The first tag, DUPLICATE_SET_SIZE_TAG (DS), " +
+            "indicates the size of the duplicate set. The smallest possible DS value is 2 which occurs when two " +
+            "reads map to the same portion of the reference only one of which is marked as duplicate. The second " +
+            "tag, DUPLICATE_SET_INDEX_TAG (DI), represents a unique identifier for the duplicate set to which the " +
+            "record belongs. This identifier is the index-in-file of the representative read that was selected out " +
+            "of the duplicate set.", optional = true)
+    public boolean TAG_DUPLICATE_SET_MEMBERS = false;
+
     @Option(doc = "If true remove 'optical' duplicates and other duplicates that appear to have arisen from the " +
             "sequencing process instead of the library preparation process, even if REMOVE_DUPLICATES is false. " +
             "If REMOVE_DUPLICATES is true, all duplicates are removed and this option is ignored.")
@@ -177,6 +182,7 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
     private SortingCollection<ReadEndsForMarkDuplicates> fragSort;
     private SortingLongCollection duplicateIndexes;
     private SortingLongCollection opticalDuplicateIndexes;
+    private SortingCollection<RepresentativeReadIndexer> representativeReadIndicesForDuplicates;
 
     private int numDuplicateIndices = 0;
     static private final long NO_SUCH_INDEX = Long.MAX_VALUE; // needs to be large so that that >= test fails for query-sorted traversal
@@ -258,6 +264,22 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
         long recordInFileIndex = 0;
         long nextOpticalDuplicateIndex = this.opticalDuplicateIndexes != null && this.opticalDuplicateIndexes.hasNext() ? this.opticalDuplicateIndexes.next() : NO_SUCH_INDEX;
         long nextDuplicateIndex = (this.duplicateIndexes.hasNext() ? this.duplicateIndexes.next() : NO_SUCH_INDEX);
+
+        // initialize variables for optional representative read tagging
+        CloseableIterator<RepresentativeReadIndexer> representativeReadIterator = null;
+        RepresentativeReadIndexer rri = null;
+        int representativeReadIndexInFile = -1;
+        int duplicateSetSize = -1;
+        int nextRepresentativeIndex = -1;
+        if (TAG_DUPLICATE_SET_MEMBERS) {
+            representativeReadIterator = this.representativeReadIndicesForDuplicates.iterator();
+            if (representativeReadIterator.hasNext()) {
+                rri = representativeReadIterator.next();
+                nextRepresentativeIndex = rri.readIndexInFile;
+                representativeReadIndexInFile = rri.representativeReadIndexInFile;
+                duplicateSetSize = rri.setSize;
+            }
+        }
 
         final ProgressLogger progress = new ProgressLogger(log, (int) 1e7, "Written");
         final CloseableIterator<SAMRecord> iterator = headerAndIterator.iterator;
@@ -343,6 +365,28 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
                 }
             }
 
+            // Tag any read pair that was in a duplicate set with the duplicate set size and a representative read name
+            if (TAG_DUPLICATE_SET_MEMBERS) {
+                final boolean needNextRepresentativeIndex = recordInFileIndex > nextRepresentativeIndex;
+                if (needNextRepresentativeIndex && representativeReadIterator.hasNext()) {
+                    rri = representativeReadIterator.next();
+                    nextRepresentativeIndex = rri.readIndexInFile;
+                    representativeReadIndexInFile = rri.representativeReadIndexInFile;
+                    duplicateSetSize = rri.setSize;
+                }
+                final boolean isInDuplicateSet = recordInFileIndex == nextRepresentativeIndex ||
+                        (sortOrder == SAMFileHeader.SortOrder.queryname &&
+                                recordInFileIndex > nextDuplicateIndex);
+                if (isInDuplicateSet) {
+                    if (!rec.isSecondaryOrSupplementary() && !rec.getReadUnmappedFlag()) {
+                        if (TAG_DUPLICATE_SET_MEMBERS) {
+                            rec.setAttribute(DUPLICATE_SET_INDEX_TAG, representativeReadIndexInFile);
+                            rec.setAttribute(DUPLICATE_SET_SIZE_TAG, duplicateSetSize);
+                        }
+                    }
+                }
+            }
+
             // Output the record if desired and bump the record index
             recordInFileIndex++;
             if (this.REMOVE_DUPLICATES            && rec.getDuplicateReadFlag()) continue;
@@ -357,6 +401,9 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
         iterator.close();
 
         this.duplicateIndexes.cleanup();
+        if (TAG_DUPLICATE_SET_MEMBERS) {
+            this.representativeReadIndicesForDuplicates.cleanup();
+        }
 
         reportMemoryStats("Before output close");
         out.close();
@@ -508,8 +555,8 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
                             if (pairedEnds.read2ReferenceIndex == pairedEnds.read1ReferenceIndex &&
                                     pairedEnds.read2Coordinate == pairedEnds.read1Coordinate &&
                                     pairedEnds.orientation == ReadEnds.RF) {
-                                pairedEnds.orientation = ReadEnds.FR;
-                            }
+                                        pairedEnds.orientation = ReadEnds.FR;
+                                    }
                         } else {
                             pairedEnds.read2ReferenceIndex = pairedEnds.read1ReferenceIndex;
                             pairedEnds.read2Coordinate = pairedEnds.read1Coordinate;
@@ -600,18 +647,35 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
      * @return an array with an ordered list of indexes into the source file
      */
     private void generateDuplicateIndexes(final boolean useBarcodes, final boolean indexOpticalDuplicates) {
+        int entryOverhead;
+        if (TAG_DUPLICATE_SET_MEMBERS) {
+            // Memory requirements for RepresentativeReadIndexer:
+            // three int entries + overhead: (3 * 4) + 4 = 16 bytes 
+            entryOverhead = 16;
+        }
+        else {
+            entryOverhead = SortingLongCollection.SIZEOF;
+        }
         // Keep this number from getting too large even if there is a huge heap.
-        int maxInMemory = (int) Math.min((Runtime.getRuntime().maxMemory() * 0.25) / SortingLongCollection.SIZEOF, (double) (Integer.MAX_VALUE - 5));
-        // If we're also tracking optical duplicates, cut maxInMemory in half, since we'll need two sorting collections
+        int maxInMemory = (int) Math.min((Runtime.getRuntime().maxMemory() * 0.25) / entryOverhead, (double) (Integer.MAX_VALUE - 5));
+        // If we're also tracking optical duplicates, reduce maxInMemory, since we'll need two sorting collections
         if (indexOpticalDuplicates) {
-            maxInMemory /= 2;
+            maxInMemory /= ((entryOverhead + SortingLongCollection.SIZEOF) / entryOverhead);
             this.opticalDuplicateIndexes = new SortingLongCollection(maxInMemory, TMP_DIR.toArray(new File[TMP_DIR.size()]));
         }
         log.info("Will retain up to " + maxInMemory + " duplicate indices before spilling to disk.");
         this.duplicateIndexes = new SortingLongCollection(maxInMemory, TMP_DIR.toArray(new File[TMP_DIR.size()]));
+        if (TAG_DUPLICATE_SET_MEMBERS) {
+            final RepresentativeReadIndexerCodec representativeIndexCodec = new RepresentativeReadIndexerCodec();
+            this.representativeReadIndicesForDuplicates = SortingCollection.newInstance(RepresentativeReadIndexer.class,
+                    representativeIndexCodec,
+                    new RepresentativeReadComparator(),
+                    maxInMemory,
+                    TMP_DIR);
+        }
 
         ReadEndsForMarkDuplicates firstOfNextChunk = null;
-        final List<ReadEndsForMarkDuplicates> nextChunk = new ArrayList<ReadEndsForMarkDuplicates>(200);
+        final List nextChunk = new ArrayList<ReadEndsForMarkDuplicates>(200);
 
         // First just do the pairs
         log.info("Traversing read pair information and detecting duplicates.");
@@ -621,13 +685,17 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
             } else {
                 if (nextChunk.size() > 1) {
                     markDuplicatePairs(nextChunk);
+                    if (TAG_DUPLICATE_SET_MEMBERS) addRepresentativeReadIndex(nextChunk);
                 }
                 nextChunk.clear();
                 nextChunk.add(next);
                 firstOfNextChunk = next;
             }
         }
-        if (nextChunk.size() > 1) markDuplicatePairs(nextChunk);
+        if (nextChunk.size() > 1) {
+            markDuplicatePairs(nextChunk);
+            if (TAG_DUPLICATE_SET_MEMBERS) addRepresentativeReadIndex(nextChunk);
+        }
         this.pairSort.cleanup();
         this.pairSort = null;
 
@@ -661,6 +729,7 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
         log.info("Sorting list of duplicate records.");
         this.duplicateIndexes.doneAddingStartIteration();
         if (this.opticalDuplicateIndexes != null) this.opticalDuplicateIndexes.doneAddingStartIteration();
+        if (TAG_DUPLICATE_SET_MEMBERS) this.representativeReadIndicesForDuplicates.doneAdding();
     }
 
     private boolean areComparableForDuplicates(final ReadEndsForMarkDuplicates lhs, final ReadEndsForMarkDuplicates rhs, final boolean compareRead2, final boolean useBarcodes) {
@@ -692,6 +761,42 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
         this.duplicateIndexes.add(bamIndex);
         ++this.numDuplicateIndices;
     }
+
+    private void addRepresentativeReadOfDuplicateSet(final long representativeReadIndexInFile, final int setSize, final long read1IndexInFile) {
+        final RepresentativeReadIndexer rri = new RepresentativeReadIndexer();
+        rri.representativeReadIndexInFile = (int) representativeReadIndexInFile;
+        rri.setSize = setSize;
+        rri.readIndexInFile = (int) read1IndexInFile;
+        this.representativeReadIndicesForDuplicates.add(rri);
+    }
+
+    /**
+     * Takes a list of ReadEndsForMarkDuplicates objects and identify the representative read based on
+     * quality score. For all members of the duplicate set, add the read1 index-in-file of the representative
+     * read to the records of the first and second in a pair. This value becomes is used for
+     * the 'DI' tag.
+     *
+     * @param list
+     */
+    private void addRepresentativeReadIndex(final List<ReadEndsForMarkDuplicates> list) {
+        short maxScore = 0;
+        ReadEndsForMarkDuplicates best = null;
+
+        /** All read ends should have orientation FF, FR, RF, or RR **/
+        for (final ReadEndsForMarkDuplicates end : list) {
+            if (end.score > maxScore || best == null) {
+                maxScore = end.score;
+                best = end;
+            }
+        }
+
+        // for read name (for representative read name), add the last of the pair that was examined
+        for (final ReadEndsForMarkDuplicates end : list) {
+            addRepresentativeReadOfDuplicateSet(best.read1IndexInFile, list.size(), end.read1IndexInFile);
+            addRepresentativeReadOfDuplicateSet(best.read1IndexInFile, list.size(), end.read2IndexInFile);
+        }
+    }
+
 
     /**
      * Takes a list of ReadEndsForMarkDuplicates objects and removes from it all objects that should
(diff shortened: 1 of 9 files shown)


nh13:
I would like to code review this if that would be acceptable.  But before I do that, I wanted to discuss a few other options for implementing of this feature.  I am concerned about the added memory of storing the read name given it is an arbitrary long string.

Off the top of my head, I can think of a few other ways:
1. Store the index into the original file of the representative (see `read1IndexInFile` or `read2IndexInFile`).  This could be done in a new `ReadEnds` structure or in a separate `SortingLongCollection`.  There would need to be a way to cover the read name (buffer reads?) if the representative is *after* the given record.
2. For each set of duplicates, give it a unique # (order?), and for each read, store the hash of the representative read in addition to the duplicate set #.  This would store two numbers, probably a `long` and an `int`.   Again, we would need a way to retrieve the representative's read name.
3. We could get really clever, but perhaps unnecessarily complicated, and store the full representative read name in the earliest alignment (coordinate order) for each duplicate set. The rest of the reads in that duplicate set would store the hash of the read name, and/or the duplicate set # (see (2).   This require only the first read (in the reference) to store the read name, limiting it to one read name per duplicate set stored.  We can also store the duplicate set size for each duplicate set (in the first record), to know how long we need to keep the duplicate set # (or read name hash) to read name mapping in memory. 

I think given all the effort in limiting memory, and `MarkDuplicates` being a bottleneck for many applications (ex. whole genome), we should be careful about adding either extra memory or extra computation to this tool.  I think having this information is extremely valuable, and so outputting the `RR` tag (not yet in the spec) will be super valuable and thus should be viewed as a production feature.  Thank-you for the effort!

yfarjoun:
We'd love to have your review and feedback. 

Regarding your initial comments, I was seeing RR as a debugging feature, not production because I wasn't sure of it's utility. it's good to hear that you consider it an important enough feature that you would likely run it by default. 

I really like option 3. and given that we already are calculating and storing the set-size, it doesn't seem difficult to remove the read-name from memory when needed.

lhogstrom:
Thanks Nils,

For your third suggestion, we would still need a sorting collection that stores the representative read name, correct? The benefit being that it would just contain just one read name per duplicate set instead of one read name per duplicate read (as I have it now). Yossi brought up that it might be

nh13:
For the third end, store the representative read name in the `ReadEnds` object of the read with the earliest genomic start position for each duplicate set.  You don't need a sorting collection.

lhogstrom:
Ok, also see my edited comment above storage size.

nh13:
You can absolutely store the read name in a tag per read with (3).  You only assign it to the tag when you output, not while you are processing/duplicate-marking.  For any duplicate set, the record with the smallest start position stores the read name and duplicate set # (and set size).  For other records in that duplicate set, they will come later, and so you can simply store the duplicate set # for those records.  Once all records for a given duplicate set have been seen (output), discard the read name.

lhogstrom:
"For the third end, store the representative read name in the ReadEnds object of the read with the earliest genomic start position for each duplicate set."

The problem is that you don't have any read name information during generateDuplicateIndexes() unless you retain the read names for all pairs in their ReadEnds object. I can't think of another way of getting and retaining the RR name at that stage without adding another iteration over the SAM file.

nh13:
I think you're right.  

There's probably a way to use a `SortingLongCollection` to store the indexes of the representatives and then do some buffering on output to fill in the `RR` tag to ensure that earlier records can retrieve the read name of the representative when it is later in the input file.

yfarjoun:
Thinking about this more....

Part of the reason that I was thinking of this as a non-production option is that it would almost double the BAM size as each (duplicate) read would now have a tag with a read-name in it.

But going on your ideas of hashes, how about this idea: for each non duplicate read (that has a duplicate set), store the read-name hash in the read itself (RH tag?) and then store the hashed readname of the representative read in each duplicates (under the RR tag?). This would mean that the ReadEnds object only grows by a single long (the hash) and there's no need for storing read names, or buffering. The only possible problem I forsee is hash-collisions, but there shouldn't be too many of them. For whole genome, 1e9 reads with a 64-bit hash gives probability of about (1e9)^2 / 1e19  = 1/10 collisions) so there will be perhaps 1 collision in a big WGS file.  In reality there will be less reads (unique reads with duplicates sets) than that and so the probability will be smaller.

lhogstrom:
If I am an end user and I want to know the representative read name for a duplicate set after running this tool, two possible alternatives could look like this:

1) Under Yossi's idea, I could take the hash assigned to my duplicate set and inspect the RH tag of records until I find a match
2) If all members of a duplicate set are labeled with a single unique #, then all but one of them has duplicate SAM flag. I could find the representative read name by inspecting the record that is in the set, but does not have the duplicate SAM flag turned on. 

Both scenarios seem like they would require the user to do a record search to find the RR name. The second scenario doesn't require labeling all records with an RH tag and you don't have to worry about collisions. Yossi, are there other benefits for saving the hash of a read name that I am missing?

lhogstrom:
I talked with Yossi today, we agreed that any unique identifier would be sufficient for labeling the members of duplicate set. This would allow users to find a representative read by finding the member of the set that does not have a duplicate SAM flag. 

Here is a proposal for changing the output for this feature:
-instead of RR tag have a DI tag (which stands for duplicate set index)
-The value of DI for all members of the duplicate set will be the index-in-file of the representative read (e.g. 'best.read1IndexInFile'). The value assigned to any duplicate set therefore will depend on the sorting order of the BAM. 
-All members of the duplicate set (including the representative read) will be tagged with a DI tag
-the name of the optional flag would change from TAG_REPRESENTATIVE_READ to TAG_DUPLICATE_SET_MEMBERS
-the DS tag (duplicate set size) will continue to be written at the sam time as the DI tag

Implementation details: 
-The ReadEndsForMarkDuplicates for mark duplicates object will be modified to hold an additional integer, the duplicate set size.  
-A sorting collection will continue to hold the duplicate set index, the duplicate set size, and read1Coordinate of each duplicate record. (Could possibly create a hashmap from duplicate set index to duplicate set size, though it is not clear how to write this to disk with Picard's existing serialization tools.)
-The DI and DS tags would be retrieve from the sorting collection and written to the output during the last record iteration. 

Thoughts on this revised plan? This extra overhead associated with this plan would be 4 bytes per read pair (held in the ReadEndsForMarkDuplicates objects) as well as 12 bytes per record that would be stored in a sorting collection. This would be more lean than handling and writing strings.

yfarjoun:
- if it's all the same, I think it would be good to tag the READ1 (set) and the READ2 (set) with a different value. The main reason for this is that it would enable en implementation of MDWMC that mimmicks this implementation without further overhead.

- then DS would be the number of occurances of a DI tag value in a file (rather than that over 2)

I like this plan! 👍

lhogstrom:
@nh13, what do you think of this revised spec? I'm hoping to crank it out in the next week.

nh13:
@lhogstrom @yfarjoun I would argue for READ1 and READ2 to have the _same_ value, since I care about duplicate molecules, not duplicate read ends.  Feel free to partition after.  In fact, @tfenne and I agree that a duplicate set iterator that returns a set with both ends of a pair is more useful for analysis in many cases.

lhogstrom:
This most recent commit tags both ends of a pair with the same duplicate set index. I agree with Nils and Tim that this makes sense given how we agreed to define members of a duplicate set above. 

Yossi and I have discussed cases where we may want to think of a 'duplicate set' that is not dependent on read pairs. This would be the case for fragments and single end read data that are not handled by this feature. @yfarjoun, in the interest of moving this tagging effort along, could the paired end vs. single end duplicate set definition be visited again at a later time? 

If so, @nh13 could you begin review of the changes? I'll look into the steps needed to change the SAM spec.

nh13:
@lhogstrom I am going to be a bit slow with my review as I have some other higher priority work.  I am committed to giving you a review though.

fleharty:
@nh13 Just a friendly ping.  @lhogstrom has left, so if everyone is okay with it, I'd like to take over this PR.

lhogstrom:
I'm still trying to keep my eye on this, but happy to have some extra help thanks Mark :)

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -230,6 +235,22 @@ protected int doWork() {
         long nextOpticalDuplicateIndex = this.opticalDuplicateIndexes != null && this.opticalDuplicateIndexes.hasNext() ? this.opticalDuplicateIndexes.next() : NO_SUCH_INDEX;
         long nextDuplicateIndex = (this.duplicateIndexes.hasNext() ? this.duplicateIndexes.next() : NO_SUCH_INDEX);
 
+        // initialize variables for optional representative read tagging
+        CloseableIterator<RepresentativeReadIndexer> representativeReadInterator = null;
+        RepresentativeReadIndexer rni = null;
+        int representativeReadIndexInFile = -1;
```
fleharty: It's not clear to me what the values -1 indicate here.

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -447,6 +493,8 @@ private void buildSortedReadEndLists(final boolean useBarcodes) {
                         // a copy since pairedEnds will be modified when the mate comes along.
                         pairedEnds = fragmentEnd.clone();
                         tmp.put(pairedEnds.read2ReferenceIndex, key, pairedEnds);
+                        /*pairedEnds = buildReadEnds(header, indexForRead, rec, useBarcodes);
```
fleharty: Comment should be removed.

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -480,7 +528,7 @@ private void buildSortedReadEndLists(final boolean useBarcodes) {
                                     pairedEnds.read2Coordinate == pairedEnds.read1Coordinate &&
                                     pairedEnds.orientation == ReadEnds.RF) {
                                 pairedEnds.orientation = ReadEnds.FR;
-                            }
+                                }
```
fleharty: Space

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -578,11 +626,26 @@ private void generateDuplicateIndexes(final boolean useBarcodes, final boolean i
             maxInMemory /= 2;
             this.opticalDuplicateIndexes = new SortingLongCollection(maxInMemory, TMP_DIR.toArray(new File[TMP_DIR.size()]));
         }
+        // If we're are tracking representative reads, adjust maInMemory
```
fleharty: spelling, maxInMemory

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -578,11 +626,26 @@ private void generateDuplicateIndexes(final boolean useBarcodes, final boolean i
...
+        if (TAG_DUPLICATE_SET_MEMBERS) {
+            // Memory requirements for RepresentativeReadIndexer:
+            // three int entries + overhead: (3 * 4) + 4
+            maxInMemory = (int) (Runtime.getRuntime().maxMemory() * 0.25) / 16;
```
fleharty: Where do these numbers come from?

src/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -578,11 +626,26 @@ private void generateDuplicateIndexes(final boolean useBarcodes, final boolean i
...
+                    representativeIndexCodec,
+                    new RepresentativeReadComparator(),
+                    maxInMemory,
+                    TMP_DIR);
+        }
 
         ReadEndsForMarkDuplicates firstOfNextChunk = null;
-        final List<ReadEndsForMarkDuplicates> nextChunk = new ArrayList<ReadEndsForMarkDuplicates>(200);
+        final List nextChunk;
+        nextChunk = new ArrayList<ReadEndsForMarkDuplicates>(200);
```
fleharty: Can 200 be made into a variable, and why is 200 chosen?

src/tests/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeReadIndexTest.java
```diff
@@ -0,0 +1,99 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2014 The Broad Institute
```
fleharty: Year

src/tests/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeeadIndexTester.java
```diff
@@ -0,0 +1,124 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2014 The Broad Institute
```
fleharty: Year

src/java/picard/sam/markduplicates/util/RepresentativeReadIndexerCodec.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009 The Broad Institute
```
fleharty: Year

src/java/picard/sam/markduplicates/util/StringCodec.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009 The Broad Institute
```
fleharty: Year

src/java/picard/sam/markduplicates/util/ReadEndsForMarkDuplicates.java
```diff
@@ -34,10 +34,10 @@
     What do we need to store you ask?  Well, we need to store:
        - byte: orientation
        - short: libraryId, readGroup, tile, x, y, score
-       - int: read1ReferenceIndex, read1Coordinate, read2ReferenceIndex, read2Coordinate
+       - int: read1ReferenceIndex, read1Coordinate, read2ReferenceIndex, read2Coordinate, duplicateSetSize
        - long: read1IndexInFile, read2IndexInFile
      */
-    protected static final int SIZE_OF = (1 * 1) + (5 * 2) + (4 * 4) + (8 * 2) + 1
+    protected static final int SIZE_OF = (1 * 1) + (5 * 2) + (5 * 4) + (8 * 2) + 1
```
fleharty: I have no idea where these numbers come from.

src/java/picard/sam/markduplicates/util/StringCodec.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    public String decode() {
+        final String str;
+        try {
+            str = in.readUTF();
+            return str;
+        } catch (final IOException ioe) {
+            throw new PicardException("Exception writing ReadEnds to file.", ioe);
```
fleharty: writing? or reading?

src/java/picard/sam/markduplicates/util/StringCodec.java
```diff
@@ -0,0 +1,70 @@
+/*
```
fleharty: Is StringCodec used?  It's not clear to me what it is used for.

src/tests/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeeadIndexTester.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * This class is an extension of AbstractMarkDuplicatesCommandLineProgramTester used to test MarkDuplicatesWithMateCigar with SAM files generated on the fly.
+ * This performs the underlying tests defined by classes such as see AbstractMarkDuplicatesCommandLineProgramTest and MarkDuplicatesWithMateCigarTest.
+ */
+public class MarkDuplicatesTagRepresentativeeadIndexTester extends AbstractMarkDuplicatesCommandLineProgramTester {
+
+    //final public Map<String, String> readToRepReadMap = new HashMap<>(); // map to contain representative read names for each record
```
fleharty: remove comment

src/tests/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeReadIndexTest.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+package picard.sam.markduplicates;
+
+import org.testng.annotations.Test;
+
+/**
+ * This class defines the individual test cases to run. The actual running of the test is done
+ * by MarkDuplicatesWithMateCigarTester (see getTester).
+ * @author nhomer@broadinstitute.org
```
fleharty: Author

src/tests/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeReadIndexTest.java
```diff
@@ -0,0 +1,99 @@
...
+ * THE SOFTWARE.
+ */
+
+package picard.sam.markduplicates;
+
+import org.testng.annotations.Test;
+
+/**
+ * This class defines the individual test cases to run. The actual running of the test is done
+ * by MarkDuplicatesWithMateCigarTester (see getTester).
```
fleharty: Do you mean, MarkDuplicatesTagRepresentativeeadIndexTester?

src/tests/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeeadIndexTester.java
```diff
@@ -0,0 +1,124 @@
...
+
+        addArg("TAGGING_POLICY=All");
+        addArg("TAG_DUPLICATE_SET_MEMBERS=true");
+    }
+
+    @Override
+    protected CommandLineProgram getProgram() { return new MarkDuplicates(); }
+
+    @Override
+    public void test() {
```
fleharty: Is there a better name for this test?

src/tests/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeeadIndexTester.java
```diff
@@ -0,0 +1,124 @@
...
+            // Ensure the program output the same number of records as were read in
+            Assert.assertEquals(outputRecords, this.getNumberOfRecords(), ("saw " + outputRecords + " output records, vs. " + this.getNumberOfRecords() + " input records"));
+
+            // Check the values written to metrics.txt against our input expectations
+            final MetricsFile<DuplicationMetrics, Comparable<?>> metricsOutput = new MetricsFile<DuplicationMetrics, Comparable<?>>();
+            try{
+                metricsOutput.read(new FileReader(metricsFile));
+            }
+            catch (final FileNotFoundException ex) {
+                System.err.println("Metrics file not found: " + ex);
```
fleharty: Should you maybe throw this exception after catching it?

lhogstrom:
Thanks @fleharty, these comments are helpful, I'll try to address these points later this week.

src/java/picard/sam/markduplicates/MarkDuplicates.java
lhogstrom: I am using '-1' as the null value for integers. This seems to be the convention in other methods like in ReadEnds:
https://github.com/broadinstitute/picard/blob/master/src/main/java/picard/sam/markduplicates/util/ReadEnds.java 

The purpose was to initialize these integers once before we begin to iterate of over each record in dowork(). These variables will only be assigned to real values for a record if the 'TAG_DUPLICATE_SET_MEMBERS' flag is on. Otherwise, they retain their null value.

src/java/picard/sam/markduplicates/util/ReadEndsForMarkDuplicates.java
lhogstrom: The first series of entries are the number primitives of each type that will be stored in the object multiplied by the number of bytes needed for each type: 
http://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html

Here is info I was reading about the overhead of 8 bytes listed here: 
http://stackoverflow.com/questions/258120/what-is-the-memory-consumption-of-an-object-in-java

I'm not sure how the extra 13 bytes were calculated before.

yfarjoun:
@lhogstrom will you be working on this PR again?

lhogstrom:
@yfarjoun, yes I've rebased with the gradle build on my local branch and I am working to get the unit tests running again.

src/java/picard/sam/markduplicates/MarkDuplicates.java
lhogstrom: I don't know the motivation for this particular value. It's a good question, but I'm not aiming to change the original behavior with this feature I'm working on.

lhogstrom:
@fleharty and I discussed, back to him.

src/test/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeReadIndexTest.java
```diff
@@ -0,0 +1,99 @@
...
+ * by MarkDuplicatesWithMateCigarTester (see getTester).
+ * @author hogstrom@broadinstitute.org
+ */
+public class MarkDuplicatesTagRepresentativeReadIndexTest extends AbstractMarkDuplicatesCommandLineProgramTest {
+    protected MarkDuplicatesTagRepresentativeReadIndexTester getTester() {
+        return new MarkDuplicatesTagRepresentativeReadIndexTester();
+    }
+
+    @Test
+    public void testRepresentativeReadTag() {
```
fleharty: Could you add a brief comment about what this test does.

fleharty: and what it expects

src/test/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeReadIndexTester.java
```diff
@@ -0,0 +1,126 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package picard.sam.markduplicates;
+
+import htsjdk.samtools.DuplicateScoringStrategy;
```
fleharty: Unused import

src/test/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeReadIndexTester.java
```diff
@@ -0,0 +1,126 @@
...
+import htsjdk.samtools.DuplicateScoringStrategy.ScoringStrategy;
```
fleharty: Unused import

src/test/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeReadIndexTester.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.Map;
+
+/**
+ * This class is an extension of AbstractMarkDuplicatesCommandLineProgramTester used to test MarkDuplicatesWithMateCigar with SAM files generated on the fly.
+ * This performs the underlying tests defined by classes such as see AbstractMarkDuplicatesCommandLineProgramTest and MarkDuplicatesWithMateCigarTest.
+ * @author hogstrom@broadinstitute.org
+ */
+public class MarkDuplicatesTagRepresentativeReadIndexTester extends AbstractMarkDuplicatesCommandLineProgramTester {
+
+    //final public Map<String, String> readToRepReadMap = new HashMap<>(); // map to contain representative read names for each record
```
fleharty: Remove this line

src/test/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeReadIndexTester.java
```diff
@@ -0,0 +1,126 @@
...
+    final public Map<Integer, Integer> expectedRepresentativeIndexMap = new HashMap<>();
+    final public Map<String, Integer> expectedSetSizeMap = new HashMap<>();
+    public boolean testRepresentativeReads = false;
+
+    public MarkDuplicatesTagRepresentativeReadIndexTester() {
+
```
fleharty: extra line

src/test/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeReadIndexTester.java
```diff
@@ -0,0 +1,126 @@
...
+            // Read the output and check the duplicate flag
+            int outputRecords = 0;
+            int indexInFile = 0;
+            final SamReader reader = SamReaderFactory.makeDefault().open(getOutput());
+            for (final SAMRecord record : reader) {
+                outputRecords++;
+
+                final String key = samRecordToDuplicatesFlagsKey(record);
+                if (!this.duplicateFlags.containsKey(key)) {
+                    System.err.println("DOES NOT CONTAIN KEY: " + key);
```
fleharty: Instead of printing to stderr, would it be better to throw an exception?

src/test/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeReadIndexTester.java
```diff
@@ -0,0 +1,126 @@
...
+                }
+                indexInFile+=1;
+            }
+            CloserUtil.close(reader);
+
+            // Ensure the program output the same number of records as were read in
+            Assert.assertEquals(outputRecords, this.getNumberOfRecords(), ("saw " + outputRecords + " output records, vs. " + this.getNumberOfRecords() + " input records"));
+
+            // Check the values written to metrics.txt against our input expectations
+            final MetricsFile<DuplicationMetrics, Comparable<?>> metricsOutput = new MetricsFile<DuplicationMetrics, Comparable<?>>();
```
fleharty: I think you don't need to explicitly state your template classes, and can just do,
new MetricsFile<>();

src/test/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeReadIndexTester.java
```diff
@@ -0,0 +1,126 @@
...
+            try{
+                metricsOutput.read(new FileReader(metricsFile));
+            }
+            catch (final FileNotFoundException ex) {
+                throw new PicardException("Metrics file not found: " + ex);
+            }
+            Assert.assertEquals(metricsOutput.getMetrics().size(), 1);
```
fleharty: It's not immediately clear to me why this should be 1, could you put in a brief comment?

yfarjoun:
@lhogstrom could you find a moment to respond to marks comments?

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -600,18 +646,36 @@ private ReadEndsForMarkDuplicates buildReadEnds(final SAMFileHeader header, fina
...
+            // three int entries + overhead: (3 * 4) + 4 = 16 bytes 
+            entryOverhead = 16;
+        }
+        else {
+            entryOverhead = SortingLongCollection.SIZEOF;
+        }
         // Keep this number from getting too large even if there is a huge heap.
-        int maxInMemory = (int) Math.min((Runtime.getRuntime().maxMemory() * 0.25) / SortingLongCollection.SIZEOF, (double) (Integer.MAX_VALUE - 5));
-        // If we're also tracking optical duplicates, cut maxInMemory in half, since we'll need two sorting collections
+        int maxInMemory = (int) Math.min((Runtime.getRuntime().maxMemory() * 0.25) / entryOverhead, (double) (Integer.MAX_VALUE - 5));
```
lhogstrom: I don’t know why these values were chosen. I was assuming that it is a heuristic to specify that memory overhead of the duplicate indices should not to exceed 25%.

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -693,6 +762,40 @@ private void addIndexAsDuplicate(final long bamIndex) {
...
+            if (end.score > maxScore || best == null) {
+                maxScore = end.score;
+                best = end;
+            }
+        }
+
+        // for read name (for representative read name), add the last of the pair that was examined
+        for (final ReadEndsForMarkDuplicates end : list) {
+            addRepresentativeReadOfDuplicateSet(best.read1IndexInFile, list.size(), end.read1IndexInFile);
+            addRepresentativeReadOfDuplicateSet(best.read1IndexInFile, list.size(), end.read2IndexInFile);
```
lhogstrom: No, I think we should have only one value that marks all members of the duplicate set. I think the read1 index-in-file of the representative read is a reasonable choice. I think this is better than having two separate indexes for the first and second pair. I’ll be more explicit about this in the description of this function.

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -621,13 +685,17 @@ private void generateDuplicateIndexes(final boolean useBarcodes, final boolean i
             } else {
                 if (nextChunk.size() > 1) {
                     markDuplicatePairs(nextChunk);
+                    if (TAG_DUPLICATE_SET_MEMBERS) addRepresentativeReadIndex(nextChunk);
```
lhogstrom: My thought is that since it is only repeated once and is only 2 lines, that abstracting it would make it a little less readable and one wouldn't gain too much having a new function. Not a big deal to me either way though!

lhogstrom:
@fleharty Thanks again for your review. I apologize for the delay on the revisions.

lhogstrom:
@fleharty @yfarjoun checking in about the potential for merging this :)

fleharty:
@lhogstrom Thanks for checking, I will finish the review later today.

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -8,7 +8,7 @@
  * in the Software without restriction, including without limitation the rights
  * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
  * copies of the Software, and to permit persons to whom the Software is
- * furnished to do so, subject to the following conditions:
+ * furrished to do so, subject to the following conditions:
```
fleharty: furnished

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -417,7 +418,7 @@ protected int doWork() {
     /**
      * package-visible for testing
      */
-    long numOpticalDuplicates() { return ((long) this.libraryIdGenerator.getOpticalDuplicatesByLibraryIdMap().getSumOfValues()); } // cast as long due to returning a double
+    long numOpticalDuplicates() { return ((long) this.libraryIdGenerator.getOpticalDuplicatesByLibraryIdMap().getSumOfValues()); } // cast as long due to returring a double
```
fleharty: returning

lhogstrom:
@fleharty @yfarjoun Thanks for all of your comments and improvements Mark. Rebased and ready to merge.

sooheelee:
I need to mention this new feature in the MarkDuplicates documentation. This is a lot of information to digest. Can I get a summary of the feature from someone (@lhogstrom @yfarjoun @fleharty?) by meeting face-to-face for 5 minutes?

lhogstrom:
@sooheelee I will message you to set up a time to chat.

fleharty:
👍  Merge away!

kbergin:
To clean up this repo, we are deleting branches that have been merged or closed. You have a short period of time to restore this branch if needed. Thanks!

PR #568, opened 2016-06-17
remove what seems to be a spurious tagging of each read with a Mark…

yfarjoun:
…Duplicates tag

@tfenne this seems to be your doing...am I missing something? I'm not sure why this line was there...


yfarjoun:
There's a test that makes sure that this behaviour is maintained...so I'm guessing that this was desirable at some point in the past. As far as I can tell there is zero information in this tag and it probably adds quite a bit of foot-print to the files... I propose we remove it and with it the test...but I would like to understand what (if anything) it is used for.

yfarjoun:
so. A little digging shows that this was added by @alecw in 2012... https://github.com/broadinstitute/picard/commit/f5864bf8437a70cab2b38d504715cc6808b41962

I think I understand what is going on. The PG records do not specify which of the readgroups they operated on. so if you merge several files that had different programs run on them you will not know which of the PR chains are relevant for each read in the combined result. OK.

PR #570, opened 2016-06-20
Fixing bug in GatherVcfs.

nh13:
If the VCF header spans multiple Gzip blocks and there are no variants,
then an Exception could be thrown:
- java.lang.IllegalStateException: Could not read available bytes from BlockCompressedInputStream

Not adding tests since the fix is obvious, and the issue breaks a production pipeline.  Thanks for reviewing: @yfarjoun @jacarey @ktibbett.

### DIFF
--- a/src/java/picard/vcf/GatherVcfs.java
+++ b/src/java/picard/vcf/GatherVcfs.java
@@ -210,7 +210,7 @@ public class GatherVcfs extends CommandLineProgram {
                     final BlockCompressedInputStream blockIn = new BlockCompressedInputStream(in, false);
                     boolean lastByteNewline = true;
 
-                    while (in.available() > 0) {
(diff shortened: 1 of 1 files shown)


yfarjoun:
Thanks for this @nh13. We also saw this problem and didn't get around to fixing it. 

Regarding the testing, I think it would be good to have a test put so that we do not regress.... either that, or at least put in an issue so that we record what needs to be done....

nh13:
It is a hard test to write, since it needs a particular VCF, and I cannot share the one that caused the problem due to IP issues.  I did as you said and made a new issue: https://github.com/broadinstitute/picard/issues/571

nh13:
@yfarjoun all tests pass, ready for merge!

PR #576, opened 2016-06-24
squid:S2063 - Comparators should be Serializable

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule squid:S2063 - Comparators should be "Serializable"

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S2063

M-Ezzat


vivster7:
We are not serializing the data structures that use these Comparators and it is unlikely we will have to. Implementing Serializable seems to add unnecessary baggage to the Comparators. When the need arises, we can add serialization, but until then, its just code clutter.

Closing this PR.

Issue #572, opened 2016-06-22
Bug in MarkDuplicates when using ASSUME_SORT_ORDER


yfarjoun:
In order to use Picard's Markduplicates in queryname mode (which marks supplementary reads and secondary reads like the canonical read) the input must be in "queryname" order, or the flag "ASSUME_SORT_ORDER=queryname" should be set (if the reads known to be query-name grouped, but not queryname sorted). The latter will fail if the reads are not actually **sorted** by queryname since the program uses that sort order in the output writer which then fails to write records that are out of order....

Thanks for the user for reporting. 

Working on fixing this.

vdauwera:
@yfarjoun I see you closed your PR without merging. Does that mean this is no longer considered a bug, or it is but it needs a different approach for fixing?

PR #582, opened 2016-06-28
Fix MarkDuplicates Bug on query-sorted input

yfarjoun:
- fixed bug #572  in MarkDuplicates that precluded the use of data that was queryname _grouped_ but not _sorted_
- added test covering this example.


yfarjoun:
fixes #572

yfarjoun:
@nh13 are you available to take a look at this?

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -234,7 +234,9 @@ protected int doWork() {
         final SAMFileHeader.SortOrder sortOrder = header.getSortOrder();
 
         final SAMFileHeader outputHeader = header.clone();
-
+        if(ASSUME_SORTED || ASSUME_SORT_ORDER != null) {
+            outputHeader.setSortOrder(SAMFileHeader.SortOrder.unsorted);
+        }
```
nh13: why is the output always unsorted?  I thought that `MarkDuplicates` can still take in and output coordinate sorted BAMs?  Also `if(` -> `if (`.

nh13: Also `ASSUME_SORTED` means `coordinate`, so please either correct this or explain to me what's going on here.

nh13:
@yfarjoun I think that setting the output to `queryname` if `ASSUME_SORTED` is true or `ASSUME_SORTED_ORDER` is set (regardless of value) is wrong.  Could you comment?

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
yfarjoun: Ah, yes. you're right. I was thinking about the query-sorted too much....

yfarjoun: On your first note, the problem is that if the input is not query-sorted, but rather it is query-grouped, you need to say "ASSUME_SORT_ORDER=queryname" but the output will be the same order as the input...and the writer will complain that the order is incorrect....

yfarjoun:
Thanks @nh13. I'm not sure what the correct behavior should be. Right now (in this PR), if you state an assumed sort-order, the output header will be "unsorted" and the output will have the same order as the input. The reason for this is the distinction between ordered-by and grouped-by. Since htsjdk support for group-by is minimal, I co-opted ASSUME_SORT_ORDER=queryname to *imply* GO=queryname, but if I put SO=queryname in the header, then I have two options:

 1. assume sorted and fail when it isn't.
 2. needlessly sort the output.

So, instead, I set the output to SO=unsorted (since if you had to ASSUME_SORT_ORDER=anything it means that the input isn't actually sorted that way...right?) and then I can output in the same order it was input.

 I guess there's another option: modify the htsjdk writer to allow presorted=true and no validation on the sort-order.

 ideas are welcome.

yfarjoun:
going to make the change so that if ASSUME_SORT_ORDER is not null, the output sort order is UNSORTED.

yfarjoun:
After thinking about this, I do not see the value of this PR.

PR #588, opened 2016-07-07
Adding a page to help explain and transform Base Qualities.

nh13:
I have found this useful when I want to convert between ASCII and integer-values for base qualities.

### DIFF
--- a/explain-qualities.html
+++ b/explain-qualities.html
@@ -1,0 +1,88 @@
+---
+---
+<!doctype html>
+<html>
+{% include head.html title="Explain Base Qualities" %}
+
+<script type="text/javascript">
+    // minimum value, maximum value, ascii shift
+	var qualValues = [
+		[0,  93, 33], // Sanger
+		[-5, 62, 64], // Solexa
+		[0,  40, 64], // Illumina 1.3+
+		[2,  40, 64], // Illumina 1.5+
+		[0,  93, 33]  // Illumina 1.8+
+	];
+	var resultPrefix = "<strong>Result: </strong>"
+    function convert(which) {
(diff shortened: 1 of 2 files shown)


tfenne:
Seems reasonable to me @nh13.  A little hard to tell without seeing it live on the site.

vdauwera:
Will take a look later today and copy it to my fork of gh-pages so we can see what it looks like live.

nh13:
@vdauwera that would be great to see!

vdauwera:
See here: http://vdauwera.github.io/picard/explain-qualities.html

I think the layout could be improved to make it more user-friendly. And maybe add some brief instructions.

nh13:
@vdauwera any suggestions on the layout and instructions, or want to submit a PR to this branch?  I am not an expert on UI.

vdauwera:
I'll take a stab at it.

nh13:
I am in your debt.

vdauwera:
Working on it now, sorry it took a while

vdauwera:
@nh13 Have a look at http://vdauwera.github.io/picard/explain-qualities.html

Do we maybe want to increase the size of the input field to allow for longer strings/reads? Could use a wrapped text box.

nh13:
@vdauwera looking good!  I would say having the text field be at least 150 characters long (typical runs are 100-200 at the least) would be great.  I agree that a text area or a wrapped text box would be better.  If you make that change and squash into two commits (one for me, one for you), we could probably merge.

vdauwera:
OK, done. Added text area and also set the word break property of the result paragraph to break-all to make sure the result/output wraps nicely. I think this is good to go.

PR #591, opened 2016-07-08
Replacing deprecated method.

kasont:


nh13:
@kasont probably good to get a thumbs up from someone else, even on such a trivial change.

yfarjoun:
👍

PR #592, opened 2016-07-08
Replacing deprecated method entire use.

kasont:
Replaced entire use of getChr() for getContig().


yfarjoun:
looks good to me. 👍 

I'll merge in a few days unless someone objects.

PR #595, opened 2016-07-11
Kt fingerprintchecker fix

kasont:


kasont:
There shouldn't be five commits anyone know how to remove this?

PR #596, opened 2016-07-11
Kt fingerprintchecker fix

kasont:


src/main/java/picard/fingerprint/FingerprintChecker.java
```diff
@@ -151,8 +147,8 @@ public void setpLossofHet(final double pLossofHet) {
             if (samples == null) samples = ctx.getSampleNames();
 
             if (isUsableSnp(ctx)) {
-                final HaplotypeBlock h = this.haplotypes.getHaplotype(ctx.getChr(), ctx.getStart());
-                final Snp snp = this.haplotypes.getSnp(ctx.getChr(), ctx.getStart());
+                final HaplotypeBlock h = this.haplotypes.getHaplotype(ctx.getContig(), ctx.getStart());
```
jsotobroad: why is this changing?

nh13: `getChr` is deprecated?

kasont: @jsotobroad @nh13 The method getChr() is depreciated along and it was changed along all classes that use it. It simply wasn't changed yet to match its original form to make sure the replacement of GenotypeReader wasn't hindered by it. I changed it after looking deeper into getChr() which only calls getContig() so no error would have occurred anyway.

jsotobroad:
👍

PR #597, opened 2016-07-12
Make RnaSeqMetricsCollector more easily extensible.

alecw:
- Make some private members public.

- Add ctor that takes an instance of metrics object, so subclass can substitute a subclass of RnaSeqMetrics.

- Move some inline code into a protected method so it can be shared by subclass.

### DIFF
--- a/src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
+++ b/src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
@@ -36,7 +36,7 @@ public class RnaSeqMetricsCollector extends SAMRecordMultiLevelCollector<RnaSeqM
     private final int minimumLength;
     private final StrandSpecificity strandSpecificity;
     private final double rrnaFragmentPercentage;
-    private final Long ribosomalInitialValue;
+    protected final Long ribosomalInitialValue;
 
     final private Set<Integer> ignoredSequenceIndices;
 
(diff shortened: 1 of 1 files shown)


yfarjoun:
@meganshand would you like to take a look at this as well? This seems to be related to stuff you are working on...

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
```diff
@@ -95,21 +95,34 @@ public RnaSeqMetricsCollector(final Set<MetricAccumulationLevel> accumulationLev
         return ignoredSequenceIndices;
     }
 
-    private class PerUnitRnaSeqMetricsCollector implements PerUnitMetricCollector<RnaSeqMetrics, Integer, SAMRecord> {
+    protected class PerUnitRnaSeqMetricsCollector implements PerUnitMetricCollector<RnaSeqMetrics, Integer, SAMRecord> {
+
+        protected final RnaSeqMetrics metrics;
```
yfarjoun: hmmm. It's nice to have a final metrics object, no? perhaps this can be done with a getMetrics that is overrided by the inherited class.

yfarjoun:
looks good to me @alecw. The note about final is open for discussion.

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
alecw: Hi @yfarjoun ,

I'm not sure I understand the question.  The metrics member remains final in this PR.  All I did was make it protected, because otherwise it wasn't accessible in a subclass in a different package.

It isn't feasible (at least not easily) to have an overridable getMetrics(), because the subclass isn't fully initialized until after the base class ctor has returned, and as you can see, the metrics member is accessed in the base class ctor.  I tried an overridable getMetrics() initially, and settled on what I did after discovering that getMetrics() wouldn't work.  Ultimately I think what I've done is cleaner, because there aren't 2 metrics objects for each PerUnitCollector to create confusion.

-Alec

yfarjoun: :+1: thanks for the explanation!

PR #599, opened 2016-07-14
UMI Aware Mark Duplicates with rudimentary error correction

fleharty:
This is a tool for performing MarkDuplicates in a UMI aware manner.  It uses a simple graph based algorithm to group UMIs into duplicate sets.  The error correction model is a bit naive, but this is intended to be a first iteration, and a more sophisticated algorithm will be submitted in the near future.

### DIFF
--- a/src/main/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
+++ b/src/main/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
@@ -32,10 +32,7 @@ import htsjdk.samtools.SAMFileWriterFactory;
 import htsjdk.samtools.SAMRecord;
 import htsjdk.samtools.SAMRecordDuplicateComparator;
 import htsjdk.samtools.SAMTag;
-import htsjdk.samtools.util.IOUtil;
-import htsjdk.samtools.util.IterableAdapter;
-import htsjdk.samtools.util.Log;
-import htsjdk.samtools.util.ProgressLogger;
+import htsjdk.samtools.util.*;
 import picard.PicardException;
 import picard.cmdline.CommandLineProgramProperties;
 import picard.cmdline.programgroups.Testing;
@@ -120,10 +117,7 @@ public class SimpleMarkDuplicatesWithMateCigar extends MarkDuplicates {
         final SAMRecordDuplicateComparator comparator = new SAMRecordDuplicateComparator(Collections.singletonList(headerAndIterator.header));
         comparator.setScoringStrategy(this.DUPLICATE_SCORING_STRATEGY);
 
-        final DuplicateSetIterator iterator = new DuplicateSetIterator(headerAndIterator.iterator,
-                headerAndIterator.header,
-                false,
-                comparator);
+        final CloseableIterator<DuplicateSet> iterator = getDuplicateSetIterator(headerAndIterator, comparator);
 
         // progress logger!
         final ProgressLogger progress = new ProgressLogger(log, (int) 1e6, "Read");
@@ -232,5 +226,12 @@ public class SimpleMarkDuplicatesWithMateCigar extends MarkDuplicates {
         finalizeAndWriteMetrics(libraryIdGenerator);
 
         return 0;
+    }
+
+    protected CloseableIterator<DuplicateSet> getDuplicateSetIterator(final SamHeaderAndIterator headerAndIterator, final SAMRecordDuplicateComparator comparator) {
+        return new DuplicateSetIterator(headerAndIterator.iterator,
+                    headerAndIterator.header,
+                    false,
+                    comparator);
     }
 }
--- a/src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
+++ b/src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
@@ -1,0 +1,249 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 The Broad Institute
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package picard.sam.markduplicates;
+
+import htsjdk.samtools.DuplicateSet;
+import htsjdk.samtools.DuplicateSetIterator;
+import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.util.CloseableIterator;
+import picard.PicardException;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static java.util.stream.Collectors.counting;
+
+/**
+ * Created by fleharty on 5/23/16.
+ */
+public class UmiAwareDuplicateSetIterator implements CloseableIterator<DuplicateSet> {
+    private final DuplicateSetIterator wrappedIterator;
+    private Iterator<DuplicateSet> nextSetsIterator;
+    private final int editDistanceToJoin;
+    private final boolean addInferredUmi;
+    private final String umiTag;
+    private final String inferredUmiTag;
+
+    public UmiAwareDuplicateSetIterator(final DuplicateSetIterator wrappedIterator, final int editDistanceToJoin, final boolean addInferredUmi, final String umiTag, final String inferredUmiTag) {
+        this.wrappedIterator = wrappedIterator;
+        this.editDistanceToJoin = editDistanceToJoin;
+        this.addInferredUmi = addInferredUmi;
+        this.umiTag = umiTag;
+        this.inferredUmiTag = inferredUmiTag;
+        nextSetsIterator = Collections.emptyIterator();
+    }
+
+    @Override
+    public void close() {
+        wrappedIterator.close();
+    }
+
+    @Override
+    public boolean hasNext() {
+        return nextSetsIterator.hasNext() || wrappedIterator.hasNext();
+    }
+
+    @Override
+    public DuplicateSet next() {
+        if(!nextSetsIterator.hasNext())
+            process(wrappedIterator.next());
+
+        return nextSetsIterator.next();
+    }
+
+    // Takes a duplicate set and breaks it up into possible smaller sets according to the UMI,
+    // and updates nextSetsIterator to be an iterator on that set of DuplicateSets.
+    private void process(final DuplicateSet set) {
+
+        List<SAMRecord> records = set.getRecords();
+
+        // If any records are missing the UMI_TAG proceed as if there were no UMIs
+        // and return nextSetsIterator without breaking it up into smaller sets.
+        for(SAMRecord rec : records) {
+            if(rec.getStringAttribute(umiTag) == null) {
+                nextSetsIterator = Collections.singleton(set).iterator();
+                return;
+            }
+        }
+
+        // Count all the uniquely observed UMI sequences
+        Map<String, Long> umiFrequencies = records.stream()
+                        .collect(Collectors.groupingBy(p -> p.getStringAttribute(umiTag),
+                                counting()));
+
+        UmiGraph umiGraph = new UmiGraph();
+
+        // Add each of the uniquely observed UMI sequences to the graph
+        int firstGroup = 0;
+        for (Map.Entry<String, Long> umiFrequency : umiFrequencies.entrySet()) {
+            umiGraph.addNode(new UmiSequence(umiFrequency.getKey(), umiFrequency.getValue()));
+        }
+        umiGraph.processNodes();
+
+        Map<Integer, List<UmiSequence>> umisByDuplicateSet = umiGraph.getUmisByDuplicateSet();
+
+        // Construct DuplicateSetList
+        List<DuplicateSet> duplicateSetList = new ArrayList<>();
+        for(int i = 0;i < umiGraph.duplicateSetCounter;i++) {
+            DuplicateSet e = new DuplicateSet();
+            duplicateSetList.add(e);
+        }
+
+        for(SAMRecord rec : records) {
+            String umi = records.get(records.indexOf(rec)).getStringAttribute(umiTag);
+
+            // Figure out which group this belongs to
+            int recordGroup = umiGraph.node.get(umi).duplicateSet;
+            duplicateSetList.get(recordGroup).add(records.get(records.indexOf(rec)));
+        }
+
+        // Optionally add the inferred Umi
+        if(addInferredUmi) {
+            for(DuplicateSet ds : duplicateSetList) {
+                // For each duplicate set identify the most common UMI
+                List<SAMRecord> recordsFromDuplicateSet = ds.getRecords();
+
+                // Assign the inferred UMI to all reads in the current group
+                for(SAMRecord record : recordsFromDuplicateSet) {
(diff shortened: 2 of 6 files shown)


fleharty:
@nh13 Would you be able to review this?

src/main/java/picard/sam/markduplicates/SimpleMarkDuplicatesWithMateCigar.java
```diff
@@ -233,4 +227,11 @@ protected int doWork() {
 
         return 0;
     }
+
+    protected CloseableIterator<DuplicateSet> getDuplicateSetIterator(SamHeaderAndIterator headerAndIterator, SAMRecordDuplicateComparator comparator) {
```
yfarjoun: final parameters to function

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+ */
+public class UmiAwareDuplicateSetIterator implements CloseableIterator<DuplicateSet> {
+    private DuplicateSetIterator wrappedIterator;
+    private Iterator<DuplicateSet> nextSetsIterator;
+    private List<List<Integer>> adjacencyList;
+    private List<Integer> groups;
+    int editDistanceToJoin;
+    boolean addInferredUmi;
+
+    public UmiAwareDuplicateSetIterator(final DuplicateSetIterator wrappedIterator, final int editDistanceToJoin, boolean addInferredUmi) {
```
yfarjoun: final boolean

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+    // and updates nextSetsIterator to be an iterator on that set of DuplicateSets.
+    private void process(final DuplicateSet set) {
+
+        List<SAMRecord> records = set.getRecords();
+
+        groups = new ArrayList<>();
+        adjacencyList = new ArrayList<>();
+
+        // If any records don't have the RX attribute... don't do anything special
+        for(int i = 0;i < records.size();i++) {
```
yfarjoun: spaces inside the for loop

yfarjoun: you shuold be using 
```
for(rec:records) {
...
}
```
bu this hwole part looks strange...why is there an empty else? what is this loop trying to do?

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+            }
+        }
+
+        // Sort records by RX tag
+        Collections.sort(records, new Comparator<SAMRecord>() {
+            @Override
+            public int compare(final SAMRecord lhs, final SAMRecord rhs) {
+                if(lhs == null || rhs == null) {
+                    return 0;
+                } else if(lhs.getAttribute("RX") == null || rhs.getAttribute("RX") == null) {
```
yfarjoun: space between `if` and `{`

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+                    return 0;
+                }
+                else {
+                    return ((String) lhs.getAttribute("RX")).compareTo((String) rhs.getAttribute("RX"));
```
yfarjoun: use getAttributeAsString

yfarjoun: it should tidy this section up, since you might be able to avoid the if/else

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+                }
+            }
+        });
+
+        int n = records.size();
+
+        // Locate records that have identical UMI sequences
+        List<String> uniqueObservedUMIs = new ArrayList<String>();
+
+        uniqueObservedUMIs.add((String) records.get(0).getAttribute("RX"));
```
yfarjoun: the barcode tag should be an input parameter.

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+        for(int i = 1;i < n;i++) {
+            // If the records differ (after sorting) we have a new duplicate set.
+            if(!records.get(i).getAttribute("RX").equals(records.get(i-1).getAttribute("RX"))) {
```
yfarjoun: this is just a set of strings...why not use a Set?

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+            }
+        });
+
+        int n = records.size();
+
+        // Locate records that have identical UMI sequences
+        List<String> uniqueObservedUMIs = new ArrayList<String>();
+
+        uniqueObservedUMIs.add((String) records.get(0).getAttribute("RX"));
+        for(int i = 1;i < n;i++) {
```
yfarjoun: java 8 syntax can really clean this up.

src/main/java/picard/sam/markduplicates/UmiAwareMarkDuplicatesWithMateCigar.java
```diff
@@ -0,0 +1,82 @@
...
+import htsjdk.samtools.DuplicateSet;
+import htsjdk.samtools.DuplicateSetIterator;
+import htsjdk.samtools.SAMRecordDuplicateComparator;
+import htsjdk.samtools.util.*;
+import picard.cmdline.CommandLineProgramProperties;
+import picard.cmdline.Option;
+import picard.cmdline.programgroups.Alpha;
+
+/**
+ * This is a simple tool to mark duplicates using the DuplicateSetIterator, DuplicateSet, and SAMRecordDuplicateComparator.
```
yfarjoun: update comment here.

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+            if(!records.get(i).getAttribute("RX").equals(records.get(i-1).getAttribute("RX"))) {
+                uniqueObservedUMIs.add((String) records.get(i).getAttribute("RX"));
+            }
+        }
+
+        // Construct Adjacency List of UMIs that are close
+        for(int i = 0;i < uniqueObservedUMIs.size();i++) {
+            adjacencyList.add(i, new ArrayList<Integer>());
+            groups.add(i, 0);
+            for(int j = 0;j < uniqueObservedUMIs.size();j++) {
```
yfarjoun: stream.filter.map.foreach

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+import java.util.Map;
+
+/**
+ * Created by fleharty on 5/23/16.
+ */
+public class UmiAwareDuplicateSetIterator implements CloseableIterator<DuplicateSet> {
+    private DuplicateSetIterator wrappedIterator;
+    private Iterator<DuplicateSet> nextSetsIterator;
+    private List<List<Integer>> adjacencyList;
+    private List<Integer> groups;
```
yfarjoun: can you move this to being a local variable?

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Created by fleharty on 5/23/16.
+ */
+public class UmiAwareDuplicateSetIterator implements CloseableIterator<DuplicateSet> {
+    private DuplicateSetIterator wrappedIterator;
```
yfarjoun: final

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+    private Iterator<DuplicateSet> nextSetsIterator;
+    private List<List<Integer>> adjacencyList;
```
yfarjoun: this is also used in two methods...could it be locally defined and passed to the second method?

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+    private List<Integer> groups;
+    int editDistanceToJoin;
+    boolean addInferredUmi;
```
yfarjoun: final

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+
+/**
+ * Created by fleharty on 5/23/16.
+ */
+public class UmiAwareDuplicateSetIterator implements CloseableIterator<DuplicateSet> {
+    private DuplicateSetIterator wrappedIterator;
+    private Iterator<DuplicateSet> nextSetsIterator;
+    private List<List<Integer>> adjacencyList;
+    private List<Integer> groups;
+    int editDistanceToJoin;
```
yfarjoun: final

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+
+        uniqueObservedUMIs.add((String) records.get(0).getAttribute("RX"));
+        for(int i = 1;i < n;i++) {
+            // If the records differ (after sorting) we have a new duplicate set.
+            if(!records.get(i).getAttribute("RX").equals(records.get(i-1).getAttribute("RX"))) {
+                uniqueObservedUMIs.add((String) records.get(i).getAttribute("RX"));
+            }
+        }
+
+        // Construct Adjacency List of UMIs that are close
```
yfarjoun: seems like a good candidate for extracting a method

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+                for(SAMRecord record : recordsFromDuplicateSet) {
+                    record.setAttribute("RI", maxEntry.getKey());
+                }
+            }
+        }
+
+        nextSetsIterator = duplicateSetList.iterator();
+    }
+
+    private void mergeGroups(final int index, final int group) {
```
yfarjoun: this is recursive, please add a comment to that effect and explain why it will definitely terminate.

yfarjoun: (easier, I think, if groups would be an argument to this function rather than a global variable)

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+        if(groups.get(index) == 0) {
+            groups.set(index, group);
+            for (int i = 0; i < adjacencyList.get(index).size(); i++) {
+                mergeGroups(adjacencyList.get(index).get(i), group);
+            }
+        }
+    }
+
+    private int getEditDistance(String s1, String s2) {
+        if(s1 == null && s2 == null) {
```
yfarjoun: what if one is null and the other isn't?

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+            return 0;
+        }
+        if(s1.length() != s2.length()) {
```
yfarjoun: NPE danger

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
```diff
@@ -0,0 +1,225 @@
...
+    private void mergeGroups(final int index, final int group) {
+        if(groups.get(index) == 0) {
+            groups.set(index, group);
+            for (int i = 0; i < adjacencyList.get(index).size(); i++) {
+                mergeGroups(adjacencyList.get(index).get(i), group);
+            }
+        }
+    }
+
+    private int getEditDistance(String s1, String s2) {
```
yfarjoun: finals

src/main/java/picard/sam/markduplicates/UmiAwareMarkDuplicatesWithMateCigar.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private final Log log = Log.getInstance(UmiAwareMarkDuplicatesWithMateCigar.class);
+
+    /** Stock main method. */
+    public static void main(final String[] args) {
+        new UmiAwareMarkDuplicatesWithMateCigar().instanceMainWithExit(args);
+    }
+
+    @Override
+    protected CloseableIterator<DuplicateSet> getDuplicateSetIterator(SamHeaderAndIterator headerAndIterator, SAMRecordDuplicateComparator comparator) {
```
yfarjoun: finals

src/main/java/picard/sam/markduplicates/UmiAwareMarkDuplicatesWithMateCigar.java
```diff
@@ -0,0 +1,82 @@
...
+
+public class UmiAwareMarkDuplicatesWithMateCigar extends SimpleMarkDuplicatesWithMateCigar {
+
+    @Option(shortName = "EDIT_DISTANCE_TO_JOIN",
+            doc = "This option specifies the edit distance of UMIs to join", optional = true)
+    public int EDIT_DISTANCE_TO_JOIN = 1;
+
+
+    @Option(shortName = "ADD_INFERRED_UMI",
+            doc = "This option adds the inferred UMI to the bam in the RI tag", optional = true)
```
yfarjoun: RI should also be a parameter.

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
fleharty: What is NPE?

yfarjoun:
quick first pass @fleharty.

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
fleharty: The else was an appendage, that I've removed, but the for loop is to deal with the case where any of the records don't have a UMI.

yfarjoun: does something like:
```java
if(records.stream
   .map(v->v.getAttribute("RX")
   .filter(s->s!=null)
   .isEmpty()) return;
```
work? (I am writing this from memory, so there might be some parts missing/misspelled)

fleharty: This particular stream doesn't work, but I may be able to use something like this elsewhere.

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
fleharty: I was able to get this to work, but I kept it out of the code because I think it made it less readable.  Any suggestions on a better way to use streams?

            observedUmis.stream().map(observedUmis::indexOf)
                    .filter(adjacencyListIndex2 -> getEditDistance(observedUmis.get(adjacencyListIndex), observedUmis.get(adjacencyListIndex2)) <= editDistanceToJoin)
                    .forEach(adjacencyListIndex2 -> adjacencyList.get(adjacencyListIndex).add(adjacencyListIndex2));

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
fleharty: I cleaned this up in a few ways, but perhaps not in the Java 8 way... I also tried using a stream here, but it was even uglier.

fleharty:
@yfarjoun Back to you, thanks for the review.  I wasn't able to improve the readability with the Java 8 syntax, it seemed to make it messier... but perhaps you have some suggestions.

src/main/java/picard/sam/markduplicates/UmiAwareDuplicateSetIterator.java
yfarjoun: no, I guess it's fine. I didn't realize that there are two lists here...this is actually a little concerning since it relies on a connection between the two lists...and there's no check that the lists are properly connected. I'd recommend using a map instead of a list in one of the containers to alleviate this.

fleharty:
@yfarjoun I made some significant changes to how the code works by getting rid of the recursion and the two lists that had a funny dependency on them.  I think this makes the code a lot more readable.  I'm not making a commit though because I now fail two of my tests, and I'm trying to figure out why.

yfarjoun:
great! thanks for letting me know. if its different enough, consider
closing the PR and opening a new one.

fleharty:
Closing this pull request because I've made extensive changes to the code, I'll create a new PR.

PR #603, opened 2016-07-15
Using predefined version of MQ (two different kinds) and SA.

yfarjoun:
A small PR that makes use of predefined strings rather than using "SA" and "MQ". 

Aesthetic only. Has no performance effects.


ktibbett:
@yfarjoun : 👍

PR #606, opened 2016-07-22
Optionally writing compressed FASTQs per read group.

vinjana:


vinjana:
Same option as COMPRESS_OUTPUTS/GZIP in IlluminaBasecallsToFastq.

PR #607, opened 2016-07-22
SamToFastq: Optionally writing compressed FASTQs per read group.

vinjana:


src/main/java/picard/sam/SamToFastq.java
```diff
@@ -133,6 +133,9 @@
             "If the original read is shorter than CLIPPING_MIN_LENGTH then the original read length will be maintained.")
     public int CLIPPING_MIN_LENGTH = 0;
 
+    @Option(shortName = "GZIP", doc = "Compress output FASTQ files using gzip and append a .gz extension to the file names.", optional = false)
+    public Boolean COMPRESS_OUTPUTS = false;
```
nh13: This is confusing for when `OUTPUT_PER_RG` is `false`, since you can specify `*.fastq.gz` and it will automatically write a compressed FASTQ even if `COMPRESS_OUTPUTS=false`.  Perhaps change it to `COMPRESS_PER_RG_OUTPUTS`?  Also, I would omit the short name, or have it include `PER_RG` (`PER_RG_GZIP`)?

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -286,6 +289,9 @@ private File makeReadGroupFile(final SAMReadGroupRecord readGroup, final String
         fileName = IOUtil.makeFileNameSafe(fileName);
         if (preExtSuffix != null) fileName += preExtSuffix;
         fileName += ".fastq";
+        if (COMPRESS_OUTPUTS) {
```
nh13:
```
fileName += COMPRESS_PER_RG_OUTPUTS ? ".fastq.gz" : ".fastq";
```

src/main/java/picard/sam/SamToFastq.java
nh13: Consider adding a `mutex` for this argument:
`mutex = {"FASTQ", "SECOND_END_FASTQ", "UNPAIRED_FASTQ"}`

nh13:
@vinjana thanks for the PR, and I definitely see the utility of this method, so please address the comments and I'll take a look again.

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -84,20 +84,25 @@
...
-    @Option(shortName = "FU", doc = "Output FASTQ file for unpaired reads; may only be provided in paired-FASTQ mode", optional = true, mutex = {"OUTPUT_PER_RG"})
+    @Option(shortName = "FU", doc = "Output FASTQ file for unpaired reads; may only be provided in paired-FASTQ mode", optional = true,
+            mutex = {"OUTPUT_PER_RG", "COMPRESS_OUTPUTS_PER_RG"})
     public File UNPAIRED_FASTQ;
 
     @Option(shortName = "OPRG", doc = "Output a FASTQ file per read group (two FASTQ files per read group if the group is paired).",
             optional = true, mutex = {"FASTQ", "SECOND_END_FASTQ", "UNPAIRED_FASTQ"})
     public boolean OUTPUT_PER_RG;
 
+    @Option(shortName = "GZRG", doc = "Compress output FASTQ files per read group using gzip and append a .gz extension to the file names.",
```
nh13: `GZPRG`? since it it looks like `*PRG` is a good convention?

nh13:
@vinjana one more comment (sorry) and the PR is good to go.

src/main/java/picard/sam/SamToFastq.java
vinjana: Maybe GZOPRG?

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -285,7 +290,7 @@ private File makeReadGroupFile(final SAMReadGroupRecord readGroup, final String
         }
         fileName = IOUtil.makeFileNameSafe(fileName);
         if (preExtSuffix != null) fileName += preExtSuffix;
-        fileName += ".fastq";
+        fileName += COMPRESS_OUTPUTS_PER_RG ? ".fastq.gz" : ".fastq";
```
yfarjoun: is this all that's needed? will the output then be compressed?

vinjana:
Yes. The filename is matched and if ending an .gz the output is compressed.

yfarjoun:
Do you mind adding a test that this is working as designed? i.e. that the results are indeed compressed.

yfarjoun:
@nh13 Do you approve of this PR? If so, can you do the honors?

yfarjoun:
looks like a rebase went very badly...please fix and resubmit.

nh13:
@vinjana could you try and rebasing again?  It looks like you pulled in a lot of commits.

vinjana:
I'll try.

src/main/java/picard/sam/SamToFastq.java
vinjana: As it is currently, the new option only allows to set the compression for read-group files, as only for these the suffix name cannot be chosen on the command line. By contrast, if the (UNPAIRED|SECOND_END|FIRST_END)_FASTQ is set to a filename with a suffixed `.gz`, the output file will be gzipped automatically. So the current solution is just minimally invasive.

But of course I can add a USE_COMPRESSED_OUTPUT option, but then conflicts between the output file names chosen explicitly with or without `.gz` suffix and the new option need to be resolved somehow.

If the filename is suffixed with `.gz` and USE_COMPRESSED_OUTPUT==false (the default), I would compress, for backward compatibility.

More problematic is the second situation, if the filename is *not* suffixed with `.gz` but USE_COMPRESSED_OUTPUT==true.

We could compress, which would kind of break backward compatibility in the sense that the filename is not decisive, but as the user chooses the new option he/she apparently (?) is aware what he/she is doing and really wants to have compression, no matter what the output file suffix is.

We could also leave the output uncompressed, which -- in total -- would implement a logic of the kind: filename suffix has higher priority than compression toggle.

I leave that choice to you guys.

yfarjoun: My thought was that you'd add "USE_COMPRESSED_OUTPUT" and that will determine the file suffixes.

yfarjoun: looked at this code with fresh view. I think what you did is good.

yfarjoun:
I rebased this here #942

PR #616, opened 2016-08-03
adding settings.gradle file

lbergelson:
this sets the project name so that jars will have consistent names
without the settings.gradle file, the project name is taken from the folder name which is not ideal


lbergelson:
@droazen please review

droazen:
:+1: merging

PR #618, opened 2016-08-04
creating a picard.jar when running ./gradlew

### LINKED ISSUE #617: Build should produce a real file named picard.jar
See requests on https://github.com/broadinstitute/picard/issues/601.

lbergelson:
I added a currentJar target which is invoked by default if you run `/.gradlew`

It copies the most recent shadowJar into build/libs/picard-current.jar

fixes #617

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -69,7 +69,7 @@ group = 'com.github.broadinstitute'
 
 defaultTasks 'all'
 
-task all(dependsOn: ['jar', 'distZip', 'documentAll'])
+task all(dependsOn: ['jar', 'distZip', 'documentAll', 'shadowJar', 'currentJar'])
 
 jar {
     manifest {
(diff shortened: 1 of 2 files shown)


lbergelson:
@droazen @nh13 This creates a real copy of the most recent shadowJar in the build/libs folder named picard.jar.  I'll let you two decide the correct name / location of it.

nh13:
@lbergelson thank-you thank-you, I know this seems trivial but it is a big help.  I should probably learn about gradle at some point. thank-you!

lbergelson:
@nh13 No problem.  Happy to help.  Sorry it's taken so long to get all this stuff into better shape.

build.gradle
```diff
@@ -86,6 +86,12 @@ if (JavaVersion.current().isJava8Compatible()) {
     }
 }
 
+task currentJar(type: Copy){
```
droazen: Think this should happen for `shadowJar` as well, since that is the main command for building Picard according to the README.

build.gradle
```diff
@@ -69,7 +69,7 @@ group = 'com.github.broadinstitute'
```
droazen: Can you update the `java -jar picard*-all.jar` line in the README to use `picard.jar` instead, and also mention that a jar with the full version in its name is also emitted.

droazen:
Two comments, back to @lbergelson

PR #620, opened 2016-08-06
Expose the INCLUDE_UNPAIRED option for CollectSequencingArtifactMetrics

nh13:
in CollectMultipleMetrics.

For fragment runs, CollectSequencingArtifactMetrics does not do anything useful unless this option is set to true.

@tfenne or @yfarjoun want to take a look?

### DIFF
--- a/src/main/java/picard/analysis/CollectMultipleMetrics.java
+++ b/src/main/java/picard/analysis/CollectMultipleMetrics.java
@@ -88,8 +88,13 @@ public class CollectMultipleMetrics extends CommandLineProgram {
             "</pre>" +
             "<hr />";
     public static interface ProgramInterface {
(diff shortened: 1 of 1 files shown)


src/main/java/picard/analysis/CollectMultipleMetrics.java
```diff
@@ -88,8 +88,12 @@
             "</pre>" +
             "<hr />";
     public static interface ProgramInterface {
+        default  SinglePassSamProgram makeInstance(final String outbase, final String outext, final File input, final File reference,
```
yfarjoun: Can you comment that the default implementations "forgets" the value of `includeUnpaired`, and calls the version of makeInstance that doesn't use it? It's clear if you read it carefully, but can be missed by a casual observer.

yfarjoun:
one small comment but otherwise LGTM 👍

yfarjoun:
thanks for humoring me.... 👍

PR #622, opened 2016-08-08
Update the NM and MD tags when clipping overlapping inserts in

nh13:
MergeBamAlignment.

The bug comes from changing the cigar but not updating the NM and MD tags.

I modified the clipping test to check the MD and NM tags for FR reads.  The MD and NM tags for the clipped versus not-clipped should be "76" vs. "56" (MD) and from "1" vs. "0" (NM).  For the latter, I inserted a mismatch into the clipped bases for the clipped reads.

@yfarjoun or @ktibbett would you care to review (git blame tells me @ktibbett last touched this in 2011!)?


nh13:
@bradtaylor any chance you could review this today?  I am blocked by this bug.

nh13:
Thank-you @yfarjoun!

Issue #625, opened 2016-08-10
[feature request] MergeBamAlignment: when merging from GRCh38 alignments, retain `AH:*` tag that notes which contigs were ALT-handled


sooheelee:
This tag is overwritten during merging by MergeBamAlignment but would be a good idea to retain.

It is in the header `@SQ` lines and denotes if a contig was ALT-handled by BWA-MEM when aligning to GRCh38.

See [Tutorial#8017](http://gatkforums.broadinstitute.org/gatk/discussion/8017)'s [section 3](http://gatkforums.broadinstitute.org/gatk/discussion/8017#3). Specifically, look at the section titled **How can I tell if a BAM was aligned with ALT-handling?** to see example screenshots of what the `AH` tag looks like.

yfarjoun:
good point soohee!! thanks for figuring this out. I was wondering why the
AH:* tags are missing.

yfarjoun:
@tfenne @nh13 @ktibbett The problem is that MBA takes the dictionary from the input reference file (or rather the nearby dictionary file). 

I would like to propose to take it from the input aligned files instead. Do you see a problem with this idea for a fix?

tfenne:
@yfarjoun I think that will create problems unfortunately.  Unless bwa is pulling through the other fields that are in the dictionary, which seems unlikely.  The resulting BAMs would likely have `@SQ` lines that lost their `MD`, `AS` and `UR` tags, which would kinda suck.  Maybe the thing to do is to merge the sequence dictionaries and take the union of tags between them?

yfarjoun:
Totally suck. Yes. it seems that bwa only picks up LN....OK. merging the
tags. Thanks for the comment.

yfarjoun:
will be fixed by #682

PR #627, opened 2016-08-14
Make MergeBamAlignment honor tags in aligned input

yfarjoun:
- Changing MergeBamAlignment so that it merges the dictionaries of the aligned bam and the reference instead of taking the reference dict file. This is since bwa-mem (latests versions) add a AH:\* to the sequence records of "Alternate Haplotypes" and this is lost in translation otherwise.
- testing that this works.
- fixes issue #625


yfarjoun:
@tfenne would you be available to review this?

src/main/java/picard/sam/SamAlignmentMerger.java
```diff
@@ -158,6 +141,70 @@ public void mergeAlignment(final File referenceFasta) {
         }
     }
 
+    @Override
+    protected SAMSequenceDictionary getDictionaryForMergedBam() {
+        SAMSequenceDictionary finalDict = null;
+        if (alignedSamFile != null && !alignedSamFile.isEmpty()) {
+            for (final File alignedFile : alignedSamFile) {
```
tfenne: Couldn't you just `map()` to dictionaries, then `foreach()` to compare each to the first dictionary, in ~2 lines?

src/main/java/picard/sam/SamAlignmentMerger.java
```diff
@@ -158,6 +141,70 @@ public void mergeAlignment(final File referenceFasta) {
...
+                        throw new UnsupportedOperationException("Cannot use input files with different dictionaries.");
+                    }
+                }
+            }
+        } else {
+            final SeparateEndAlignmentIterator mergingIterator = new SeparateEndAlignmentIterator(this.read1AlignedSamFile, this.read2AlignedSamFile, referenceFasta);
+            finalDict = mergingIterator.getHeader().getSequenceDictionary();
+        }
+
+        SAMSequenceDictionary referenceDict = SAMSequenceDictionaryExtractor.extractDictionary(referenceFasta);
```
tfenne: I would be tempted to do this _first_ and then compare the other dictionaries from the SAM files to this one.

src/main/java/picard/sam/SamAlignmentMerger.java
```diff
@@ -158,6 +141,70 @@ public void mergeAlignment(final File referenceFasta) {
...
+        if (referenceDict == null) {
+            throw new PicardException("No sequence dictionary found for " + referenceFasta.getAbsolutePath() +
+                    ".  Use CreateSequenceDictionary.jar to create a sequence dictionary.");
+        }
+        return mergeDictionaries(finalDict, SAMSequenceDictionaryExtractor.extractDictionary(referenceFasta));
+    }
+
+    private SAMSequenceDictionary mergeDictionaries(SAMSequenceDictionary dict1, SAMSequenceDictionary dict2) {
```
tfenne: Would it make sense to have this be a static method in SAMSequenceDictionary instead?

src/main/java/picard/sam/SamAlignmentMerger.java
```diff
@@ -158,6 +141,70 @@ public void mergeAlignment(final File referenceFasta) {
...
+        for (final SAMSequenceRecord sequence : finalDict.getSequences()) {
+            final int sequenceIndex = sequence.getSequenceIndex();
+            final Set<String> allTags = new HashSet<>();
+
+            for (SAMSequenceRecord seq : Arrays.asList(
+                    dict1.getSequence(sequenceIndex),
+                    dict2.getSequence(sequenceIndex))) {
+
+                final Set<String> dictTags = seq
+                        .getAttributes().parallelStream()
```
tfenne: Using a parallelStream here is likely a) slower and b) going to cause MBA to use unexpectedly more CPU cores.  Doing a hashmap lookup is almost certainly faster than switching threads.

src/main/java/picard/sam/SamAlignmentMerger.java
```diff
@@ -158,6 +141,70 @@ public void mergeAlignment(final File referenceFasta) {
...
+                        .map(Map.Entry::getKey)
+                        .collect(Collectors.toSet());
+                allTags.addAll(dictTags);
+            }
+
+            for (final String tag : allTags) {
+                final String value1 = dict1.getSequence(sequenceIndex).getAttribute(tag);
+                final String value2 = dict2.getSequence(sequenceIndex).getAttribute(tag);
+
+                if (value1 != null && value2 != null && !value1.equals(value2)) {
```
tfenne: This worries me a little.  I wonder if this shouldn't also check for VALIDATION_STRINGENCY == STRICT?  What if, e.g., the one next to the reference has a `UR` tag in pointing to the original source and an aligner inserts a `UR` tag pointing at the reference file it used?  They'll be different and then no more merging for you.

src/main/java/picard/sam/SamAlignmentMerger.java
```diff
@@ -158,6 +141,70 @@ public void mergeAlignment(final File referenceFasta) {
...
+            throw new PicardException("No sequence dictionary found for " + referenceFasta.getAbsolutePath() +
+                    ".  Use CreateSequenceDictionary.jar to create a sequence dictionary.");
+        }
+        return mergeDictionaries(finalDict, SAMSequenceDictionaryExtractor.extractDictionary(referenceFasta));
+    }
+
+    private SAMSequenceDictionary mergeDictionaries(SAMSequenceDictionary dict1, SAMSequenceDictionary dict2) {
+
+        final SAMSequenceDictionary finalDict = new SAMSequenceDictionary(dict1.getSequences());
+        for (final SAMSequenceRecord sequence : finalDict.getSequences()) {
```
tfenne: So you make a copy of the dictionary, but then go through and modify the sequences in the dictionary, effectively modifying the source dictionary entries.  Why not just construct new records in the loop since you're setting all the attributes anyway?

src/main/java/picard/sam/SamAlignmentMerger.java
```diff
@@ -158,6 +141,70 @@ public void mergeAlignment(final File referenceFasta) {
...
+            final int sequenceIndex = sequence.getSequenceIndex();
+            final Set<String> allTags = new HashSet<>();
+
+            for (SAMSequenceRecord seq : Arrays.asList(
```
tfenne: This whole loop to get the unique keys could be:
```java
final Set<String> tags = new HashSet<>(dict1.getAttributes().stream().map(e -> e.getKey()).collect(Collectors.toSet());
tags.addAll(dict2.getAttributes().stream().map(e -> e.getKey()).collect(Collectors.toSet())

File: src/main/java/picard/sam/SamAlignmentMerger.java [hunk 8]
```diff
@@ -370,4 +417,37 @@ protected boolean isContaminant(final HitsForInsert hits) {
...
+                compareValues(sequenceRecord1.getSequenceLength(), sequenceRecord2.getSequenceLength()) &&
+                compareValues(sequenceRecord1.getSpecies(), sequenceRecord2.getSpecies()) &&
+                compareValues(sequenceRecord1.getAssembly(), sequenceRecord2.getAssembly()) &&
+                compareValues(sequenceRecord1.getAttribute("M5"), sequenceRecord2.getAttribute("M5")) &&
+                compareValues(sequenceRecord1.getAttribute("UR"), sequenceRecord2.getAttribute("UR")) &&
+                compareValues(sequenceRecord1.getAttribute("AH"), sequenceRecord2.getAttribute("AH"));
+
+    }
+
+    private <T> boolean compareValues(final T v1, final T v2) {
```
tfenne: `Objects.equals()` does this.

src/main/java/picard/sam/SamAlignmentMerger.java
```diff
@@ -370,4 +417,37 @@ protected boolean isContaminant(final HitsForInsert hits) {
 
     // Accessor for testing
     public boolean getForceSort() {return this.forceSort; }
+
+    private boolean compareSequenceDictionaries(final SAMSequenceDictionary d1, final SAMSequenceDictionary d2) {
```
tfenne: I think this would be better off in `SAMSequenceDictionary` also.  There are already methods there for comparing sequence dictionaries and it would be nice to keep them all in one place.

tfenne:
@yfarjoun Back to you.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -285,6 +253,7 @@ public void mergeAlignment(final File referenceFasta) {
         final SamReader unmappedSam = SamReaderFactory.makeDefault().referenceSequence(referenceFasta).open(this.unmappedBamFile);
 
         final CloseableIterator<SAMRecord> unmappedIterator = unmappedSam.iterator();
+        this.header.setSequenceDictionary(getDictionaryForMergedBam());
```
yfarjoun: I  had to move it from there since the input isn't yet known at that point.

src/main/java/picard/sam/SamAlignmentMerger.java
yfarjoun: sure.

src/main/java/picard/sam/SamAlignmentMerger.java
yfarjoun: ok. thanks.

src/main/java/picard/sam/SamAlignmentMerger.java
yfarjoun: would it be better to have a special case for UR?

src/main/java/picard/sam/SamAlignmentMerger.java
yfarjoun: :blush: I copied the code from somewhere else and removed the messaging part...

src/main/java/picard/sam/SamAlignmentMerger.java
yfarjoun: good point. htsjdk! here we come.

yfarjoun:
thanks @tfenne

src/main/java/picard/sam/SamAlignmentMerger.java
tfenne: I'd be more inclined to go the opposite way and assert that the name, LN and MD tags that really should match, and just warn about mismatches on other attributes.

src/main/java/picard/sam/SamAlignmentMerger.java
yfarjoun: yes, this was a bug.

src/main/java/picard/sam/SamAlignmentMerger.java
yfarjoun: Not sure I understand this comment. I cannot compare the SAM file dictionaries to the reference one since they might be different, but I do require that all the SAM file dictionaries are the same, and then I merge one of them with the reference dict....I could reverse the order (check that the reference dictionary exisits before checking that the sam files are the same), if that's what you mean.

yfarjoun:
closing in favor of #682

PR #628, opened 2016-08-15
DuplicationMetrics are now mergeable.

nh13:
@yfarjoun want to take a look since you wrote `MergeableMetricBase`? 

Also, I am not sure if it is too late, but the metric names in `IndependentReplicateMetric` (the first class to use `MergeableMetricBase`) are camel-case rather than upper-case+underscore, with the latter being the convention of all other picard metrics that I know of.

### DIFF
--- a/src/main/java/picard/analysis/MergeableMetricBase.java
+++ b/src/main/java/picard/analysis/MergeableMetricBase.java
@@ -20,8 +20,9 @@
  * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
+ *
  */
-package picard.analysis.replicates;
+package picard.analysis;
 
 import htsjdk.samtools.metrics.MetricBase;
 
--- a/src/main/java/picard/sam/DuplicationMetrics.java
+++ b/src/main/java/picard/sam/DuplicationMetrics.java
@@ -24,59 +24,81 @@
 
 package picard.sam;
 
-import htsjdk.samtools.metrics.MetricBase;
 import htsjdk.samtools.util.Histogram;
+import picard.analysis.MergeableMetricBase;
 
 /**
  * Metrics that are calculated during the process of marking duplicates
  * within a stream of SAMRecords.
  */
-public class DuplicationMetrics extends MetricBase {
+public class DuplicationMetrics extends MergeableMetricBase {
     /** The library on which the duplicate marking was performed. */
+    @MergeByAssertEquals
     public String LIBRARY;
 
     /**
      * The number of mapped reads examined which did not have a mapped mate pair,
      * either because the read is unpaired, or the read is paired to an unmapped mate.
      */
+    @MergeByAdding
     public long UNPAIRED_READS_EXAMINED;
 
     /** The number of mapped read pairs examined. (Primary, non-supplemental) */
+    @MergeByAdding
     public long READ_PAIRS_EXAMINED;
 
     /** The number of reads that were either secondary or supplementary */
+    @MergeByAdding
     public long SECONDARY_OR_SUPPLEMENTARY_RDS;
 
     /** The total number of unmapped reads examined. (Primary, non-supplemental) */
+    @MergeByAdding
     public long UNMAPPED_READS;
 
     /** The number of fragments that were marked as duplicates. */
+    @MergeByAdding
     public long UNPAIRED_READ_DUPLICATES;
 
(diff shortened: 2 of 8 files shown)


src/main/java/picard/sam/DuplicationMetrics.java
```diff
@@ -24,62 +24,84 @@
 
 package picard.sam;
 
-import htsjdk.samtools.metrics.MetricBase;
 import htsjdk.samtools.util.Histogram;
+import picard.analysis.replicates.MergeableMetricBase;
```
yfarjoun: if MergableMetrics is going to be used more broadly, I think we should move it to htsjdk. Or atleast out of the  replicate package...

yfarjoun:
regarding CamelCasing in replicates...you are right. I don't think it's too late as it's marked "ALPHA"...that was the point of doing that...right?

yfarjoun:
it's not entirely clear why you are doing this...is there going to be a scatter-gather functionality to MarkDuplicates soon?? that would be really nice... 😄

src/main/java/picard/sam/DuplicationMetrics.java
```diff
@@ -24,62 +24,84 @@
...
     /** The estimated number of unique molecules in the library based on PE duplication. */
+    @NoMergingIsDerived
     public Long ESTIMATED_LIBRARY_SIZE;
 
     /**
      * Fills in the ESTIMATED_LIBRARY_SIZE based on the paired read data examined where
      * possible and the PERCENT_DUPLICATION.
      */
-    public void calculateDerivedMetrics() {
+    @Override
```
yfarjoun: The java-doc from the deprecated version is useful here, no?

src/main/java/picard/sam/DuplicationMetrics.java
```diff
@@ -24,62 +24,84 @@
...
     }
 
     /**
+     * Fills in the ESTIMATED_LIBRARY_SIZE based on the paired read data examined where
+     * possible and the PERCENT_DUPLICATION.
+     *
+     * Deprecated, use {@link #calculateDerivedFields()} instead.
+     */
+    @Deprecated
+    public void calculateDerivedMetrics() {
```
yfarjoun: Can you change the few other places (in picard) where the newly deprecated function is called?

yfarjoun:
a few small comments, and it's good to go!

src/main/java/picard/sam/DuplicationMetrics.java
nh13: Will move to `picard.analysis`.  Moving to htsjdk is a good idea, but lets get a few more uses of this class under our belt so we feel good about what it provides.

src/main/java/picard/sam/DuplicationMetrics.java
nh13: There there, (lines 81-84).

src/main/java/picard/sam/DuplicationMetrics.java
nh13: Done.

nh13:
@yfarjoun the motivation for this PR (sorry for not including it) is to enable `MarkDuplicates` to be run independently on different sets of reads from the same library and then be able to merge the output metric files as though `MarkDuplicates` was run across all reads.

nh13:
@yfarjoun fixups complete, and see my comment about moving the `MergeableMetricBase` to htsjdk a little later once we have used it a few times successfully.  One final 👍  and I will squash and merge.

src/main/java/picard/sam/DuplicationMetrics.java
yfarjoun: oh, oops.

src/main/java/picard/analysis/MergeableMetricBase.java
```diff
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2016 The Broad Institute
+ * Copyright (c) 2016 Fulcrum Genomics
```
yfarjoun: I don't think so.

yfarjoun:
you moved the MergeableMetricsBase in a not-so-careful way and broke the relevant tests...please rename using IntelliJ ... 😄

src/main/java/picard/analysis/MergeableMetricBase.java
nh13: Silly intellij, thanks for spotting this.

nh13:
@yfarjoun I actually moved with Intellij if you can believe it.

nh13:
@yfarjoun one last look (I have squashed).

nh13:
@yfarjoun I am taking your earlier "a few small comments, and it's good to go" to mean I am good to go.

PR #629, opened 2016-08-15
Make WgsMetrics mergeable.

nh13:
@yfarjoun want to take a look?

### DIFF
--- a/src/main/java/picard/analysis/CollectWgsMetrics.java
+++ b/src/main/java/picard/analysis/CollectWgsMetrics.java
@@ -24,22 +24,16 @@
 package picard.analysis;
 
 import htsjdk.samtools.SAMFileHeader;
+import htsjdk.samtools.SAMSequenceRecord;
 import htsjdk.samtools.SamReader;
 import htsjdk.samtools.SamReaderFactory;
 import htsjdk.samtools.filter.SamRecordFilter;
 import htsjdk.samtools.filter.SecondaryAlignmentFilter;
-import htsjdk.samtools.metrics.MetricBase;
 import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.reference.ReferenceSequence;
 import htsjdk.samtools.reference.ReferenceSequenceFileWalker;
-import htsjdk.samtools.util.Histogram;
-import htsjdk.samtools.util.IOUtil;
-import htsjdk.samtools.util.IntervalList;
-import htsjdk.samtools.util.Log;
-import htsjdk.samtools.util.ProgressLogger;
-import htsjdk.samtools.util.QualityUtil;
-import htsjdk.samtools.util.SamLocusIterator;
-import htsjdk.samtools.util.SequenceUtil;
+import htsjdk.samtools.util.*;
+import picard.PicardException;
 import picard.cmdline.CommandLineProgram;
 import picard.cmdline.CommandLineProgramProperties;
 import picard.cmdline.Option;
@@ -55,7 +49,6 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
-import java.util.stream.IntStream;
 
 import static picard.cmdline.StandardOptionDefinitions.MINIMUM_MAPPING_QUALITY_SHORT_NAME;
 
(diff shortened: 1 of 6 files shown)


src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,241 @@
...
+    public static class WgsMetrics extends MergeableMetricBase {
+
+        @NoMergingIsDerived
+        protected final Histogram<Integer> depthHistogram;
+
+        @NoMergingIsDerived
+        protected final Histogram<Integer> baseQHistogram;
+
+        @NoMergingIsDerived
+        protected final int coverageCap;
```
yfarjoun: shouldn't this be "assertEqual"?

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,241 @@
     private static final double LOG_ODDS_THRESHOLD = 3.0;
 
     /** Metrics for evaluating the performance of whole genome sequencing experiments. */
-    public static class WgsMetrics extends MetricBase {
+    public static class WgsMetrics extends MergeableMetricBase {
+
+        @NoMergingIsDerived
```
yfarjoun: you are abusing the annotation here...I guess we should just rename the annotation "NoMerging" since in this case, it's not actually derived but rather manually merged.

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,241 @@
...
+            final double totalWithExcludes    = thisTotalWithExcludes + otherTotalWithExcludes;
+
+            if (0 < totalWithExcludes) PCT_EXC_DUPE     = (PCT_EXC_DUPE * thisTotalWithExcludes + otherMetric.PCT_EXC_DUPE * otherTotalWithExcludes) / totalWithExcludes;
+            if (0 < totalWithExcludes) PCT_EXC_MAPQ     = (PCT_EXC_MAPQ * thisTotalWithExcludes + otherMetric.PCT_EXC_MAPQ * otherTotalWithExcludes) / totalWithExcludes;
+            if (0 < totalWithExcludes) PCT_EXC_UNPAIRED = (PCT_EXC_UNPAIRED * thisTotalWithExcludes + otherMetric.PCT_EXC_UNPAIRED * otherTotalWithExcludes) / totalWithExcludes;
+            if (0 < totalWithExcludes) PCT_EXC_BASEQ    = (PCT_EXC_BASEQ * thisTotalWithExcludes + otherMetric.PCT_EXC_BASEQ * otherTotalWithExcludes) / totalWithExcludes;
+            if (0 < totalWithExcludes) PCT_EXC_OVERLAP  = (PCT_EXC_OVERLAP * thisTotalWithExcludes + otherMetric.PCT_EXC_OVERLAP * otherTotalWithExcludes) / totalWithExcludes;
+            if (0 < totalWithExcludes) PCT_EXC_CAPPED   = (PCT_EXC_CAPPED * thisTotalWithExcludes + otherMetric.PCT_EXC_CAPPED * otherTotalWithExcludes) / totalWithExcludes;
```
yfarjoun: there could be new capped values now...you need to add the histograms first and then cap the histogram itself.

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,241 @@
...
+            // Thus:  totalWithExcludes = total / (1 - PCT_EXC_TOTAL)
+            // Proof: Exercise is left to the reader.
+            final long thisMetricTotal        = (long) depthHistogram.getSum();
+            final long otherMetricTotal       = (long) otherMetric.depthHistogram.getSum();
+            final long total                  = thisMetricTotal + otherMetricTotal;
+            final long thisTotalWithExcludes  = (long) (thisMetricTotal / (1.0 - PCT_EXC_TOTAL));
+            final long otherTotalWithExcludes = (long) (otherMetricTotal / (1.0 - otherMetric.PCT_EXC_TOTAL));
+            final double totalWithExcludes    = thisTotalWithExcludes + otherTotalWithExcludes;
+
+            if (0 < totalWithExcludes) PCT_EXC_DUPE     = (PCT_EXC_DUPE * thisTotalWithExcludes + otherMetric.PCT_EXC_DUPE * otherTotalWithExcludes) / totalWithExcludes;
```
yfarjoun: shouldn't these lines (309-315) be in calculateDerivedFields?

yfarjoun: I guess it can't be done currently. 

what I had envisioned for cases like this is that we would have private or protected members (so they do not get output in the metrics) that are `@MergedByAdding`. I think I have an implementation of that somewhere, but it's dependent on some changes that I didn't get around to doing (yet!!)...

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,241 @@
...
+        protected final Histogram<Integer> depthHistogram;
+
+        @NoMergingIsDerived
+        protected final Histogram<Integer> baseQHistogram;
+
+        @NoMergingIsDerived
+        protected final int coverageCap;
+
+        @NoMergingIsDerived
+        protected final int sampleSize;
```
yfarjoun: Same here. should be a global parameter, thus should be the same for all metrics.

yfarjoun: Also, could you use a more informative name for this, it's a bit misleading (I know the docs explain it)

src/main/java/picard/analysis/CollectWgsMetrics.java
yfarjoun: is there a reason not to have a single if statement here?

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,241 @@
...
+            if (0 < totalWithExcludes) PCT_EXC_OVERLAP  = (PCT_EXC_OVERLAP * thisTotalWithExcludes + otherMetric.PCT_EXC_OVERLAP * otherTotalWithExcludes) / totalWithExcludes;
+            if (0 < totalWithExcludes) PCT_EXC_CAPPED   = (PCT_EXC_CAPPED * thisTotalWithExcludes + otherMetric.PCT_EXC_CAPPED * otherTotalWithExcludes) / totalWithExcludes;
+            if (0 < totalWithExcludes) PCT_EXC_TOTAL    = (totalWithExcludes - total) / totalWithExcludes;
+
+            // this is currently not necessary, but here in case any new fields are added.
+            super.merge(other);
+
+            // merge the histograms
+            this.depthHistogram.addHistogram(otherMetric.depthHistogram);
+            if (baseQHistogram != null && otherMetric.baseQHistogram != null) this.baseQHistogram.addHistogram(otherMetric.baseQHistogram);
```
yfarjoun: why the distinction between depthHistogram and BaseQHistogram?

src/main/java/picard/analysis/CollectWgsMetrics.java
yfarjoun: forget this comment...I understand now...

yfarjoun: The intervals over which you are scattering have to be distinct. This need to be written in a comment somewhere. Perhaps in the special constructors?

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,241 @@
...
+            if (0 < totalWithExcludes) PCT_EXC_BASEQ    = basesExcludedByBaseq        / totalWithExcludes;
+            if (0 < totalWithExcludes) PCT_EXC_OVERLAP  = basesExcludedByOverlap      / totalWithExcludes;
+            if (0 < totalWithExcludes) PCT_EXC_CAPPED   = basesExcludedByCapping      / totalWithExcludes;
+            if (0 < totalWithExcludes) PCT_EXC_TOTAL    = (totalWithExcludes - total) / totalWithExcludes;
+
+            calculateDerivedFields();
+        }
 
         /** The number of non-N bases in the genome reference over which coverage will be evaluated. */
+        @NoMergingIsDerived
```
yfarjoun: This can be mergeByAdding, no?

yfarjoun: it doesn't have to be, but it clarifies (perhaps) that the different teritorries have to be non-overlapping

src/main/java/picard/analysis/CollectWgsMetrics.java
yfarjoun: would it be onerous to include an IntervalList for each metric and make sure that they have zero intersection before merging? I worry that folks will try to collect this metric over random subsets of the data rather than distinct regions...

yfarjoun:
tests are failing...review complete.

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: It's derived in the the `merge` method but not in the `calculateDerivedFields`.  Perhaps `@ManualMerging` instead?

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: I could turn my sample size really low since I don't care about baseq until I run the final `calculateDerivedFields`.  Het sensitivity is slow for larger sample sizes, especially when not needed.  I'll add a `@NoMerging` annotation.

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: It's calculated in `calculateDerivedFields`, which can be called post-merging as well as normally within the tool, so I thought it best to let `calculateDerivedFields` handle it and annotate it `NoMergingIsDerived`.

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: The private/protected members would need to be written to disk, and so would be exposed to the end-user, which is ugly.  My use case is collecting the metrics across multiple computers, so I want to only utilize what we have stored on disk.  In this case, we need to do some custom merging.

About the if statement, yeah, that can be done 👍

nh13: Also, we could add support for merging histograms (`hist.addHistogram(otherHisto)`) but I am not going to tackle that in this PR.

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: `MergeableMetricBase` can't access protected members.

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: Done.

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: `INCLUDE_BQ_HISTOGRAM` is `false` by default, so may not be calculated or in the resulting metrics file.

nh13:
@yfarjoun thanks for the review, I think there should be some discussion about exactly which annotations  `MergeableMetricBase` should include.  Please take a look in the updated commits.

nh13:
@yfarjoun I wrote a tool locally that merges these metrics, but the only catch is that the interval list accompanying each metrics file needs to exist separately.  I am not in favor of adding in the interval list into the metrics file, as you should have had them for each invocation of `CollectWgsMetrics`. Thanks for a final review!

src/main/java/picard/analysis/MergeableMetricBase.java
```diff
@@ -45,18 +45,33 @@
...
 
+    /** Metrics that are not merged, but are subsequently derived from other metrics, for example by
+     * {@link #calculateDerivedFields()}. */
     @Retention(RetentionPolicy.RUNTIME)
     @Target(ElementType.FIELD)
     protected @interface NoMergingIsDerived {}
 
+    /** Metrics that are merged manually in the {@link #merge(MergeableMetricBase)} ()}. Typically these metrics need
+     * access to both metrics being merged. */
+    @Retention(RetentionPolicy.RUNTIME)
```
yfarjoun: what is the functional difference between these two annotations?

src/test/java/picard/analysis/TheoreticalSensitivityTest.java
```diff
@@ -128,7 +128,7 @@ public void testCentralLimitTheorem() throws Exception {
         //test whether the number of elements within one standard deviation agrees with the normal distribution
         final List<Double> thresholds = Arrays.asList(mu - sigma, mu + sigma);
 
-        //sums is 1 x sampleSize, thresholds is a 2-vector, so proportions is 1 x 2
+        //sums is 1 x theoreticalHetSensitivitySampleSize, thresholds is a 2-vector, so proportions is 1 x 2
```
yfarjoun: I don't think this should have changed here. since you still have sampleSize in the code above.

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -275,12 +468,90 @@ protected int doWork() {
...
+                                            final long basesExcludedByBaseq,
+                                            final long basesExcludedByOverlap,
+                                            final long basesExcludedByCapping,
+                                            final int coverageCap,
+                                            final Histogram<Integer> baseQHistogram,
+                                            final int theoreticalHetSensitivitySampleSize) {
+        final double total              = depthHistogram.getSum();
+        final double totalWithExcludes  = total + basesExcludedByDupes + basesExcludedByMapq + basesExcludedByPairing + basesExcludedByBaseq + basesExcludedByOverlap + basesExcludedByCapping;
+
+        final double pctExcludedByMapq    = (0 == totalWithExcludes) ? 0 : (basesExcludedByMapq         / totalWithExcludes);
```
yfarjoun: you have a dislike to a single if statement?

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,268 @@
...
+                PCT_EXC_DUPE     = (PCT_EXC_DUPE * thisTotalWithExcludes + otherMetric.PCT_EXC_DUPE * otherTotalWithExcludes) / totalWithExcludes;
+                PCT_EXC_MAPQ     = (PCT_EXC_MAPQ * thisTotalWithExcludes + otherMetric.PCT_EXC_MAPQ * otherTotalWithExcludes) / totalWithExcludes;
+                PCT_EXC_UNPAIRED = (PCT_EXC_UNPAIRED * thisTotalWithExcludes + otherMetric.PCT_EXC_UNPAIRED * otherTotalWithExcludes) / totalWithExcludes;
+                PCT_EXC_BASEQ    = (PCT_EXC_BASEQ * thisTotalWithExcludes + otherMetric.PCT_EXC_BASEQ * otherTotalWithExcludes) / totalWithExcludes;
+                PCT_EXC_OVERLAP  = (PCT_EXC_OVERLAP * thisTotalWithExcludes + otherMetric.PCT_EXC_OVERLAP * otherTotalWithExcludes) / totalWithExcludes;
+                PCT_EXC_CAPPED   = (PCT_EXC_CAPPED * thisTotalWithExcludes + otherMetric.PCT_EXC_CAPPED * otherTotalWithExcludes) / totalWithExcludes;
+                PCT_EXC_TOTAL    = (totalWithExcludes - total) / totalWithExcludes;
+            }
+
+            // this is currently not necessary, but here in case any new fields are added.
```
yfarjoun: already needed in the derived class.

yfarjoun: also genome territory.

src/main/java/picard/analysis/MergeableMetricBase.java
yfarjoun: I guess `@NoMerging` should be something like `@NoMergingKeepsValue`

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,268 @@
...
+        @MergingIsManual
+        protected final Histogram<Integer> depthHistogram;
+
+        /** The count of bases observed with a given base quality. */
+        @MergingIsManual
+        protected final Histogram<Integer> baseQHistogram;
+
+        /** The maximum depth/coverage to consider. */
+        @NoMerging
+        protected final int coverageCap;
```
yfarjoun: for sanity's sake, this should be equal, no?

yfarjoun: meaning MergeByAdding

yfarjoun: argh. MergeByAssertEquals is what I really mean.

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,268 @@
     private static final double LOG_ODDS_THRESHOLD = 3.0;
 
     /** Metrics for evaluating the performance of whole genome sequencing experiments. */
-    public static class WgsMetrics extends MetricBase {
+    public static class WgsMetrics extends MergeableMetricBase {
+
+        /** The intervals over which this metric was computed. */
+        @NoMerging
+        protected IntervalList intervals;
```
yfarjoun: I"m confused. didn't you say that you do not want to include this here? 

looks like you are merging this manually...so why "NoMerging"?

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,268 @@
...
+                throw new PicardException("Sample size is required when a baseQ histogram is given when deriving metrics.");
+            }
+
+            final long[] depthHistogramArray = new long[coverageCap+1];
+            for (final Histogram.Bin<Integer> bin : depthHistogram.values()) {
+                final int depth = Math.min((int) bin.getIdValue(), coverageCap);
+                depthHistogramArray[depth] += bin.getValue();
+            }
+
+            GENOME_TERRITORY = (long) depthHistogram.getSumOfValues();
```
yfarjoun: so this is derived, not MergeByAdding...

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,268 @@
...
+            PCT_EXC_BASEQ    = pctExcludedByBaseq;
+            PCT_EXC_OVERLAP  = pctExcludedByOverlap;
+            PCT_EXC_CAPPED   = pctExcludedByCapping;
+            PCT_EXC_TOTAL    = pctExcludeTotal;
+
+            calculateDerivedFields();
+        }
 
         /** The number of non-N bases in the genome reference over which coverage will be evaluated. */
+        @MergeByAdding
```
yfarjoun: seems to be derived.

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,268 @@
...
+                throw new PicardException("Depth histogram is required when deriving metrics.");
+            }
+
+            // Union the intervals over which bases are called.  They should have no overlaps!
+            // NB: interval lists are already uniqued.
+            final long genomeTerritory = this.intervals.getBaseCount() + otherMetric.intervals.getBaseCount();
+            this.intervals.addall(otherMetric.intervals.getIntervals());
+            this.intervals = this.intervals.uniqued();
+            if (this.intervals.getBaseCount() != genomeTerritory) {
+                throw new PicardException("Trying to merge metrics calculated on intervals that overlap.");
```
yfarjoun: s/metrics/WgsMetrics/

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,268 @@
...
+            }
+
+            if (coverageCap != otherMetric.coverageCap) {
```
yfarjoun: if you `@MergeByAssertEquals` you can avoid this.

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -136,68 +129,268 @@
...
+
+            PCT_EXC_MAPQ     = pctExcludedByMapq;
+            PCT_EXC_DUPE     = pctExcludedByDupes;
+            PCT_EXC_UNPAIRED = pctExcludedByPairing;
+            PCT_EXC_BASEQ    = pctExcludedByBaseq;
+            PCT_EXC_OVERLAP  = pctExcludedByOverlap;
+            PCT_EXC_CAPPED   = pctExcludedByCapping;
+            PCT_EXC_TOTAL    = pctExcludeTotal;
+
+            calculateDerivedFields();
```
yfarjoun: do you need this in every constructor?

yfarjoun:
I"m confused about your last comment. you say that you are not in favor, but you put the intervalList in the metric...are you asking to remove it? I think that the sanity checks that it enables are good....

yfarjoun:
back you you @nh13

nh13:
@yfarjoun I am happy to have the interval list be a protected field in the `WgsMetrics` class.  I do not favor having the intervals printed to the resulting metrics file (on disk).  Hopefully that makes sense.

src/main/java/picard/analysis/MergeableMetricBase.java
nh13: Changed `@NoMerging` to `@NoMergingKeepsValue`

src/test/java/picard/analysis/TheoreticalSensitivityTest.java
nh13: IntelliJ ftw.

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: I can declare and set them in one line 👍 .

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: 👍

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: Good catch, I'll use the `MergingIsManual` annotation.

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: Fixed.

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: Fixed.

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: Updated.

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: Fixed.

src/main/java/picard/analysis/CollectWgsMetrics.java
nh13: The sub-classes call this constructor, so it is not necessary in any other constructor, unless that constructor does not call this constructor, which only happens when `depthHistogram` is not available.

nh13:
@yfarjoun back to you. I incorporated all your changes.

yfarjoun:
thanks for this @nh13. I hope that a tool for merging WGS metrics is in the works...but if not, we can write one too!

yfarjoun:
👍

nh13:
@yfarjoun thanks for the guidance here. I am rebased and am going to wait for tests to pass.

PR #633, opened 2016-08-19
Added the ability to include duplicate reads in CollectSequencingArtifactMetrics

tfenne:
Off by default, but I wanted to be able to view all reads vs. duplicates and didn't really want to have to rewrite the entire BAM just to toggle the duplicate flags.  I found this useful because I use CSAM not only to detect artifact metrics but as a general source of information on error rates after screening out bases by quality and from sites thought to be variant.


tfenne:
Another tiny PR for your @yfarjoun .

yfarjoun:
👍  after tests pass.

tfenne:
Thanks!  While I have your attention ... would you be interested in a PR that produced another output file for CSAM that just gave rolled up error rates by base substitution type?  I.e. a rate for A>C, A>G, A>T, etc.?  Right now I'm post-processing to create those, but I'm finding them very useful.

yfarjoun:
That would be nice. Could that go in the summary file or would it be a
different file?

tfenne:
Likely a different file I think, as it would make sense to output a small number of fields per error type.

yfarjoun:
sounds good.

PR #639, opened 2016-08-26
Added check for REFERENCE_SEQUENCE to avoid NPE in CollectSequencingA…

tfenne:
…rtifactMetrics


tfenne:
@yfarjoun This has been driving me crazy!  Because it inherits REFERENCE_SEQUENCE from a parent class, and it's optional up there, CSA doesn't give an error message if it's null, it just NPEs.  Trivial change to add a check for it and produce a nicer error message.

yfarjoun:
👍

nh13:
@yfarjoun I think I was assigned this in error.

yfarjoun:
Oops. Thanks.

PR #643, opened 2016-09-02
Modified CheckFingerprint to be able to take a VCF as INPUT

gbggrant:
Modified CheckFingerprint to be able to take a VCF as INPUT (as opposed to a SAM/BAM).
Added specialized method in FingerprintChecker for that purpose
Generalized certain methods to handle VCF vs BAM/SAM as input.


src/main/java/picard/cmdline/programgroups/Fingerprinting.java
```diff
@@ -0,0 +1,12 @@
+package picard.cmdline.programgroups;
+
+import picard.cmdline.CommandLineProgramGroup;
+/**
+ * @author nhomer
```
yfarjoun: @nhomer?

src/main/java/picard/cmdline/programgroups/Fingerprinting.java
```diff
@@ -0,0 +1,12 @@
+package picard.cmdline.programgroups;
```
yfarjoun: do we normally license these files?

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -33,48 +34,51 @@
...
 )
 public class CheckFingerprint extends CommandLineProgram {
 
-    static final String USAGE_DETAILS = "Computes a fingerprint from the supplied SAM/BAM file and " +
+    static final String USAGE_DETAILS = "Computes a fingerprint from the supplied input file (SAM/BAM or VCF) file and " +
             "compares it to the expected fingerprint genotypes provided. The key output is a LOD score " +
             "which represents the relative likelihood of the sequence data originating from the same " +
             "sample as the genotypes vs. from a random sample.  Two outputs are produced: (1) a summary " +
-            "metrics file that gives metrics related single read group (lane or index within a lane) " +
+            "metrics file that gives metrics at the single sample level (if the input was a VCF) or at the read " +
```
yfarjoun: vcf's can have more than one sample. please inform here what happens when N!=1 (both 0 and >1)

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -33,48 +34,51 @@
...
+            "level (lane or index within a lane) (if the input was a SAM/BAM) " +
             "versus a set of known genotypes for the expected sample, and (2) a detail metrics file that " +
             "contains an individual SNP/Haplotype comparison within a fingerprint comparison.  The two " +
             "files may be specified individually using the SUMMARY_OUTPUT and DETAIL_OUTPUT options.  " +
             "Alternatively the OUTPUT option may be used instead to give the base of the two output " +
             "files, with the summary metrics having a file extension '" + CheckFingerprint.FINGERPRINT_SUMMARY_FILE_SUFFIX + "' " +
             "and the detail metrics having a file extension '" + CheckFingerprint.FINGERPRINT_DETAIL_FILE_SUFFIX + "'.";
 
-    @Option(shortName=StandardOptionDefinitions.INPUT_SHORT_NAME, doc = "Input SAM or BAM file.")
+    @Option(shortName=StandardOptionDefinitions.INPUT_SHORT_NAME, doc = "Input file SAM/BAM or VCF.  (Single Sample)")
```
yfarjoun: would it be difficult to make it work for a multi-sample vcf while you're at it?

yfarjoun: I see that there are several places where a single EXPECTED_SAMPLE_ALIAS is assumed...still, in the case that a SAMPLE_ALIAS is provided, it could be used for both the  INPUT and the GENOTYPES files.

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -89,10 +93,10 @@
             "any number of genotypes; CheckFingerprint will use only those that are usable for fingerprinting.")
     public File GENOTYPES;
 
-    @Option(optional=true, doc = "If using VCF format genotypes, this parameter can be used to specify which sample's " +
-            "genotypes to use from the VCF file.  If not supplied the sample name from the BAM read group header " +
-            "is used instead.")
-    public String SAMPLE_ALIAS;
+    @Option(optional=true, doc = "This parameter can be used to specify which sample's genotypes to use from the " +
```
yfarjoun: you are breaking backwards compatibility, could you make SAMPLE_ALIAS the short name for the argument?

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -129,37 +134,65 @@ protected int doWork() {
...
-        SequenceUtil.assertSequenceDictionariesEqual(SAMSequenceDictionaryExtractor.extractDictionary(INPUT), checker.getHeader().getSequenceDictionary(), true);
+        String observedSampleAlias = null;
+        final boolean isBamOrSamFile = isBamOrSamFile(INPUT);
+        if (isBamOrSamFile) {
+            SequenceUtil.assertSequenceDictionariesEqual(SAMSequenceDictionaryExtractor.extractDictionary(INPUT), SAMSequenceDictionaryExtractor.extractDictionary(GENOTYPES), true);
+            SequenceUtil.assertSequenceDictionariesEqual(SAMSequenceDictionaryExtractor.extractDictionary(INPUT), checker.getHeader().getSequenceDictionary(), true);
 
-        // If sample alias isn't supplied, assume it's the one from the INPUT file's RGs
-        if (SAMPLE_ALIAS == null) {
+            // Verify that there's only one sample in the SAM/BAM.
```
yfarjoun: this verification should only happen if EXPECTED_SAMPLE_ALIAS is null, right?

yfarjoun: the if statement below needs to be moved up.

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -129,37 +134,65 @@ protected int doWork() {
...
+                    EXPECTED_SAMPLE_ALIAS,
+                    IGNORE_READ_GROUPS);
+        } else {            // Input is a VCF
+            // Note that FingerprintChecker.loadFingerprints() verifies that the VCF's Sequence Dictionaries agree with that of the Haplotye Map File
+
+            // Verify that there is only one sample in the VCF
+            final VCFFileReader fileReader = new VCFFileReader(INPUT, false);
+            final VCFHeader fileHeader = fileReader.getFileHeader();
+            if (fileHeader.getNGenotypeSamples() != 1) {
+                throw new PicardException("INPUT VCF file must not contain data from multiple samples.");
```
yfarjoun: a vcf could also have getNGenotypeSamples()==0...so your error message will be incorrect.

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -217,12 +252,25 @@ else if (!SAMPLE_ALIAS.equals(rec.getSample())) {
...
         }
 
         summaryFile.write(outputSummaryMetricsFile);
         detailsFile.write(outputDetailMetricsFile);
 
         return 0;
     }
+
+    protected String[] customCommandLineValidation() {
+        IOUtil.assertFileIsReadable(INPUT);
```
yfarjoun: oh, I see, you are validating Readability here...

yfarjoun: maybe move all the IO Asserts here for consistency?

yfarjoun: nevermind. just put a comment that INPUT asserted to be Readable in  customCommandLineValidation()

src/main/java/picard/fingerprint/CrosscheckReadGroupFingerprints.java
```diff
@@ -50,7 +50,7 @@
 @CommandLineProgramProperties(
         usage = "Checks if all read groups within a set of BAM files appear to come from the same individual",
```
yfarjoun: actually, this seems to be incorrect: "a set of..." program only takes a single file as input.

src/main/java/picard/fingerprint/FingerprintChecker.java
```diff
@@ -527,7 +526,7 @@ public IntervalList getLociToGenotype(final Collection<Fingerprint> fingerprints
                 final Fingerprint combinedFp = new Fingerprint(specificSample, f, null);
                 fingerprintsByReadGroup.values().forEach(combinedFp::merge);
 
-                final FingerprintResults results = new FingerprintResults(f, specificSample);
+                final FingerprintResults results = new FingerprintResults(f, specificSample, specificSample);
```
yfarjoun: it's a little odd to have both the readgroup and the sample be the sample_alias. wouldn't it be better to have the readgroup be null or the empty string in this case?

src/main/java/picard/fingerprint/FingerprintResults.java
```diff
@@ -29,27 +29,31 @@
...
         this.readGroup = readGroup;
+        this.sampleAlias = sampleAlias;
     }
 
     public void addResults(final MatchResults matchResults) {
         this.matchResults.add(matchResults);
     }
 
-    public File getSamFile() { return samFile; }
+    public File getInputFile() { return inputFile; }
```
yfarjoun: how is this used? what is this for?

src/main/java/picard/fingerprint/FingerprintResults.java
```diff
@@ -29,27 +29,31 @@
...
-    private final String readGroup;
+    private final File inputFile;
+    private final String readGroup;         // null if the input is a VCF.
+    private final String sampleAlias;
     private final SortedSet<MatchResults> matchResults = new TreeSet<>();
 
-    public FingerprintResults(final File samFile, final String readGroup) {
-        this.samFile = samFile;
+    public FingerprintResults(final File inputFile, final String readGroup, final String sampleAlias) {
+        this.inputFile = inputFile;
```
yfarjoun: why do you needs to know the inputFile here?

yfarjoun: nevermind. I see. you were just changing the samFile to inputFile...OK.

src/main/java/picard/fingerprint/FingerprintResults.java
yfarjoun: nm

yfarjoun:
Back to you @gbggrant. 

comments:
0. what does this addition do that crossCheckGenotypes doesn't? is it the creation of a fingerprint metric? Does this supersede CrossCheckGenotypes? should it be deprecated now?
1. A test or two would be good. I can help you come up with some case examples, and a haplotype_database to use
2. other comments in-line

src/main/java/picard/cmdline/programgroups/Fingerprinting.java
gbggrant: no mhre

src/main/java/picard/cmdline/programgroups/Fingerprinting.java
gbggrant: Evidently not, but I'll add the license

src/main/java/picard/fingerprint/CheckFingerprint.java
gbggrant: Yeah - that's a really good point and suggestion (I was considering reverting the name).

src/main/java/picard/fingerprint/CheckFingerprint.java
gbggrant: Yes.  But I envision a case where the sample alias in the input VCF is not the same as that in the genotypes file, so I will supply an OBSERVED_SAMPLE_ALIAS parameter.

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -129,37 +134,65 @@ protected int doWork() {
             outputSummaryMetricsFile = new File(OUTPUT + FINGERPRINT_SUMMARY_FILE_SUFFIX);
         }
 
-        IOUtil.assertFileIsReadable(INPUT);
```
gbggrant: I think I'll put it back even though it is checked in CustomCommandLineValidation - it's a pretty quick check.

src/main/java/picard/fingerprint/CheckFingerprint.java
gbggrant: No.  I still want to validate that there's only one sample in the bam, so that's done in the for loop starting at line 154.  If there's more than one sample in the BAM, then an exception will be thrown, otherwise, observedSampleAlias will be defined.

The if statement at 165 checks if the user set the EXPECTED_SAMPLE_ALIAS to null and if so sets it to the observedSampleAlias I set as a result of the for loop above.

src/main/java/picard/fingerprint/CheckFingerprint.java
gbggrant: True, so true.

src/main/java/picard/fingerprint/CheckFingerprint.java
gbggrant: I put a copy back where it was.  Needed to validate readability here since this is done before doWork is hit.

src/main/java/picard/fingerprint/CrosscheckReadGroupFingerprints.java
gbggrant: Hmm.  Not my code, but looks like it takes a list of bams as an input and uses them

src/main/java/picard/fingerprint/FingerprintChecker.java
gbggrant: Yeah, I agree.  That's what the original code was doing, so I left it that way.  I like your idea of changing it to null

src/main/java/picard/fingerprint/CheckFingerprint.java
gbggrant: done.

gbggrant:
Thanks @yfarjoun I think I have addressed all of your comments. 
0.  This mostly is useful for the arrays workflow where we have an input single-sample VCF (the output of the autocall) and we want to check that against the fingerprint vcf AND generate the metrics file. 
0a.  I don't think that CrossCheckGenotypes can be deprecated because it does an all-pairs comparison and spits out best matches.
1.  I would appreciate some help on examples - particularly on how to handle what haplotype db to use.

src/main/java/picard/fingerprint/CheckFingerprint.java
yfarjoun: got it. so it's the same result, but in a different way. OK.

gbggrant:
@yfarjoun I have added unit tests for this.  Could you please take another look?

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
```diff
@@ -19,6 +21,11 @@
     private final Snp snp = new Snp("test", "chr1", 1, (byte) 'A', (byte) 'C', maf, Collections.singletonList("dummy"));
     private final HaplotypeBlock hb = new HaplotypeBlock(maf);
 
+    private static final double DELTA = 1e-6;
+
+    private static final File TEST_DATA_DIR = new File("testdata/picard/fingerprint/");
+    private static final File TEST_HAPLOTYPE_DATABASE_MAP = new File(TEST_DATA_DIR, "Homo_sapiens_assembly19.haplotype_database.txt");
```
yfarjoun: can you change the name to make it clear that this is a small DB?

gbggrant:
@yfarjoun Have changed vcfs to be uncompressed.  Changed name of Haplotype database file.  Also reverted an earlier change of mine (the addition of the column 'OBSERVED_SAMPLE_ALIAS' to the summary metrics file which I didn't think was needed in retrospect.

testdata/picard/fingerprint/NA12892.fp.vcf
```diff
@@ -0,0 +1,102 @@
...
+##contig=<ID=GL000222.1,length=186861>
+##contig=<ID=GL000200.1,length=187035>
+##contig=<ID=GL000193.1,length=189789>
+##contig=<ID=GL000194.1,length=191469>
+##contig=<ID=GL000225.1,length=211173>
+##contig=<ID=GL000192.1,length=547496>
+##contig=<ID=NC_007605,length=171823>
+##fileDate=Wed Sep 14 12:02:21 EDT 2016
+##reference=file:///seq/references/Homo_sapiens_assembly19/v1/Homo_sapiens_assembly19.fasta
+##source=Genotypes downloaded from gap.broadinstitute.org by DownloadGenotypes.jar
```
yfarjoun: redact?

testdata/picard/fingerprint/NA12891.vcf
```diff
@@ -0,0 +1,102 @@
...
+##INFO=<ID=AC,Number=A,Type=Integer,Description="Allele count in genotypes, for each ALT allele, in the same order as listed">
+##INFO=<ID=AF,Number=A,Type=Float,Description="Allele Frequency, for each ALT allele, in the same order as listed">
+##INFO=<ID=AN,Number=1,Type=Integer,Description="Total number of alleles in called genotypes">
+##INFO=<ID=DP,Number=1,Type=Integer,Description="Approximate read depth; some reads may have been filtered">
+##Illumina manifest file name=PsychChip_v1-1_15073391_A1.csv
```
yfarjoun: needs redacting?

testdata/picard/fingerprint/NA12891.vcf
```diff
@@ -0,0 +1,102 @@
+##fileformat=VCFv4.1
```
yfarjoun: really 4.1?

gbggrant: Really

gbggrant:
Okay @yfarjoun Have redacted information from vcf files.
- FingerprintChecker wants indexed vcfs, and I don't know an easy way to index on the fly, so have left the index files checked in.
- VCF format remains 4.1 - we are creating another ticket to see why this is set and if it is appropriate or can be revved.

yfarjoun:
I think you need to re-write the vcf and enable indexing on the fly for this.

yfarjoun:
you changed the vcfs, did you re-index them?

gbggrant:
Yes, I did re-index them (the binaries were updated).
Will look into creating the index on the fly for the test.

gbggrant:
@yfarjoun have removed the index files and created a utility class to create indexed vcfs on the fly and clean up after.

gbggrant:
Thanks @yfarjoun

PR #644, opened 2016-09-02
Added options to CollectOxoGMetrics and CollectSequencingArtifactMetrics to include/exclude non-PF reads.

tfenne:
I've been trying to tie out results between CollectOxoGMetrics and CollectSequencingArtifactMetrics, which I believe should produce near-identical values for a specific subset of metrics, if given the same inputs.  I was getting results that were different enough (off by Q=3-4 for several contexts) that I tracked down what was going on.  The difference is that CollectOxoGMetrics includes non-PF reads and CollectSequencingArtifactMetrics doesn't.  Neither exposed an option to control this.

I've added options to both, but retained the original behaviour of each by default.

### DIFF
--- a/src/main/java/picard/analysis/CollectOxoGMetrics.java
+++ b/src/main/java/picard/analysis/CollectOxoGMetrics.java
@@ -130,6 +130,9 @@ public class CollectOxoGMetrics extends CommandLineProgram {
             doc = "The maximum insert size for a read to be included in analysis. Set of 0 to allow unpaired reads.")
     public int MAXIMUM_INSERT_SIZE = 600;
 
+    @Option(shortName = "NON_PF", doc = "Whether or not to include non-PF reads.")
+    public boolean INCLUDE_NON_PF_READS = true;
+
     @Option(doc = "When available, use original quality scores for filtering.")
     public boolean USE_OQ = true;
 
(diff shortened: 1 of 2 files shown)


tfenne:
@yfarjoun You might be interested in this.

fleharty:
Could we get a simple test of behavior in CollectSequencingArtifactMetrics since it already has some simple tests?

I don't see any tests for CollectOxoGMetrics at all, do we want to leave it that way?  Or should we consider adding tests to that?

@tfenne back to you

yfarjoun:
While it's true that more tests would be good. I wouldn't gate this simple commit on tests. CollectOxoGMetrics can be considered deprecated as SequencingArtifacts supersedes it, and it is tested. The default behavior regarding the PF issue was (I think) correct and thus the tests were OK. 

@tfenne, do you think that using the non-pf reads is superior?

fleharty:
@tfenne @yfarjoun 
Fair enough, this is a simple PR, and other than the lack of tests it all looks good to me.
👍

tfenne:
@yfarjoun I'm not sure to be honest.  I do see fairly different results in some datasets, which surprised me given the attempt to remove unbiased errors from the calculations.  I mostly just wanted to be able to tie the two out and convince myself there wasn't some bigger reason for the divergence.

On CollectOxoG being effectively deprecated - I've heard feedback from more than one group that it's more difficult to interpret the output of CSAM, and not clear how to map key Q score values from CollectOxoG to values produced by CSAM.  Even I (as the author of CollectOxoG) needed ~1.5+ hours to piece it all together, and even then I still find it kind of confusing!!  Perhaps it would be useful to add an optional extra output to CSAM that re-created the Oxog output file exactly?

yfarjoun:
ConvertSequencingArtifactstoOxoG does this in a commandline program. It
should be fairly easy to take the interesting parts from this CLP and put
them into CSAM...

yfarjoun:
this got 👍  I thought. merge away!

Issue #342, opened 2015-11-18
in GC bias show metrics for deduped and non-deduped reads 


yfarjoun:

nh13:
@yfarjoun could this be another "level" and therefore modify or extend `MultiLevelCollector` and `PerUnitMetricCollector`?  It may then be useful for other tools?

yfarjoun:
Interesting idea. Though it would be nice to have the plots contain both
data, no?

nh13:
In some cases it would, like for GC bias plots, but the same could be said for any metric level.

yfarjoun:
@jabbarish another interesting issue, I think.

zstudent:
As far as we understand it, CollectGcBiasMetrics calculates GC bias statistics for different levels (all reads, sample, library, read group) and produces two output txt files (details and summary) with statistics and pdf file with data plots. Current version doesn't check whether a read is a duplicate or not at any level. The key point is to collect information about the duplicate flag for all levels (one level is selected per execution) and plot two batches of charts (one includes all reads according to level and the other includes only non-duplicates according to level). 

We propose several changes in this metric:
1. New command line option NON_DUPLICATES that has 'false' default value and true if user would like to calculate GC bias statistics for both cases (for all reads and for non-duplicates).
2. If that option is true output files will contain additional information for non-duplicates case with extra prefix.
3. New unit test to validate the results. 

Also, the proposed changes contain solution for the other issue https://github.com/broadinstitute/picard/issues/342: there is an exception while processing reads, because of symbol '*' in SEQ field. SAM/BAM format specification allows this, but CollectGcBiasMetrics can not process such situation correctly. 

We propose to omit reads with '*' is SEQ field and log message. Also, we added new unit test to check correctness. 

All proposed improvements do not effect the performance.

@yfarjoun if this sounds good to you, we will create PR with this changes.

jabbarish:
As far as we understand, CollectGcBiasMetrics calculates GC bias statistics for different levels (all reads, sample, library, read group) and produces two output txt files (details and summary) with statistics and pdf file with data plots. Current version doesn't check whether a read is a duplicate or not at any level. The key point is to collect information about a duplicate flag for all levels (one level is selected per execution) and plot two batch of charts (one includes all reads according to level and the other includes only non-duplicates according to level). 

We suggest several changes in this metric:
1. New command line option NON_DUPLICATES that false default and true if user would like to calculate GC bias statistics for both cases (for all reads and for non-duplicates).
[...]

Issue #138, opened 2014-12-18
Picard CollectGcBiasMetrics does not handle reads without stored bases (*) gracefully


nh13:
http://sourceforge.net/p/samtools/mailman/message/33164920/

vdauwera:
Emailed the user for a data snippet.

visivas:
This issue is not yet resolved. I am willing to share a data snippet for you to reproduce the error. Please let me know how I can upload the BAM file. Also the reference is streptomyces rapamycinicus available from NCBI http://www.ncbi.nlm.nih.gov/nuccore/NC_022785.1 (I can also upload that if you need).

vdauwera:
@visivas, thanks for being willing to share your data. Please follow the instructions given at https://www.broadinstitute.org/gatk/guide/article?id=1894 

If you could please include the reference genome in the same data upload that would make it much easier for us, thanks.

visivas:
Alright, I have uploaded all the required files as a tar.gz file (GcBiasTroubleShoot.tar.gz). Please let me know if you need anything else. Thanks for working on this issue.

vdauwera:
Thanks @visivas, we'll try to look at this soon.

vdauwera:
To recap the issue, the user was running CollectGcBiasMetrics on some bacterial genomes but it was failing with error: 

    Exception in thread "main" net.sf.samtools.SAMException: Exception counting
    mismatches for read MAC1-WDL30341:66:000000000-ABTGC:1:2111:13507:4770 2/2
    0b aligned read.

User's input file validates with no errors (with IGNORE= MATE_NOT_FOUND since the input is a snippet).

But running GATK PrintReads to subset the culprit read fails with error:

    ##### ERROR MESSAGE: SAM/BAM file SAMFileReader{/Users/vdauwera/codespace/sandbox/GcBiasTroubleShoot/AS_ATCC.bam} is malformed: the BAM file has a read with no stored bases (i.e. it uses '*')

I'm not sure what Picard's position is on reads with no stored bases. Looks like ValidateSamFile is letting it go but CollectGcBiasMetrics chokes on it. I haven't specifically tested whether that is the actual culprit, but it seems likely since the error occurs when counting bases and the tool reports seeing 0b. 

@nh13 Would you care to opine? Should ValidateSamFile catch this?

yfarjoun:
From the sam spec:

SEQ: segment SEQuence. This field can be a ‘*’ when the sequence is not
stored.

[...]

PR #646, opened 2016-09-08
Fix for issues 342 and 138: introduced ALSO_IGNORE_DUPLICATES option

jabbarish:
Fix for issue https://github.com/broadinstitute/picard/issues/342:

CollectGcBiasMetrics calculates GC bias statistics for different levels (all reads, sample, library, read group) and produces two output txt files (details and summary) with statistics and pdf file with data plots. Current version doesn't check whether a read is a duplicate or not at any level but it maybe useful to compare two plots with and without duplicates. The main idea is to implement reading with filtering by duplicate flag. 

We propose the following changes in this metric: 
New command line option ALSO_IGNORE_DUPLICATES that false by default and true if user would like to calculate GC bias statistics for both cases (for all reads and for non-duplicates). Enabling this option will extend current functionality and allow to get two plots per level at the same time: the first one is the usual and the second one excludes duplicates. If that option is true the output files will contain additional information for non-duplicates case with extra prefix.

Fix for issue https://github.com/broadinstitute/picard/issues/138:

There is an exception while processing reads, because of symbol '*' in SEQ field. SAM/BAM format specification allows this but CollectGcBiasMetrics can not process such situation correctly.

 We propose to omit reads with '*' is SEQ field and print a log message.
All proposed improvements do not effect the performance and include acceptance tests to validate the results.


src/main/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -179,14 +188,18 @@ protected void acceptRead(final SAMRecord rec, final ReferenceSequence ref) {
     @Override
     protected void finish() {
         multiCollector.finish();
+        writeResultsToFiles(OUTPUT, SUMMARY_OUTPUT, CHART_OUTPUT);
+    }
+
+    private void writeResultsToFiles(final File OUTPUT, final File SUMMARY_OUTPUT, final File CHART_OUTPUT) {
```
yfarjoun: can this method be static?

src/main/java/picard/analysis/GcBiasMetricsCollector.java
```diff
@@ -102,70 +108,111 @@ public PerUnitGcBiasMetricsCollector(final String sample, final String library,
...
+        /////////////////////////////////////////////////////////////////////////////
+        @Override
+        public void addMetricsToFile(final MetricsFile<GcBiasMetrics, Integer> file) {
+            addGcDataToFile(file, this.gcData, "");
+            if (IS_NON_DUPLICATES) {
+                addGcDataToFile(file, this.gcDataNonDups, NON_DUPS_PREFIX);
+            }
+        }
+
+        private void updTotalClusters(final SAMRecord rec, final Map<String, GcObject> gcData) {
```
yfarjoun: updateTotalClusters is better then updTotalClusters (I think)

src/main/java/picard/analysis/GcBiasMetricsCollector.java
```diff
@@ -102,70 +108,111 @@ public PerUnitGcBiasMetricsCollector(final String sample, final String library,
...
         /////////////////////////////////////////////////////////////////////////////
         //Takes each record and sends them to addRead to calculate gc metrics for
         // that read for each accumulation level
         /////////////////////////////////////////////////////////////////////////////
+        @Override
         public void acceptRecord(final GcBiasCollectorArgs args) {
             final SAMRecord rec = args.getRec();
-            final String type;
+            if (rec.getReadBases().length == 0) {
+                log.warn("Omitting read " + rec.getReadName() + " with '*' in SEQ field.");
```
yfarjoun: this can get old very quickly if there are no reads with SEQ bases...limit at 100 warnings please.

src/main/java/picard/analysis/GcBiasMetricsCollector.java
```diff
@@ -178,11 +225,8 @@ private double sum(final int[] values) {
             return total;
         }
 
-        /////////////////////////////////////////////////////////////////////////////
-        //Called to add metrics to the output file for each level of collection
-        // these metrics are used for graphing gc bias in R script
-        /////////////////////////////////////////////////////////////////////////////
-        public void addMetricsToFile(final MetricsFile<GcBiasMetrics, Integer> file) {
+        private void addGcDataToFile(final MetricsFile<GcBiasMetrics, Integer> file, final Map<String, GcObject> gcData,
```
yfarjoun: is this a static function now?

src/main/resources/picard/analysis/gcBias.R
```diff
@@ -68,7 +68,12 @@ for (k in 1:(num.plots)){
     else{
     	if(accLevel == "Sample"){datasetName <- summaryMetrics[k, "SAMPLE"]}
     	if(accLevel == "Library"){datasetName <- summaryMetrics[k, "LIBRARY"]}
-    	else(datasetName <- summaryMetrics[k, "READ_GROUP"])}
+        if(accLevel == "Read Group"){datasetName <- summaryMetrics[k, "READ_GROUP"]}
+        if(accLevel == "Non-Duplicates All Reads"){datasetName <- summaryMetrics[k, "ALL_READS"]}
+        if(accLevel == "Non-Duplicates Sample"){datasetName <- summaryMetrics[k, "SAMPLE"]}
+        if(accLevel == "Non-Duplicates Library"){datasetName <- summaryMetrics[k, "LUBRARY"]}
```
yfarjoun: typo LUBRARY->LIBRARY

src/test/java/picard/analysis/CollectGcBiasMetricsTest.java
```diff
@@ -274,17 +276,94 @@ public void runGcBias (final File input, final File outfile, final File detailsO
...
+     * @throws IOException
+     */
+    @Test
+    public void runNonDupsComparisonTest() throws IOException {
+        final File inputFileWithDuplicates = new File("testdata/picard/metrics/chrMReads.sam");
+        final File detailsOutfile = File.createTempFile("test", ".gc_bias_detail_metrics");
+        final File summaryOutfile = File.createTempFile("test", ".gc_bias_summary_metrics");
+        detailsOutfile.deleteOnExit();
+        summaryOutfile.deleteOnExit();
+
```
yfarjoun: so many newlines!!

src/main/java/picard/analysis/GcBiasMetricsCollector.java
```diff
@@ -102,70 +108,111 @@ public PerUnitGcBiasMetricsCollector(final String sample, final String library,
...
+                prefix = this.sample;
+                gcData.put(prefix, new GcObject());
+            } else {
+                prefix = allReads;
+                gcData.put(prefix, new GcObject());
+            }
+            return gcData;
+        }
+
+        private void addReadToGcData(final SAMRecord rec, final Map<String, GcObject> gcData,
```
yfarjoun: I think that it would be better to change the metric class and add a field that is something like 
```
Boolean INCLUDES_DUPLICATES
```
Instead of using Strings for figuring this out.

src/main/java/picard/analysis/GcBiasMetricsCollector.java
jabbarish: fixed

src/main/java/picard/analysis/GcBiasMetricsCollector.java
jabbarish: done

src/main/java/picard/analysis/GcBiasMetricsCollector.java
jabbarish: If we understand your comment correctly, adding new boolean constant is not necessary because MultiLevelCollector already has boolean field IS_NON_DUPLICATES that implements logic you are talking about. According to our changes method addReadToGcData() might be called twice: to provide old logic and to add extra data without duplicates. That's why we use String as method argument, it is just a prefix, that will be added to output file.  We slightly refactored the method, though, just cleaned up duplicate code.

src/main/java/picard/analysis/GcBiasMetricsCollector.java
jabbarish: no, it uses a lot of instance members of both outer and inner classes.

src/main/resources/picard/analysis/gcBias.R
jabbarish: fixed

src/test/java/picard/analysis/CollectGcBiasMetricsTest.java
jabbarish: fixed

src/main/java/picard/analysis/GcBiasMetricsCollector.java
yfarjoun: I meant changing the metric class, not the collector.

src/main/java/picard/analysis/GcBiasMetricsCollector.java
yfarjoun: OK. I think that it could be made static, but concede that it would be beyond the scope of this PR.

ekazachkova:
We corrected logic to support new command line option ALSO_IGNORE_DUPLICATES: new constructor in GcBiasMetricsCollector with one additional argument has been created, all our MultiLevelCollector changes (there was a static boolean field IS_NON_DUPLICATES) have been removed. Also, `addToGcData` and `addGcDataToFile` methods no longer receive String as argument: we added enum class GroupsType to get rid of strings concatenation that you are talking about. Moreover, we made some little changes to clean up code.

yfarjoun:
@ZLyanov, I'm not seeing the change I requested: that INCLUDES_DUPLICATES will be a (new) field in the metric rather than a modification of the group value. parsing strings is not what I want to do when I read metric files. So I'd like there to be another field in the metric, perhaps called READS_USED and it will have two options: UNIQUE and ALL.

ZLyanov:
@yfarjoun, sorry for misunderstanding. Now we see what you are talking about. To make sure that we don't have any disagreements, please, look through the attached metrics file. Does it match the expected result? Is new metrics file format suitable?
[example_CollectGcBiasMetrics_details_output.txt](https://github.com/broadinstitute/picard/files/579239/example_CollectGcBiasMetrics_details_output.txt)

yfarjoun:
Exactly! thanks!

ZLyanov:
@yfarjoun done.  Check please.

src/test/java/picard/analysis/CollectGcBiasMetricsTest.java
```diff
@@ -274,17 +276,92 @@ public void runGcBias (final File input, final File outfile, final File detailsO
...
+                Assert.assertEquals(summary.ALIGNED_READS, 3);
+                Assert.assertEquals(summary.AT_DROPOUT, 79.180328);
+                Assert.assertEquals(summary.GC_DROPOUT, 12.28901);
+                Assert.assertEquals(summary.GC_NC_0_19, 0.0);
+                Assert.assertEquals(summary.GC_NC_20_39, 0.0);
+                Assert.assertEquals(summary.GC_NC_40_59, 1.246783);
+                Assert.assertEquals(summary.GC_NC_60_79, 0.0);
+                Assert.assertEquals(summary.GC_NC_80_100, 0.0);
+            }
+            if (summary.ACCUMULATION_LEVEL.equals("All Reads") && summary.READS_USED.equals("ALL")) { //ALL_READS level
```
yfarjoun: You have static variables for these strings now...please use them.

src/test/java/picard/analysis/CollectGcBiasMetricsTest.java
```diff
@@ -274,17 +276,92 @@ public void runGcBias (final File input, final File outfile, final File detailsO
...
+        detailsOutfile.deleteOnExit();
+        summaryOutfile.deleteOnExit();
+
+        runGcBias(inputFileWithDuplicates, REFERENCE_FILE_2, summaryOutfile, detailsOutfile, true);
+
+        final MetricsFile<GcBiasSummaryMetrics, Comparable<?>> outputSummary = new MetricsFile<>();
+        outputSummary.read(new FileReader(summaryOutfile));
+
+        for (final GcBiasSummaryMetrics summary : outputSummary.getMetrics()) {
+            if (summary.ACCUMULATION_LEVEL.equals("All Reads") && summary.READS_USED.equals("UNIQUE")) { //ALL_READS level for case without duplicates
```
yfarjoun: here too (static strings)

src/test/java/picard/analysis/CollectGcBiasMetricsTest.java
```diff
@@ -274,17 +276,92 @@ public void runGcBias (final File input, final File outfile, final File detailsO
...
+        summaryOutfile.deleteOnExit();
+        detailsOutfile.deleteOnExit();
+
+        runGcBias(input, REFERENCE_FILE_2, summaryOutfile, detailsOutfile, false);
+
+        final MetricsFile<GcBiasSummaryMetrics, Comparable<?>> output = new MetricsFile<>();
+        output.read(new FileReader(summaryOutfile));
+
+        for (final GcBiasSummaryMetrics metrics : output.getMetrics()) {
+            if (metrics.ACCUMULATION_LEVEL.equals("All Reads")) { //ALL_READS level
```
yfarjoun: and here..

yfarjoun:
Thanks! 👍

PR #661, opened 2016-09-21
Added in an summary error rate by substitution type to CollectSequencingArtifactMetrics

tfenne:
@yfarjoun This is the PR I promised a few weeks ago.  The result is a new metrics file from CollectSequencingArtifactMetrics with six rows that look something like this:
```
REF_BASE  ALT_BASE  SUBSTITUTION  REF_COUNT  ALT_COUNT  SUBSTITUTION_RATE
A         C         A>C           207358     28         0.00014
A         G         A>G           207358     52         0.00025
A         T         A>T           207358     36         0.00017
C         A         C>A           181950     20         0.00011
C         G         C>G           181950     14         0.00008
C         T         C>T           181950     17         0.00009
```

I've been finding this super-useful to just get a summary of the different kinds of errors that are predominant in a library before diving into _how_ of _where_ they are coming from.

**Please note:** I haven't touched the tests yet.  If you're willing, I'd like you to take a first pass look at the code _first_.  Once you're happy with the calculations and the output format, I will extend the tests to cover the new functionality.  I just don't want to do that up front and then have to fix them all if you ask for some change that affects everything!

### DIFF
--- a/src/main/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
+++ b/src/main/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
@@ -4,36 +4,26 @@ import htsjdk.samtools.AlignmentBlock;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMReadGroupRecord;
 import htsjdk.samtools.SAMRecord;
-import htsjdk.samtools.filter.AggregateFilter;
-import htsjdk.samtools.filter.AlignedFilter;
-import htsjdk.samtools.filter.DuplicateReadFilter;
-import htsjdk.samtools.filter.FailsVendorReadQualityFilter;
-import htsjdk.samtools.filter.InsertSizeFilter;
-import htsjdk.samtools.filter.MappingQualityFilter;
-import htsjdk.samtools.filter.NotPrimaryAlignmentFilter;
-import htsjdk.samtools.filter.SamRecordFilter;
+import htsjdk.samtools.filter.*;
 import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.reference.ReferenceSequence;
-import htsjdk.samtools.util.IOUtil;
-import htsjdk.samtools.util.IntervalList;
-import htsjdk.samtools.util.IntervalListReferenceSequenceMask;
-import htsjdk.samtools.util.StringUtil;
+import htsjdk.samtools.util.*;
 import picard.PicardException;
 import picard.analysis.SinglePassSamProgram;
+import picard.analysis.artifacts.SequencingArtifactMetrics.BaitBiasDetailMetrics;
+import picard.analysis.artifacts.SequencingArtifactMetrics.BaitBiasSummaryMetrics;
+import picard.analysis.artifacts.SequencingArtifactMetrics.PreAdapterDetailMetrics;
+import picard.analysis.artifacts.SequencingArtifactMetrics.PreAdapterSummaryMetrics;
 import picard.cmdline.CommandLineProgramProperties;
 import picard.cmdline.Option;
 import picard.cmdline.programgroups.Metrics;
 import picard.util.DbSnpBitSetUtil;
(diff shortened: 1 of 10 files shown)


src/main/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
```diff
@@ -332,13 +324,42 @@ protected void finish() {
...
         baitBiasSummaryMetricsFile.write(baitBiasSummaryOut);
+
+        // Calculate the summary error rates - it's CRITICAL that the other files are written out
+        // first as this code modifies the pre-adapter detail metrics!
+        if (!preAdapterDetailMetricsFile.getMetrics().isEmpty()) {
+            final List<PreAdapterDetailMetrics> in = preAdapterDetailMetricsFile.getMetrics();
+            in.forEach(m -> {
+                if (m.REF_BASE == 'G' || m.REF_BASE == 'T') {
+                    m.REF_BASE = (char) SequenceUtil.complement((byte) m.REF_BASE);
+                    m.ALT_BASE = (char) SequenceUtil.complement((byte) m.ALT_BASE);
```
yfarjoun: shouldn't the counts be negated if you complement the bases? I'm confused about what this is trying to do otherwise...

yfarjoun: given that the summands are themselves differences of other tallies...this is hard to understand the math..can you add a comment explaining why this is the right thing to do?

yfarjoun:
looks like a useful metric to have...still a bit confused about the details of complementing...

tfenne:
@yfarjoun I'll add a comment.  The complementing is to reduce the number of output rows and consolidate things that are the same kind of substitution into a single row.  So it collapses `G>*` and `T>*` into substitutions into the complementary `C>*` and `A>*` substitutions.  E.g. `T>C` gets summed with, and reported under, `A>G`.

It just gives fewer rows and makes it clearer where the errors are clustered.  If that doesn't make sense let me know and I'll try again with the explanation.

yfarjoun:
So is this possible because in effect only one of these (in each case) will
have a non-zero rate (as the other should be negative and thus assumed to
be "not the problem"? and then by summing them you'll get half the entries
to look at and it will be more clear what _type_ of conversion has large
rate (if any)?

tfenne:
I think you're overthinking things.  It's not trying to pick up an artifact, just summarize the rate of different kinds of substitutions that we're seeing.  And since most of what we're looking at after filtering for relatively high base quality are substitutions prior to sequencing, they show up on both strands (as, e.g. paired C>A and G>T substitutions) regardless of whether the top-strand reference base happens to be a C or a G.  It's the same reason that the lego plots you often see used to breakdown cancer mutations use only 6 substitutions instead of 12.

yfarjoun:
👍

tfenne:
@yfarjoun Just pushed an update that includes extending the tests to cover this new functionality.  I think this is ready for merge once tests pass if your 👍 is still good.

src/test/java/picard/analysis/artifacts/CollectSequencingArtifactMetricsTest.java
```diff
@@ -32,11 +32,12 @@ public String getCommandLineProgramName() {
     @BeforeTest
     public void setUp() throws IOException {
         globalTempOutputDir = IOUtil.createTempDir("artifactMetrics.", ".tmp");
+        System.out.println("Writing to: " + globalTempOutputDir.getAbsolutePath());
     }
 
     @AfterTest
     public void tearDown() throws IOException {
-        IOUtil.deleteDirectoryTree(globalTempOutputDir);
+//        IOUtil.deleteDirectoryTree(globalTempOutputDir);
```
yfarjoun: if this isn't needed please remove the code (is it not needed?)

tfenne: Oops, that shouldn't have been committed like that.  Sorry!

yfarjoun:
small request and then your good to go!

Issue #518, opened 2016-04-19
MergeBamAlignment: Wrong orientation for reads with secondary alignments


fedxa:
I have an aligned sam file with secondary alignments reported.  The MErgeBamAlignment tool returns all the secondary alignments in the same orientation, as the primary, though they are aligned to different strands.

In the following example, the first alignment line in the merged file is wrongly reverse complemented

Command:
`java -jar picard.jar MergeBamAlignment UNMAPPED=u.sam ALIGNED=a.sam O=m.sam R=yeast.fa R1_TRIM=1`

Unaligned file:

```
@HD VN:1.5  SO:queryname
@RG ID:WT_S3    SM:WT_S3    LB:WT_S3    PL:illumina PU:C7WN7ACXX.ACAGTG.8   CN:UCR
SN279:506:C7WN7ACXX:8:2312:20390:58438  4   *   0   0   *   *   0   0   AAAATTCACCTAAAGATAAAAACTGGAGGCACCATCAATAGATCGGAAGAG =@1??D2AC?F??<A+<<3AHA:BB+<E;7C?8C<G9*:9B4??F0DA?F( RG:Z:WT_S3
```

Aligned file (sorry, I am lazy to fix this -- the first NT of the read was clipped by the pipeline), first line is the secondary alignement to + strand, second line is the primary alignment to - strand

```
@HD VN:1.0  SO:coordinate
@SQ SN:2-micron LN:6318
@SQ SN:chrI LN:230218
@SQ SN:chrII    LN:813184
@SQ SN:chrIII   LN:316620
@SQ SN:chrIV    LN:1531933
@SQ SN:chrIX    LN:439888
@SQ SN:chrM LN:85779
@SQ SN:chrV LN:576874
@SQ SN:chrVI    LN:270161
@SQ SN:chrVII   LN:1090940
@SQ SN:chrVIII  LN:562643
@SQ SN:chrX LN:745751
@SQ SN:chrXI    LN:666816
@SQ SN:chrXII   LN:1078177
@SQ SN:chrXIII  LN:924431
@SQ SN:chrXIV   LN:784333
@SQ SN:chrXV    LN:1091291
@SQ SN:chrXVI   LN:948066
[...]

PR #663, opened 2016-09-23
Fix for MergeBamAlignment when a read is mapped to both strands

meganshand:
Addresses #518 

When a single read maps to both strands, cloning the unaligned read is not deep enough because the reverse complement of the bases will be taken. I changed `clone` to `deepCopy` which makes a deep copy of the attributes of a SamRecord (which clone does not).


src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -409,7 +410,7 @@ public void mergeAlignment(final File referenceFasta) {
                         final SAMRecord matePrimary = isRead1 ? r2Primary : r1Primary;
 
                         for (final SAMRecord supp : supplementals) {
-                            final SAMRecord out = clone(sourceRec);
+                            final SAMRecord out = sourceRec.deepCopy();
```
nh13: Do supplementals always need to be deep copied?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -351,9 +351,10 @@ public void mergeAlignment(final File referenceFasta) {
 
             // See if there are alignments for current unaligned read or read pair.
             if (nextAligned != null && rec.getReadName().equals(nextAligned.getReadName())) {
-                // If there are multiple alignments for a read (pair), then the unaligned SAMRecord must be cloned
-                // before copying info from the aligned record to the unaligned.
-                final boolean clone = nextAligned.numHits() > 1 || nextAligned.hasSupplementalHits();
+                // If there are multiple alignments for a read (pair), then the unaligned SAMRecord must be deep copied
+                // before copying info from the aligned record to the unaligned. Cloning in htsjdk is not deep enough
+                // since the bases can be reverse complemented in the case where one read aligns to both strands.
```
nh13: Perhaps say "...since later (below) the bases can be...".  I was confused if it meant prior to or after this statement.

yfarjoun:
@meganshand can you push the first commit to a different branch so that travis runs on it and will show that it's currently failing? you can then post the direct link to that travis build here.

src/main/java/picard/sam/AbstractAlignmentMerger.java
yfarjoun: perhaps it would be better to do a deep copy at the step before (and if) the bases get reverse complemented? is that possible?

yfarjoun: The only place in the code that I could find that reverse complements is ::makeReadUnmaped() is that the problem? Perhaps a better solution would be to add a boolean flag to this method that will copy the bases rather than do an in-place reverse complementing of the bases and qualities?

meganshand: @nh13 You're right, the SAM spec says that a chimeric alignment is a set of linear alignments and linear alignments must be on the same strand. If we assume that, I can change it back to clone for supplementals.

@yfarjoun makeReadUnmapped() isn't the only problem, it's also in AbstractAlignmentMerger::setValuesFromAlignment(). This means that we need a deep copy before calling transferAlignmentInfoToPairedRead() or transferAlignmentInfoToFragment(), the way it is written now. Additionally, you won't know you need a deep copy until you've seen every secondary alignment that came from the original read.

meganshand:
I pushed the first commit to a new branch and GitHub helpfully added a red x above! But the failing test is here: https://travis-ci.org/broadinstitute/picard/builds/162276946

src/main/java/picard/sam/AbstractAlignmentMerger.java
meganshand: done

src/main/java/picard/sam/AbstractAlignmentMerger.java
yfarjoun: hmmm. I don't think that we should make that assumption. I don't think that supplementary alignments need to be on the same strand as the primary. it doesn't even make sence since they can be on different contigs...IMHO, the right way to avoid deep-copying all the supplementals would be to modify reverseComplement() (unfortunately since it it in htsjdk) and provide it with an optional boolean flag that will make new arrays of bases and qualities instead of reversing in place. this will mean that we don't actually do a deep copy, but that we protect other copies of the bases and qualities from modification.

meganshand: @yfarjoun You think that supplementals can be on different strands even if that violates the SAM spec? Also it doesn't make intuitive sense to me how a supplemental alignment could biologically come from different strands, whereas secondary alignments are mutually exclusive biologically, so it's clear how they can be on different strands.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -353,8 +353,9 @@ public void mergeAlignment(final File referenceFasta) {
             if (nextAligned != null && rec.getReadName().equals(nextAligned.getReadName())) {
                 // If there are multiple alignments for a read (pair), then the unaligned SAMRecord must be deep copied
                 // before copying info from the aligned record to the unaligned. Cloning in htsjdk is not deep enough
-                // since the bases can be reverse complemented in the case where one read aligns to both strands.
-                final boolean deepCopy = nextAligned.numHits() > 1 || nextAligned.hasSupplementalHits();
+                // since later (below) the bases can be reverse complemented in the case where one read aligns to both strands.
+                final boolean deepCopy = nextAligned.numHits() > 1;
+                final boolean clone = nextAligned.hasSupplementalHits();
```
yfarjoun: I don't understand this logic...

src/main/java/picard/sam/AbstractAlignmentMerger.java
yfarjoun: I must be missing something in the sam spec. 

> Chimeric alignment: An alignment of a read that **cannot** be represented as a linear alignment.

What part of the spec are you reading?

meganshand: @yfarjoun I was getting this idea from the very next sentence which describes a Chimeric alignment as a "set of linear alignments", but I was misinterpreting that. I talked to Chris and I see how supplementary alignments can be on opposite strands (and apparently this is fairly common in some technologies), either from errors or inversion events.

meganshand:
@yfarjoun @nh13 I updated this to use the new reverseComplement in htsjdk. Could you please take another look?

yfarjoun:
thanks! 👍

PR #664, opened 2016-09-30
Total clusers is now a long in GcBiasSummaryMetrics.

nh13:
Previously it was an integer, which could be an issue if we have many
clusters.


yfarjoun:
👍

PR #665, opened 2016-09-30
QualityYieldMetrics may overflow with many reads.

nh13:
Changing TOTAL_READS and PF_READS to a long instead of an int.


yfarjoun:
👍

PR #670, opened 2016-10-09
removed documentation about intel deflater

maarten-k:
Since the Intel deflater is not functional anymore, i removed the links and the documentation about this topic. It might be nice to have some info on other compression optimization options.


yfarjoun:
@vdauwera could you review this, or assign someone to do so?

vdauwera:
Sorry I missed this. Will do now.

faq.html
```diff
@@ -213,16 +213,10 @@
 					</div>
 				</div>
 				
-				<div class="faq" onclick="toggle('faq26')">
-					<strong>Q:</strong> What is IntelDeflater and how do I control it?
```
vdauwera: I think we should keep the documentation line on this but modify the text to reflect the fact that ID has been taken out into a separate repo (https://github.com/Intel-HLS/GKL), and specify when this happened, so people using older version don't get too confused.

intel-deflater.html
```diff
@@ -1,34 +0,0 @@
----
----
-<!doctype html>
-<html>
-	{% include head.html title="Intel Deflater" %}
-	<body>
-		<div class="wrapper">
-			{% include header.html %}
-
-			<h2>Intel Deflater</h2>
```
vdauwera: We should keep this but update it to reflect how IntelDeflater is used from the GKL package -- and possibly generalize it to mention other compression/decompression capabilities available from GKL are used in Picard. 

@lbergelson, do you know who would be in the best position to update/write about this?

vdauwera:
Alternatively we could remove the docs as originally proposed and plan to put updated content into a wiki page on here since it's developer-oriented.

intel-deflater.html
lbergelson: @vdauwera I think @gspowley would probably be the best person to write an update for this.

vdauwera:
Ok, thanks @lbergelson. I copied the old content to a [new wiki page](https://github.com/broadinstitute/picard/wiki/IntelDeflater---(de)compression-capabilites-from-Intel-GKL) which will need to be updated, preferably with @gspowley's help.

@maarten-k Could you please update your pull request to modify the content in the FAQs to point to the wiki page linked above instead of deleting the mention altogether? The `intel-deflater.html` page itself can be deleted.

gspowley:
@vdauwera, I can help update the wiki page. A lot has changed since we moved to GKL.

maarten-k:
@vdauwera I updated the docs

vdauwera:
Great, thanks @maarten-k

PR #658, opened 2016-09-19
Include low quality bases in depth and base quality histograms for theoretical het sensitivity

takutosato:
We now store two depth histograms: `unfilteredDepthHistogram` and `highQualityDepthHistogram`. The former includes all but quality 2 bases, and the latter includes bases with quality greater than or equal to `MINIMUM_BASE_QUALITY` (default 20). Base quality histogram adopts the 'unfiltered' rule. With this change we hope to improve theoretical het sensitivity estimates.

Also included in this PR: in `TargetMetricsCollector.java`, we used to double-count the zero-bin of coverageDistribution when the target had no coverage (once in line 624 and again in 630). This did not affect the metrics (as far as I can tell) but over-counted the zero-depth bin in the depth histogram. I fixed the bug and added a test: `testCoverageHistogram` in `CollectHsMetricsTest`.

(edit: 10/21/16, tsato - removed outdated notes)

PR #671, opened 2016-10-11
Fix bug that causes CollectHsMetrics (and friends) to double-count...

tfenne:
…0-coverage sites in 0-coverage targets.

### DIFF
--- a/src/main/java/picard/analysis/directed/TargetMetricsCollector.java
+++ b/src/main/java/picard/analysis/directed/TargetMetricsCollector.java
@@ -619,10 +619,7 @@ public abstract class TargetMetricsCollector<METRIC_TYPE extends MultilevelMetri
                 final boolean hasCoverage = c.hasCoverage();
                 final int[] targetDepths = c.getDepths();
 
-                if (!hasCoverage) {
-                    zeroCoverageTargets++;
-                    coverageDistribution[0] += c.getDepths().length;
-                }
+                if (!hasCoverage) zeroCoverageTargets++;
 
                 for (final int depth : targetDepths) {
                     if (0 < depth) totalCoverage += depth;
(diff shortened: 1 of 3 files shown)


tfenne:
Calling @yfarjoun, @kbergin and @davidbenjamin (and maybe @ktibbett).  It looks like when the theoretical het sensitivity code went in a bug got introduced that double-counted 0-coverage bases from 0-coverage targets when calculating the coverage distribution which is used for both the histogram in the metrics file _and_ the het sensitivity calculations.

Looking at it fresh, I think actually the `calculateTargetCoverageMetrics()` method is very confusing and error-prone.  It essentially calculates three different coverage distributions/arrays in parallel with a bunch of subtly different conditionals.   I'm positive it would be better long term to refactor to either calculate a single array from which all the resulting things are derived, or separate the array calculations so they are not interleaved with one another, and thus easier to follow.

If you look at the _before_ code you can see that on line 624 the distribution was being incremented for zero coverage targets.  And then on line 630 it was being incremented for _all_ bases regardless of whether the target has coverage or not.  

The result is that:
1. The zero coverage bin is not the same fraction of bases as `1-PCT_TARGET_BASES_1X` (it should be)
2. The sum of the histogram of bases does not equal the `TARGET_TERRITORY` (it should).

I've updated the tests to assert the target territory is equal, but this passed even before my changes because the current tests are very minimal.

yfarjoun:
I think that @takutosato PR #658 fixes this problem (though it makes the issue you raise of the complication of the code worse)

yfarjoun:
@tfenne can you verify that the PR that was just merged #658 actually solves this bug and therefore this PR can be closed?

yfarjoun:
@tfenne I'm closing this as I suspect that #658 resolves the same issue reported here. please feel free to reopen (and fix the conflicts) If I'm wrong.

PR #673, opened 2016-10-14
Added a public accessor for the per-target coverage information.

tfenne:
I find that I'm using the `HsMetricCollector` and related classes programmatically a lot, and it would be really nice to be able to access the per-base and per-target coverage information that is accumulated without having to write it out to a file and parse it back in again.  This PR adds a one-line accessor to make the information available to other classes.

### DIFF
--- a/src/main/java/picard/analysis/directed/TargetMetricsCollector.java
+++ b/src/main/java/picard/analysis/directed/TargetMetricsCollector.java
@@ -58,14 +58,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.lang.reflect.Field;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


src/main/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -406,6 +399,11 @@ public void setBaitSetName(final String name) {
             this.metrics.PROBE_SET = name;
         }
 
+        /** Returns the accumulated coverage per target. */
+        public Map<Interval,Coverage> getCoverageByTarget() {
```
yfarjoun: -Space after comma, 
-Note that the map is still *mutable* since it may change "from within" due to another call to acceptRecord. Could you add a comment to that effect?

otherwise, 👍

tfenne:
Done @yfarjoun - will merge one tests go green again.

PR #676, opened 2016-10-18
Yf change bacterial contamination to filter

yfarjoun:
This PR changes the bahavior of MergeBamAlignment when encountering a read that looks like bacterial contamination (soft-clipped on both sides, and few unclipped bases left)

- previous behavior: unmap read (and mate) and reset position
- new behavior: filter (0x200 flag) read and mate, and leave mapping and other tags alone.


yfarjoun:
@tfenne could you please review this? this is in response to your comments on hts-spec...

tfenne:
@yfarjoun The code changes look fine to me.  I'm confused though as I thought I was against the idea of setting the `fails vendor quality` flag for these kinds of reads?

yfarjoun:
hmmm. perhaps I misunderstood you. 

so let's have the discussion again:

0x4: 
pro: it is stating a correct fact that we think that the segment is not mapped where the aligner put it and we have no better place for it
con: we lose the original mapping information

0x200:
pro: it retains the mapping information while having the effect of keeping the segment from being used downstream
con: traditionally this flag is used by the machine and not changed by downstream analysis

tfenne:
I think the real downside of `0x200` (non-PF) is that unless any/all tools that rely on that flag are updated to also look at the `FT` tag, things are going to get lost.  Imagine someone hands you a BAM that has these reads with `0x200` set, and you decide you want to re-map the reads, so you remove all the mapping information and excess tags, and start over.  But if you weren't careful then all these reads will still be set as non-PF and might get skipped over.

I think the previous method of making the read unmapped is more in line with reality - you're saying you don't trust these very short/poor quality mappings, so you want to ignore them.

yfarjoun:
The revert-sam problem is indeed compelling....

the problem I'm faced with is the loss of the original data. Perhaps
there's a solution that will allow us to save the mapping data...

PR #681, opened 2016-11-01
update htsjdk to 2.7.0

yfarjoun:
also:
- removed usage of deprecated code
- added some java8 sugar

### DIFF
--- a/src/main/java/picard/sam/AbstractAlignmentMerger.java
+++ b/src/main/java/picard/sam/AbstractAlignmentMerger.java
@@ -36,7 +36,6 @@ import htsjdk.samtools.SAMProgramRecord;
 import htsjdk.samtools.SAMRecord;
 import htsjdk.samtools.SAMRecordCoordinateComparator;
 import htsjdk.samtools.SAMRecordQueryNameComparator;
-import htsjdk.samtools.SAMRecordUtil;
 import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.SAMSequenceRecord;
 import htsjdk.samtools.SAMTag;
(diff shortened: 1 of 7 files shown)


src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -663,7 +662,7 @@ protected void setValuesFromAlignment(final SAMRecord rec, final SAMRecord align
         // If it's on the negative strand, reverse complement the bases
         // and reverse the order of the qualities
         if (rec.getReadNegativeStrandFlag()) {
-            SAMRecordUtil.reverseComplement(rec);
+            rec.reverseComplement();
```
meganshand: Because the default changed, in order for rec.reverseComplement() to remain an in-place operation (instead of making a copy), this should be rec.reverseComplement(true); Even though, I'm planning on changing this in #663, it's probably better to keep the behavior the same as it currently is in this PR.

There's also a usage of SamRecordUtil.reverseComplement in RevertSam. Could you also update that one to rec.reverseComplement(true)?

meganshand:
@yfarjoun back to you

src/main/java/picard/sam/AbstractAlignmentMerger.java
yfarjoun: SamRecordUtil is @Depracated...should I really be updating it?

yfarjoun:
done (though I didn't change SamRecordUtils)

meganshand:
@yfarjoun Sorry, I meant can you update it in RevertSam? There is a usage there.

yfarjoun:
oh, ok. done.

meganshand:
@yfarjoun Thank you! 👍

Issue #433, opened 2016-02-11
MarkDuplicates throws error using cram files


dekling:
Get the following error when using cram file for this tool:

Exception in thread "main" htsjdk.samtools.cram.CRAMException: A reference source is required for CRAM files
    at htsjdk.samtools.CRAMIterator.<init>(CRAMIterator.java:86)
    at htsjdk.samtools.CRAMFileReader.<init>(CRAMFileReader.java:130)
    at htsjdk.samtools.CRAMFileReader.<init>(CRAMFileReader.java:146)
    at htsjdk.samtools.CRAMFileReader.<init>(CRAMFileReader.java:152)
    at htsjdk.samtools.SamReaderFactory$SamReaderFactoryImpl.open(SamReaderFactory.java:273)
    at picard.sam.markduplicates.util.AbstractMarkDuplicatesCommandLineProgram.openInputs(AbstractMarkDuplicatesCommandLineProgram.java:205)
    at picard.sam.markduplicates.MarkDuplicates.buildSortedReadEndLists(MarkDuplicates.java:307)
    at picard.sam.markduplicates.MarkDuplicates.doWork(MarkDuplicates.java:155)
    at picard.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:209)
    at picard.cmdline.PicardCommandLine.instanceMain(PicardCommandLine.java:95)
    at picard.cmdline.PicardCommandLine.main(PicardCommandLine.java:105)
-bash:gsa3:/humgen/gsa-scr1/dkling 622 $ 

Here is the server command line and test files:
-bash:gsa3:/humgen/gsa-scr1/dkling 621 $ java -jar picard.jar MarkDuplicates I=/humgen/gsa-scr1/dkling/cram/pruned_snippet.cram O=/humgen/gsa-scr1/dkling/cram/pruned_snippet_md.bam M=/humgen/gsa-scr1/dkling/cram/pruned_snippet_md.txt R=/humgen/gsa-scr1/pub/bundle/2.8/b37/human_g1k_v37.fasta.gz

Picard jar is in gsa3:/humgen/gsa-scr1/dkling 

-Thanks

P.S.  It worked fine with a bam file: /humgen/gsa-scr1/dkling/cram/pruned_snippet.bam

yfarjoun:
This is going to be a problem with all picard tools. I think that the best
solution would be to add the REFERENCE to the default SamReaderFactory, but
[...]

Issue #431, opened 2016-02-05
picard-1.141 Cram Output for FastqToSam


vcaropr1:
I have been messing with the cram implementation into picard and was wondering if the output of FastqToSam can indeed be a cram file or not.  I know for SamFormatConverter the output can be a cram file, but when trying the same method FastqToSam seems to output a bam sized file with a .cram ending.  The code below is what I ran:

java -jar picard FastqToSam 
FASTQ=read1.fastq 
FASTQ2=read2.fastq
SORT_ORDER=coordinate
OUTPUT=sample.cram
SAMPLE_NAME= sample_name
PLATFORM_UNIT= read_index
LIBRARY_NAME=library
PLATFORM=ILLUMINA

Again, I am using 1.141 and not sure if this was resolved in 2.0.1 or if this capability isn't there.  

Thanks!

cmnbroad:
I'm not sure what the official support status for writing (or reading) CRAM in picard tools is, but it looks like most of them are still using makeSAMOrBAMWriter, which isn't really equipped for writing CRAM files since doing so requires a reference. It also has a [bug](https://github.com/samtools/htsjdk/issues/366) that results in behavior exactly as you describe (writes BAM encoded contents even if the output file has a .cram extension). You can verify this is whats happening by running "od -c -N 4 youroutput.cram". An actual CRAM file produced by htsjdk will output "0000000   C   R   A   M" but for BAM will produce less unrecognizable compressed output like "0000000 037 213  \b 004".

vcaropr1:
Yup you are definitely right.  This file created is just a bam file with a .cram extension.

Issue #366, opened 2015-12-23
CollectWgsMetrics should double-check that sequence dictionaries match


eitanbanks:
It is possible for a user to pass in e.g. the hg19 reference for the REFERENCE_SEQUENCE argument with a bam aligned to hg38.  No error or warning is generated until/unless we hit a read that aligns off the end of the reference sequence (although it doesn't really; we're just using the wrong reference).
We should put in a super quick sanity check on the dictionaries to make sure they match (at least somewhat).

akiezun:
see also https://github.com/broadinstitute/gatk/pull/1379

nsoranzo:
:+1: The `ArrayIndexOutOfBoundsException` error is quite obscure for the user.

golharam:
This also happens when using a BAM file aligned with GRCh37 (1000 Genomes), but REFERENCE_SEQUENCE is given hg19.  I don't think its obscure.   If the sequence dictionaries don't match, and error should be reported.

kbergin:
@sophiacrennan has added a check for this in #1610 . Closing this issue!

Issue #361, opened 2015-12-10
MarkDuplicates not working with cram input file


mmoisse:
When I try to dedup a cram file I get an error telling me a reference genome should be supplied, although one of my arguments is the reference genome.

```
[Thu Dec 10 11:01:48 CET 2015] picard.sam.markduplicates.MarkDuplicatesWithMateCigar 
  INPUT=[/home/test/test.cram] OUTPUT=/home/test/test.dedup.cram 
  METRICS_FILE=/home/test/test.dedup_metrix.txt VERBOSITY=DEBUG
  REFERENCE_SEQUENCE=/home/test/Homo_sapiens.GRCh38.dna_sm.primary_assembly.fa.gz    
  MINIMUM_DISTANCE=-1 SKIP_PAIRS_WITH_NO_MATE_CIGAR=true BLOCK_SIZE=100000 
  REMOVE_DUPLICATES=false ASSUME_SORTED=false 
  DUPLICATE_SCORING_STRATEGY=TOTAL_MAPPED_REFERENCE_LENGTH 
  PROGRAM_RECORD_ID=MarkDuplicates 
  PROGRAM_GROUP_NAME=MarkDuplicatesWithMateCigar
  READ_NAME_REGEX=[a-zA-Z0-9]+:[0-9]:([0-9]+):([0-9]+):([0-9]+).* 
  OPTICAL_DUPLICATE_PIXEL_DISTANCE=100 QUIET=false 
  VALIDATION_STRINGENCY=STRICT COMPRESSION_LEVEL=5 
  MAX_RECORDS_IN_RAM=500000 CREATE_INDEX=false CREATE_MD5_FILE=false 
  GA4GH_CLIENT_SECRETS=client_secrets.json
[Thu Dec 10 11:01:48 CET 2015] Executing as test@test on Linux 3.16.0-53-generic amd64; 
  OpenJDK 64-Bit Server VM 1.8.0_45-internal-b14; Picard version: 2.0.1
  (fe51f432e4b918385037e6d15e958f0298203e28_1449251266) IntelDeflater
[Thu Dec 10 11:01:48 CET 2015] picard.sam.markduplicates.MarkDuplicatesWithMateCigar 
  done. Elapsed time: 0,00 minutes.
  Runtime.totalMemory()=251658240
To get help, see http://broadinstitute.github.io/picard/index.html#GettingHelp
Exception in thread "main" htsjdk.samtools.cram.CRAMException: 
A reference source is required for CRAM files
[...]

PR #683, opened 2016-11-03
CRAM format support: fix for issues 433, 431, 361, 366

ekazachkova:
1. Current version of Picard Tools has a bug: if you pass in a filename with a .cram extension, makeSAMOrBAMWriter will create a BAM file with a .cram extension and return a BAMFileWriter, which is probably not what was intended. This bug is described in htsjdk issue[#366](https://github.com/broadinstitute/picard/issues/433#issuecomment-183141150), discussed in picard issue[#433](https://github.com/broadinstitute/picard/issues/433) and manifested in issue[#431](github.com/broadinstitute/picard/issues/431) (FastqToSam)
2.  Also, there is a bug in MarkDuplicates, it cannot get CRAM file as input, doesn't pass reference sequence directly to SamReaderFactory, despite the fact that the CommandLineProgram has such an input parameter: issue [#433](https://github.com/broadinstitute/picard/issues/433), issue[#361](https://github.com/broadinstitute/picard/issues/361). The same bug is in SplitSamByLibrary.

This PR provides the following changes:

1.  `makeWriter` method is used instead of `makeSamOrBamWriter` in  
 - AddOrReplaceReadGroups (+documentation changes)
 - CleanSam (+documentation changes, licence update)
 - FixMateInformation (+documentation changes, licence update)
 - MarkDuplicates (+documentation changes, licence update)
 - MergeSamFiles (+documentation changes, licence update)
 - PositionBasedDownsampleSam (+documentation changes, licence update)
 - ReorderSam (+documentation changes, licence update)
 - SortSam (+documentation changes, licence update)
 - RevertOriginalBaseQualitiesAndAddMateCigar (+documentation changes)
 - GatherBamFiles
 - IlluminaBasecallsToSam (+documentation changes, licence update)
 - FastqToSam (+documentation changes, licence update)
 - SimpleMarkDuplicatesWithMateCigar (+documentation changes, licence update)
 - SplitSamByLibrary (+documentation changes, licence update)
 - FilterSamReads (+documentation changes, licence update)
 - MarkDuplicatesWithMateCigar (+documentation changes, licence update)
 - MarkIlluminaAdapters (+documentation changes, licence update)
 - ReplaceSamHeader (+documentation changes, licence update)
 - SetNmMdAndUqTags (+documentation changes)
 - AbstractAlignmentMerger (licence update)
2. make AbstractMarkDuplicatesCommandLineProgram and SplitSamByLibrary classes call  SamReaderFactory#referenceSource
3. provide .cram extension support for GatherBamFiles, SplitSamByLibrary and FixMateInformation classes
4. extend test suite for CRAM support: expand FilterSamReadsTest and create new CramCompatibilityTest class


src/main/java/picard/sam/ReplaceSamHeader.java
```diff
@@ -53,8 +53,8 @@
         programGroup = SamOrBam.class
 )
 public class ReplaceSamHeader extends CommandLineProgram {
-    static final String USAGE_SUMMARY = "Replaces the SAMFileHeader in a SAM or BAM file.  ";
-    static final String USAGE_DETAILS = "This tool makes it possible to replace the header of a SAM or BAM file with the header of another" +
+    static final String USAGE_SUMMARY = "Replaces the SAMFileHeader in a SAM, BAM or CRAM file.  ";
+    static final String USAGE_DETAILS = "This tool makes it possible to replace the header of a SAM, BAM or CRAM  file with the header of another" +
             "file, or a header block that has been edited manually (in a stub SAM file). The sort order (@SO) of the two input files must " +
             "be the same.<br /><br />" +
             "Note that validation is minimal, so it is up to the user to ensure that all the elements referred to in the SAMRecords " +
```
yfarjoun: need to change doc for INPUT and HEADER

yfarjoun: also, coudl you make use of  replaceCramHeader if it's a cram file? (or open an issue to do this)

cmnbroad: The CRAM format specifies that each container in a CRAM stream contain a [value](https://github.com/samtools/htsjdk/blob/fbba5364e1809de071bc479f30e4e2c8b17f5bbe/src/main/java/htsjdk/samtools/cram/structure/Container.java#L60) indicating that container's offset with its containing stream (!). I think there maybe one for each slice as well. If resetting the header displaces the rest of the stream, these values will be off. AFAIK, htsjdk doesn't actually use these values for anything currently, but it does set them. So generic block-type moves are probably not a good idea unless we implement a generalized method that knows how and when to repair these (which BTW we could use for other things).

ekazachkova: Should we replace in the docs any mention of "SAM" with "SAM, BAM or CRAM", only for relevant metrics, of course? In the example above for INPUT: "SAM file from which SAMRecords will be read." should it be "SAM, BAM or CRAM file from which SAMRecords will be read."?

ekazachkova:
We would like to add in this PR CRAM support for DownsampleSam class.

src/main/java/picard/sam/ReplaceSamHeader.java
yfarjoun: I think that if it says "sam" you can leave it alone, or change it to "sam, bam or cram". 
if it already says "sam or bam", then please fix it.

ekazachkova: Done.

src/main/java/picard/sam/ReorderSam.java
```diff
@@ -117,7 +117,7 @@ protected int doWork() {
 
         log.info("Writing reads...");
         if (in.hasIndex()) {
-            final SAMFileWriter out = new SAMFileWriterFactory().makeSAMOrBAMWriter(outHeader, true, OUTPUT);
+            final SAMFileWriter out = new SAMFileWriterFactory().makeWriter(outHeader, true, OUTPUT, REFERENCE_SEQUENCE);
```
cmnbroad: For some reason (probably because it was a convenient way to override the optional attribute for REFERENCE for this tool) ReorderSam has a duplicate command line arg for reference. There is one in the ReorderSam class called REFERENCE, and one inherited from CommandLineProgram called REFERENCE_SEQUENCE. Both of them are used in the tool code, but only REFERENCE is properly populated by the command line parser. This code should be fixed so the variable usage is consistent. It might be best to just remove the local REFERENCE variable and rely on the inherited REFERENCE_SEQUENCE one, and just have the code throw if the argument is not populated. I think that would retain the behavior and remove the ambiguity.

src/test/java/picard/sam/CramCompatibilityTest.java
```diff
@@ -0,0 +1,319 @@
...
+                            "REFERENCE=" + REFERENCE_FILE + " " +
+                        "ALLOW_CONTIG_LENGTH_DISCORDANCE=true " +
+                        "ALLOW_INCOMPLETE_DICT_CONCORDANCE=true "
+                },
+                {"picard.sam.SortSam", "SORT_ORDER=coordinate"},
+        };
+    }
+
+    @Test(dataProvider = "programms")
+    public void writingMetricsShouldWriteCRAMWhenCRAMWithReferenceIsInput(String program, String exParameters)
```
cmnbroad: I think it would be clearer to not use the phrase metrics in these method names, as none of these are what I think of as metrics per se.

ekazachkova:
This update includes the following changes:
- refactoring CramCompatibilityTest class: each test corresponds to one of the cases: program works with aligned cram correctly, program fails if aligned cram without reference and program works with unmapped cram correctly.
- minor changes according to your comments in CramCompatibilityTest class
- changes in ReorderSam class: we have removed cmd option REFERENCE to avoid ambiguity
- test data have been regenerated
- minor refactoring in FilterSamReadsTest class: this test removes .bai files now
@cmnbroad Please check

sooheelee:
What is the status on CRAM support in Picard tools @cmnbroad? We have a forum member needing this for DownsampleSam at <https://gatkforums.broadinstitute.org/gatk/discussion/comment/41175>.

cmnbroad:
@sooheelee https://github.com/broadinstitute/picard/pull/817 affects one part of this, and I have been waiting for that to go in, but it took quite a bit longer than I expected (my apologies to @ekazachkova). Thats finally merged now, but this branch is quite far behind master, and needs to be rebased and conflicts resolved. I expect we'll pick this up again now that https://github.com/broadinstitute/picard/pull/817 is in.

PolinaBevad:
@cmnbroad Sorry for delay with these changes!
Some of the tests fell on new master, so we made several changes to solve this problem.
The main reason for tests to fall is that access level modifier of `REFERENCE_SEQUENCE` was changed from public to protected in `CommandLineProgram` class. 

* `CommandLineProgram` class and `FilterSamReads` class (which extends `CommandLineProgram` class) are in different packages. This leads to changing of access level modifier of overriden field `REFERENCE_SEQUENCE` from protected to private in class-inheritor.
Our solution: we added package-private setter for `REFERENCE_SEQUENCE` to `FilterSamReads`.

* Problems also occurred in class `AbstractMarkDuplicatesCommand` (method `openInputs()`):
when reader is created there is check which depends on flag `EAGERLY_DECODE`. Flag is never False in this case, so `REFERENCE_SEQUENCE` is never called.
Our solution: we added invoking of `REFERENCE_SEQUENCE` in both cases (when `EAGERLY_DECODE` is True or False).

* We had to make changes in class `CramCompatibilityTest` (method `launchProgram()`):
check “if (reference != null)” does not work for `ReorderSam` class as it contains `REFERENCE`, not just `REFERENCE_SEQUENCE`.
Our solution: we added one more check specially for `ReorderSam`.

cmnbroad:
@PolinaBevad Thanks for the update. This branch is pretty far behind master now, and there have definitely been some changes made there that will affect. Before I start another review pass I think we should get this rebased onto current master, and preferably squashed down to a single commit. Once we have that, with tests passing, I will start a review pass as soon as I possibly can. Thanks!

Ksenomorfa:
@cmnbroad  Thanks for comment! Rebase is done and also all changes are squashed to one commit.

PolinaBevad:
@cmnbroad Thanks for comment! Rebase is done and also all changes are squashed to one commit.

cmnbroad:
@PolinaBevad Thanks for rebasing this. Before I do a more detailed review, the main thing we need to do is to get the tests refactored. Rather than having all of the CRAM tests in a single test file, and just checking that the output is a CRAM, the CRAM tests for a given tool should live with the existing tests for that tool; should replicate an existing test; and should validate the results the same way the existing test does. 

I would suggest the following: To validate that CRAM reading works, each CRAM read test should use a CRAM file that has been converted from an existing BAM test file for that tool using samtools. Then the same test can be run using BAM and CRAM inputs (perhaps though a data provider), and the results should verified to be the same (i.e., if its a metrics tool, the resulting metrics should be the same for the BAM and the CRAM). For CRAM write tests, it should be sufficient to reproduce an existing test that writes a BAM, only with CRAM output, and validate that the results are the same.

I'm not sure how many of these tools don't have existing tests. For those cases, the tests can live in the CRAM compatibility test file, but the test strategy should be the same (verify that a BAM, and a CRAM generated by converting that BAM to CRAM using samtools, produce the same results).

If you have questions about this, or want to have me review a couple of these before doing them all, I'm happy to do so. We're grateful for the help! Thx.

marchoeppner:
Just out of curiosity, is this ever going to get fixed? I mean, gatk4 now has a cram-compatible replacement for MarkDuplicates, albeit one that is not considered stable yet. But still... would be nice if Picard had consistent support for CRAM :-<

yfarjoun:
@cmnbroad can this be closed now?

cmnbroad:
@yfarjoun I think closing it makes sense -  its out of date, and presumably its superseded by your branch. But it reminds me that you should also include the `makeSAMorBAMWriter` -> `makeWriter` change as part of enabling cram writing.

yfarjoun:
actually...reading through it...I don't see what's wrong with it...it has tests and everything. can I rebase this and we can consider merging it?

cmnbroad:
@yfarjoun Good point. IIRC there was some weirdness with the way the tests were incorporated, but there is a lot of work in here that would be great to salvage if you can.

yfarjoun:
commit absorbed into a separate PR: #1356

Issue #432, opened 2016-02-07
CreateSequenceDictionary stalls indefinitely with large genomes


cacampbell:
Running create sequence dictionary results in an indefinitely running process (at 3 days now -- just to see how long it would go) with the following output:

[Thu Feb 04 10:10:14 PST 2016] picard.sam.CreateSequenceDictionary REFERENCE=[large conifer species] OUTPUT=[large conifer species].dict VERBOSITY=DEBUG QUIET=false MAX_RECORDS_IN_RAM=40000000    TRUNCATE_NAMES_AT_WHITESPACE=true NUM_SEQUENCES=2147483647 VALIDATION_STRINGENCY=STRICT COMPRESSION_LEVEL=5 CREATE_INDEX=false CREATE_MD5_FILE=false GA4GH_CLIENT_SECRETS=client_secrets.json
[Thu Feb 04 10:10:14 PST 2016] Executing as [user]@bigmem3 on Linux 3.13.0-36-generic amd64; Java HotSpot(TM) 64-Bit Server VM 1.8.0_20-b26; Picard version: 2.0.1() JdkDeflater

the fasta file for that genome is around 40GB

Using samtools 1.3 dict creates a sequence dictionary relatively quickly (about an hour)

Sorry if this has been discussed / resolved elsewhere

tfenne:
FYI it's likely that this is less related to the raw size of the genome in bytes or bases, and more to the number of contigs in the genome.  Unless it has changed CreateSequenceDictionary caches all the `@SQ` records in memory before writing them to disk, and if there are hundreds of thousands, it will end up running low on memory and thrashing in garbage collection.

yfarjoun:
if any of the contigs is greater in size than MAX_INT, there will be a
problem since we seem to be using int for iterating the reference.

tfenne:
Though to be clear - a single sequence of length greater than max int will cause problems in many places, since the SAM spec defines position fields as ints.

cacampbell:
Regarding contig count -- yeah this seems to be the case for me

jmarshall:
See also [this samtools-help thread](https://sourceforge.net/p/samtools/mailman/message/31531210/) regarding 2<sup>31</sup>-1 vs. 2<sup>32</sup>-1 for maximum contig sizes.

jabbarish:
@cacampbell  could you please clarify how many contigs your reference file contains? Do you allocate additional memory for the picard process or use the default value?

[...]

PR #687, opened 2016-11-17
Fix for issue #432 CreateSequenceDictionary stalls indefinitely with large genomes 

SilinPavel:
We introduce two pull requests to Htsjdk (https://github.com/samtools/htsjdk/pull/744) and Picard Tools to resolve the issue https://github.com/broadinstitute/picard/issues/432 .

Original version of CreateSequenceDictionary creates and stores all reference sequences entirely in memory in a SAMHeader object and then writes it to a file.

In order to check sequence's names for uniqueness, they are added to a HashSet. This might lead to "Out of memory" exceptions for files with a large number of sequences .

**Changes in Htsjdk:**

In our implementation we introduce a new "on fly" codec SAMSequenceDictionaryCodec, which encodes each sequence and directly writes it to the Dictionary file. This allows us not to store sequences in memory.

**Changes in Picard Tools:**

CreateSequenceDictionary now uses new SAMSequenceDictionaryCodec to write dictionary file.

SortingCollection is used instead of a HashSet to verify the sequences names uniqueness , this allows us not to keep all the names in memory since identical names will be stored in the collection one after another and we can find duplicates in one pass. 

Please pay attention, that now CreateSequenceDictionary doesn't write HD line, since we think that it is not required, SequenceDictionary must contain only SQ line, not HD. If our assumption is wrong, correct us, please.

 

 

So, if we look at the plot "GC Times" from JMC: FlightRecorder, we will see that standard version of CreateSequenceDictionary have a big trouble with a large amount of records (several millions):

**The plot "GC Times" for a  Standard version:**
 
selection_041
selection_043

**The plot "GC Times" for a fixed version:**
selection_040
selection_042

How we can see, now CreateSequenceDictionary processes a big reference without troubles, for example time of execution for file from issue https://github.com/broadinstitute/picard/issues/432 is:

```
real 762.83
user 626.94
sys 67.02
```

Full GC reports (jmc flight recorder):
[JMC FR logs.zip](https://github.com/broadinstitute/picard/files/597773/JMC.FR.logs.zip)

### Checklist

- [x] Code compiles correctly
- [ ] New tests covering changes and new functionality
- [x] All tests passing

### DIFF
--- a/src/main/java/picard/sam/CreateSequenceDictionary.java
+++ b/src/main/java/picard/sam/CreateSequenceDictionary.java
@@ -140,11 +140,12 @@ public class CreateSequenceDictionary extends CommandLineProgram {
                     getReferenceSequenceFile(REFERENCE, TRUNCATE_NAMES_AT_WHITESPACE);
             SAMSequenceDictionaryCodec samDictCodec = new SAMSequenceDictionaryCodec(writer);
 
+            samDictCodec.encodeHeaderLine(false);
             // read reference sequence one by one and write its metadata
             ReferenceSequence refSeq;
             while ((refSeq = refSeqFile.nextSequence()) != null) {
                 final SAMSequenceRecord samSequenceRecord = makeSequenceRecord(refSeq);
-                samDictCodec.encodeSQLine(samSequenceRecord);
+                samDictCodec.encodeSequenceRecord(samSequenceRecord);
                 sequenceNames.add(refSeq.getName());
             }
         } catch (FileNotFoundException e) {
--- a/src/main/java/picard/sam/CreateSequenceDictionary.java
+++ b/src/main/java/picard/sam/CreateSequenceDictionary.java
@@ -39,10 +39,10 @@ import picard.PicardException;
 import picard.cmdline.CommandLineProgram;
 import picard.cmdline.CommandLineProgramProperties;
 import picard.cmdline.Option;
-import picard.cmdline.StandardOptionDefinitions;
 import picard.cmdline.programgroups.Fasta;
+import picard.cmdline.StandardOptionDefinitions;
 
-import java.io.*;
+import java.io.File;
 import java.math.BigInteger;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
@@ -71,7 +71,6 @@ public class CreateSequenceDictionary extends CommandLineProgram {
             "" +
             "</pre>" +
             "<hr />";
-
     // The following attributes define the command-line arguments
 
     @Option(doc = "Input reference fasta or fasta.gz", shortName = StandardOptionDefinitions.REFERENCE_SHORT_NAME)
--- a/src/main/java/picard/sam/CreateSequenceDictionary.java
+++ b/src/main/java/picard/sam/CreateSequenceDictionary.java
@@ -42,7 +42,7 @@ import picard.cmdline.Option;
 import picard.cmdline.programgroups.Fasta;
 import picard.cmdline.StandardOptionDefinitions;
 
-import java.io.File;
+import java.io.*;
 import java.math.BigInteger;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
@@ -135,7 +135,7 @@ public class CreateSequenceDictionary extends CommandLineProgram {
 
         // SortingCollection is used to check uniqueness of sequence names
         final SortingCollection<String> sequenceNames = makeSortingCollection();
-        try (Writer writer = makeWriter()) {
+        try (BufferedWriter writer = makeWriter()) {
             final ReferenceSequenceFile refSeqFile = ReferenceSequenceFileFactory.
                     getReferenceSequenceFile(REFERENCE, TRUNCATE_NAMES_AT_WHITESPACE);
             SAMSequenceDictionaryCodec samDictCodec = new SAMSequenceDictionaryCodec(writer);
@@ -168,13 +168,15 @@ public class CreateSequenceDictionary extends CommandLineProgram {
         return 0;
     }
 
-    private Writer makeWriter() throws FileNotFoundException {
-        return new AsciiWriter(this.CREATE_MD5_FILE ?
+    private BufferedWriter makeWriter() throws FileNotFoundException {
+        return new BufferedWriter(
+                new AsciiWriter(this.CREATE_MD5_FILE ?
                         new Md5CalculatingOutputStream(
                                 new FileOutputStream(OUTPUT, false),
                                 new File(OUTPUT.getAbsolutePath() + ".md5")
                         )
                         : new FileOutputStream(OUTPUT)
+                )
         );
     }
 
--- a/src/main/java/picard/sam/CreateSequenceDictionary.java
+++ b/src/main/java/picard/sam/CreateSequenceDictionary.java
@@ -23,34 +23,32 @@
  */
 package picard.sam;
 
-import htsjdk.samtools.SAMFileHeader;
-import htsjdk.samtools.SAMFileWriter;
-import htsjdk.samtools.SAMFileWriterFactory;
-import htsjdk.samtools.SAMSequenceDictionary;
+import htsjdk.samtools.SAMSequenceDictionaryCodec;
 import htsjdk.samtools.SAMSequenceRecord;
 import htsjdk.samtools.reference.ReferenceSequence;
 import htsjdk.samtools.reference.ReferenceSequenceFile;
 import htsjdk.samtools.reference.ReferenceSequenceFileFactory;
+import htsjdk.samtools.util.AsciiWriter;
+import htsjdk.samtools.util.CloseableIterator;
+import htsjdk.samtools.util.IOUtil;
+import htsjdk.samtools.util.Md5CalculatingOutputStream;
+import htsjdk.samtools.util.RuntimeIOException;
+import htsjdk.samtools.util.SortingCollection;
 import htsjdk.samtools.util.StringUtil;
 import picard.PicardException;
 import picard.cmdline.CommandLineProgram;
 import picard.cmdline.CommandLineProgramProperties;
 import picard.cmdline.Option;
-import picard.cmdline.programgroups.Fasta;
 import picard.cmdline.StandardOptionDefinitions;
+import picard.cmdline.programgroups.Fasta;
 
-import java.io.File;
+import java.io.*;
 import java.math.BigInteger;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
 
 /**
- * Create a SAM/BAM file from a fasta containing reference sequence.  The output SAM file contains a header but no
- * SAMRecords, and the header contains only sequence records.
+ * Create a .dict file from a fasta containing reference sequence. dict file contains only sequence records.
  */
 @CommandLineProgramProperties(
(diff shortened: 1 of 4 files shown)


vdauwera:
Hi @SilinPavel, this is going to be really useful, so thanks for working on it! 

I'm interested to know if this fix also might address performance problems we've seen reported about genomes with very large contigs. Specifically, we had a report from a user working with Triticum aestivum (wheat bread), which has a contig that is longer than 512M bases. This crashes Picard BuildBamIndex as well as bam indexing functionality in GATK3. Do you know off the top of your head if that would be related to this?

SilinPavel:
Hello @vdauwera, we are happy to hear this is going to be useful!
We do not think this pull request might address the problem of large contigs performance.It rather is directed to fix one particular problem of writting sequence dictionary seq by seq. Anyway we are ready to investigate the large contigs performance issues, e.g. Picard BuildBamIndex crashing while handling Triticum aestivum BAM. 
Is it possible to provide us with additional information and the data that helps to reporoduce the problem? We will much appreciate this.
Thank you!

vdauwera:
Hi @SilinPavel, the genome reference is available at ftp://ftp.ensemblgenomes.org/pub/plants/release-33/fasta/triticum_aestivum/dna/ but I'm not sure which of the files is the correct one; I'll find out. I think that should be all you need to reproduce the BuildBamIndex issue.

vdauwera:
FYI the reference to use for the large contig issue is ftp://ftp.ensemblgenomes.org/pub/plants/release-33/fasta/triticum_aestivum/dna/Triticum_aestivum.TGACv1.dna_sm.toplevel.fa.gz.

Thanks for looking into this! It's going to make a lot of people who are traditionally underserved very happy if you can address this problem.

vdauwera:
@yfarjoun or @lbergelson would you care to review this, or find someone who can? Feel free to ignore the bit of conversation where I jumped in.

yfarjoun:
needs to be rebased on a version of picard that revs htsjdk so that the changes in htsjdk are here.

lbergelson:
@SilinPavel We just released htsjdk 2.8.0 that includes your first pull request.  It would be good to update this pr to bump the htsjdk version in `build.gradle` from 2.7.0 -> 2.8.0.

SilinPavel:
@yfarjoun , @lbergelson done. Have a look, please!

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -134,33 +132,53 @@ protected int doWork() {
...
+            }
+        } catch (FileNotFoundException e) {
+            throw new PicardException("File " + OUTPUT.getAbsolutePath() + " not found");
+        } catch (IOException e) {
+            throw new PicardException("Can't write to or close output file " + OUTPUT.getAbsolutePath());
+        }
+
+        // check uniqueness of sequences names
+        final CloseableIterator<String> iterator = sequenceNames.iterator();
+        String first = iterator.hasNext() ? iterator.next() : "";
```
yfarjoun: I think it would be more clear to have 
```
if(!iterator.hasNext()) return 0;
String first = iterator.next();
```
?

yfarjoun: also, s/first/current/

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -134,33 +132,53 @@ protected int doWork() {
...
-        samWriter.close();
-        return 0;
-    }
 
-    /**
-     * Read all the sequences from the given reference file, and convert into SAMSequenceRecords
-     * @param referenceFile fasta or fasta.gz
-     * @return SAMSequenceRecords containing info from the fasta, plus from cmd-line arguments.
-     */
-    public SAMSequenceDictionary makeSequenceDictionary(final File referenceFile) {
```
yfarjoun: you are deleting a public method here. 

can you keep it around (even though it isn't used here any more) and @Deprecate it? Picard is also used as a library and this is a change in API.

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -134,33 +132,53 @@ protected int doWork() {
...
+            throw new PicardException("Can't write to or close output file " + OUTPUT.getAbsolutePath());
+        }
+
+        // check uniqueness of sequences names
+        final CloseableIterator<String> iterator = sequenceNames.iterator();
+        String first = iterator.hasNext() ? iterator.next() : "";
+        while (iterator.hasNext()) {
+            final String next = iterator.next();
+            if (first.equals(next)) {
+                OUTPUT.delete();
```
yfarjoun: I think it makes more sense to write a temporary file and move it into place when successful than to write a file and then possibly delete it. For example, you might not have permissions to delete files...

yfarjoun: also, will the .md5 file get written in this case? If so you probably want to delete it.

yfarjoun: (please test...)

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -196,4 +214,59 @@ private String md5Hash(final byte[] bytes) {
         }
         return s;
     }
+
+    private SortingCollection<String> makeSortingCollection() {
+        final String parent = System.getProperty("java.io.tmpdir") + "/" + System.getProperty("user.name");
```
yfarjoun: please use IOUtil.createTempDir

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -196,4 +214,59 @@ private String md5Hash(final byte[] bytes) {
...
+            throw new IllegalStateException("Could not create tmpdir: " + tmpDir.getAbsolutePath());
+        }
+        tmpDir.deleteOnExit();
+        // 256 byte for one name, and 1/10 part of all memory for this, rough estimate
+        long maxNamesInRam = Runtime.getRuntime().maxMemory() / 256 / 10;
+        return SortingCollection.newInstance(
+                String.class,
+                new StringCodec(),
+                String::compareTo,
+                maxNamesInRam > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int) maxNamesInRam,
```
yfarjoun:
```
(int) Math.min(maxNamesInRam, Integer.MAX_VALUE)
```

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -196,4 +214,59 @@ private String md5Hash(final byte[] bytes) {
         }
         return s;
     }
+
+    private SortingCollection<String> makeSortingCollection() {
+        final String parent = System.getProperty("java.io.tmpdir") + "/" + System.getProperty("user.name");
+        final String child = getClass().getSimpleName();
+        final File tmpDir = new File(parent, child);
+        IOUtil.deleteDirectoryTree(tmpDir);
```
yfarjoun: deleting random directory trees on the path is not a nice way to behave....please only delete what you create, after you create it, and be sure that it didn't exist before creating it.....

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -134,33 +132,53 @@ protected int doWork() {
...
+        final SortingCollection<String> sequenceNames = makeSortingCollection();
+        try (BufferedWriter writer = makeWriter()) {
+            final ReferenceSequenceFile refSeqFile = ReferenceSequenceFileFactory.
+                    getReferenceSequenceFile(REFERENCE, TRUNCATE_NAMES_AT_WHITESPACE);
+            SAMSequenceDictionaryCodec samDictCodec = new SAMSequenceDictionaryCodec(writer);
+
+            samDictCodec.encodeHeaderLine(false);
+            // read reference sequence one by one and write its metadata
+            ReferenceSequence refSeq;
+            while ((refSeq = refSeqFile.nextSequence()) != null) {
```
yfarjoun: Please write this as a for loop. I understand what its doing, but I find it annoying to read.

src/main/java/picard/sam/CreateSequenceDictionary.java
SilinPavel: I brought it back and added Deprecated annotation, and I put it in the beginning of the class to other public methods.

src/main/java/picard/sam/CreateSequenceDictionary.java
SilinPavel: ReferenceSequence is not Iterable. It doesn't have any methods to get some Iterable (f.e. FastaSequenceIndex). One possible solution is to make it Iterable. Another way is to write for loop in the following way (which is worse): 
````
for (ReferenceSequence refSeq = refSeqFile.nextSequence(); refSeq != null; refSeq = refSeqFile.nextSequence()) {
       final SAMSequenceRecord samSequenceRecord = makeSequenceRecord(refSeq);
       samDictCodec.encodeSequenceRecord(samSequenceRecord);
       sequenceNames.add(refSeq.getName());
}
````

File: src/main/java/picard/sam/CreateSequenceDictionary.java [hunk 8]
```diff
@@ -156,7 +184,10 @@ protected int doWork() {
 
         // check uniqueness of sequences names
         final CloseableIterator<String> iterator = sequenceNames.iterator();
-        String first = iterator.hasNext() ? iterator.next() : "";
+
+        if(!iterator.hasNext()) return 0;
+
+        String first = iterator.next();
```
yfarjoun: rename `first` to `current` please

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -216,21 +247,15 @@ private String md5Hash(final byte[] bytes) {
...
-        final String parent = System.getProperty("java.io.tmpdir") + "/" + System.getProperty("user.name");
-        final String child = getClass().getSimpleName();
-        final File tmpDir = new File(parent, child);
-        IOUtil.deleteDirectoryTree(tmpDir);
-        if (!tmpDir.mkdirs()) {
-            throw new IllegalStateException("Could not create tmpdir: " + tmpDir.getAbsolutePath());
-        }
-        tmpDir.deleteOnExit();
+        final String name = getClass().getSimpleName();
+        final File tmpDir = IOUtil.createTempDir(name, null);
```
yfarjoun: tmpDir still need to be deleted. please call tmpDir.deleteOnExit

testdata/picard/reference/csd_dict.dict
```diff
@@ -0,0 +1,9 @@
+@HD	VN:1.5
+@SQ	SN:chr1	LN:101	M5:bd01f7e11515bb6beda8f7257902aa67	UR:file:/home/pavel/git/epm-cmbi/picard-pho/testdata/picard/reference/test.fasta
```
yfarjoun: please remove references to personal location of files from repository (since it doesn't matter, replace with dummy path)

yfarjoun:
back to you @SilinPavel

SilinPavel:
@yfarjoun Done! Have a look please!

yfarjoun:
Thank-you for this @SilinPavel !!

👍

ZLyanov:
@vdauwera Could you create an issue for the problem with the BuildBamIndex, please? Also, it would be great to take a look at the bug reports from the users.  And the last question: could you provide us with any sample bam files to help us reproduce the crash?  With regards, Zaal.

vdauwera:
Sure, will do.

PR #689, opened 2016-11-17
The molecular indexes should be joined with hyphens rather than

nh13:
concatenated.

@tfenne and @yfarjoun Since https://github.com/samtools/hts-specs/pull/119 hasn't been merged yet, I'd like to change this behavior.  In many applications it is important to know from which read (i.e. R1/R2) each molecular index belongs.


tfenne:
Can I give more than 1 👍👍👍👍👍👍 ?

src/main/java/picard/illumina/ClusterDataToSamConverter.java
```diff
@@ -74,6 +74,7 @@
     // TODO: - https://github.com/samtools/hts-specs/pull/119
     private String molecularIndexTag          = "RX";
     private String molecularIndexQualityTag   = "QX";
+    private String molecularIndexDelimiter    = "-";
```
yfarjoun: should these all be final?

src/main/java/picard/illumina/ClusterDataToSamConverter.java
```diff
@@ -178,10 +179,10 @@ private SAMRecord createSamRecord(final ReadData readData, final String readName
 
         if (!molecularIndexes.isEmpty()) {
             if (!this.molecularIndexTag.isEmpty()) {
-                sam.setAttribute(this.molecularIndexTag, String.join("", molecularIndexes));
+                sam.setAttribute(this.molecularIndexTag, String.join(molecularIndexDelimiter, molecularIndexes));
             }
             if (!this.molecularIndexQualityTag.isEmpty()) {
-                sam.setAttribute(this.molecularIndexQualityTag, String.join("", molecularIndexQualities));
+                sam.setAttribute(this.molecularIndexQualityTag, String.join(molecularIndexDelimiter, molecularIndexQualities));
```
yfarjoun: given that `-` is a legal quality character, splitting the quality string will involve splitting the bases first and then using the resulting lengths to split the quality string, I wonder if it would be clearer without the delimiter in the quality string. Not a request, just a opening for discussion.

nh13: I really want the lengths of the strings to be equal if possible, including the delimiter.

src/main/java/picard/illumina/ClusterDataToSamConverter.java
nh13: `molecularIndexDelimiter` can be yes, but the others can be overwritten with appropriate setter methods.

nh13:
@yfarjoun one last 👍 and I can merge this.

yfarjoun:
I concede to you as you work with multiple UMI's more than I. 👍

PR #695, opened 2016-12-05
Speed up loading DBSNP file in CollectSequencingArtifactMetrics with …

nh13:
…an IntervalList.

This also adds logging to the loading of DBSNP, which is useful if we
have a large set of regions or analyzing the whole genome.

@yfarjoun, @droazen, or @lbergelson care to take a look?

### DIFF
--- a/src/main/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
+++ b/src/main/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
@@ -18,12 +18,12 @@ import picard.cmdline.CommandLineProgramProperties;
 import picard.cmdline.Option;
 import picard.cmdline.programgroups.Metrics;
 import picard.util.DbSnpBitSetUtil;
+import picard.util.VariantType;
 
 import java.io.File;
 import java.util.*;
 import java.util.stream.Collectors;
 
-import static com.sun.tools.doclets.formats.html.markup.HtmlStyle.header;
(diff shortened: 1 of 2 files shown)


src/main/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
```diff
@@ -192,10 +194,16 @@ protected void setup(final SAMFileHeader header, final File samFile) {
...
+            intervalMask = new IntervalListReferenceSequenceMask(intervalList);
+
+            if (DB_SNP != null) {
+                IOUtil.assertFileIsReadable(DB_SNP);
+                dbSnpMask = new DbSnpBitSetUtil(DB_SNP, header.getSequenceDictionary(), EnumSet.noneOf(VariantType.class), intervalList, Optional.of(log));
+            }
+        }
+        else if (DB_SNP != null) {
             IOUtil.assertFileIsReadable(DB_SNP);
             dbSnpMask = new DbSnpBitSetUtil(DB_SNP, header.getSequenceDictionary());
```
yfarjoun: why not log this case too?

yfarjoun:
👍 looks fine. address my logging comment if you like.

src/main/java/picard/analysis/artifacts/CollectSequencingArtifactMetrics.java
nh13: Good idea, will do.

nh13:
Merging after tests pass

Issue #688, opened 2016-11-17
CollectIlluminaLaneMetrics does not handle read structures with molecular indexes


nh13:
The read structure `75T8B75T` works but `8M67B75T` does not.  The error I get is:
```
Exception in thread "main" picard.PicardException: Don't have both phasing and prephasing values for tile
	at picard.illumina.parser.TileMetricsUtil.getTilePhasingValues(TileMetricsUtil.java:116)
	at picard.illumina.parser.TileMetricsUtil.parseTileMetrics(TileMetricsUtil.java:95)
	at picard.illumina.CollectIlluminaLaneMetrics$IlluminaLaneMetricsCollector.readLaneTiles(CollectIlluminaLaneMetrics.java:110)
	at picard.illumina.CollectIlluminaLaneMetrics$IlluminaLaneMetricsCollector.collectLaneMetrics(CollectIlluminaLaneMetrics.java:123)
	at picard.illumina.CollectIlluminaLaneMetrics.doWork(CollectIlluminaLaneMetrics.java:91)
	at picard.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:208)
	at picard.cmdline.PicardCommandLine.instanceMain(PicardCommandLine.java:95)
	at picard.cmdline.PicardCommandLine.main(PicardCommandLine.java:105)
```

nh13:
@yfarjoun assigning this to you since I want to ask if you have seen this in your production environment, given you demultiplex and collect metrics using Picard.

nh13:
This also doesn't work with skips: `1S74T8B75T`

yfarjoun:
We are aware that it doesn't work with molecular barcodes, but we do not
have the priority to fix it...

Would you like to contribute a fix?

Y

nh13:
An obvious hack is to convert the skip and molecular barcode segments to a template segment (hint: that's my current workaround).  I may contribute a fix if it becomes a major issue going forward for me (it's not).  It's good to have this documented as a bug so that if other folks have the same problem they know they aren't the only ones.

vdauwera:
@nh13 @yfarjoun Should this have been closed when #697 went in?

yfarjoun:
not quite. this workaround doesn't resolve the underlying fundamental
problem...

nh13:
it doesn't handle skips and barcodes when present in both the index and in-line in the reads

PR #697, opened 2016-12-07
Support in-line skips and molecular barcodes.

nh13:
Solves issue https://github.com/broadinstitute/picard/issues/688

The phasing and pre-phasing metrics need all template cycles, including those that are marked as skips or molecular barcodes when the latter two are in-line in the reads.  I have not solved the case where we have both skips/molecular-barcodes in the index reads AND inline in the templates.  I added an option (`INCLUDE_SKIPS_AND_MOLECULAR_BARCODES`) that converts each skip or molecular barcode descriptor to a template descriptor for the purposes of `CollectIlluminaLaneMetrics`.

@yfarjoun want to take a look?

### DIFF
--- a/src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
+++ b/src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
@@ -33,13 +33,13 @@ import picard.cmdline.CommandLineProgramProperties;
 import picard.cmdline.Option;
 import picard.cmdline.StandardOptionDefinitions;
 import picard.cmdline.programgroups.Illumina;
-import picard.illumina.parser.ReadStructure;
-import picard.illumina.parser.Tile;
-import picard.illumina.parser.TileMetricsUtil;
+import picard.illumina.parser.*;
 
 import java.io.File;
 import java.io.FileNotFoundException;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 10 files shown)


src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
```diff
@@ -84,10 +84,28 @@
...
     protected int doWork() {
         final MetricsFile<MetricBase, Comparable<?>> laneMetricsFile = this.getMetricsFile();
         final MetricsFile<MetricBase, Comparable<?>> phasingMetricsFile = this.getMetricsFile();
+
+        // If skips or molecular identifiers are part of the read structure and also part of the non-index reads (line),
+        // then convert them template descriptors for our purposes.
+        if (INCLUDE_SKIPS_AND_MOLECULAR_BARCODES) {
+            final List<ReadDescriptor> descriptors = READ_STRUCTURE.descriptors.stream().map(descriptor -> {
+                if (descriptor.type == ReadType.Skip || descriptor.type == ReadType.MolecularIndex) {
+                    return new ReadDescriptor(descriptor.length, ReadType.Template);
```
yfarjoun: in my manual tests (modifying the provider in IlluminaLaneMetricsCollectorTest.java) I don't find that having 6T70T instead of  6M70M avoids the error...so I don't see how this will work. (a test would prove me wrong...)

src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
```diff
@@ -84,10 +84,28 @@
     @Option(doc = ReadStructure.PARAMETER_DOC, shortName = "RS")
     public ReadStructure READ_STRUCTURE;
 
+    @Option(doc = "Include skips and molecular barcodes for both phasing and pre-phasing values.  Use this if the either of the former are in-line in the read(s) (not in index reads).")
+    public boolean INCLUDE_SKIPS_AND_MOLECULAR_BARCODES = false;
```
nh13: Ah, I thought I was going to skate by without writing tests, but now I see `IlluminaLaneMetricsCollectorTest`.  Shame on me.

src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
nh13: I have to merge `6T70T` into `76T`.  Will do.

nh13: So the user would provide `6M70M`, the tool would convert that to `76T`, and all is peachy.

src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
```diff
@@ -84,10 +84,28 @@
...
         final MetricsFile<MetricBase, Comparable<?>> laneMetricsFile = this.getMetricsFile();
         final MetricsFile<MetricBase, Comparable<?>> phasingMetricsFile = this.getMetricsFile();
+
+        // If skips or molecular identifiers are part of the read structure and also part of the non-index reads (line),
+        // then convert them template descriptors for our purposes.
+        if (INCLUDE_SKIPS_AND_MOLECULAR_BARCODES) {
+            final List<ReadDescriptor> descriptors = READ_STRUCTURE.descriptors.stream().map(descriptor -> {
+                if (descriptor.type == ReadType.Skip || descriptor.type == ReadType.MolecularIndex) {
+                    return new ReadDescriptor(descriptor.length, ReadType.Template);
+                }
```
nh13: ?

nh13:
To do:
- [ ] add a test (or two)

nh13:
@yfarjoun I am wondering if this should just be scrapped, and folks should parse the `RunInfo.xml` for Illumina sequencers to get the read structure without user-specified skips and molecular barcodes.  What do you think?  If so, I'll change this PR to just add some more documentation about using this tool with molecular barcodes and skips.

yfarjoun:
can we parse the RunInfo.xml within this tool instead?

nh13:
I suppose we could by default (READ_STRUCTURE is optional).  I'll do that.

nh13:
@yfarjoun If `READ_STRUCTURE` is omitted (now optional), we parse the `RunInfo.xml`.

yfarjoun:
https://www.google.com/url?sa=i&rct=j&q=&esrc=s&source=images&cd=&cad=rja&uact=8&ved=0ahUKEwig6om-y-fQAhXBwFQKHQ9MB-gQjRwIBw&url=http%3A%2F%2Fwww.memegen.com%2Fmeme%2Fn8ywrt&psig=AFQjCNFrFqBs6sTt4x27wWWf_EXRkZ9FvA&ust=1481389203725577

yfarjoun:
image

PR #701, opened 2016-12-08
Small changes to how MBA works with bacterial contamination

yfarjoun:
### Description
MBA currently unmaps reads that are suspect of being Bacterial contamination. There are severl limitations in the implementation:
1. The (old) mapping information is discarded when converted to CRAM format
2. The FT: tag (into which a note is placed regarding the change made was confusing and not in the SAM spec
3. The log files explode when there are many reads that look like bacterial contamination.

This PR solves each of these as follows:

1. The old mapping is placed into a PA tag with encoding like the AS tag
2. instead of FT it now uses the CO tag (COmment) with a `|` separator incase that there already was a comment
3. limit number of warnings to 1000, with one extra warning at 1000 stateing that no more warnings to that effect will be posted.

----

#### Content

#### Review


src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -192,7 +183,11 @@ void close() {
      * @param primaryAlignmentSelectionStrategy What to do when there are multiple primary alignments, or multiple
      *                                          alignments but none primary, for a read or read pair.
      * @param addMateCigar                      True if we are to add or maintain the mate CIGAR (MC) tag, false if we are to remove or not include.
+<<<<<<< 115e723047df28931869e91197c8bce053361463
      * @param unmapContaminantReads             If true, identify reads having the signature of contamination from a foreign organism (i.e. mostly clipped bases),
+=======
+     * @param filterContaminantReads            If true, identify reads having the signature of contamination from a foreign organism (i.e. mostly clipped bases),
+>>>>>>> - added the old alignment information to the unmapped reads as an extra tag
```
nh13: merge garbage

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -488,14 +483,15 @@ public void mergeAlignment(final File referenceFasta) {
...
-     * @param refSeqWalker a ReferenceSequenceWalker that will be used to traverse the reference
+     * @param record              the record to be fixed
+     * @param refSeqWalker        a ReferenceSequenceWalker that will be used to traverse the reference
      * @param isBisulfiteSequence a flag indicating whether the sequence came from bisulfite-sequencing which would imply a different
-     * calculation of the NM tag.
-     *
-     * No return value, modifies the provided record.
+     *                            calculation of the NM tag.
+     *                            <p>
+     *                            No return value, modifies the provided record.
```
nh13: merge garbage

yfarjoun:
sorry about that @nh13 and thanks for looking. I've separated the changes into two commits, 

The first contains the "meat" and the second is only aesthetic changes.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -555,13 +560,37 @@ private void transferAlignmentInfoToFragment(final SAMRecord unaligned, final SA
...
-            log.warn("Record looks like foreign contamination; making unmapped: " + aligned);
-            // NB: for reads that look like contamination, just set unmapped flag and zero MQ but keep other flags as-is.
-            // this maintains the sort order so that downstream analyses can use them for calculating evidence
-            // of contamination vs other causes (e.g. structural variants)
+            if (numForeignContaminantWarnings < MAX_FOREIGN_CONTAM_WARN) {
+                log.warn("Record looks like foreign contamination ; making unmapped: " + aligned);
+                numForeignContaminantWarnings++;
+            }
+
+            if (numForeignContaminantWarnings == MAX_FOREIGN_CONTAM_WARN) {
```
nh13: `else if`

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -86,6 +87,7 @@
     public static final int MAX_RECORDS_IN_RAM = 500000;
 
     private static final char[] RESERVED_ATTRIBUTE_STARTS = {'X', 'Y', 'Z'};
+    private static final int MAX_FOREIGN_CONTAM_WARN = 1_000;
```
nh13: since the output in the tag is `cross-species` perhaps we change `foreign` to `cross-species` throughout?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -555,13 +560,37 @@ private void transferAlignmentInfoToFragment(final SAMRecord unaligned, final SA
...
+                log.warn("Record looks like foreign contamination ; making unmapped: " + aligned);
+                numForeignContaminantWarnings++;
+            }
+
+            if (numForeignContaminantWarnings == MAX_FOREIGN_CONTAM_WARN) {
+                log.warn("No more warnings regarding foreign contamination; limit reached.");
+            }
+
+            // For reads that look like contamination, set unmap. And reset the mapping information as
+            // keeping it violates the sam-spec and is not supported by some formats (CRAM)
```
nh13: I disagree, if it is set to unmapped, then it is perfectly fine if the other fields are set.  From the spec:

    However, an unmapped segment may also have an ordinary coordinate such that it 
    can be placed at a desired position after sorting."

nh13: Also, perhaps CRAM should support this then?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -555,13 +560,37 @@ private void transferAlignmentInfoToFragment(final SAMRecord unaligned, final SA
...
+                    ((Integer) aligned.getMappingQuality()).toString(),
+                    aligned.getStringAttribute(SAMTag.NM.name()))+";");
+
             unaligned.setReadUnmappedFlag(true);
             unaligned.setMappingQuality(SAMRecord.NO_MAPPING_QUALITY);
-            unaligned.setAttribute(SAMTag.FT.name(), "Cross-species contamination");
+            unaligned.setReferenceIndex(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX);
+            unaligned.setAlignmentStart(0);
+            unaligned.setCigar(null);
+            unaligned.setCigarString(SAMRecord.NO_ALIGNMENT_CIGAR);
```
nh13: Do SV callers use the fact that the record is unmapped but has coordinates?  The previous comment makes me concerned that we could be breaking callers relying on this behavior. 👎

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -633,12 +660,13 @@ protected void clipForOverlappingReads(final SAMRecord read1, final SAMRecord re
...
-            if ((!isReservedTag(attr.tag) || this.attributesToRetain.contains(attr.tag)) && !this.attributesToRemove.contains(attr.tag)) {
-                rec.setAttribute(attr.tag, attr.value);
-            }
-        }
+        // Copy over any non-reserved attributes.  attributesToRemove overrides attributesToRetain.
+        alignment.getAttributes().stream()
+                .filter(attr ->
+                        (!isReservedTag(attr.tag) || this.attributesToRetain.contains(attr.tag)) &&
+                                !this.attributesToRemove.contains(attr.tag))
+                .forEach(attr -> rec.setAttribute(attr.tag, attr.value));
```
nh13: Is this really clearer?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -747,9 +775,13 @@ protected void updateCigarForTrimmedOrClippedBases(final SAMRecord rec, final SA
         }
     }
 
-    protected SAMSequenceDictionary getSequenceDictionary() { return this.sequenceDictionary; }
+    protected SAMSequenceDictionary getSequenceDictionary() {
+        return this.sequenceDictionary;
+    }
```
nh13: what's the point of these changes, here and below?

src/main/java/picard/sam/AbstractAlignmentMerger.java
yfarjoun: I talked to the folks who develop CRAM and they seem to want a change in SAM spec for that...but I'm a reasonable person...how about I (optionally) copy the mapping information to a tag and not remove it from the record itself?

yfarjoun:
removed all the "pointless" changes and enabled various strategies for dealing with the alignment information

yfarjoun:
back to you @nh13

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -156,6 +160,29 @@ void close() {
         }
     }
 
+    public enum UNMAPPING_READ_STRATEGY {
+        // Leave on record, and copy to tag
+        COPY_TO_TAG(false,true),
```
nh13: space

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -156,6 +160,29 @@ void close() {
...
+        // Leave on record, but do not create additional tag
+        DO_NOT_CHANGE(false, false),
+        // Add tag with information, and remove from standard fields in record
+        MOVE_TO_TAG(true, true);
+
+        private final boolean resetMappingInformation, populatePATag;
+
+        UNMAPPING_READ_STRATEGY(boolean resetMappingInformation, boolean populatePATag) {
```
nh13: final

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -192,8 +219,10 @@ void close() {
      * @param primaryAlignmentSelectionStrategy What to do when there are multiple primary alignments, or multiple
      *                                          alignments but none primary, for a read or read pair.
      * @param addMateCigar                      True if we are to add or maintain the mate CIGAR (MC) tag, false if we are to remove or not include.
-     * @param unmapContaminantReads             If true, identify reads having the signature of contamination from a foreign organism (i.e. mostly clipped bases),
+     * @param unmapContaminantReads             If true, identify reads having the signature of cross-species contamination (i.e. mostly clipped bases),
      *                                          and mark them as unmapped.
+     * @param unmappingReadsStrategy            An enum desribing how to deal with reads whose mapping information is being removed (currently this happens due to cross-species
```
nh13: - typo in describing.
-  "reads whose mapping information is being removed" -> are being removed?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -205,7 +234,8 @@ public AbstractAlignmentMerger(final File unmappedBamFile, final File targetBamF
                                    final SAMFileHeader.SortOrder sortOrder,
                                    final PrimaryAlignmentSelectionStrategy primaryAlignmentSelectionStrategy,
                                    final boolean addMateCigar,
-                                   final boolean unmapContaminantReads) {
+                                   final boolean unmapContaminantReads,
+                                   final UNMAPPING_READ_STRATEGY unmappingReadsStrategy) {
```
nh13: How does unmapContaminantReads and unmappingReadsStrategy relate now?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -555,17 +588,58 @@ private void transferAlignmentInfoToFragment(final SAMRecord unaligned, final SA
...
+            if (numCrossSpeciesContaminantWarnings < MAX_CROSS_SPECIES_CONTAM_WARN) {
+                log.warn("Record looks like cross-species contamination ; unmapping read: " + aligned);
+                numCrossSpeciesContaminantWarnings++;
+            } else if (numCrossSpeciesContaminantWarnings == MAX_CROSS_SPECIES_CONTAM_WARN) {
+                log.warn("No more warnings regarding cross-species contamination; limit reached.");
+            }
+
+            // For reads that look like contamination, set unmap, and copy the mapping information to
+            // a tag since CRAM seems to throw away the mapping information for
+            // unaligned reads.
```
nh13: Move to method documentation?  Perhaps also include in the description of the enum.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -555,17 +588,58 @@ private void transferAlignmentInfoToFragment(final SAMRecord unaligned, final SA
...
+
+            // Save the old alignment information in a "PA" Tag in case it is needed.
+
```
nh13: newline

src/main/java/picard/sam/AbstractAlignmentMerger.java
nh13: Probably need a comment that unmappingReadsStrategy is ignored if unmapContaminantReads is true

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -86,6 +87,7 @@
     public static final int MAX_RECORDS_IN_RAM = 500000;
 
     private static final char[] RESERVED_ATTRIBUTE_STARTS = {'X', 'Y', 'Z'};
+    private static final int MAX_CROSS_SPECIES_CONTAM_WARN = 1_000;
```
nh13: This works?

src/test/java/picard/sam/MergeBamAlignmentTest.java
```diff
@@ -1351,6 +1349,7 @@ private void doMergeAlignment(final File unmappedBam, final List<File> alignedBa
         }
         if (unmapContaminantReads != null) {
             args.add("UNMAP_CONTAMINANT_READS=" + unmapContaminantReads);
+            args.add("UNMAPPED_READ_STRATEGY=MOVE_TO_TAG");
```
nh13: Should the other strategies be tested?

src/main/java/picard/sam/AbstractAlignmentMerger.java
yfarjoun: yes!

yfarjoun:
back to you @nh13 with more tests. Thanks for your time and comments.

yfarjoun:
@nh13 are the latest changes acceptable?

yfarjoun:
@tfenne would you be available to review this? This is blocking a call-set that you also might be interested in... 😄

src/main/java/picard/sam/AbstractAlignmentMerger.java
tfenne: Since Java 8.  Syntax isn't supported in scala though (but there you can write `1e3.toInt`).

tfenne: Also, is there really value in even 1,000 warnings?  I would think that after about 20-100 I'd want them to stop cluttering up my log file.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -156,6 +160,29 @@ void close() {
         }
     }
 
+    public enum UNMAPPING_READ_STRATEGY {
```
tfenne: Why is this all upper case?  Enums are classes, so this should really be `UnmappingReadStrategy`.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -156,6 +160,29 @@ void close() {
...
+        // Leave on record, and copy to tag
+        COPY_TO_TAG(false, true),
+        // Leave on record, but do not create additional tag
+        DO_NOT_CHANGE(false, false),
+        // Add tag with information, and remove from standard fields in record
+        MOVE_TO_TAG(true, true);
+
+        private final boolean resetMappingInformation, populatePATag;
```
tfenne: Should be `populatePaTag`

src/main/java/picard/sam/AbstractAlignmentMerger.java
tfenne: So ... since this breaks the API and callers will have to add the new parameter anyway, why not just do away with the `unmapContaminantReads` parameter?  Also, since `unmappingReadsStrategy` only applies to highly-clipped likely-contaminant reads, could that be reflected in the name.  E.g. `contaminantUnmappingStrategy`?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -537,10 +568,20 @@ private HitsForInsert nextAligned() {
         return null;
     }
 
+    private int numCrossSpeciesContaminantWarnings = 0;
+
     /**
      * Copies alignment info from aligned to unaligned read, clips as appropriate, and sets PG ID.
      * May also un-map the resulting read if the alignment is bad (e.g. no unclipped bases).
      *
+     * depending on the value of unmappingReadsStrategy this function will potentially
```
tfenne: Capitalize depending.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -537,10 +568,20 @@ private HitsForInsert nextAligned() {
...
+     *
+     * - reset the mapping information (MOVE_TO_TAG)
+     * - copy the original mapping information to a tag "PA" (MOVE_TO_TAG, or COPY_TO_TAG)
+     * a third possibility for unmappingReadsStrategy is DO_NOT_CHANGE
```
tfenne: `- do nothing (DO_NOT_CHANGE)`

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -555,17 +596,55 @@ private void transferAlignmentInfoToFragment(final SAMRecord unaligned, final SA
...
+            if (unmappingReadsStrategy.isPopulatePATag()) {
+                unaligned.setAttribute("PA", encodeMappingInformation(aligned));
+            }
+
+
+            if (unmappingReadsStrategy.isResetMappingInformation()) {
+                unaligned.setReferenceIndex(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX);
+                unaligned.setAlignmentStart(0);
+                unaligned.setCigar(null);
+                unaligned.setCigarString(SAMRecord.NO_ALIGNMENT_CIGAR);
```
tfenne: Should it also make an attempt to clean out optional tags that are mapping related?  E.g. `SA`?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -555,17 +596,55 @@ private void transferAlignmentInfoToFragment(final SAMRecord unaligned, final SA
...
+                log.warn("Record looks like cross-species contamination ; unmapping read: " + aligned);
+                numCrossSpeciesContaminantWarnings++;
+            } else if (numCrossSpeciesContaminantWarnings == MAX_CROSS_SPECIES_CONTAM_WARN) {
+                log.warn("No more warnings regarding cross-species contamination; limit reached.");
+            }
+
+            if (unmappingReadsStrategy.isPopulatePATag()) {
+                unaligned.setAttribute("PA", encodeMappingInformation(aligned));
+            }
+
```
tfenne: Extra newline.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -555,17 +596,55 @@ private void transferAlignmentInfoToFragment(final SAMRecord unaligned, final SA
...
+
+            if (unmappingReadsStrategy.isResetMappingInformation()) {
+                unaligned.setReferenceIndex(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX);
+                unaligned.setAlignmentStart(0);
```
tfenne: Please use `SAMRecord.NO_ALIGNMENT_START`

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -555,17 +596,55 @@ private void transferAlignmentInfoToFragment(final SAMRecord unaligned, final SA
...
+                unaligned.setCigar(null);
+                unaligned.setCigarString(SAMRecord.NO_ALIGNMENT_CIGAR);
+            }
+
             unaligned.setReadUnmappedFlag(true);
+            // Unmapped read cannot have non-zero mapping quality and remain valid
             unaligned.setMappingQuality(SAMRecord.NO_MAPPING_QUALITY);
-            unaligned.setAttribute(SAMTag.FT.name(), "Cross-species contamination");
+
```
tfenne: Extra newline.

yfarjoun:
Thanks for the comments and suggestions @tfenne. Regarding the "one enum to rule them all". I am trying to avoid API changes at the CLI level and willing to suffer API changes at the code level. My thoughts are that if you currently use MBA with options UNMAP_CONTAMINANT_READS=true the behavior shouldn't change.  

I think that it make sense to keep this design because there might be other reasons to unmap reads, and the question of what to do with unmapped reads is separate from the question of _whether_ to unmap them in the first place. 

For example, we have recently found that STAR aligner is overzealous in its mapping and it (sometimes) maps adapter dimers to the reference. I am considering putting in an option to unmap such reads and the same question arrises: how to deal with the information in the reads that are being unmapped.

src/main/java/picard/sam/AbstractAlignmentMerger.java
yfarjoun: hmmm. any ideas on how to store it elsewhere? Should it be copied to PA as well (with some special formatting) or yet another new tag...?

tfenne: I was thinking that it would just null out those attributes.  But I see your point that other values are placed into your `PA` tag.

tfenne:
@yfarjoun ultimately it's up to you.  My 2c is that `MergeBamAlignment` and the implementing classes have already become too complicated, and I dislike the fact that if I use them elsewhere (e.g. using SamAlignmentMerger directly) that I have to figure out and supply ~20 constructor parameters many of which have sensible defaults, but those only exist in MergeBamAlignment.

yfarjoun:
I agree with your points about the complexity of MBA, but I do not think
that I can fix it in this PR.

Ideally, Java would have named parameters like scala and default
values...but, alas it doesn't.

yfarjoun:
@tfenne I tried to address all your comments. I avoided breaking the API by including default constructors that use the "DO_NOT_CHANGE" enum value. I hope that this is a reasonable compromise.

yfarjoun:
@tfenne This is waiting for a final OK from you...

tfenne:
@yfarjoun Didn't realize you were waiting on me.  👍

PR #702, opened 2016-12-08
CollectIlluminaBasecallingMetrics supports an alternative barcodes

nh13:
... directory.

If ExtractIlluminaBarcodes is run with a different output directory
(OUTPUT_DIR) than the basecalling directory,
CollectIlluminaBasecallingMetrics will fail.  Add an option BARCODES_DIR
to specify the alternate barcodes directory.


nh13:
@yfarjoun can you take a quick peek as I'd like this to go out in the next release.  Without this change, we cannot run more than one analysis that extracts barcodes differently since it would write to the original basecalls directory.  This also is a problem if the basecalls directory is read-only.

PR #704, opened 2016-12-09
Show HTSJDK version in stderr when running

yfarjoun:
### Description

Reports the htsjdk version that the jar was compiled with at the top of a run.

----

#### Content

#### Review


yfarjoun:
I don't know how to test this. I tested manually, ideas are welcome.

lbergelson:
annoyingly it doesn't seem like assigned PR reviews show up in any of the lists of assigned things on github... sorry I didn't see this

src/main/java/picard/cmdline/CommandLineParser.java
```diff
@@ -235,9 +238,9 @@ private String getUsagePreamble() {
 
         if (null != this.programVersion && 0 < this.programVersion.length()) {
             usagePreamble += "Version: " + getVersion() + "\n";
+            usagePreamble += "HTSJDK Version: " + getHtsjdkVersion() + "\n";
```
lbergelson: Instead of adding calls to getHtsjdkVersion everywhere that getVersion is called, I would change getVersion() to return a string that includes both the picard version and the htsjdk version and expose a separate getPicardVersion method that does what the old getVersion did. Otherwise it will end up being inconsistently reported.

lbergelson: Alternatively, you should probably update the version reporting on line 325 and 497 as well as the definition of `CommandLineProgram.getVersion()`

gradlew
```diff
@@ -161,4 +161,9 @@ function splitJvmOpts() {
 eval splitJvmOpts $DEFAULT_JVM_OPTS $JAVA_OPTS $GRADLE_OPTS
 JVM_OPTS[${#JVM_OPTS[*]}]="-Dorg.gradle.appname=$APP_BASE_NAME"
 
+# by default we should be in the correct project dir, but when run from Finder on Mac, the cwd is wrong
```
lbergelson: Did you mean to update the gradle wrapper in this pr?  It's not a bad idea to do so, so I think it's fine to keep the update here.

lbergelson:
@yfarjoun This is a surprisingly annoying problem that you're tackling.  I had never really looked into it before, but it turns out the current solution for even getting the picard version is extremely brittle.  It depends on looking up the implementation version in manifest file.  However, a jar file can only have a single manifest, which means that anyone packaging an uber jar with picard in it will likely provide there own implementation version, which picard will happily report as it's own version number.  

There's a solution to this, you can provide per-package versions in the manifest file if you build it carefully.  However, since the java packaging scheme expects company/package, and picard and htsjdk are top level packages, these are interpreted as company names and can't be assigned package version (as far as I can tell by experimenting with munging that manifest files by hand...)

Your solution of just embedding a new value in the manifest seems like a reasonable one, but it's brittle if the manifest get replaced during repackaging.  It might be better if we embed different uniquely named resource files that include the versions in order to survive repackaging by library consumers.  If we were going to do that we should probably make the change at the htsjdk as well as in picard.

nh13:
It seems we have lost this in the move to gradle.

The previous ant build wrote the version to a file:
https://github.com/samtools/htsjdk/blob/f871ec9f121bb5de8bdd31b4b282394ad1960376/build.xml#L45-L46
https://github.com/samtools/htsjdk/blob/f871ec9f121bb5de8bdd31b4b282394ad1960376/build.xml#L64-L71

Then Picard could pick it up:
https://github.com/broadinstitute/picard/blob/9a00c87b7ffdb01cfb5a0d6e76556146196babb8/build.xml#L119-L133
and add it as an attribute in the manifest:
https://github.com/broadinstitute/picard/blob/9a00c87b7ffdb01cfb5a0d6e76556146196babb8/build.xml#L327

lbergelson:
@nh13 Thanks for pointing that out. Why add it to the manifest instead of reading it directly from the file?

nh13:
Having it in the manifest makes it easy to check the version # in any Picard JAR you are given.  I am not sure though what you mean by reading it directly from the file.  It does read it from the file.  Or do you mean in the PicardCommandLineProgramClassThingy?

lbergelson:
I meant why bake it into the manifest during the build instead of just reading the it from the htsjdk version file at runtime?  If you bake it into the manifest then you're still vulnerable to the manifest being overwritten, which is the problem you're trying to avoid by including a version file instead of reading it from the htsjdk manifest.

nh13:
I don't recall the reason for the decision, but _it just works_.  Note that the properties file needs to be baked into the htsjdk jar and picard jar for your suggestion to work.  Anyhow, I was just trying to show that the functionality was there previously for a properties file.  I have nothing else of use to add.

yfarjoun:
Presumably one can also add random files to the jar, as we do with the R
scripts...so the only issue will be to find them from the CommandlineParser
afterwards...hrm

yfarjoun:
I don't think that my approach in this PR this is the right one. I'll let someone else tackle this.

PR #707, opened 2016-12-12
DO NOT MERGE

nh13:
Example implementation supporting marking illumina adapters containing inline molecular barcodes to start a discussion of how to deal with them with regards to adapter marking.  

If the molecular barcodes are at the start of R1 and R2, then we will read `<adapter><mol-barcode><read>`.  This implementation does not support the case where the barcode is only on one end of a pair (easily fixed).  Before writing tests, I want to make sure there aren't other cases I am missing.

@yfarjoun @tfenne want to chime in here?

### DIFF
--- a/src/main/java/picard/illumina/MarkIlluminaAdapters.java
+++ b/src/main/java/picard/illumina/MarkIlluminaAdapters.java
@@ -54,6 +54,7 @@ import picard.util.ClippingUtility;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import static picard.util.IlluminaUtil.IlluminaAdapterPair;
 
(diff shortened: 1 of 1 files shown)


yfarjoun:
this seems reasonable, though I wonder if there's a better, more generic way. 

one of the problems with Picard's identification of adapters is that it need an explicit length on the indexes. We use 8 Ns, but some sample indexes are of different length I wonder if there's a way to make an annotation for an expandable region which could then contain sample indexes of various lengths or, alternatively UMIs....

nh13:
@yfarjoun I suppose we would need to know _for each read_ how many bases are between the end of the adapter and the start of the template bases.  Perhaps another solution is to add a SAM tag (ducks) to each read that gives the number of bases removed the start of the read (to use for the other end).  This would allow us to customize the adapter pair on a per-read basis.  I am assuming in this case, that inline sample and molecular barcodes are removed prior to adapter trimming.  And yes, I have seen data where the number of molecular barcode bases differs between source molecules within a sample.

yfarjoun:
I think it would be good to start with a constant (not per-read) modification of the adapter sequence. I would advocate for a way to add different number of bases to read1 and read2 and also to expand/contract (!) the read1/read2 stretch of Ns that is the stand-in for the sample-barcode. this seems especially important given that we are using the Broad's convention of 8 basepair sample-barcode whereas the "standard" illumina ones have 6 basepairs. Also, if folks put UMI inline with the sample barcode (and risk swapping UMIs...CAVAET EMPTOR) they will need to change the number of "N"s in that stretch.

nh13:
Closing since this is not a priority.

kbergin:
To clean up this repo, we are deleting branches that have been merged or closed. You have a short period of time to restore this branch if needed. Thanks!

Issue #708, opened 2016-12-12
Overlapping reads are not clipped in MergeBamAlignment if soft-clipping is present


nh13:
If a record (ex. 127M14S below) already has soft-clipping and needs to be further soft-clipped, the calculation for "from where" to soft-clip in the read does not take into account the already soft-clipped bases.  It calculates the amount one read goes past the other based on the soft-clipped ends (or starts), but uses the full read length (not the aligned portion of the read).

See:
https://github.com/broadinstitute/picard/blob/master/src/main/java/picard/sam/AbstractAlignmentMerger.java#L608-L609
https://github.com/broadinstitute/picard/blob/master/src/main/java/picard/sam/AbstractAlignmentMerger.java#L614-L615

genome: hg38DH
test SAM records:
```
READ    83  chr1    114713831   60  22S119M =   114713831   -119    NCTCTTCCGATCTGACCTCAATTATTATTGATGGCAAATACACAGAGGAAGCCTTCGCCTGTCCTCATGTATTGGTCTCTCATGGCACTGTACTCTTCTTGTCCAGCTGTATCCAGTATGTCCAACAAACAGGTTTCACCA   #{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{
READ    163 chr1    114713831   60  127M14S =   114713831   119 TATTATTGATGGCAAATACACAGAGGAAGCCTTCGCCTGTCCTCATGTATTGGTCTCTCATGGCACTGTACTCTTCTTGTCCAGCTGTATCCAGTATGTCCAACAAACAGGTTTCACCATATATAACTAGATCGGAAGAGC   {{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{{
```

vdauwera:
Looks like this was fixed by #709. Reopen if I'm wrong.

PR #709, opened 2016-12-12
Fix to handle overlapped read clipping in MergeBamAlignment when clipping already present

tfenne:
### Description

Fixed the clipping of overlapping reads which is currently broken in the case where the reads _already_ contain soft-clipping.  If the reads contain softclipping at their 3' ends the overlapping read clipping should _add_ to that soft-clipping in order to stop the reads alignment's extending past each other's 5' ends.  Prior to the change, the existing soft-clipping was ignored, and hence the result was incorrect.

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/sam/AbstractAlignmentMerger.java
+++ b/src/main/java/picard/sam/AbstractAlignmentMerger.java
@@ -58,6 +58,8 @@ import picard.PicardException;
 
 import java.io.File;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 
 /**
@@ -590,11 +592,10 @@ public abstract class AbstractAlignmentMerger {
      * Checks to see whether the ends of the reads overlap and soft clips reads
      * them if necessary.
      */
-    protected void clipForOverlappingReads(final SAMRecord read1, final SAMRecord read2) {
+    protected static void clipForOverlappingReads(final SAMRecord read1, final SAMRecord read2) {
         // If both reads are mapped, see if we need to clip the ends due to small
         // insert size
         if (!(read1.getReadUnmappedFlag() || read2.getReadUnmappedFlag())) {
-
             if (read1.getReadNegativeStrandFlag() != read2.getReadNegativeStrandFlag()) {
                 final SAMRecord pos = (read1.getReadNegativeStrandFlag()) ? read2 : read1;
                 final SAMRecord neg = (read1.getReadNegativeStrandFlag()) ? read1 : read2;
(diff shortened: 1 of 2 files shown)


src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -590,11 +592,10 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
      * Checks to see whether the ends of the reads overlap and soft clips reads
      * them if necessary.
      */
-    protected void clipForOverlappingReads(final SAMRecord read1, final SAMRecord read2) {
+    protected static void clipForOverlappingReads(final SAMRecord read1, final SAMRecord read2) {
```
nh13: while we're here, can this be private?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -605,23 +606,35 @@ protected void clipForOverlappingReads(final SAMRecord read1, final SAMRecord re
...
             }
         }
+    }
+
+    /** Returns the number of soft-clipped bases until a non-soft-clipping element is encountered. */
+    private static int lengthOfSoftClipping(Iterator<CigarElement> iterator) {
+        int clipped = 0;
+        while (iterator.hasNext()) {
+            final CigarElement elem = iterator.next();
+            if (elem.getOperator() != CigarOperator.SOFT_CLIP) break;
```
nh13: I think that this will do the wrong thing if we have hard clips (ex. `120M12S8H`), namely break too early.  I think that's the only one though.

nh13:
Original bug report: https://github.com/broadinstitute/picard/issues/708

nh13:
I can confirm this fixes the test case on #708

src/main/java/picard/sam/AbstractAlignmentMerger.java
tfenne: No, it needs to be protected or package level for the tests to be able to access it.

src/main/java/picard/sam/AbstractAlignmentMerger.java
tfenne: Ah, good catch.  I've fixed that.  The only remaining issue is that it will lose the hard-clipping due to the issue here:
    https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/samtools/util/CigarUtil.java#L46

The cigar will be correctly soft-clipped, but the hard-clip will get lost.  I'm not inclined to try and fix that at the moment unless you think it's critical.

tfenne:
@nh13 Second commit pushed to handle the hard-clipping.  It's not ideal though, so do you want to take a second look?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -630,8 +630,8 @@ private static int lengthOfSoftClipping(Iterator<CigarElement> iterator) {
         int clipped = 0;
         while (iterator.hasNext()) {
             final CigarElement elem = iterator.next();
-            if (elem.getOperator() != CigarOperator.SOFT_CLIP) break;
-            clipped = elem.getLength();
+            if (elem.getOperator() != CigarOperator.SOFT_CLIP && elem.getOperator() != CigarOperator.HARD_CLIP) break;
+            if (elem.getOperator() == CigarOperator.SOFT_CLIP) clipped = elem.getLength();
```
nh13: I think this is fine that we will lose the hard clip, since methods in `CigarUtil` also note this.  Perhaps add a TODO here and where this function is used that it doesn't handle hard clips?

nh13:
@tfenne one minor comment about adding a note about losing the hard clips and then merge at will.

tfenne:
@ktibbett Do you want to take a peek at this at all since it affects `MergeBamAlignment`?  It is a critical fix for anyone who is using UMIs inline in the template reads.  Without this, if you have reads that are already soft clipped (because of adapters being identified), then the clipping back of reads that run off the insert doesn't work correctly.  @nh13 and I would really like to get this merged ASAP but figure you should at least be aware of it.

PR #712, opened 2016-12-14
Default output for CreateSequenceDictionary

magicDGS:
### Description

`CreateSequenceDictionary` requires always to specified the output name, but GATK programs assumes that the file name is the FASTA name without the extension and ending in __.dict__. Thus, I included the following changes:
* Optional output argument with updated documentation for default behavior.
* Sets the output name as in the GATK framework (port of https://github.com/broadinstitute/gatk/pull/2243) if no one is provided.
* Test for default output name.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/sam/CreateSequenceDictionary.java
+++ b/src/main/java/picard/sam/CreateSequenceDictionary.java
@@ -23,6 +23,7 @@
  */
 package picard.sam;
 
+import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.SAMSequenceDictionaryCodec;
 import htsjdk.samtools.SAMSequenceRecord;
@@ -32,6 +33,7 @@ import htsjdk.samtools.reference.ReferenceSequenceFileFactory;
 import htsjdk.samtools.util.AsciiWriter;
 import htsjdk.samtools.util.CloseableIterator;
 import htsjdk.samtools.util.IOUtil;
+import htsjdk.samtools.util.Log;
 import htsjdk.samtools.util.Md5CalculatingOutputStream;
 import htsjdk.samtools.util.RuntimeIOException;
 import htsjdk.samtools.util.SortingCollection;
@@ -50,6 +52,7 @@ import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -146,6 +146,14 @@ public SAMSequenceDictionary makeSequenceDictionary(final File referenceFile) {
         if (URI == null) {
             URI = "file:" + REFERENCE.getAbsolutePath();
         }
+        if (OUTPUT == null) {
+            // determine the name for the dict file in the same way as CachingIndexedFastaSequenceFile.checkAndCreate
```
yfarjoun: You seem to be referencing a GATK class, but this is Picard...so rather confusing. 

Also, it look s as though this is a good target for refactoring. Could you please move this to a function in ReferenceSequenceFileFactory?

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -146,6 +146,14 @@ public SAMSequenceDictionary makeSequenceDictionary(final File referenceFile) {
...
+            final String name = REFERENCE.getName();
+            final String fastaExt = ReferenceSequenceFileFactory.FASTA_EXTENSIONS.stream()
+                    .filter(name::endsWith).findFirst().orElseGet(() -> "");
+            OUTPUT = new File(REFERENCE.getParentFile(),
+                    REFERENCE.getName().replace(fastaExt, IOUtil.DICT_FILE_EXTENSION));
```
yfarjoun: I think that this has a bug (also the implementation in GATK, but I'm reveiwing this one.)

to show I'm wrong please have a test with the following a fasta files:

`break.txt.txt -> break.txt.dict` (rather than `break.dict.fasta` or `break.dict.dict`)
`break.fasta.fasta -> break.fasta.dict`

yfarjoun: Also, in this case, please emit a info or warning message informing the user of the choice of output name.

magicDGS: Yes, it is a bug.

magicDGS: Added info logging for the output file in the case of automatically detected.

src/main/java/picard/sam/CreateSequenceDictionary.java
magicDGS: Removed the mention to the GATK class, and added a package-protected method with a TODO to pass to HTSJDK.

magicDGS:
I did a PR in HTSJDK for this, @yfarjoun. But in the meanwhile I implemented the solution here with a package protected method (maybe will be better to make it private first, to avoid usages and try to keep them into HTSJDK).

yfarjoun:
Thanks. Looks good now. Please add a ticket the replace this implementation with the HTSJDK one when it goes in.

yfarjoun:
👍

magicDGS:
Thnaks!

PR #713, opened 2016-12-14
New approach to splitting on sub-group in EstimateLibraryComplexity 

SilinPavel:
We introduce additional improvements in ELC algorithm, which are directed to increase diversity of hash value of PairedReadSequences.

# Problem
About the potential problem:
If we look at the group which contains about 100k PairedReadSequence and many of them have similar parts, according to current algorithm we will have reads which are similar to many other reads, and in this case we will lose main profit of current approach.

image
In this example using current approach all 4 reads would be considered similar and eventually would undergo complete check, while actually they can have a lot of difference in other parts.

For example, this file is a group which contains PRS which are very similar in the one of parts but aren't duplicates
[0_241500.zip](https://github.com/broadinstitute/picard/files/652042/0_241500.zip)

It is a list of PRS, you can unzip this file and decode it by PairedReadSequencesCodec from EsimateLibraryComplexity.

# Solving
We want to propose a new approach to  calculate "hashes" that would result in another splitting on sub-groups:

image

The main idea is build one "hash" of read from bases located within equal distance from each other.  
This approach provides more complex structure for the ﻿"hashes" and splitting on groups becomes more uniform.
Our new algorithm processes the group from the file above in less than a minute while old version of splitting on groups can not cope with this at all.
**Please note that it's case very unusual but can provide big trouble if you find it.**

**Before:**
image

**After:**
image

### DIFF
--- a/src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
+++ b/src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
@@ -42,6 +42,9 @@ class ElcHashBasedDuplicatesFinder extends ElcDuplicatesFinder {
 
     // We split each read on several parts, each part represent set of nucleotides,
     // which behind each other on numberOfHashesInGroup.
+    // f. i.:
+    // 1 2 3 4 1 2 3 4 1 2 3 4
+    // A C A T T A C G G A T T
     private int numberOfHashesInGroup = -1; //initial value
     private int minReadLenInGroup = Integer.MAX_VALUE; //initial value
 
@@ -78,12 +81,12 @@ class ElcHashBasedDuplicatesFinder extends ElcDuplicatesFinder {
         }
     }
 
-    private Set<PairedReadSequence> getSimilarReads(PairedReadSequence lhs) {
+    private Set<PairedReadSequence> getSimilarReads(final PairedReadSequence pattern) {
         final Set<PairedReadSequence> toCheck = new HashSet<>();
-        for (int[] hashesForRead: new int[][]{lhs.hashes1, lhs.hashes2}){
+        for (int[] hashesForRead: new int[][]{pattern.hashes1, pattern.hashes2}) {
             for (int hash : hashesForRead) {
                 List<PairedReadSequence> readsWithSameHash = readsByHashInGroup.get(hash);
-                // if size == 1, then this list contains only lhs read
+                // if size == 1, then this list contains only pattern read
                 if (readsWithSameHash.size() > 1) {
                     toCheck.addAll(readsWithSameHash);
                 }
@@ -106,7 +109,7 @@ class ElcHashBasedDuplicatesFinder extends ElcDuplicatesFinder {
             int[][] readHashValues = {prs.hashes1, prs.hashes2};
             for (int[] readHashValue : readHashValues) {
                 for (int key : readHashValue) {
-                    List<PairedReadSequence> dupCandidates
+                    final List<PairedReadSequence> dupCandidates
                             = readsByHashInGroup.computeIfAbsent(key, k -> new ArrayList<>());
                     dupCandidates.add(prs);
                 }
--- a/src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
+++ b/src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
@@ -40,15 +40,24 @@ import static picard.sam.markduplicates.EstimateLibraryComplexity.PairedReadSequ
  */
 class ElcHashBasedDuplicatesFinder extends ElcDuplicatesFinder {
 
-    // We split each read on several parts, each part represent set of nucleotides,
-    // which behind each other on numberOfHashesInGroup.
+    // We split the bases of each read into interspersed parts, so that each part spans the entire length of the read.
+    // e.g. breaking a read of length 12 into 4 parts would be as follows:
     // f. i.:
     // 1 2 3 4 1 2 3 4 1 2 3 4
     // A C A T T A C G G A T T
-    private int numberOfHashesInGroup = -1; //initial value
-    private int minReadLenInGroup = Integer.MAX_VALUE; //initial value
 
-    private Map<Integer, List<PairedReadSequence>> readsByHashInGroup;
+    // number of parts we split reads in
+    // -1 because we want to find biggest number of hashes in this group
+    // calculated formula : ((minReadLength - minIdenticalBases) * maxDiffRate) + 1
+    // when minReadLength = Math.min(Math.min(prs.read1.length, prs.read2.length), maxReadLength)
+    private int numberOfHashesInGroup = -1;
+
+    // minimal read length in group, this variable necessary for splitting reads on part,
+    // see EstimateLibraryComplexity.PairedReadSequence.getHashes()
+    // initial value Integer.MAX_VALUE
+    private int minReadLenInGroup = Integer.MAX_VALUE;
+
+    private final Map<Integer, List<PairedReadSequence>> readsByHashInGroup;
 
     ElcHashBasedDuplicatesFinder(double maxDiffRate, int maxReadLength, int minIdenticalBases,
                                  OpticalDuplicateFinder opticalDuplicateFinder) {
@@ -181,9 +190,9 @@ class ElcHashBasedDuplicatesFinder extends ElcDuplicatesFinder {
         int errors = 0;
         final int minReadLength = minLength(read1, read2);
 
-        for (int k = 0; k < numberOfHashesInGroup; ++k) {
-            if (hashes1[k] != hashes2[k]) {
-                errors += compareHashes(read1, read2, k);
+        for (int hashNumber = 0; hashNumber < numberOfHashesInGroup; ++hashNumber) {
+            if (hashes1[hashNumber] != hashes2[hashNumber]) {
+                errors += compareHashes(read1, read2, hashNumber);
                 if (errors > maxErrors) {
                     return errors;
                 }
@@ -202,13 +211,15 @@ class ElcHashBasedDuplicatesFinder extends ElcDuplicatesFinder {
      *
      * @return errors number for current part of the read
      */
-    private int compareHashes(byte[] read1, byte[] read2, int k) {
+    private int compareHashes(byte[] read1, byte[] read2, int hashNumber) {
         int errors = 0;
-        int position = minIdenticalBases + k;
+        //shift position corresponding to hash we inspect
+        int position = minIdenticalBases + hashNumber;
         while (position < minReadLenInGroup) {
             if (read1[position] != read2[position]) {
                 errors++;
             }
+            // shift position to next base in the hash
             position += numberOfHashesInGroup;
         }
         return errors;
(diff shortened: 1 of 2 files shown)


src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
```diff
@@ -40,13 +40,17 @@
  */
 class ElcHashBasedDuplicatesFinder extends ElcDuplicatesFinder {
 
-    // We split each read on several parts and each part has length hashLength.
-    // hashLength = (min read length) / (number of errors + 1)
-    private int hashLength = Integer.MAX_VALUE;
+    // We split each read on several parts, each part represent set of nucleotides,
+    // which behind each other on numberOfHashesInGroup.
```
yfarjoun: comment is unclear. please reword

src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
```diff
@@ -58,92 +62,85 @@ void searchDuplicates(List<PairedReadSequence> sequences, Histogram<Integer> dup
...
             }
-
             dupSet.addAll(dupes);
             fillHistogram(duplicationHisto, opticalHisto, lhs, dupes);
         }
     }
 
+    private Set<PairedReadSequence> getSimilarReads(PairedReadSequence lhs) {
+        final Set<PairedReadSequence> toCheck = new HashSet<>();
+        for (int[] hashesForRead: new int[][]{lhs.hashes1, lhs.hashes2}){
```
yfarjoun: space ) {

src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
```diff
@@ -58,92 +62,85 @@ void searchDuplicates(List<PairedReadSequence> sequences, Histogram<Integer> dup
...
                     dupes.add(rhs);
                 }
             }
-
             dupSet.addAll(dupes);
             fillHistogram(duplicationHisto, opticalHisto, lhs, dupes);
         }
     }
 
+    private Set<PairedReadSequence> getSimilarReads(PairedReadSequence lhs) {
```
yfarjoun: why `lhs` for the name of the variable? please use something more fitting. also, `final`

src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
```diff
@@ -58,92 +62,85 @@ void searchDuplicates(List<PairedReadSequence> sequences, Histogram<Integer> dup
...
             int[][] readHashValues = {prs.hashes1, prs.hashes2};
             for (int[] readHashValue : readHashValues) {
                 for (int key : readHashValue) {
-                    List<PairedReadSequence> dupCandidates = readsByHash.get(key);
-
-                    if (dupCandidates == null) {
-                        dupCandidates = new ArrayList<>();
-                        readsByHash.put(key, dupCandidates);
-                    }
+                    List<PairedReadSequence> dupCandidates
```
yfarjoun: final

src/main/java/picard/sam/markduplicates/EstimateLibraryComplexity.java
```diff
@@ -210,26 +207,31 @@ public static int getSizeInBytes() {
...
-
-                // use custom hash to avoid using System.arraycopy
+        // Split read by numberOfHashes parts and hash each part
+        // For instance:
+        //        0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19
+        // read = A C G T A C G T A C G  T  A  C  G  T  A  C  G  T
+        // numberOfHashes = 5
+        // skippedBases = 1
+        // minReadLength = 15
+        // So, method returns hashValues with 5 hash value
```
yfarjoun: unless I am misunderstanding something, the example in the comment does not make use of skippedBases=1

src/main/java/picard/sam/markduplicates/EstimateLibraryComplexity.java
```diff
@@ -210,26 +207,31 @@ public static int getSizeInBytes() {
...
+        // first value calculated from read[0], read[5], read[10]
+        // second value calculated from read[1], read[6], read[11]
+        // etc.
+        // chars from 15 to 19 position is a tail, see compareTails() in ElcHashBasedDuplicatesFinder
+        private int[] getHashes(byte[] read, int numberOfHashes, int skippedBases, int minReadLength) {
+            int[] hashValues = new int[numberOfHashes];
```
yfarjoun: final

yfarjoun:
What format is the attached file? I couldn't read it properly.

is the problem due to the bases at the ends of the read being of lower quality (and being effectively random)? And so in the first approach you would get through the batch of hashes in the beginning and then land on the "bad" bases at the end? 

I wonder if your example bam has especially long reads and this is the source of the problem. I think that ELC was written at a time when reads were short and references were scarce. Now, reads are long and references are abundant...perhaps it makes sense to limit the number of bases compared for the purpose of declaring reads "duplicates"...

SilinPavel:
@yfarjoun I added some correction, please check!

src/main/java/picard/sam/markduplicates/EstimateLibraryComplexity.java
SilinPavel: @yfarjoun correct, it's my bad, I changed it.

src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
SilinPavel: I added a example, is it enough?

SilinPavel:
@yfarjoun 
1) You can read this file by PairedReadCodec from ELC, but you have to unzip it before. 
2) Actually, we don't know which nature this problem has, but we can research it. If you can provide us some directions of researching, it would be great. However, your assumption makes sense. In the new approach, we "diluted" hashes with nucleotides that are in different parts of a read.
3) In our case reads in the file has a length approximately equals to 150bp and also have the same length (I mean that our sample don't has reads with a length with 100 and 150)

src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
```diff
@@ -40,13 +40,20 @@
  */
 class ElcHashBasedDuplicatesFinder extends ElcDuplicatesFinder {
 
-    // We split each read on several parts and each part has length hashLength.
-    // hashLength = (min read length) / (number of errors + 1)
-    private int hashLength = Integer.MAX_VALUE;
+    // We split each read on several parts, each part represent set of nucleotides,
```
yfarjoun: I would word it as follows: "We split the bases of each read into interspersed parts, so that each part spans the entire length of the read. e.g. breaking a read of length 12 into 4 parts would be as follows:"

src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
```diff
@@ -40,13 +40,20 @@
...
+    // which behind each other on numberOfHashesInGroup.
+    // f. i.:
+    // 1 2 3 4 1 2 3 4 1 2 3 4
+    // A C A T T A C G G A T T
+    private int numberOfHashesInGroup = -1; //initial value
```
yfarjoun: rather than "initial value" please comment about the purpose of this variable. in this case I think it is the number of different groups the bases are split into"

src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
```diff
@@ -40,13 +40,20 @@
...
+    private int minReadLenInGroup = Integer.MAX_VALUE; //initial value
```
yfarjoun: again, please comment about the purpose of this variable.

src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
```diff
@@ -40,13 +40,20 @@
...
+
+    private Map<Integer, List<PairedReadSequence>> readsByHashInGroup;
```
yfarjoun: can be final, right?

src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
```diff
@@ -204,7 +202,24 @@ private int compareReadToRead(byte[] read1, int[] hashes1, byte[] read2, int[] h
      *
      * @return errors number for current part of the read
      */
-    private int compareHashes(byte[] read1, byte[] read2, int start, int stop) {
+    private int compareHashes(byte[] read1, byte[] read2, int k) {
```
yfarjoun: what is `k`?

src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
```diff
@@ -58,92 +65,85 @@ void searchDuplicates(List<PairedReadSequence> sequences, Histogram<Integer> dup
...
             int[][] readHashValues = {prs.hashes1, prs.hashes2};
             for (int[] readHashValue : readHashValues) {
                 for (int key : readHashValue) {
-                    List<PairedReadSequence> dupCandidates = readsByHash.get(key);
-
-                    if (dupCandidates == null) {
-                        dupCandidates = new ArrayList<>();
-                        readsByHash.put(key, dupCandidates);
-                    }
+                    final List<PairedReadSequence> dupCandidates
```
yfarjoun: this functions seems to be useless....I must be missing something. you are creating `dupCandidates` and adding prs to it...but then you return void and the object is cleaned up...what's going on ?

SilinPavel: Actually, it's not cleaned because now it is in readsByHashInGroup map, and this list will contain all reads with identical hashes, it means that this reads will be checked for duplication.

SilinPavel:
@yfarjoun done, take a look, please!

src/main/java/picard/sam/markduplicates/ElcHashBasedDuplicatesFinder.java
yfarjoun: got it. I missed the part about computeIfAbsent. TIL

yfarjoun:
@jacarey could you also take a look here?

jacarey:
:+1:

yfarjoun:
OK. two reviewers and all looks well!

Merging!

yfarjoun:
thanks @SilinPavel

SilinPavel:
@yfarjoun Great! Thank you too!

sooheelee:
Hey @SilinPavel. Can you explain how this new hashing impacts estimating library complexity? Is this about efficiency or better estimates of library complexity?

I'm updating the release notes to be more user-friendly. Thanks.

yfarjoun:
this is a speedup with no effect on the results, if I recall correctly.

SilinPavel:
@sooheelee  
As @yfarjoun said, this thing has no effect on library complexity estimating, it's just speedup feature.
Sorry for the long respond!

PR #725, opened 2017-01-09
Added '.genotype_concordance' into the output VCF name.

tfenne:
### Description

Change to genotype concordance when `OUTPUT_VCF` is true, to change the output filename to include `.genotype_concordance`.  The other output files have more meaningful names so if you use `O=foo.concordance` you end up with `foo.concordance.genotype_concordance_summary_metrics` etc. which seems redundant.  But if you use `O=foo` you get `foo.vcf.gz` out.  This is a problem if, like me, you tend to run things like `picard CollectGenotypeConcordanceMetrics OUTPUT_VCF=true TV=truth.vcf CV=foo.vcf.gz O=foo`.

#### Review

### DIFF
--- a/src/main/java/picard/vcf/GenotypeConcordance.java
+++ b/src/main/java/picard/vcf/GenotypeConcordance.java
@@ -160,9 +160,10 @@ public class GenotypeConcordance extends CommandLineProgram {
     private final Log log = Log.getInstance(GenotypeConcordance.class);
     private final ProgressLogger progress = new ProgressLogger(log, 10000, "checked", "variants");
 
-    public static final String SUMMARY_METRICS_FILE_EXTENSION = ".genotype_concordance_summary_metrics";
-    public static final String DETAILED_METRICS_FILE_EXTENSION = ".genotype_concordance_detail_metrics";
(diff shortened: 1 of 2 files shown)


src/main/java/picard/vcf/GenotypeConcordance.java
```diff
@@ -381,7 +381,7 @@ private boolean indexExists(final File vcf) {
     /** Gets the variant context writer if the output VCF is to be written, otherwise empty. */
     private Optional<VariantContextWriter> getVariantContextWriter(final VCFFileReader truthReader, final VCFFileReader callReader) {
         if (OUTPUT_VCF) {
-            final File outputVcfFile = new File(OUTPUT + ".vcf.gz");
+            final File outputVcfFile = new File(OUTPUT + ".genotype_concordance.vcf.gz");
```
nh13: Can you make the extension a static member beside the other extensions?

nh13:
@tfenne one quick suggestion, but no objection other than fixing the tests.

tfenne:
@nh13 Your suggest makes sense; if only the original author had done that ;)

nh13:
I always reserve the right to correct myself and admit I am wrong 🤕

PR #732, opened 2017-01-14
CollectRnaSeqMetrics can infer the strand specificity.

nh13:
We examine the first N reads and compare the transcription strand (from
the gene annotations) to the strand of the mapped reads.  More details
are described in StrandSpecificityUtil.inferStrandSpecificity.

@tfenne or @yfarjoun care to take a look?

### DIFF
--- a/src/main/java/picard/analysis/StrandSpecificityUtil.java
+++ b/src/main/java/picard/analysis/StrandSpecificityUtil.java
@@ -1,0 +1,196 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 Nils Homer
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package picard.analysis;
+
+
+import htsjdk.samtools.AlignmentBlock;
+import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.SamPairUtil;
+import htsjdk.samtools.SamReader;
+import htsjdk.samtools.util.Interval;
+import htsjdk.samtools.util.Log;
+import htsjdk.samtools.util.OverlapDetector;
+import picard.annotation.Gene;
+import picard.annotation.LocusFunction;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+
+/** A utility class for inferring strand specificity for RNA-Seq experiments. */
+public class StrandSpecificityUtil {
+
+    public enum StrandSpecificity {NONE, FIRST_READ_TRANSCRIPTION_STRAND, SECOND_READ_TRANSCRIPTION_STRAND}
+
+    /**
+     * This method can be used to infer the strand specificity of an RNA-Seq experiment.
+     *
+     * The strand specificity refers to how the sequencing was performed relative to the strand of transcription.  The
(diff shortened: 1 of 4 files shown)


src/main/java/picard/analysis/StrandSpecificityUtil.java
```diff
@@ -0,0 +1,196 @@
...
+     *
+     * The strand specificity refers to how the sequencing was performed relative to the strand of transcription.  The
+     * strand specificity of sequencing is determined from mapped reads while the strand of transcription is determined from
+     * a set of genes whose transcription strand and regions are known.
+     *
+     * There are three possible outcomes:
+     * 1. No strand specificity - occurs when there is no correlation between the strands of the reads and the strand of transcription.
+     * 2. First read is on the transcription strand.
+     * 3. First read is on the opposite strand of transcription.
+     * For paired end reads, we assume the second read is on the opposite strand as that of the first (forward/reverse orientation).
```
tfenne: I see below that it is tested.  So perhaps reword to say that you filter rather than assume.

src/main/java/picard/analysis/StrandSpecificityUtil.java
```diff
@@ -0,0 +1,196 @@
...
+                            break;
+                    }
+                }
+
+                return false;
+            });
+
+            // Only examine reads that overlap an exon
+            if (overlapsExon) {
+                // We would need both read pairs to check multiple genes, and therefore we ignore such cases and call
```
tfenne: I don't understand this comment.

src/main/java/picard/analysis/StrandSpecificityUtil.java
```diff
@@ -0,0 +1,196 @@
...
+                Arrays.fill(locusFunctions, 0, locusFunctions.length, LocusFunction.INTERGENIC);
+
+                // Assign the function for each transcripts
+                for (final Gene gene : overlappingGenes) {
+                    for (final Gene.Transcript transcript : gene) {
+                        transcript.assignLocusFunctionForRange(alignmentBlock.getReferenceStart(), locusFunctions);
+                    }
+                }
+
+                // Tally the function of each base in the alignment block.
```
tfenne: No tallying, just checking

src/main/java/picard/analysis/StrandSpecificityUtil.java
```diff
@@ -0,0 +1,196 @@
...
+                return false;
+            });
+
+            // Only examine reads that overlap an exon
+            if (overlapsExon) {
+                // We would need both read pairs to check multiple genes, and therefore we ignore such cases and call
+                // them unclassified later.
+                if (overlappingGenes.stream().map(Gene::isPositiveStrand).distinct().count() == 1) {
+                    final boolean negativeReadStrand            = rec.getReadNegativeStrandFlag();
+                    final boolean negativeTranscriptionStrand   = overlappingGenes.iterator().next().isNegativeStrand();
```
tfenne: You should probably limit yourself to reads that only overlap a single gene (or transcripts from a single gene).  There are places in the genome (albeit few) where genes are overlapped with one on each strand.

src/main/java/picard/analysis/StrandSpecificityUtil.java
```diff
@@ -0,0 +1,196 @@
...
+
+        final int numReads = numReadsExamined; // OK IntelliJ
+        log.ifPresent(l -> {
+            l.info("Examined " + numReads + " reads to infer strand specificity.");
+            l.info("Fraction of reads consistent with read one on the transcription strand:     " + fractionFirstReadTranscriptionStrand);
+            l.info("Fraction of reads consistent with read one not on the transcription strand: " + fractionSecondReadTranscriptionStrand);
+            l.info("Unclassified reads: " + fractionUnclassified);
+        });
+
+        if (fractionFirstReadTranscriptionStrand >= minFraction && fractionSecondReadTranscriptionStrand >= minFraction) {
```
tfenne: I feel like you require minFraction > 0.5, and that this makes this clause impossible.

src/main/java/picard/analysis/StrandSpecificityUtil.java
tfenne: I see above that you are limiting to where overlapped genes are on the same strand.  Maybe comment there why, and that is makes this safe?

src/main/java/picard/analysis/StrandSpecificityUtil.java
```diff
@@ -0,0 +1,196 @@
...
+                }
+                numReadsExamined += 1;
+            }
+        }
+
+        // NB: these two values may not add up to one if some reads cannot be classified due to overlapping multiple
+        // genes that have different transcription strands
+        double fractionFirstReadTranscriptionStrand  = (readOneAgree + readTwoDisagree) / (double) numReadsExamined;
+        double fractionSecondReadTranscriptionStrand = (readOneDisagree + readTwoAgree) / (double) numReadsExamined;
+        double fractionUnclassified = (numReadsExamined - readOneAgree - readOneDisagree - readTwoAgree - readTwoDisagree) / (double) numReadsExamined;
```
tfenne: Looking at the loop as it's presently constructed, I don't see a way for this to be anything other than 0 since you always add each read to one of the four counts.  What am I missing?

src/main/java/picard/analysis/StrandSpecificityUtil.java
```diff
@@ -0,0 +1,196 @@
...
+
+        if (fractionFirstReadTranscriptionStrand >= minFraction && fractionSecondReadTranscriptionStrand >= minFraction) {
+            throw new IllegalStateException("Could not infer strand specificity as both + ("
+                    + fractionFirstReadTranscriptionStrand
+                    + ") and - ("
+                    + fractionSecondReadTranscriptionStrand
+                    + ") are >= the threshold '" + minFraction + "'");
+        }
+        else if (fractionUnclassified >= minFraction) {
+            throw new IllegalStateException("Could not infer strand specificity as the number of unclassified reads ("
```
tfenne: I don't think this can happen - and even if it does, should it really throw an exception and not just return `NONE`?

PR #733, opened 2017-01-14
Adding metrics for estimating strand specificity.

nh13:
Added two metrics that measure the fraction of observed reads that
support R1 being on the strand of transcription versus the opposite
strand of transcription.  For paired end reads, R2 is assumed to be on
the opposite strand of that of R1.

### DIFF
--- a/src/main/java/picard/analysis/RnaSeqMetrics.java
+++ b/src/main/java/picard/analysis/RnaSeqMetrics.java
@@ -66,6 +66,36 @@ public class RnaSeqMetrics extends MultilevelMetrics {
     /** Number of aligned reads that are mapped to the incorrect strand.  0 if library is not strand-specific. */
     public long INCORRECT_STRAND_READS;
 
+    /** The number of reads that support the model where R1 is on the strand of transcription and R2 is on the
+     * opposite strand.
+     */
+    public long NUM_R1_TRANSCRIPT_STRAND_READS;
+
+    /**
+     * The fraction of reads that support the model where R2 is on the strand of transcription and R1 is on the opposite
+     * strand.
+     */
+    public long NUM_R2_TRANSCRIPT_STRAND_READS;
+
+    /**
+     * The fraction of reads for which the transcription strand model could not be inferred.
+     */
+    public long NUM_UNEXPLAINED_READS;
+
+
+    /** The fraction of reads that support the model where R1 is on the strand of transcription and R2 is on the
+     * opposite strand.  For unpaired reads, it is the fraction of reads that are on the transcription strand (out of all
+     * the reads).
+     */
+    public double PCT_R1_TRANSCRIPT_STRAND_READS;
+
+    /**
+     * The fraction of reads that support the model where R2 is on the strand of transcription and R1 is on the opposite
+     * strand.  For unpaired reads, it is the fraction of reads that are on opposite strand than that of the the
+     * transcription strand (out of all the reads).
+     */
+    public double PCT_R2_TRANSCRIPT_STRAND_READS;
+
     /** Fraction of PF_ALIGNED_BASES that mapped to regions encoding ribosomal RNA, RIBOSOMAL_BASES/PF_ALIGNED_BASES */
     public Double PCT_RIBOSOMAL_BASES;
 
--- a/src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
+++ b/src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
@@ -1,10 +1,6 @@
 package picard.analysis.directed;
 
-import htsjdk.samtools.AlignmentBlock;
-import htsjdk.samtools.SAMFileHeader;
-import htsjdk.samtools.SAMReadGroupRecord;
-import htsjdk.samtools.SAMRecord;
-import htsjdk.samtools.SAMSequenceRecord;
+import htsjdk.samtools.*;
 import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.util.CoordMath;
 import htsjdk.samtools.util.Histogram;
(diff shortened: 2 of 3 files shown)


src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
```diff
@@ -232,25 +228,39 @@ public void acceptRecord(SAMRecord rec) {
...
+                    } else {
+                        ++metrics.INCORRECT_STRAND_READS;
+                    }
                 }
-            }
 
+                if (!rec.getReadPairedFlag() || SamPairUtil.getPairOrientation(rec) == SamPairUtil.PairOrientation.FR) {
+                    final boolean firstReadNegativeStrand;
+                    // Get the strand of the first read.
+                    if (!rec.getReadPairedFlag() || rec.getFirstOfPairFlag()) firstReadNegativeStrand = rec.getReadNegativeStrandFlag();
```
tfenne: There are a couple of things here that seem a little off to me:
1) You're looking at first of pair strand for your test, but you only really know whether the read you have in hand (maybe R1 or R2) overlaps an exon.
2) You're only looking at R1 strand, but you're single-counting it for fragments and double-counting it for pairs.

nh13:
@tfenne care to take one more look?

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
```diff
@@ -232,25 +228,60 @@ public void acceptRecord(SAMRecord rec) {
...
+                    final int leftMostAlignedBase, rightMostAlignedBase;
+                    if (rec.getReadPairedFlag()) {
+                        if (rec.getMateUnmappedFlag()) {
+                            properOrientation   = false; // mate is unmapped!
+                            leftMostAlignedBase = rightMostAlignedBase = 0;
+                        }
+                        else {
+                            properOrientation    = SamPairUtil.getPairOrientation(rec) == SamPairUtil.PairOrientation.FR;
+                            leftMostAlignedBase  = Math.min(rec.getAlignmentStart(), rec.getMateAlignmentStart());
+                            rightMostAlignedBase = Math.max(rec.getAlignmentEnd(), SAMUtils.getMateAlignmentEnd(rec)); // requires the mate cigar
```
tfenne: I think you could just use the inferred insert size here to avoid requiring the mate cigar.

nh13: I think this is true for `FR` reads, but not `FF` reads.  Unfortunately, Picard/htsjdk computes insert size as 5' to 5' rather than left-most to right-most bases.

nh13:
@yfarjoun it would be great to have a look by someone else, either yourself, or someone you suggest?

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
```diff
@@ -232,25 +228,60 @@ public void acceptRecord(SAMRecord rec) {
...
+                    }
+                    else {
+                        properOrientation    = true; // can only be false for paired end reads
+                        leftMostAlignedBase  = rec.getAlignmentStart();
+                        rightMostAlignedBase = rec.getAlignmentEnd();
+                    }
+
+                    if (properOrientation && CoordMath.encloses(gene.getStart(), gene.getEnd(), leftMostAlignedBase, rightMostAlignedBase)) {
+                        if (negativeReadStrand == negativeTranscriptionStrand) {
+                            ++metrics.PCT_FIRST_READ_TRANSCRIPTION_STRAND_READS;
```
yfarjoun: Perhaps you could ++ private (integer) members of the metrics class and then divide them to create the double value? This will solve the problem below (`finish()`) and also avoid annoying rounding errors due to incrementing a double multiple times. Faster too (increment on integers seems to have a 50%-75% speedup on doubles)...

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
yfarjoun: for example use the inferred read length if the mate cigar is missing...

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
```diff
@@ -232,25 +228,60 @@ public void acceptRecord(SAMRecord rec) {
...
+                        }
+                        else {
+                            ++metrics.PCT_SECOND_READ_TRANSCRIPTION_STRAND_READS;
+                        }
+                    }
```
yfarjoun: how about you also count the number/percent of reads for which you couldn't infer the transcription strand?

src/test/java/picard/analysis/CollectRnaSeqMetricsTest.java
```diff
@@ -212,11 +217,86 @@ else if (metrics.LIBRARY.equals("bar")) {
...
+        final File metricsFile = File.createTempFile("tmp.", ".rna_metrics");
+        metricsFile.deleteOnExit();
+
+        final String[] args = new String[] {
+                "INPUT=" +               samFile.getAbsolutePath(),
+                "OUTPUT=" +              metricsFile.getAbsolutePath(),
+                "REF_FLAT=" +            getRefFlatFile(sequence).getAbsolutePath(),
+                "RIBOSOMAL_INTERVALS=" + rRnaIntervalsFile.getAbsolutePath(),
+                "STRAND_SPECIFICITY=SECOND_READ_TRANSCRIPTION_STRAND",
+                "IGNORE_SEQUENCE=" +ignoredSequence
```
yfarjoun: space after `+`

src/test/java/picard/analysis/CollectRnaSeqMetricsTest.java
```diff
@@ -212,11 +217,86 @@ else if (metrics.LIBRARY.equals("bar")) {
...
+
+        // Reads that start *before* the gene: not counted
+        builder.addPair("pair_prior_1", sequenceIndex, 45, 50, false, false, "50M", "50M", true, false, -1);
+        builder.addPair("pair_prior_2", sequenceIndex, 49, 50, false, false, "50M", "50M", true, false, -1);
+
+        // Read that is enclosed in the gene, but one end does not map: not counted
+        builder.addPair("read_one_end_unmapped", sequenceIndex, 50, 51, false, true, "50M", null, false, false, -1);
+
+        // Reads that are enclosed in the gene, paired and frag: one count per template
+        builder.addFrag("frag_enclosed_forward", sequenceIndex, 150, false);
```
yfarjoun: These have null cigars...so how many bases get counted towards coverage?

src/test/java/picard/analysis/CollectRnaSeqMetricsTest.java
```diff
@@ -212,11 +217,86 @@ else if (metrics.LIBRARY.equals("bar")) {
...
+        final String ignoredSequence = "chrM";
+
+        // Create some alignments that hit the ribosomal sequence, various parts of the gene, and intergenic.
+        final SAMRecordSetBuilder builder = new SAMRecordSetBuilder(true, SAMFileHeader.SortOrder.coordinate);
+        // Set seed so that strandedness is consistent among runs.
+        builder.setRandomSeed(0);
+        final int sequenceIndex = builder.getHeader().getSequenceIndex(sequence);
+
+        // Reads that start *before* the gene: not counted
+        builder.addPair("pair_prior_1", sequenceIndex, 45, 50, false, false, "50M", "50M", true, false, -1);
```
yfarjoun: could you please add:

1. secondary alignments
2. supplementary alignments
3. records that do not have MC tag
4. duplicate reads.
5. ignored sequence reads
6. non-PF reads

src/test/java/picard/analysis/CollectRnaSeqMetricsTest.java
yfarjoun: answered my own question. 36.

src/test/java/picard/analysis/CollectRnaSeqMetricsTest.java
```diff
@@ -212,11 +217,86 @@ else if (metrics.LIBRARY.equals("bar")) {
...
+                "IGNORE_SEQUENCE=" +ignoredSequence
+        };
+        Assert.assertEquals(runPicardCommandLine(args), 0);
+
+        final MetricsFile<RnaSeqMetrics, Comparable<?>> output = new MetricsFile<RnaSeqMetrics, Comparable<?>>();
+        output.read(new FileReader(metricsFile));
+
+        final RnaSeqMetrics metrics = output.getMetrics().get(0);
+        Assert.assertEquals(metrics.PF_ALIGNED_BASES, 722);
+        Assert.assertEquals(metrics.PF_BASES, 576);
```
yfarjoun: is it just me or is it strange that PF_BASES < PF_ALIGNED_BASES?

src/test/java/picard/analysis/CollectRnaSeqMetricsTest.java
```diff
@@ -212,11 +217,86 @@ else if (metrics.LIBRARY.equals("bar")) {
...
+
+        // Read that starts *after* the gene: not counted
+        builder.addPair("pair_after_1", sequenceIndex, 545, 550, false, false, "50M", "50M", true, false, -1);
+        builder.addPair("pair_after_2", sequenceIndex, 549, 550, false, false, "50M", "50M", true, false, -1);
+
+        final File samFile = File.createTempFile("tmp.collectRnaSeqMetrics.", ".sam");
+        samFile.deleteOnExit();
+
+        final SAMFileWriter samWriter = new SAMFileWriterFactory().makeSAMWriter(builder.getHeader(), false, samFile);
+        for (final SAMRecord rec: builder.getRecords()) samWriter.addAlignment(rec);
```
yfarjoun: Streams for the win?!?

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
```diff
@@ -232,25 +228,60 @@ public void acceptRecord(SAMRecord rec) {
...
+                            properOrientation   = false; // mate is unmapped!
+                            leftMostAlignedBase = rightMostAlignedBase = 0;
+                        }
+                        else {
+                            properOrientation    = SamPairUtil.getPairOrientation(rec) == SamPairUtil.PairOrientation.FR;
+                            leftMostAlignedBase  = Math.min(rec.getAlignmentStart(), rec.getMateAlignmentStart());
+                            rightMostAlignedBase = Math.max(rec.getAlignmentEnd(), SAMUtils.getMateAlignmentEnd(rec)); // requires the mate cigar
+                        }
+                    }
+                    else {
```
nh13: Done

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
```diff
@@ -232,25 +228,60 @@ public void acceptRecord(SAMRecord rec) {
...
+                        properOrientation    = true; // can only be false for paired end reads
+                        leftMostAlignedBase  = rec.getAlignmentStart();
+                        rightMostAlignedBase = rec.getAlignmentEnd();
+                    }
+
+                    if (properOrientation && CoordMath.encloses(gene.getStart(), gene.getEnd(), leftMostAlignedBase, rightMostAlignedBase)) {
+                        if (negativeReadStrand == negativeTranscriptionStrand) {
+                            ++metrics.PCT_FIRST_READ_TRANSCRIPTION_STRAND_READS;
+                        }
+                        else {
```
nh13: Done

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
nh13: Done

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
nh13: I'll just make some public `long` members and we can all have more metrics.

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
```diff
@@ -277,6 +308,12 @@ public void finish() {
             if (metrics.CORRECT_STRAND_READS > 0 || metrics.INCORRECT_STRAND_READS > 0) {
                 metrics.PCT_CORRECT_STRAND_READS = metrics.CORRECT_STRAND_READS/(double)(metrics.CORRECT_STRAND_READS + metrics.INCORRECT_STRAND_READS);
             }
+
+            final double readsExamined = metrics.PCT_FIRST_READ_TRANSCRIPTION_STRAND_READS + metrics.PCT_SECOND_READ_TRANSCRIPTION_STRAND_READS;
+            if (0 < readsExamined) {
+                metrics.PCT_FIRST_READ_TRANSCRIPTION_STRAND_READS /= readsExamined;
```
nh13: Moot now, but I suppose it can happen, but you really shouldn't call it twice.

src/test/java/picard/analysis/CollectRnaSeqMetricsTest.java
nh13: Same as the rest of the code.  Keeping it for now.  I am not in love with Java streams yet 👍

src/test/java/picard/analysis/CollectRnaSeqMetricsTest.java
nh13: Fixed here and elsewhere (copy paste)

src/test/java/picard/analysis/CollectRnaSeqMetricsTest.java
nh13: I did not add 5 since that is tested in the other tests.

nh13: FYI, duplicate reads are OK.

nh13: NB: previously were were including supplementary reads when calculating `CORRECT_STRAND_READS` and `INCORRECT_STRAND_READS`.  Now we are not.

src/test/java/picard/analysis/CollectRnaSeqMetricsTest.java
nh13: I had to set the read length on the builder, the call to `add*` didn't check the cigar versus the bases length (yikes).

nh13:
@yfarjoun thanks for the helpful comments, take another look and let me know what you think.

tfenne:
@nh13 I just realized that somewhere between my review and @yfarjoun's review something changed that I disagree with.  Previously the two percentages were calculated using the sum of the `NUM_FIRST_READ_TRANSCRIPTION_STRAND_READS` and `NUM_SECOND_READ_TRANSCRIPTION_STRAND_READS` alone.  Now the denominator also rolls in `NUM_UNEXPLAINED_TRANSCRIPTION_STRAND_READS`.

I think this is wrong because what that represents is reads that aren't properly paired or in the expected `FR` orientation, so we can't really assay them for strandedness.  I think it should go back to just using the two values for the denominator.

Also, while I'm here, could we change `FIRST_READ` to `R1` and `SECOND_READ` to `R2` for brevity please?

nh13:
@tfenne thanks for catching this, I'll update.

src/main/java/picard/analysis/RnaSeqMetrics.java
```diff
@@ -66,19 +66,43 @@
...
 
+    /** The number of reads that support the model where R1 is on the strand of transcription and R2 is on the
+     * opposite strand.  For unpaired reads, it is the fraction of read that are on the transcription strand (out of all
+     * the reads).
+     */
+    public long NUM_FIRST_READ_TRANSCRIPTION_STRAND_READS;
+
+    /**
+     * The fraction of reads that support the model where R2 is on the strand of transcription and R1 is on the opposite
+     * strand.  For unpaired reads, it is the fraction of read that are on opposite strand than that of the the
```
nh13: Removing unpaired description as it is just plain confusing.

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
```diff
@@ -226,9 +226,13 @@ public void acceptRecord(SAMRecord rec) {
                 }
             }
 
+            if (metrics.PF_ALIGNED_BASES > metrics.PF_BASES) {
+                throw new IllegalStateException("HUH?");
```
nh13: Removed.

nh13:
@yfarjoun and @tfenne, thanks for the comments.  I feel I have all of the resolved, so one final 👍  and I can merge after tests pass.

tfenne:
👍 from me. @yfarjoun ?

yfarjoun:
actually, given that oracle DB has problems with long names...could you shorten the length of the metrics names to 30 or less?

yfarjoun:
sorry for the late comment. (I was starting to think about the needed DB changes..)

nh13:
@yfarjoun I shortened them to 30 characters or less.  👍 , then I can rebase and once tests pass merge?

PR #747, opened 2017-02-07
Allow an optional file extension for CollectIlluminaLaneMetrics.

nh13:
@lbergelson can I trouble you for a quick review?  I was getting annoyed that the metrics coming out of this tool don't have a ".txt" or similar extension.  This change keeps the current behavior.

### DIFF
--- a/src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
+++ b/src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
@@ -88,6 +88,9 @@ public class CollectIlluminaLaneMetrics extends CommandLineProgram {
     @Option(doc = ReadStructure.PARAMETER_DOC + "\nIf not given, will use the RunInfo.xml in the run directory.", shortName = "RS", optional = true)
     public ReadStructure READ_STRUCTURE;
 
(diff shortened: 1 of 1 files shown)


src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
```diff
@@ -115,7 +118,7 @@ protected int doWork() {
             }
         }
 
-        IlluminaLaneMetricsCollector.collectLaneMetrics(RUN_DIRECTORY, OUTPUT_DIRECTORY, OUTPUT_PREFIX, laneMetricsFile, phasingMetricsFile, READ_STRUCTURE);
+        IlluminaLaneMetricsCollector.collectLaneMetrics(RUN_DIRECTORY, OUTPUT_DIRECTORY, OUTPUT_PREFIX, laneMetricsFile, phasingMetricsFile, READ_STRUCTURE, FILE_EXTENSION == null ? "" : FILE_EXTENSION);
```
lbergelson: Would it make more sense to initialize `FILE_EXTENSION` to `""` instead of `null` and then avoid this null check?  (although I suppose someone could perversely pass `null` in on the command line, but maybe those sorts of people deserver what they get?)   Or is it null to avoid printing a default of `""` on the commandline? )

nh13: For future reference, I took my inspiration from `CollectMultipleMetrics`

PR #784, opened 2017-04-04
Fix CollectHsMetrics and CollectTargetedPcrMetrics to correctly handle…

tfenne:
… read pairs that are fully overlapped when CLIP_OVERLAPPING_READS=true.

### Description

We noticed a significant discrepancy between running CollectHsMetrics on an aligned BAM file, and the same BAM file after running it through fgbio's ClipOverlappingReads.  The results are expected to be the same, but mean target coverage was different by ~10% on the sample I was looking at.  I've diagnosed the problem - it's that a) the clipping code in HTSJDK sets reads to be unmapped but doesn't otherwise alter them if the request is to clip the entire read and b) TargetMetricsCollector never re-checks the the unmapped flag, and since the cigar and position are left on the read, it consumes the all the bases from these reads where it should consume none.

I've added a re-check of whether the read is unmapped and fixed the test that should have caught this in the first place.

#### Content

#### Review


tfenne:
@ktibbett If you have the time it would be good for you to take a look at this - or at least the implications of this.  Or could you nominate someone to take a look please?

@nh13 Want to add a review since the clipping changes were made by you originally?

ktibbett:
@jacarey , would you please take a quick look at this as well?

jacarey:
👍

Issue #768, opened 2017-03-06
GenotypeConcordance dies if both samples are no call


apregier:
### Affected tool(s)
`GenotypeConcordance OUTPUT_VCF=true`

### Affected version(s)
- [x] Latest public release version [2.9.0-1-gf5b9f50-SNAPSHOT]

### Description 
At first I thought I was running into the spanning deletion bug, because the first time I hit it was at a site with a * allele.  But, even after I removed all sites with those alleles, I got the following stacktrace:
```
Exception in thread "main" java.lang.IllegalArgumentException: BUG: Cannot add a no call allele to a variant context [TA*, .] this=[VC Unknown @ chr1:53707-53708 Q584.04 of type=INDEL alleles=[TA*, .] attr={CONC_ST=[EMPTY]} GT=[[call ./. DP 1 AD 1,0],[truth ./. DP 1 AD 1,0]]
        at htsjdk.variant.variantcontext.VariantContext.validateAlleles(VariantContext.java:1344)
        at htsjdk.variant.variantcontext.VariantContext.validate(VariantContext.java:1297)
        at htsjdk.variant.variantcontext.VariantContext.<init>(VariantContext.java:401)
        at htsjdk.variant.variantcontext.VariantContextBuilder.make(VariantContextBuilder.java:494)
        at htsjdk.variant.variantcontext.VariantContextBuilder.make(VariantContextBuilder.java:488)
        at picard.vcf.GenotypeConcordance.writeVcfTuple(GenotypeConcordance.java:454)
        at picard.vcf.GenotypeConcordance.lambda$doWork$0(GenotypeConcordance.java:333)
        at java.util.Optional.ifPresent(Optional.java:159)
        at picard.vcf.GenotypeConcordance.doWork(GenotypeConcordance.java:333)
        at picard.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:205)
        at picard.cmdline.PicardCommandLine.instanceMain(PicardCommandLine.java:94)
        at picard.cmdline.PicardCommandLine.main(PicardCommandLine.java:104)
```
This only seems to be an issue when writing an output vcf - I was able to get the metrics files when I ran with `OUTPUT_VCF=false`.

#### Steps to reproduce
It should reproduce if you run GenotypeConcordance OUTPUT_VCF=true for any two vcfs where a site has a './.' genotype for both the truth and call samples.

#### Expected behavior
[...]

PR #785, opened 2017-04-04
Fix in GenotypeConcordance for writing VCFs with no-call genotypes

meganshand:
### Description

This addresses #768. If the truth or call VCF has a no call genotype, GenotypeConcordance can still calculate the metrics, but breaks if you try to write a new VCF. This fixes the variant context that is written, to get the allele from the site rather than the genotypes.
----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/vcf/GenotypeConcordance.java
+++ b/src/main/java/picard/vcf/GenotypeConcordance.java
@@ -432,9 +432,18 @@ public class GenotypeConcordance extends CommandLineProgram {
             final List<Allele> truthAlleles = alleles.truthAlleles();
             final List<Allele> callAlleles  = alleles.callAlleles();
 
+            // Get the alleles present at this site for both samples to use for the output variant context.
+            final Set<Allele> siteAlleles = new HashSet<>();
+            if (truthContext != null) {
+                siteAlleles.addAll(truthContext.getAlleles());
+            }
+            if (callContext != null) {
+                siteAlleles.addAll(callContext.getAlleles());
+            }
+
             // Initialize the variant context builder
             final VariantContext initialContext = (callContext == null) ? truthContext : callContext;
-            builder = new VariantContextBuilder(initialContext.getSource(), initialContext.getContig(), initialContext.getStart(), initialContext.getEnd(), Collections.emptyList());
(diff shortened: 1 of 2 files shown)


src/main/java/picard/vcf/GenotypeConcordance.java
```diff
@@ -442,8 +442,17 @@ private void writeVcfTuple(final VcfTuple tuple, final VariantContextWriter writ
             addToGenotypes(genotypes, truthContext, TRUTH_SAMPLE, OUTPUT_VCF_TRUTH_SAMPLE_NAME, allAlleles, truthAlleles, MISSING_SITES_HOM_REF);
             addToGenotypes(genotypes, callContext, CALL_SAMPLE, OUTPUT_VCF_CALL_SAMPLE_NAME, allAlleles, callAlleles, false);
 
+            // Since they differ in the no call case, use the site (rather than the genotype) alleles for new VC
+            Set<Allele> siteAlleles = new HashSet<>();
```
nh13: can be `final`

meganshand:
@nh13 Don't you need the alleles from both the truth and the call contexts? `initialContext` would only be either `truthContext` or `callContext`. It doesn't appear that they need to be identical at this point. For example if the call context only has the reference allele, the truth could have both a ref and alt allele.

nh13:
@meganshand good point, then how about adding creating `siteAlleles` earlier and passing it to the builder's constructor?

meganshand:
Ah sure, good point.

src/main/java/picard/vcf/GenotypeConcordance.java
```diff
@@ -432,9 +432,18 @@ private void writeVcfTuple(final VcfTuple tuple, final VariantContextWriter writ
             final List<Allele> truthAlleles = alleles.truthAlleles();
             final List<Allele> callAlleles  = alleles.callAlleles();
 
+            // Since they differ in the no call case, use the site (rather than the genotype) alleles for new VC
```
nh13: I think this comment is more relevant to this PR than to the code itself, how about:
```
// Get the alleles present at this site for both samples to use for the output variant context.
```
?

nh13:
@meganshand one minor quibble from me, but once you are happy 👍  from me

PR #812, opened 2017-05-06
CollectIlluminaLaneMetric should be lenient when missing prephasing and

nh13:
phasing metrics.

I have some data that when analyzed fails with the exception below, but in some cases I would rather just a warning, and for it to continue on.  I took a look at making some tests for this, but that means mucking with the `TileMetricsOut.bin` which seems not fun.  So if this PR looks good, I'll take another shot at adding a test.

```
Exception in thread "main" picard.PicardException: Don't have both phasing and prephasing values for tile
	at picard.illumina.parser.TileMetricsUtil.getTilePhasingValues(TileMetricsUtil.java:116)
	at picard.illumina.parser.TileMetricsUtil.parseTileMetrics(TileMetricsUtil.java:95)
	at picard.illumina.CollectIlluminaLaneMetrics$IlluminaLaneMetricsCollector.readLaneTiles(CollectIlluminaLaneMetrics.java:110)
	at picard.illumina.CollectIlluminaLaneMetrics$IlluminaLaneMetricsCollector.collectLaneMetrics(CollectIlluminaLaneMetrics.java:123)
	at picard.illumina.CollectIlluminaLaneMetrics.doWork(CollectIlluminaLaneMetrics.java:91)
	at picard.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:208)
	at picard.cmdline.PicardCommandLine.instanceMain(PicardCommandLine.java:95)
	at picard.cmdline.PicardCommandLine.main(PicardCommandLine.java:105)
```


src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -111,13 +117,30 @@ public static File renderTileMetricsFileFromBasecallingDirectory(final File illu
...
                 final int prePhasingCode = IlluminaMetricsCode.getPhasingCode(descriptorIndex, IlluminaMetricsCode.PREPHASING_BASE);
 
+                float phasingValue = 0;
+                float prePhasingValue = 0;
+
                 if (!(codeMetricsMap.containsKey(phasingCode) && codeMetricsMap.containsKey(prePhasingCode))) {
-                    throw new PicardException("Don't have both phasing and prephasing values for tile");
+                    final String message = String.format(
+                            "Don't have both phasing and prephasing values for read %s cycle %s.  Phasing code was %d and prephasing code was %d.",
+                            isFirstRead ? "one" : "two", descriptorIndex+1, phasingCode, prePhasingCode
```
yfarjoun: can use tileTemplateRead.toString() here (and reverse the order of `%s` and `read` in the format string).

just a suggestion.

yfarjoun: space around operator.

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -111,13 +117,30 @@ public static File renderTileMetricsFileFromBasecallingDirectory(final File illu
...
+                    );
+                    switch (validationStringency) {
+                        case LENIENT:
+                            LOG.warn(message);
```
yfarjoun: so in this case, you'll get 0 for both phasing and pre-phasing in the case that one of them is missing. 

1. do you understand _why_ one of them could be missing? a comment would be helpful for future improvement, even if you are not going to do it yourself.
2. if one is available (of phasing and pre-phasing) do you not want to grab that one?

nh13: @yfarjoun I could not figure out _why_ one of them is missing.  I have seen other parsers of `TileMetricsOut.bin` handle this situation similarly.  Unfortunately, I don't have someone Illumina to ask.

ktibbett: This happens to us as well, and no clue as to why.  We have been modifying the TileMetricsOut.bin to add 0s, so this is much better.  Thanks @nh13!

nh13:
@yfarjoun @ktibbett just checking in on if the ball is in my court.  

The likely cause of having the missing data is actually some foreign object (ex. grain of sand) that impedes the ability to image the given tile, and so no data would exist for the tile, including phasing and pre-phasing metrics.

yfarjoun:
so that explains why both would be missing. could you change the `if` statement to only cover the case where both are missing as we have a likely explanation for that, while keeping the case that one but not the other are missing as an error?

nh13:
@yfarjoun will do.

nh13:
@yfarjoun back to you.  I snuck an extra commit where I found a typo in a test code.  Don't ask me how I found it, just that I saw the file modified on disk.  I'll squash the first two commits if you 👍  before merging.

nh13:
@yfarjoun oops forgot to add a test, doing that now.

nh13:
@yfarjoun test added, back to you.

nh13:
@yfarjoun, in 6e8bd26 I added some guards against zero-area tiles, which ignore tiles when the tile has zero cluster density.

src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
```diff
@@ -118,7 +119,7 @@ protected int doWork() {
             }
         }
 
-        IlluminaLaneMetricsCollector.collectLaneMetrics(RUN_DIRECTORY, OUTPUT_DIRECTORY, OUTPUT_PREFIX, laneMetricsFile, phasingMetricsFile, READ_STRUCTURE, FILE_EXTENSION == null ? "" : FILE_EXTENSION);
+        IlluminaLaneMetricsCollector.collectLaneMetrics(RUN_DIRECTORY, OUTPUT_DIRECTORY, OUTPUT_PREFIX, laneMetricsFile, phasingMetricsFile, READ_STRUCTURE, FILE_EXTENSION == null ? "" : FILE_EXTENSION, VALIDATION_STRINGENCY);
```
yfarjoun: `<rant>` isn't it annoying that one needs to push the validation stringency down this way? it would be so nice if there was an "engine-level" validation stringency that could be "polled" instead of this... `</rant>`

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -111,13 +117,33 @@ public static File renderTileMetricsFileFromBasecallingDirectory(final File illu
...
+                // is missing, then likely the data is corrupt.
+                if (!codeMetricsMap.containsKey(phasingCode) && !codeMetricsMap.containsKey(prePhasingCode)) {
+                    final String message = String.format(
+                            "Don't have both phasing and prephasing values for %s read cycle %s.  Phasing code was %d and prephasing code was %d.",
+                            tileTemplateRead.toString(), descriptorIndex + 1, phasingCode, prePhasingCode
+                    );
+                    switch (validationStringency) {
+                        case LENIENT:
+                            LOG.warn(message);
+                            // NB: intentionally no break
```
yfarjoun: why no break? I mean, sure it will break due to the next case, but why code this way? seems brittle.

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -111,13 +117,33 @@ public static File renderTileMetricsFileFromBasecallingDirectory(final File illu
...
+                        case SILENT:
+                            break;
+                        case STRICT:
+                            throw new PicardException(message);
+                    }
+                }
```
yfarjoun: `} else {` (too many code-bases...)

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -111,13 +117,33 @@ public static File renderTileMetricsFileFromBasecallingDirectory(final File illu
...
                 final int prePhasingCode = IlluminaMetricsCode.getPhasingCode(descriptorIndex, IlluminaMetricsCode.PREPHASING_BASE);
 
-                if (!(codeMetricsMap.containsKey(phasingCode) && codeMetricsMap.containsKey(prePhasingCode))) {
-                    throw new PicardException("Don't have both phasing and prephasing values for tile");
+                float phasingValue = 0;
+                float prePhasingValue = 0;
+
+                // If both the phasing and pre-phasing data are missing, then likely something went wrong when imaging
+                // this tile, for example a grain of sand disrupting the path of light to the sensor.  If only one of them
+                // is missing, then likely the data is corrupt.
```
yfarjoun: if data is corrupt you should throw an exception...but you modified the old if statement to be stricter, and now the case that **only** one of the codes is missing is ignored...this case should still throw.

src/test/java/picard/illumina/IlluminaLaneMetricsCollectorTest.java
```diff
@@ -77,4 +83,54 @@ public void testCollectIlluminaLaneMetrics(final String testRun, final ReadStruc
                 {"A67HY", new ReadStructure("8B8B")}
         };
     }
+
+    /** Ensures that an exception is thrown when we encounter a tile without phasing/pre-phasing metrics. */
+    @Test(expectedExceptions = PicardException.class)
```
yfarjoun: Can you add a test that is only missing phasing _**or**_ prephasing and see that even in SILENT mode it blows up? (I don't know how hard it is to cook up the data like that...)

src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
nh13: Come to fgbio to develop: https://github.com/fulcrumgenomics/fgbio/blob/master/src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala#L54-L57

In all seriousness, does Barclay have a facility for a sub-parser (a la python's argparse)?  We've just starting using in fgbio a main parser that has tool-wide options and a sub-parser for tools-specific options and it seems promising.

src/test/java/picard/illumina/IlluminaLaneMetricsCollectorTest.java
nh13: I am going to pass on this.

src/main/java/picard/illumina/parser/TileMetricsUtil.java
nh13: Fixed.

src/main/java/picard/illumina/parser/TileMetricsUtil.java
nh13: Fixed.

nh13:
@yfarjoun thanks for the look, back to you.

src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
yfarjoun: this isn't about the parser...as much as I would like there to be more "engine" wide constants like htsjdk's ASYNC...anyhow...another day.

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -123,25 +123,23 @@ public static File renderTileMetricsFileFromBasecallingDirectory(final File illu
                 // If both the phasing and pre-phasing data are missing, then likely something went wrong when imaging
                 // this tile, for example a grain of sand disrupting the path of light to the sensor.  If only one of them
                 // is missing, then likely the data is corrupt.
-                if (!codeMetricsMap.containsKey(phasingCode) && !codeMetricsMap.containsKey(prePhasingCode)) {
+                if (codeMetricsMap.containsKey(phasingCode) && codeMetricsMap.containsKey(prePhasingCode)) {
+                    phasingValue = CollectionUtil.getSoleElement(codeMetricsMap.get(phasingCode)).getMetricValue();
```
yfarjoun: Could you make `phasingValue` and `prePhasingValue` final and assign to zero in the `else` block?

nh13:
Oh how I hate merge commits.  @yfarjoun can you enable "rebase and merge"?  Merge commits really clutter they history.

Issue #786, opened 2017-04-05
Discussion around command line parser changes


cmnbroad:
We propose to make changes to Picard to also support Posix-style command line syntax in addition to the existing syntax. The new syntax would be enabled by default, but the user would have the ability to opt-out and use current syntax by using an alternate jar file, which will be provided pre-compiled with each release. In the event the user opts out, the existing command line syntax will remain unchanged and be fully backward compatible.

Following is a Posix-style command line:

> java -jar picard.jar ValidateSamFile -I input.bam -MODE SUMMARY

The equivalent Picard command line using the current syntax is:

> java -jar picard.jar ValidateSamFile I=input.bam MODE=SUMMARY

**Overview of the proposed changes:**

- Change Picard to use the command line parsers and annotations in the (BSD 3 licensed) [Barclay](https://github.com/broadinstitute/barclay) library. Barclay contains two command line parsers (a slightly modified version of the existing Picard parser, and a posix-style parser based on jopt-simple). Both parsers use the same set of argument annotation classes, also defined in Barclay.
- Change picard.jar to use the new command line parser, and add a new build target (alternate jar) that would use the existing parser.

**Picard source changes:**

- Add a dependency on Barclay.
- Change all occurrences of the @Option, @CommandLineProgramProperties, and @PositionalArguments annotations to use equivalents in the Barclay package “org.broadinstitute.barclay.argparser” package (named @Argument, @CommandLineProgramProperties, and @PositionalArguments respectively). This is a fairly easy package name change that can be done automatically with a typical IDE.
- Add a command line translator that converts old-style command lines into posix-style for purposes of test execution. Tests would continue to be written using the old style syntax for now (recognizing both could be done down the road), but the test would be run once as they currently are and once with the translator and new parser.

**Note on `overridable` Arguments**

The Picard parser currently allows a tool to duplicate and override arguments (via the @Option attribute `overridable`) that are inherited from a base class in order to change either the default value or the values of other Option attributes, such as the doc string or the “required” attribute. This is used by several tools for example to make the reference argument, which is usually optional, required for that tool. We propose to remove this feature of the argument parser, and to replace it with equivalent functionality in the form of @ArgumentCollection annotations. The @ArgumentCollection feature allows a tool to have a reference (annotated with @ArgumentCollection) to an object that itself contains one or more embedded @Arguments. This allows the tool to select the attributes of its arguments by dynamically selecting the class to instantiate, (i.e., one of RequiredReference or OptionalReference).

[...]

PR #817, opened 2017-05-17
Switch to Barclay command line parser(s)

cmnbroad:
### Description

Changes to use the Barclay command line parser(s), as discussed [here](https://github.com/broadinstitute/picard/issues/786).

There are 5 separate commits to make this somewhat easier to review:

1. Changes to Barclay package/class names, annotations, attributes, and imports only.
2. Switch to the Barclay version of the existing parser. This mostly consists of removal of the existing command line parser, and changes to CommandLineProgram to use the Barclay version of the Picard parser.
3. Use @ArgumentCollection, remove overridable. With this commit, the tests pass using the Barclay version of the Picard parser.
4. Add ability to toggle between the two Barclay parsers. This includes property-based opt-in CLP toggling (default uses the existing parser); Gradle changes to generate a second artifact that uses the Barclay parser; minor changes to tools to get the tests to pass with both parsers, and a few changes required for tools that have no tests.
5. Gradle changes for the Barclay documentation generation process. These changes are build-only, plus the addition of the @DocumentedFeature annotation (required to make a tool a doc target) to use to verify this process.

There is one tool (DownsampleSam, and three more in picard-private) that use “R” as a command line argument for something other than reference, which is problematic for the @ArgumentCollection pattern. I’ve temporarily changed these to use a different short name. If thats  not workable, its possible to keep them as is, though it will require a little bit of ugly code and a typecast.

(There is also a separate branch for [picard-private](https://github.com/broadinstitute/picard-private/tree/cn_barclay) that has corresponding changes, subject to completion of this PR).

#### Content
- [X] All tests passing on Travis

#### Review


src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -84,57 +86,74 @@
...
     private static final double LOG_ODDS_THRESHOLD = 3.0;
 
+    @Override
+    protected boolean requiresReference() { return true; }
+
+    /**
+     * @return An interval argument collection to be used for this tool. Subclasses can override this
+     * to provide an argument collection with alternative arguments or argument annotations.
+     */
+    protected IntervalArgumentCollection getIntervalArgumentCollection() {
```
tfenne: I wonder if methods like these should be called `build...` or `make...` to make it clear that they're constructing a collection object to be used, not accessing the one that is populated by Barclay?

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -84,57 +86,74 @@
...
+        return new CollectWGSMetricsIntervalArgumentCollection();
+    }
+
+    public static class CollectWGSMetricsIntervalArgumentCollection implements IntervalArgumentCollection {
```
tfenne: Could this be `CollectWgsMetricsIntervalArgumentCollection` to be consistent with the enclosing class please?

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -84,57 +86,74 @@
...
 
-    @Option(doc = "Average read length in the file. Default is 150.", optional = true)
+    @Argument(doc = "Average read length in the file. Default is 150.", optional = true)
     public int READ_LENGTH = 150;
 
-    @Option(doc = "An interval list file that contains the positions to restrict the assessment. Please note that " +
-            "all bases of reads that overlap these intervals will be considered, even if some of those bases extend beyond the boundaries of " +
-            "the interval. The ideal use case for this argument is to use it to restrict the calculation to a subset of (whole) contigs.",
-            optional = true, overridable = true)
     public File INTERVALS = null;
```
tfenne: Same here .. I think this needs to be protected.  If anyone uses this class programmatically from another tool, and sets `INTERVALS` directly, it'll get clobbered by the value from `intervalArugmentCollection` via the assignment on line 450.  Better to get a compile break than a runtime break I think.

src/main/java/picard/analysis/directed/CollectTargetedMetrics.java
```diff
@@ -60,42 +60,47 @@ protected abstract COLLECTOR makeCollector(final Set<MetricAccumulationLevel> ac
...
             "one line per target base and can grow very large. It is not recommended for use with large target sets.")
     public File PER_BASE_COVERAGE;
 
-    @Option(optional = true, doc= "The maximum distance between a read and the nearest probe/bait/amplicon for the read to be " +
+    @Argument(optional = true, doc= "The maximum distance between a read and the nearest probe/bait/amplicon for the read to be " +
             "considered 'near probe' and included in percent selected.")
     public int NEAR_DISTANCE = TargetedPcrMetricsCollector.NEAR_PROBE_DISTANCE_DEFAULT;
 
-    @Option(shortName = MINIMUM_MAPPING_QUALITY_SHORT_NAME, doc = "Minimum mapping quality for a read to contribute coverage.", overridable = true)
+    //TODO: MINIMUM_MAPPING_QUALITY and MINIMUM_BASE_QUALITY are both annotated as required,
```
tfenne: I think the intent is that they are "required" but that the requirement is met by having a default value.  I'm not sure what the right way to represent that is in Barclay, but IIRC Picard's parser treats required-ness as "required to have a value after parsing", whether that value came from a default, or came from a command line arg.  I wouldn't change these particular ones to nullable values.

src/main/java/picard/analysis/replicates/CollectIndependentReplicateMetrics.java
```diff
@@ -102,50 +102,50 @@
...
-        usage = "Estimates the rate of independent replication rate of reads within a bam. \n" +
+        summary = "Estimates the rate of independent replication rate of reads within a bam. \n" +
                 "That is, it estimates the fraction of the reads which would be marked as duplicates but " +
                 "are actually biological replicates, independent observations of the data. ",
-        usageShort = "Estimates the rate of independent replication of reads within a bam.",
+        oneLineSummary = "Estimates the rate of independent replication of reads within a bam.",
         programGroup = Alpha.class
 )
 public class CollectIndependentReplicateMetrics extends CommandLineProgram {
     private static final int DOUBLETON_SIZE = 2, TRIPLETON_SIZE = 3;
```
tfenne: I saw this line and said to myself, "this has to be a @yfarjoun class", and sure enough... 😉

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -65,45 +80,57 @@
  * 4. Implement the following static method in the concrete class:
  *
  *     public static void main(String[] argv) {
-        new MyConcreteClass().instanceMain(argv);
-    }
-
-
+ *       new MyConcreteClass().instanceMain(argv);
```
tfenne: Two things: the documentation above is now out of date (it references the old annotations).  Also it's been years since Picard required `main()` methods in it's classes, I'm assuming Barclay doesn't either?  I.e. this has been inaccurate for a long time, not just made so by this commit.

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -65,45 +80,57 @@
...
  */
 public abstract class CommandLineProgram {
-    @Option(common=true, optional=true)
+    // Picard CmdLine properties file resource path, placed at this path by the gradle build script
+    private static String PICARD_CMDLINE_PROPERTIES_FILE = "picard/picardCmdLine.properties";
+    private static String PROPERTY_USE_LEGACY_PARSER = "picard.useLegacyParser";
+    private static String PROPERTY_CONVERT_LEGACY_COMMAND_LINE = "picard.convertCommandLine";
+    private static Boolean useLegacyParser;
+
+    @Argument(common=true, optional=true)
```
tfenne: Does this get `doc` some other way or should it have something here?

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -65,45 +80,57 @@
...
     public Boolean QUIET = false;
 
-    @Option(doc = "Validation stringency for all SAM files read by this program.  Setting stringency to SILENT " +
+    @Argument(doc = "Validation stringency for all SAM files read by this program.  Setting stringency to SILENT " +
             "can improve performance when processing a BAM file in which variable-length data (read, qualities, tags) " +
             "do not otherwise need to be decoded.", common=true)
     public ValidationStringency VALIDATION_STRINGENCY = ValidationStringency.DEFAULT_STRINGENCY;
 
-    @Option(doc = "Compression level for all compressed files created (e.g. BAM and GELI).", common=true)
+    @Argument(doc = "Compression level for all compressed files created (e.g. BAM and GELI).", common=true)
```
tfenne: Another archaic doc that could be fix.  GELI is long gone.  Could you fix this please to say `e.g. BAM and .gz files`?

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -254,18 +279,30 @@ public int instanceMain(final String[] argv) {
...
-        commandLineParser = new CommandLineParser(this);
-        final boolean ret = commandLineParser.parseOptions(System.err, argv);
+        commandLineParser = getCommandLineParser();
+
+        boolean ret = false;
+        try {
+            ret = commandLineParser.parseArguments(System.err, argv);
+        } catch (CommandLineException e) {
+            // Barclay command line parser throws on parsing/argument errors
+            System.err.println(e.getMessage());
```
tfenne: Yes please to reversing these two lines!

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -254,18 +279,30 @@ public int instanceMain(final String[] argv) {
...
         }
+        REFERENCE_SEQUENCE = referenceSequence.getReferenceFile();
+
         final String[] customErrorMessages = customCommandLineValidation();
         if (customErrorMessages != null) {
             for (final String msg : customErrorMessages) {
                 System.err.println(msg);
             }
-            commandLineParser.usage(System.err, false);
+            System.err.print(commandLineParser.usage(false, false));
```
tfenne: Ditto here.

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -311,20 +383,38 @@ public void setDefaultHeaders(final List<Header> headers) {
...
+    public static String getStandardUsagePreamble(final Class<?> mainClass) {
+        return "USAGE: " + mainClass.getSimpleName() +" [options]\n\n" +
+                (hasWebDocumentation(mainClass) ?
+                        "Documentation: http://broadinstitute.github.io/picard/command-line-overview.html" +
+                                mainClass.getSimpleName() + "\n\n" :
+                        "");
+    }
+
+    /**
+     * Determine if a class has wbe documentation based on its package name
```
tfenne: wbe -> web

src/main/java/picard/cmdline/PicardCommandLine.java
```diff
@@ -241,9 +243,9 @@ private static void printUsage(final Set<Class<?>> classes, final String command
                 }
                 if (!commandListOnly) {
                     if (clazz.getSimpleName().length() >= 45) {
-                        builder.append(String.format("%s    %s    %s%s%s\n", KGRN, clazz.getSimpleName(), KCYN, property.usageShort(), KNRM));
+                        builder.append(String.format("%s    %s    %s%s%s\n", KGRN, clazz.getSimpleName(), KCYN, property.summary(), KNRM));
```
tfenne: Shouldn't this also be `oneLineSummary`?

build.gradle
```diff
@@ -81,11 +89,46 @@ jar {
...
+}
+
+javadoc {
+    // This is a hack to disable the java 8 default javadoc lint until we fix the html formatting
+    // We only want to do this for the javadoc task, not gatkDoc
     options.addStringOption('Xdoclint:none', '-quiet')
 }
 
+// Generate Picard Online Doc
+task picardDoc(type: Javadoc, dependsOn: ['cleanPicardDoc', classes]) {
```
magicDGS: Will be nice to have some kind of plugin for generating the documentation with Barclay without repeating code here and in GATK (and my own software). Is there any way to do it easier?

lbergelson: @magicDGS A gradle plugin for barclay seems like a good idea.  I've never written one, but I imagine it's probably not that difficult.  I opened a ticket to discuss it #819

lbergelson: Ack, I put that in the wrong place, see https://github.com/broadinstitute/barclay/issues/56 instead.

src/main/java/picard/util/help/HelpConstants.java
```diff
@@ -0,0 +1,69 @@
...
+     * List of "supercategory" values used for doc purposes. Every doc group name can/should be put into
+     * one of the following supercategories.
+     */
+    private final static String DOC_SUPERCAT_TOOLS = "tools";
+    private final static String DOC_SUPERCAT_UTILITIES = "utilities";
+    private final static String DOC_SUPERCAT_EXCLUDE = "exclude";
+
+    private static Map<String, String> groupToSuperCategory;
+
+    private static Map<String, String> getSuperCategoryMap() {
```
magicDGS: It will be nice to have this method and the supercategory values accessible to reuse this map. For example, if I create a new program group for a tool I can add it to the tools supercategory and like that reuse this class to also include the others without copying the code in this method and re-define the supercat constants....

I would like to have this also in GATK to re-use super-categories and read groups...

cmnbroad:
@yfarjoun Can you or review or appoint someone ? Thx.

src/main/java/picard/util/help/PicardHelpDoclet.java
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Custom Barclay-based Javadoc Doclet used for generating Picard help/documentation.
+ *
+ * NOTE: Methods in this class are intended to be called by Gradle/Javadoc only, and should not be called
+ * by methods that are used by the Picard runtime. This class has a dependency on com.sun.javadoc classes,
+ * which may not be present since they're not provided as part of the normal Picard runtime classpath.
+ */
+public class PicardHelpDoclet extends HelpDoclet {
+
+    private final static String PICArD_FREEMARKER_INDEX_TEMPLATE_NAME = "generic.index.template.html";
```
yfarjoun: why the lowercase `r` in PICArD?

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -254,18 +279,30 @@ public int instanceMain(final String[] argv) {
     */
     protected boolean parseArgs(final String[] argv) {
 
-        commandLineParser = new CommandLineParser(this);
-        final boolean ret = commandLineParser.parseOptions(System.err, argv);
+        commandLineParser = getCommandLineParser();
+
+        boolean ret = false;
```
jacarey: redundant set

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -282,13 +319,48 @@ protected boolean parseArgs(final String[] argv) {
...
+    /**
+     * @return Return the command line parser to be used.
+     */
     public CommandLineParser getCommandLineParser() {
+        if (commandLineParser == null) {
+            commandLineParser = useLegacyParser(getClass()) ?
+                        new LegacyCommandLineArgumentParser(this) :
+                        new CommandLineArgumentParser(this,
+                            Collections.EMPTY_LIST,
+                            new HashSet<>(Arrays.asList(CommandLineParserOptions.APPEND_TO_COLLECTIONS)));
```
jacarey: Collections.singletonList

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -282,13 +319,48 @@ protected boolean parseArgs(final String[] argv) {
...
+     * otherwise, false.
+     *
+     * The Barclay parser is enabled by opt-in only, via the presence of a (true-valued) boolean property
+     * "picard.useLegacyParser", either as a System property, or property in a file called "picard.properties".
+     * System property value takes precedence.
+     *
+     * @return true if the legacy parser should be used
+     */
+    public static boolean useLegacyParser(final Class<?> clazz) {
+        String legacyPropertyValue = null;
```
jacarey: can be scoped inside the if and doesn't need initialization

src/main/java/picard/util/help/HelpConstants.java
```diff
@@ -0,0 +1,69 @@
+package picard.util.help;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+public final class HelpConstants {
+
+    private HelpConstants() {};
```
jacarey: extra semicolon

src/test/java/picard/cmdline/PicardCommandLineTest.java
```diff
@@ -13,4 +23,40 @@ public void TestPicardPublic() { // this tests fails if any CLP in picard is mis
...
+    // NOTE that it does NOT test that those tools actually work correctly, only that they aren't immediately
+    // rejected with a CommandLineParserInternalException from the Barclay parser.
+    @Test
+    public void TestLaunchAllCommandLineProgramsWithBarclayParser() {
+        final ClassFinder classFinder = new ClassFinder();
+        for (final String pkg : Collections.singletonList("picard")) {
+            classFinder.find(pkg, CommandLineProgram.class);
+        }
+
+        final Map<String, Class<?>> simpleNameToClass = new HashMap<String, Class<?>>();
```
jacarey: unused variable

src/main/java/picard/cmdline/CommandLineProgram.java
pshapiro4broad: Better yet replace the expression with `Collections.singleton(CommandLineParserOptions.APPEND_TO_COLLECTIONS))` or
`EnumSet.of(CommandLineParserOptions.APPEND_TO_COLLECTIONS)`. Both assume that `CommandLineArgumentParser` doesn't side effect its arguments, which it shouldn't.

yfarjoun:
@cmnbroad this seems to be waiting for you to respond and rebase...

cmnbroad:
Yes, I expect to get back to this sometime this week.

src/main/java/picard/analysis/CollectRrbsMetrics.java
```diff
@@ -127,6 +123,20 @@
...
+    // return a custom argument collection since this tool uses a (required) argument name
+    // of "REFERENCE", not "REFERENCE_SEQUENCE"
+    @Override
+    protected ReferenceArgumentCollection getReferenceArgumentCollection() {
+        return new CollectRrbsMetricsReferenceArgumentCollection();
+    }
+
+    public static class CollectRrbsMetricsReferenceArgumentCollection implements ReferenceArgumentCollection {
+        @Argument(doc = "The reference sequence fasta file", shortName = StandardOptionDefinitions.REFERENCE_SHORT_NAME)
+        public File REFERENCE;
```
cmnbroad: Barclay doesn't directly support either of these now, but I think it could be added if we wanted to go down this path since the underlying jopt library appears to support both abbreviations and multiple names.

src/main/java/picard/cmdline/CommandLineProgram.java
cmnbroad: Fixed.

src/main/java/picard/cmdline/CommandLineProgram.java
cmnbroad: Updated with doc.

src/main/java/picard/cmdline/CommandLineProgram.java
cmnbroad: Done.

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -65,45 +80,57 @@
...
 
-    @Option(doc="Whether to create an MD5 digest for any BAM or FASTQ files created.  ", common=true)
+    @Argument(doc="Whether to create an MD5 digest for any BAM or FASTQ files created.  ", common=true)
     public boolean CREATE_MD5_FILE = Defaults.CREATE_MD5;
 
-    @Option(shortName = StandardOptionDefinitions.REFERENCE_SHORT_NAME, doc = "Reference sequence file.", common = true, optional = true, overridable = true)
+    @ArgumentCollection
+    public ReferenceArgumentCollection referenceSequence = getReferenceArgumentCollection();
+
     public File REFERENCE_SEQUENCE = Defaults.REFERENCE_FASTA;
```
cmnbroad: Yes, I should have been more explicit about why I kept that. And I see your point (articulated in a comment below), about these possibly being clobbered, but CollectMultipleMetrics sets REFERENCE_SEQUENCE directly for all of the CLPs that it uses. I've reduced the visibility to protected, and added a setReferenceSequence method to SinglePassSam so that CollectMultipleMetrics can populate this, which reduces the scope of the exposure. Let me know if you have another suggestion.

On a somewhat unrelated note: in GATK4, we had previously refactored (some) of the metrics collectors that originally came from Picard in a way thats similar to what @nh13 did [here](https://github.com/broadinstitute/picard/pull/840), only more so. For each type of metric, we factored out:

1) an accumulator class for holding the results (which is also a reducer, where applicable)
2) an argument collection class for the controlling arguments
3) the collector itself, which accepts a read and populates the accumulator

In addition to codifying the ability to set the arguments for a collector independently of it's container, it entirely separates the metrics computation from the source of the reads (which in GATK's case could be a CLP, a multiple collector CLP, a Spark CLP, or a Spark RDD in a pipeline). The CLPs are just thin shells around these components. Its obsolete now since once this Barclay PR goes through we'll use the Picard version of the collectors, but I would lobby that we adopt a similar approach similar in Picard.

src/main/java/picard/cmdline/CommandLineProgram.java
cmnbroad: Done.

src/main/java/picard/cmdline/CommandLineProgram.java
cmnbroad: Done.

src/main/java/picard/analysis/CollectWgsMetrics.java
cmnbroad: Made protected.

src/main/java/picard/analysis/CollectWgsMetrics.java
cmnbroad: Done.

src/main/java/picard/util/help/PicardHelpDoclet.java
cmnbroad: Typo. Fixed.

src/main/java/picard/cmdline/CommandLineProgram.java
cmnbroad: Fixed.

src/main/java/picard/analysis/CollectWgsMetrics.java
cmnbroad: Yeah, "make" seems like a better choice. Done.

src/main/java/picard/analysis/directed/CollectTargetedMetrics.java
cmnbroad: Barclay retains this, um, feature, and works the same way. It causes some [problems](https://github.com/broadinstitute/barclay/issues/23) though so I'd like to change it at some point.

Anyway, for now Ill make no change other than to remove my comment.

src/main/java/picard/cmdline/PicardCommandLine.java
cmnbroad: Hm, yes it should....Done.

src/main/java/picard/util/help/HelpConstants.java
cmnbroad: @magicDGS We can certainly widen the visibility (here and in GATK), assuming you mean that you want to be able to add entries to this from your custom doclet ?

src/main/java/picard/cmdline/CommandLineProgram.java
cmnbroad: Fixed.

src/main/java/picard/cmdline/CommandLineProgram.java
cmnbroad: Done.

src/main/java/picard/cmdline/CommandLineProgram.java
cmnbroad: Done.

src/main/java/picard/util/help/HelpConstants.java
cmnbroad: Done.

src/test/java/picard/cmdline/PicardCommandLineTest.java
cmnbroad: Done.

src/main/java/picard/util/help/HelpConstants.java
magicDGS: Exactly. Either add entries from my custom doclet or use as default and override if necessary some of the supercategories (by creating a new one, not directly here, to be safer).

cmnbroad:
@tfenne @jacarey @pshapiro4broad @yfarjoun Sorry for the delay on follow up. I've added one commit with changes based on all of the review comments. Once everyone is good with those, I'd like to squash all of these commits down to a single commit before I rebase (this is now about a month behind master), since its painful to keep the changes in the individual commits in sync through a rebase. Once I've rebased I'll make a separate commit with any new changes required at that point. Finally, I'll update my corresponding picard-private branch and make the PR for that.

nh13:
@cmnbroad this is great! When you have rebased, I'd like to build a copy and make sure that it "works" in my hands and some of my pipelines.  Thank-you for all this effort.

cmnbroad:
@ktibbett @yfarjoun Also, as I mentioned in the original PR comments, in this PR I had to change the "R" argument for DownsampleSam (which uses "R" for the random seed) since it collides with the REFERENCE short name). Please let me know if you guys *really* can't live with that, in which case I'll have to add a little bit of ugly coding/typecasting to preserve the existing command line arg name.

vdauwera:
@cmnbroad Re: the DownsampleSam  random seed thing, what happens if the user doesn't update their command accordingly? I think the desired behavior should be to fail immediately with an informative error message ("yo, update your command, buddy -- R means reference now; the new arg name for randmo seed is [blah]"). If it fails late or screws up silently then we'd want to put in some logic to catch the case.

cmnbroad:
@vdauwera Good question. I just tried it, and it fails early, but badly (with an exception/stacktrace), with either parser. If we keep it, I can add in some code that checks to see if the reference you specified looks more like a random seed than a filename, and fail a bit more gracefully.

cmnbroad:
@nh13 A test drive before we merge this would be great - I'll ping you when its rebased.

cmnbroad:
@nh13 This is rebased and up-to-date on current Picard as of today if you want to try it out.

cmnbroad:
@vdauwera Also, I removed the DownsampleSam issue, so this is compatible now with the existing command line.

pshapiro4broad:
Looks ok to me 👍

build.gradle
```diff
@@ -94,6 +165,46 @@ task currentJar(type: Copy){
...
+    archiveName 'picardBarclay.jar'
+}
+
+task barclayTest(type: Test) {
+    // Run all the tests using, but with the command line converter and the Barclay command line parser
+    systemProperty 'picard.convertCommandLine', 'true'
+    systemProperty 'picard.useLegacyParser', 'false'
+}
+
+test {
```
lbergelson: Could you add a `legacyTest(type: Test)` target and make test depend on both that and `barclayTest` so that it's possible to run just one or the other.

build.gradle
```diff
@@ -94,6 +165,46 @@ task currentJar(type: Copy){
...
+            fileName.replace(legacySourcePropertyFile, picardTargetPropertyFile)
+        }
+    }
+}
+
+// Create picardBarclay.jar, which is a identical to picard.jar, but contains a .properties
+// file that tells Picard to use the Barclay command line parser instead of the Picard
+// command line parser.
+//
+task barclayJar(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar)  {
```
lbergelson: Shouldn't this be `barclayShadowJar`?

build.gradle
```diff
@@ -257,6 +276,9 @@ task sourcesJar(type: Jar) {
  * This specifies what artifacts will be built and uploaded when performing a maven upload.
  */
 artifacts {
+    // Make sure the barclayJar is listed first to prevent it from being chosen as the one to be
+    // installed under the snapshot name in the local maven repository when doing "./gradlew install".
+    archives barclayJar
```
lbergelson: I think we want to upload a non-shadow jar as the barclay jar we archive so that gatk can depend on it.  We probably need to add another jar task to make that.

build.gradle
```diff
@@ -322,25 +344,3 @@ uploadArchives {
         System.out.println("Uploading version $version")
     }
 }
-
-//update static web docs
-task copyJavadoc(dependsOn: 'javadoc', type: Copy) {
```
lbergelson: Are we ok with no longer updating the picard website?

build.gradle
```diff
@@ -78,12 +107,54 @@ jar {
                 'Implementation-Vendor': 'Broad Institute',
                 'Implementation-Version': version
     }
+    from('src/main/resources/properties.templates') {
```
lbergelson: This repeated copy block is so ugly.  I'm trying to come up with a way around it but I'm not sure how to do it in a way that isn't worse.

cmnbroad: Yeah, I'm sure there is some way. Let me know if you have something.

build.gradle
```diff
@@ -78,12 +107,54 @@ jar {
...
+    }
+}
+
+tasks.withType(Javadoc) {
+    // do this for all javadoc tasks, including gatkDoc
+    options.addStringOption('Xdoclint:none')
+}
+
+javadoc {
+    // This is a hack to disable the java 8 default javadoc lint until we fix the html formatting
```
cmnbroad: Done.

build.gradle
```diff
@@ -78,12 +107,54 @@ jar {
...
+    // We only want to do this for the javadoc task, not gatkDoc
+    options.addStringOption('Xdoclint:none', '-quiet')
```
cmnbroad: Without it, the javadoc task causes the build to fail due to javadoc issues.

build.gradle
cmnbroad: Yes, makes sense.

build.gradle
```diff
@@ -94,6 +165,46 @@ task currentJar(type: Copy){
...
+
+// Create picardBarclay.jar, which is a identical to picard.jar, but contains a .properties
+// file that tells Picard to use the Barclay command line parser instead of the Picard
+// command line parser.
+//
+task barclayJar(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar)  {
+    configurations = [project.configurations.runtime]
+    from project.sourceSets.main.output
+    manifest {
+        inheritFrom project.tasks.jar.manifest
```
cmnbroad: Yeah - I think I could also move the manifest 'attributes' into that as well.

build.gradle
cmnbroad: Sure.

build.gradle
cmnbroad: @vdauwera We've discussed this in the past, but just want to confirm: once we commit this branch, the only doc generation capability we'll have for Picard will be the equivalent of gatkDoc, which will generate Barclay-style syntax, and will require a the addition of @DocumentedFeature annotations).

build.gradle
```diff
@@ -94,6 +165,46 @@ task currentJar(type: Copy){
...
+    }
+    from('src/main/resources/properties.templates') {
+        // for the default jar, we want the properties file that enables the picard parser
+        include barclaySourcePropertyFile
+        into picardTargetPropertiesPath
+        rename { String fileName ->
+            fileName.replace(barclaySourcePropertyFile, picardTargetPropertyFile)
+        }
+    }
+    exclude('META-INF/INDEX.LIST', 'META-INF/*.SF', 'META-INF/*.DSA', 'META-INF/*.RSA')
```
cmnbroad: Good catch - I don't recall what made me add this, but I don't think we need it here. Removed.

cmnbroad:
@lbergelson Ping - I'm ready to rebase this on current code once you review the gradle changes I made based on your comments.

cmnbroad:
@ktibbett @jacarey @vdauwera @nh13 @tfenne @pshapiro4broad I think all the stars are finally aligned, and I’m planning to squash and rebase this one more time, and merge, on Friday. Please let me know if there any concerns with that plan.

Note (@ktibbett @jacarey) that once this branch is merged, picard-private can’t be upgraded to a new picard build again until the corresponding picard-private [PR](https://github.com/broadinstitute/picard-private/pull/987) is also merged. In order to minimize the window of time where the two repos are out of sync, I’d like to get that PR reviewed as soon as possible. Can you assign someone to do that ? It passes tests locally, but won’t actually build on Jenkins until the this PR is merged and the Picard snapshot is available via artifactory.

build.gradle
vdauwera: @cmnbroad Sorry for the delayed response, I was out for a while then catching up. 

I can do the addition of @documentedfeature annotations. Should that be part of this pull request or one queued up to follow right away afterward? 

We do need to set up the release system to build and publish the gatkDocs on the http://broadinstitute.github.io/picard/ website, as I don't think anyone is ready yet to have Picard docs on the GATK website only.  

Github now provides a simpler option for publishing a doc website; based on [this post](https://github.com/blog/2228-simpler-github-pages-publishing) it looks like we just need to switch a repo setting and we'll be able to simply copy the docs in a `/docs` directory. I'll edit the template to work with the picard theme. 

Am I missing anything?

vdauwera:
We need a bit more wiring to migrate the docs publishing, see my comment above.

PR #818, opened 2017-05-17
Move the command line error to be printed LAST!

nh13:
@tfenne, this one's for you 🍺 

p0l0xvckvbluu

### DIFF
--- a/src/main/java/picard/cmdline/CommandLineProgram.java
+++ b/src/main/java/picard/cmdline/CommandLineProgram.java
@@ -42,10 +42,7 @@ import htsjdk.variant.variantcontext.writer.VariantContextWriterBuilder;
 import java.io.File;
 import java.net.InetAddress;
 import java.text.DecimalFormat;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 /**
  * Abstract class to facilitate writing command-line programs.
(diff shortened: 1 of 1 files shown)


yfarjoun:
Thanks! this annoys me SOO much!

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -262,10 +259,17 @@ protected boolean parseArgs(final String[] argv) {
...
         final String[] customErrorMessages = customCommandLineValidation();
         if (customErrorMessages != null) {
+            commandLineParser.usage(System.err, false);
+            if (customErrorMessages.length > 0) {
+                System.err.println(String.join("", Collections.nCopies(80, "#")));
+                System.err.println();
+            }
             for (final String msg : customErrorMessages) {
-                System.err.println(msg);
+                System.err.println("Error: " + msg + "\n");
```
tfenne: I don't think you need the trailing `\n` since you're invoking `println`

cmnbroad:
First of all - brilliant! Wish I'd thought of that.

But, I think @tfenne is right - to fully fix this we'd have to capture the PrintStream we give to the CLP onto a ByteArrayOutputStream and then write it to stderr after the CLP returns. The Barclay parser throws on errors, but I followed the same ordering in the catch clause in my PR. Anyway, if we don't do this here I'd be happy to fold it in as part of the Barclay PR, since it will save a lot of scrolling...

nh13:
@cmnbroad if you can make it happen in Barclay, I'm all for closing this.

PR #821, opened 2017-05-26
Yf rev  htsjdk

yfarjoun:
This brings picard to htsjdk 2.9.1 and fixes a small ambiguity in a constructor in one of the tests.

### DIFF
--- a/src/test/java/picard/util/FifoBufferTest.java
+++ b/src/test/java/picard/util/FifoBufferTest.java
@@ -57,8 +57,8 @@ public class FifoBufferTest {
         }
 
         // Run the input file through the FifoBuffer and back into another file
-        final Md5CalculatingInputStream in   = new Md5CalculatingInputStream(new FileInputStream(inputFile), null);
-        final Md5CalculatingOutputStream out = new Md5CalculatingOutputStream(new FileOutputStream("/dev/null"), null);
+        final Md5CalculatingInputStream in   = new Md5CalculatingInputStream(new FileInputStream(inputFile), (File)null);
+        final Md5CalculatingOutputStream out = new Md5CalculatingOutputStream(new FileOutputStream("/dev/null"), (File)null);
         final PrintStream outStream = new PrintStream(out);
         final FifoBuffer buffer = new FifoBuffer(in, outStream);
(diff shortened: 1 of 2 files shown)


ronlevine:
LGTM

ronlevine:
LGTM

src/test/java/picard/util/FifoBufferTest.java
```diff
@@ -57,8 +57,8 @@ public void test(final double megabytes) throws IOException {
         }
 
         // Run the input file through the FifoBuffer and back into another file
-        final Md5CalculatingInputStream in   = new Md5CalculatingInputStream(new FileInputStream(inputFile), null);
-        final Md5CalculatingOutputStream out = new Md5CalculatingOutputStream(new FileOutputStream("/dev/null"), null);
+        final Md5CalculatingInputStream in   = new Md5CalculatingInputStream(new FileInputStream(inputFile), (File)null);
```
ronlevine: The cast is not necessary, only one cosntructor.

src/test/java/picard/util/FifoBufferTest.java
```diff
@@ -57,8 +57,8 @@ public void test(final double megabytes) throws IOException {
...
+        final Md5CalculatingOutputStream out = new Md5CalculatingOutputStream(new FileOutputStream("/dev/null"), (File)null);
```
ronlevine: `FifoBufferTest.testFifoBuffer` should throw a NPE since this `Md5CalculatingOutputStream` constructor uses the statement `digestFile.toPath()` and `digestFile` in null. But, there are no errors in the test results. Did travis not run the tests? Since you fixed the NPE in https://github.com/samtools/htsjdk/pull/881/files you can either grab the latest `htsjdk SNAPSHOT` or cast the `null` to `Path`.

yfarjoun: I don't want to use a snapshot here, only official releases. 

Strange about the tests not failing here..I agree.

pshapiro4broad: It would be nice to update this code at some point. E.g. use `Path` not `File` and use try/catch to open the files to avoid the need to `close()`.

Issue #794, opened 2017-04-14
ValidateSamFile should return more informative error codes


vdauwera:
### Tool(s) involved
ValidateSamFile

### Description
ValidateSamFile returns 1 when a file fails validation and when the tool crashes. The tool should return different codes so that we can differentiate "tool crashed" from "file has some problems". Ideally we'd want to further distinguish "there were some WARNs" from "there were some ERRORs".

This is important for running the tool in pipelining environments like FireCloud.

ronlevine:
ValidateSamFile calls the HTSJDK methods SamFileValidator#validateSamFileSummary and SamFileValidator#validateSamFileVerbose, which return boolean values to indicate validation errors. The validation errors are stored as an enum, SAMValidationError.Type. These methods need to return the enum's ordinal. 
@vdauwera If a numerical value is returned, how would the pipeline know what it corresponds to? Should an error message be added to each SAMValidationError.Type value?

tfenne:
@ronlevine Maybe I'm missing something, but I think @vdauwera was asking for something simpler - that if the file does not validate that the tool exit with any value other than 0 or 1, so you might end up with:
  `exit(0)` == successfully read whole file and it was all valid
  `exit(1)` == the program failed to complete execution (exception, IO issue, etc.)
  `exit(2)` == successfully read the whole file but there were validation warnings
  `exit(3)` == successfully read the whole file but there were validation errors

With separate exits for validation warnings/errors being a nice to have.

ronlevine:
@tfenne I might be missing something. I am not familiar with the use cases for this tool within the pipelines. If those exit cases are as fine grained as needed, I'll go with that.

vdauwera:
Yep, @tfenne has it right, that's all we would need. Sorry I wasn't clearer in the original description. 

In the case of Cromwell we can set the execution engine to evaluate the return code intelligently, so it can correctly identify that the tool execution succeeded, yet still convey programmatically that the data file was not fully valid.

ronlevine:
HTSJDK's `SamFileValidator` needs to keep track of warnings and errors. `ValidateSamFile` can return the following error codes:
-1  - `program failed to complete execution`
0   - `program ran successfully`
1   - `validation warnings but no errors`
2  - `validation errors and possibly warnings`

ronlevine:
Changes in SamFileValidator are blocked until htsjdk > 2.9.1 is released.

ronlevine:
[...]

PR #822, opened 2017-05-29
ValidateSamFile returns informative error codes

ronlevine:
### Description
Implements #794.
`ValidateSamFile` needs to return different error codes for running in pipelining environments like FireCloud. `ValidateSamFile` will return the following:
```
exit(-1) - program failed to complete execution
exit(0) - program ran successfully
exit(1) - validation warnings but no errors
exit(2) - validation errors and possibly warnings
```
This pull request requires an HTSJDK snapshot because https://github.com/samtools/htsjdk/pull/855 has not yet been incorporated into a release. I also fixed the use of an ambiguous constructor in FifoBufferTest.java.

----

#### Content
- [X] All tests passing on Travis

#### Review
- [X] Rebase, squash and reword as applicable


ronlevine:
@yfarjoun @vdauwera Please review.

src/main/java/picard/sam/ValidateSamFile.java
```diff
@@ -133,75 +133,117 @@ public static void main(final String[] args) {
         System.exit(new ValidateSamFile().instanceMain(args));
     }
 
+    /**
+     * Return tyeps for doWork()
+     */
+    enum ReturnTypes {
+        FAILED(-1),     // failed to complete execution
+        SUCCESSFUL(0),  // ran successfully
+        WARNINGS(1),    // warnings but no errors
```
magicDGS: I would like to have a parameter to treat "warnings" as successful runs, because if not they could stop pipelines anyway even if there are no errors.

src/main/java/picard/sam/ValidateSamFile.java
```diff
@@ -133,75 +133,117 @@ public static void main(final String[] args) {
...
+        ERRORS(2);      // errors and possibly warnings
```
magicDGS: What about having also a difference between "only errors" and "errors + warnings"?

ronlevine: Is there a need?

src/main/java/picard/sam/ValidateSamFile.java
ronlevine: Is it useful to process warnings differently? Refer to @tfenne 's comment in https://github.com/broadinstitute/picard/issues/794.

src/main/java/picard/sam/ValidateSamFile.java
magicDGS: I do not have the need, but it does not add too much complexity and it may be useful in some cases to know if warnings and errors should be fixed, or just errors.

ronlevine: Will do.

src/main/java/picard/sam/ValidateSamFile.java
magicDGS: Warnings does not mean that the file is not a valid SAM file, but that it could have some problems (such as no read groups).

Although there is an option to ignore warnings, I can imagine a situation where you want to keep track of the possible problems with the file but continue the pipeline as the file does not contain "ERRORS". Although it may be not the recommended way of doing this, I guess that it is a fair use case...

src/main/java/picard/sam/ValidateSamFile.java
ronlevine: Done.

ronlevine:
@yfarjoun Can you take a look? It's now using HTSJDK 2.10.0 instead of a snapshot.

src/main/java/picard/sam/ValidateSamFile.java
```diff
@@ -133,75 +133,122 @@ public static void main(final String[] args) {
...
+        int value() {
+            return value;
+        }
+    }
+
+    /**
+     * Do the work after command line has been parsed.
+     * 
+     * @return 0 - ran successfully without warnings or errors, 1 - warnings but no errors, 2 - errors and possibly warnings,
+     * -1 - failed to complete execution
```
yfarjoun: javadoc is not accurate

src/main/java/picard/sam/ValidateSamFile.java
```diff
@@ -133,75 +133,122 @@ public static void main(final String[] args) {
...
+                } else {
+                    if (validator.getNumWarnings() > 0) {
+                        return ReturnTypes.ERRORS_WARNINGS.value();   // errors and warnings
+                    } else {
+                        return ReturnTypes.ERRORS.value();  // errors but no warnings
+                    }
+                }
+            }
+        } catch (Exception e) {
+            return ReturnTypes.FAILED.value();  // failed to complete execution
```
yfarjoun: you eat the exception here which will make the output very annoying. 

Can you log the exception message via the logger?

src/main/java/picard/sam/ValidateSamFile.java
```diff
@@ -133,75 +133,122 @@ public static void main(final String[] args) {
...
         }
-        out.flush();
 
-        return result ? 0 : 1;
+        return ReturnTypes.SUCCESSFUL.value();
```
yfarjoun: The only way you get to this line is that `result==false` even though `getNumWarnings()==0 && getNumErrors()==0`.

Since this is a inconsistent state, I would consider that a failure to execute correctly. I would log an error stating this and return FAILED

src/main/java/picard/sam/ValidateSamFile.java
ronlevine: Fixed.

src/main/java/picard/sam/ValidateSamFile.java
ronlevine: Yes. Done.

src/main/java/picard/sam/ValidateSamFile.java
ronlevine: Done.

ronlevine:
@yfarjoun Made post review changes. Please take another look.

yfarjoun:
Now that htsjdk has been rev'ed, please remove your changes to gradle and FifoPipeTest and rebase.

Then you should be good to go!

👍

PR #823, opened 2017-05-30
Added new data provider for cbcls (NovaSeq)

jacarey:
This PR adds a new IlluminaDataProvider that can handle cbcl files.  It eschews the standard data provider for several reasons:

1) Cbcls are concatenated tile files (essentially a per surface file) which was not modeled in the old provider.
2) The old provider handles lots of different (some defunct) file formats that aren't needed.
3) The old provider assumes locations are tile based, but with NovaSeq (and HiSeqX) these are pad locs and similar for all tiles.
4) It gave us the opportunity to do some performance enhancements.

There is a fair amount of refactor to allow use to use the factory machinery for the old data provider which allowed us to minimize the changes needed to add it to our CLPs.

----

#### Content

#### Review


src/main/java/picard/illumina/BarcodeMetric.java
```diff
@@ -0,0 +1,128 @@
+package picard.illumina;
+
+import htsjdk.samtools.metrics.MetricBase;
+
+/**
+ * Metrics produced by the ExtractIlluminaBarcodes program that is used to parse data in
+ * the basecalls directory and determine to which barcode each read should be assigned.
+ */
+public class BarcodeMetric extends MetricBase {
```
nh13: Since this is a breaking change, perhaps leave the inner class in `ExtractIlluminaBarcodes` and deprecate it?  Likely it will look ugly in the online documentation, but I know that myself I both rely on this class `ExtractIlluminaBarcodes` and point folks to its documentation.

src/main/java/picard/illumina/BarcodeMetric.java
```diff
@@ -0,0 +1,128 @@
...
+     */
+    public double PCT_MATCHES = 0d;
+    /**
+     * The rate of all reads matching this barcode to all reads matching the most prevalent barcode. For the
+     * most prevalent barcode this will be 1, for all others it will be less than 1 (except for the possible
+     * exception of when there are more orphan reads than for any other barcode, in which case the value
+     * may be arbitrarily large).  One over the lowest number in this column gives you the fold-difference
+     * in representation between barcodes.
+     */
+    public double RATIO_THIS_BARCODE_TO_BEST_BARCODE_PCT = 0d;
```
yfarjoun: if this needs to go into a certain DB, then the name here is too long.

src/main/java/picard/illumina/BarcodeMetric.java
```diff
@@ -0,0 +1,128 @@
...
+    public double PF_PCT_MATCHES = 0d;
+
+    /**
+     * The rate of PF reads matching this barcode to PF reads matching the most prevalent barcode. For the
+     * most prevalent barcode this will be 1, for all others it will be less than 1 (except for the possible
+     * exception of when there are more orphan reads than for any other barcode, in which case the value
+     * may be arbitrarily large).  One over the lowest number in this column gives you the fold-difference
+     * in representation of PF reads between barcodes.
+     */
+    public double PF_RATIO_THIS_BARCODE_TO_BEST_BARCODE_PCT = 0d;
```
yfarjoun: here too (DB)

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -57,117 +63,178 @@
...
+            }
 
         } catch (final IOException ioe) {
             throw new RuntimeIOException(ioe);
         }
     }
 
+    private Map<Integer, Map<Integer, File>> sortCbcls(List<File> cbcls) {
+        Map<Integer, Map<Integer, File>> sortedMap = new TreeMap<>();
+        Pattern pattern = Pattern.compile("^.+C(\\d{1,4}).+L(\\d{1,3})_(\\d).cbcl$");
```
pshapiro4broad: The result of `Pattern.compile()` is thread safe so it can be stored in a static final field to avoid doing it more than once.

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -57,117 +63,178 @@
...
+        for (File cbcl : cbcls) {
+            Matcher matcher = pattern.matcher(cbcl.getAbsolutePath());
+            if (!matcher.matches()) {
+                throw new PicardException("CBCL File " + cbcl.getAbsolutePath() + " does not match expected pattern.");
+            }
+            Integer surface = Integer.valueOf(matcher.group(3));
```
pshapiro4broad: If these values can never be null they should be stored in `int` not `Integer`.

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -57,117 +63,178 @@
...
+    }
+
     @Override
     public boolean hasNext() {
-        if (queue.isEmpty()) {
+        if (queue == null) {
             advance();
         }
-        return !queue.isEmpty();
+        return !(queue == null);
```
pshapiro4broad: Is this form preferred over `queue != null`? It seems to me that `!=` is clearer than `!(==)`

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -57,117 +63,178 @@
...
-                byte[] qualityBins = new byte[numberOfBins];
-                //each bin has a pair of 4 byte mappings
-                for (int j = 0; j < numberOfBins; j++) {
-                    headerBuffer.getInt(); // first int is "from" value, which we don't need
-                    int to = headerBuffer.getInt();
-                    qualityBins[j] = (byte) to;
-                }
+                    if (bitsPerBasecall != 2 && bitsPerBasecall != bitsPerQualityScore) {
+                        close();
+                        throw new PicardException("CBCL data not encoded in nibbles. (not currently supported)");
```
pshapiro4broad: I was going to correct this as `nybble` but apparently nibble is the accepted spelling. Maybe this changed in the last 20 years?

src/main/java/picard/illumina/parser/BaseIlluminaDataProvider.java
```diff
@@ -0,0 +1,77 @@
+package picard.illumina.parser;
+
+import java.util.Iterator;
+
+public abstract class BaseIlluminaDataProvider implements Iterator<ClusterData>, Iterable<ClusterData> {
```
pshapiro4broad: public class should have a javadoc

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -180,126 +247,124 @@ public void close() {
...
-                } catch (final IOException ioe) {
-                    throw new RuntimeIOException(ioe);
-                }
+                decodeQualityBinnedBasecall(data, read, cycle, singleByte, currentCycleData);
 
+                totalCycleCount++;
             }
         }
-        this.queue.add(data);
+        data.setPositionInfo(positionInfoIterator.next());
```
pshapiro4broad: This part of the code is confusing to me. I was trying to figure out why this was set here instead of being passed in as an argument to the `CbclData` constructor. I see that this iterator is updated both in `advance()` and in `cacheFilterAndLocs()` but it wasn't obvious to me if one of these was called from the other or not.

src/test/java/picard/illumina/IlluminaBasecallsToFastqTest.java
```diff
@@ -132,18 +137,100 @@ public void testDualBarcodes() throws Exception {
...
+        File f2 = new File(testDataDir, filename);
+        FastqReader reader1 = new FastqReader(f1);
+        List<FastqRecord> reads = new ArrayList<>();
+        FastqReader reader2 = new FastqReader(f2);
+        for (FastqRecord record : reader1) {
+            reads.add(record);
+        }
+
+        for (FastqRecord record : reader2) {
+            Assert.assertTrue(reads.contains(record));
```
pshapiro4broad: This asserts that reader1 contains all of reader2 but not the reverse. Is that sufficient?

I'm guessing there are expected differences in the files that prohibit comparing them directly?

src/test/java/picard/illumina/IlluminaBasecallsToSamTest.java
```diff
@@ -252,6 +254,54 @@ public void testDualBarcodes() throws Exception {
...
+        runNewConverterTest(1, "dualBarcode.", "barcode_double.params", 2, "151T8B8B151T", TEST_DATA_DIR_WITH_CBCLS, DUAL_CBCL_TEST_DATA_DIR);
+    }
+
+    private void runNewConverterTest(final int lane, final String jobName, final String libraryParamsFile,
+                                     final int concatNColumnFields, final String readStructure, final File baseCallsDir,
+                                     final File testDataDir) throws Exception {
+        final File outputDir = File.createTempFile(jobName, ".dir");
+        outputDir.delete();
+        outputDir.mkdir();
+        outputDir.deleteOnExit();
```
pshapiro4broad: This kind of code for tests makes me wish we were using JUnit instead of TestNG. You could get rid of a lot of this cruft by using JUnit's `TemporaryFolder` rule.

src/test/java/picard/illumina/parser/readers/CbclReaderTest.java
```diff
@@ -7,39 +7,41 @@
...
 public class CbclReaderTest {
 
-    public static final File TestDataDir = new File("testdata/picard/illumina/readerTests");
-    public static final File PASSING_CBCL_C99_1 = new File(TestDataDir, "C99.1.cbcl");
-    public static final File PASSING_CBCL_C100_1 = new File(TestDataDir, "C100.1.cbcl");
+    public static final File TestDataDir = new File("testdata/picard/illumina/readerTests/cbcls");
+    public static final File PASSING_CBCL_C99_1 = new File(TestDataDir + "/C1.1", "L001_1.cbcl");
+    public static final File PASSING_CBCL_C100_1 = new File(TestDataDir + "/C2.1", "L001_1.cbcl");
     public static final File TILE_1101_FILTER = new File(TestDataDir, "tile_1101.filter");
     public static final File TILE_1102_FILTER = new File(TestDataDir, "tile_1102.filter");
```
pshapiro4broad: This appears to be unused now.

src/test/java/picard/illumina/parser/readers/CbclReaderTest.java
```diff
@@ -7,39 +7,41 @@
...
-        filters.put(1102, TILE_1102_FILTER);
-        reader.addFilters(filters);
+        LocsFileReader locsFileReader = new LocsFileReader(new File("testdata/picard/illumina/readerTests/s_1_6.locs"));
+        List<AbstractIlluminaPositionFileReader.PositionInfo> locs = locsFileReader.toList();
+
+        final CbclReader reader = new CbclReader(Arrays.asList(PASSING_CBCL_C99_1, PASSING_CBCL_C100_1), filters, new int[]{2}, 1101, locs);
 
         int i = 0;
         while (reader.hasNext()) {
             final BclData bv = reader.next();
```
pshapiro4broad: Should this be `final CbclData ...`?

src/main/java/picard/illumina/parser/PerTileFileUtil.java
```diff
@@ -22,9 +21,9 @@ public PerTileFileUtil(final String extension, final File base,
         super(true, extension, base, faker, lane, skipEmptyFiles);
         this.fileMap = getTiledFiles(base, matchPattern);
         if (!fileMap.isEmpty()) {
-            this.tiles = Collections.unmodifiableList(new ArrayList<Integer>(this.fileMap.keySet()));
+            this.tiles = new ArrayList<>(this.fileMap.keySet());
         } else {
-            this.tiles = new ArrayList<Integer>();
+            this.tiles = new ArrayList<>();
```
pshapiro4broad: Could use `Collections.emptyList()`

src/main/java/picard/illumina/parser/ParameterizedFileUtil.java
```diff
@@ -151,7 +151,7 @@ private String makeLaneRegex(final String fileNameEndPattern, final int lane) {
     /**
      * The period separator is expected in the file extension, since some do not start with it
```
pshapiro4broad: If this is now public, it should have a better javadoc

src/main/java/picard/illumina/parser/NewIlluminaDataProvider.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Parse cbcls Illumina Basecall files, and use them the to populate
+ * ClusterData objects.
+ */
+public class NewIlluminaDataProvider extends BaseIlluminaDataProvider {
```
pshapiro4broad: Maybe this should be `CbclIlluminaDataProvider`? `new` is only valid until the next file format appears...

src/main/java/picard/illumina/parser/NewIlluminaDataProvider.java
```diff
@@ -0,0 +1,116 @@
...
+            filterFileMap.put(fileToTile(filterFile.getName()), filterFile);
+        }
+        this.reader = new CbclReader(cbcls, filterFileMap, outputMapping.getOutputReadLengths(), tileNum, locs);
+        this.outputReadStructure = outputMapping.getOutputReadStructure();
+
+        this.noMatchMetric = noMatchMetric;
+    }
+
+    @Override
+    public void close() {
```
pshapiro4broad: If `close()` is required to release resources associated with this object, maybe `BaseIlluminaDataProvider` should implement `AutoCloseable`.

src/main/java/picard/illumina/parser/NewIlluminaDataProvider.java
```diff
@@ -0,0 +1,116 @@
...
+                minMismatchDelta, minimumBaseQuality);
+
+        if (match.isMatched()) {
+            cluster.setMatchedBarcode(match.getBarcode());
+        }
+        return cluster;
+    }
+
+    private Integer fileToTile(final String fileName) {
+        final Matcher matcher = Pattern.compile("^s_\\d+_(\\d{1,5}).+").matcher(fileName);
```
pshapiro4broad: Another `Pattern.compile()` that could be done statically. Putting it in a static also provides an opportunity to document it, since most regular expressions are harder to read than they are to write.

src/main/java/picard/illumina/parser/IlluminaDataProviderFactory.java
```diff
@@ -68,10 +70,10 @@
...
      */
-    private static final Map<IlluminaDataType, List<SupportedIlluminaFormat>> DATA_TYPE_TO_PREFERRED_FORMATS = new EnumMap<IlluminaDataType, List<SupportedIlluminaFormat>>(IlluminaDataType.class);
+    private static final Map<IlluminaDataType, List<SupportedIlluminaFormat>> DATA_TYPE_TO_PREFERRED_FORMATS = new EnumMap<>(IlluminaDataType.class);
 
     static {
-        /** For types found in Qseq, we prefer the NON-Qseq file formats first.  However, if we end up using Qseqs then we use Qseqs for EVERY type it provides,
+        /* For types found in Qseq, we prefer the NON-Qseq file formats first.  However, if we end up using Qseqs then we use Qseqs for EVERY type it provides,
          * see determineFormats
          */
         DATA_TYPE_TO_PREFERRED_FORMATS.put(IlluminaDataType.BaseCalls, makeList(
```
pshapiro4broad: Why `makeList()` instead of `Arrays.asList()`?

src/main/java/picard/illumina/NewIlluminaBasecallsConverter.java
```diff
@@ -0,0 +1,392 @@
...
+
+        if (cbcls.size() == 0) {
+            throw new PicardException("No CBCL files found.");
+        }
+
+        IOUtil.assertFilesAreReadable(cbcls);
+
+        //locs
+        File locsFile = new File(basecallsDir.getParentFile(), "s.locs");
+        LocsFileReader locsFileReader = new LocsFileReader(locsFile);
```
pshapiro4broad: Seems like you'd want to put this in a try/catch to avoid not closing the reader on exception. E.g.
```
        try (LocsFileReader locsFileReader = new LocsFileReader(locsFile)) {
            while (locsFileReader.hasNext()) {
                locs.add(locsFileReader.next());
            }
        }
```

src/main/java/picard/illumina/IlluminaBasecallsToSam.java
```diff
@@ -282,28 +297,42 @@ private void initialize() {
...
-        /**
+        if (USE_NEW_CONVERTER) {
+            final MetricsFile<BarcodeMetric, Integer> metrics = getMetricsFile();
+
+            if (METRICS_FILE != null) {
+                IOUtil.assertFileIsWritable(METRICS_FILE);
+            } else {
+                log.warn("No METRIC_FILE specified. Barcode metrics will not be generated.");
+            }
+            basecallsConverter = new NewIlluminaBasecallsConverter<>(BASECALLS_DIR, BARCODES_DIR, LANE, readStructure,
```
pshapiro4broad: If we don't use the type argument to the generic, does it need to be declared as a generic? It doesn't seem like it's buying us anything unless a type is used.

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -57,117 +63,189 @@
...
+                      int tileNum, List<AbstractIlluminaPositionFileReader.PositionInfo> locs, boolean headerOnly) {
         super(outputLengths);
+        surfaceToTileToCbclMap = sortCbcls(cbcls);
+        this.filterFileMap = filterFileMap;
         cycleData = new CycleData[cycles];
-        cachedTile = new ByteBuffer[cycles];
-        currentTile = new int[cycles];
+        cachedTile = new byte[cycles][];
+        cachedTilePosition = new int[cycles];
+        for (int i = 1; i <= cycles; i++) {
```
pshapiro4broad: It looks like this could be represented as an array instead of a map.

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -180,126 +258,147 @@ public void close() {
...
+
+    public CycleData[] getCycleData() {
+        return cycleData;
+    }
+
+    public int getHeaderSize() {
+        return headerSize;
+    }
+
+    public List<File> getFilesForCycle(int i) {
```
pshapiro4broad: The way this is used from `CheckIlluminaDirectory.doWork()` is kind of awkward. That code calls `getAllTiles()` and then uses the map key to call this method. If instead `getAllTiles()` return a list of pairs of `List<TitleData>` and `List<File>` then I think you'd only need a single API. Actually it looks like the entire block in `CheckIlluminaDirectory` could be moved into this class, as `validate()` or some similar method.

src/main/java/picard/illumina/parser/NewIlluminaDataProvider.java
```diff
@@ -0,0 +1,118 @@
...
+import picard.illumina.parser.readers.CbclReader;
+
+import java.io.File;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static picard.illumina.ExtractIlluminaBarcodes.PerTileBarcodeExtractor.findBestBarcodeAndUpdateMetrics;
```
pshapiro4broad: Do we have a policy on static imports? I think in general it would be better to avoid them. In this case it looks like `PerTileBarcodeExtractor` should be promoted to a top-level class, to make it easier to work with.

My preference is to only allow them in test code, for using APIs like mockito that read better with static imports.

src/main/java/picard/illumina/BasecallsConverter.java
```diff
@@ -0,0 +1,150 @@
...
+        this.includeNonPfReads = includeNonPfReads;
+        this.bclQualityEvaluationStrategy = bclQualityEvaluationStrategy;
+        this.outputRecordClass = outputRecordClass;
+        this.factory = factory;
+
+
+        if (numProcessors == 0) {
+            this.numThreads = Runtime.getRuntime().availableProcessors();
+        } else if (numProcessors < 0) {
+            this.numThreads = Runtime.getRuntime().availableProcessors() + numProcessors;
```
pshapiro4broad: Do you want to wrap this in `Math.min(1, …);`, e.g. what if a user passes in -100?

src/main/java/picard/illumina/IlluminaBasecallsToFastq.java
```diff
@@ -270,18 +286,38 @@ private void initialize() {
...
-                sampleBarcodeFastqWriterMap, demultiplex, Math.max(1, MAX_READS_IN_RAM_PER_TILE / readsPerCluster), TMP_DIR, NUM_PROCESSORS,
-                FORCE_GC, FIRST_TILE, TILE_LIMIT, queryNameComparator,
-                new FastqRecordsForClusterCodec(readStructure.templates.length(),
-                        readStructure.sampleBarcodes.length(), readStructure.molecularBarcode.length()), FastqRecordsForCluster.class, bclQualityEvaluationStrategy,
-                this.APPLY_EAMSS_FILTER, INCLUDE_NON_PF_READS, IGNORE_UNEXPECTED_BARCODES);
+        if (USE_NEW_CONVERTER) {
+            if (METRICS_FILE != null) {
+                IOUtil.assertFileIsWritable(METRICS_FILE);
+            } else {
+                log.warn("No METRIC_FILE specified. Barcode metrics will not be generated.");
```
gbggrant: METRICS_FILE (note missing 'S')

src/main/java/picard/illumina/NewIlluminaBasecallsConverter.java
```diff
@@ -0,0 +1,393 @@
...
+     * @param includeNonPfReads        If true, will include ALL reads (including those which do not have PF set)
+     * @param ignoreUnexpectedBarcodes If true, will ignore reads whose called barcode is not found in barcodeRecordWriterMap,
+     * @param maxNoCalls               The maximum number of no calls to allow in a match.
+     * @param maxMismatches            The maximum number of mismatched calls to allow in a match.
+     * @param minMismatchDelta         The minimum mismatch difference between the best and second best matches.
+     * @param minimumBaseQuality       The minimum base quality to allow for a matching base call.
+     * @param metrics                  The metrics output for barcode metrics.
+     * @param metricsFile              The metrics file to write out the barcode metrics to.
+     */
+    public NewIlluminaBasecallsConverter(final File basecallsDir, File barcodesDir, final int lane,
```
gbggrant: 'NewIlluminaBasecallsConverter?' - thought this was flagged previously as an unimaginative name.

src/main/java/picard/illumina/NewIlluminaBasecallsConverter.java
```diff
@@ -0,0 +1,393 @@
...
+                                         final List<File> tmpDirs, final int numProcessors,
+                                         final Integer firstTile,
+                                         final Integer tileLimit,
+                                         final Comparator<CLUSTER_OUTPUT_RECORD> outputRecordComparator,
+                                         final SortingCollection.Codec<CLUSTER_OUTPUT_RECORD> codecPrototype,
+                                         final Class<CLUSTER_OUTPUT_RECORD> outputRecordClass,
+                                         final BclQualityEvaluationStrategy bclQualityEvaluationStrategy,
+                                         final boolean applyEamssFiltering, final boolean includeNonPfReads,
+                                         final boolean ignoreUnexpectedBarcodes, int maxNoCalls, int maxMismatches,
+                                         int minMismatchDelta, int minimumBaseQuality, MetricsFile<BarcodeMetric, Integer> metrics, File metricsFile) {
```
gbggrant: final, final, final

src/main/java/picard/illumina/NewIlluminaBasecallsConverter.java
```diff
@@ -0,0 +1,393 @@
...
+                    int endIndex = readStructure.sampleBarcodes.getDescriptorLengths()[i];
+                    bcStrings[i] = barcode.substring(pos, endIndex + pos);
+                    pos += endIndex;
+                }
+                this.barcodesMetrics.put(barcode, new BarcodeMetric(null, null, barcode, bcStrings));
+            }
+            barcodeWriterThreads.put(barcode, new ThreadPoolExecutorWithExceptions(1));
+        });
+
+        File laneDir = new File(basecallsDir, IlluminaFileUtil.longLaneStr(lane));
```
gbggrant: final?

src/main/java/picard/illumina/parser/IlluminaDataProviderFactory.java
```diff
@@ -202,17 +223,36 @@ public ReadStructure getOutputReadStructure() {
...
+     * @param filterFiles        A list of the pf filter files to use when creating this data provider.
+     * @param barcodesMetrics    A map of barcode to barcode metrics to output for this data provider.
+     * @param maxNoCalls         The maximum number of no calls to allow in a match.
+     * @param maxMismatches      The maximum number of mismatched calls to allow in a match.
+     * @param minMismatchDelta   The minimum mismatch difference between the best and second best matches.
+     * @param minimumBaseQuality The minimum base quality to allow for a matching base call.
      * @return An iterator for reading the Illumina basecall output for the lane specified in the ctor.
      */
-    public IlluminaDataProvider makeDataProvider() {
+    public NewIlluminaDataProvider makeDataProvider(List<File> cbcls,
```
gbggrant: Think these could all be finals

src/main/java/picard/illumina/parser/NewIlluminaDataProvider.java
gbggrant: 👍

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -180,126 +258,147 @@ public void close() {
...
-            totalRead += read;
+        try {
+            while ((read = gzipInputStream.read(uncompressedByteArray, totalRead, uncompressedByteArray.length - totalRead)) != -1) {
+                if (read == 0) break;
+                totalRead += read;
+            }
+        } catch (EOFException eofException) {
+            throw new PicardException("Unexpected end of file " + this.streamFiles[totalCycleCount].getAbsolutePath()
+                    + " this file is likely corrupt or truncated. We have read "
+                    + totalRead + "and were expecting to read "
```
gbggrant: " and"

src/main/java/picard/illumina/CheckIlluminaDirectory.java
```diff
@@ -113,30 +125,89 @@ protected int doWork() {
...
+                    Matcher tileMatcher = laneTileRegex.matcher(filterFile.getName());
+                    if (tileMatcher.matches()) {
+                        tiles.add(Integer.valueOf(tileMatcher.group(1)));
+                    }
+                }
+                IOUtil.assertFilesAreReadable(Arrays.asList(filterFiles));
+                tiles.sort(TILE_NUMBER_COMPARATOR);
+
+                //check s.locs
+                File locsFile = new File(BASECALLS_DIR.getParentFile(), "s.locs");
```
pshapiro4broad: Is there a benefit in doing that as a separate step? The file is being read from when the `LocsFileReader` constructor is called, so an IOException will be thrown.

src/main/java/picard/illumina/NewIlluminaBasecallsConverter.java
pshapiro4broad: Is there a reason why we use `final` for parameters and locals? I think `final` is great for fields (and its use on methods and classes has a different semantic meaning) but I'm not convinced it buys us much on parameters and locals. If methods are a reasonable size it shouldn't be hard to tell if the code is assigning to a parameter or field.

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -57,117 +63,189 @@
...
-                boolean pfExcluded = headerBuffer.get() == 1;
-                cycleData[i] = new CycleData(version, headerSize, bitsPerBasecall, bitsPerQualityScore, numberOfBins, qualityBins, numTiles, tileInfo, pfExcluded);
-                this.streams[i] = stream;
-                this.streamFiles[i] = bclFile;
-                byteBuffer.clear();
-                headerBuffer.clear();
+            int totalCycleCount = 0;
+
+            if (cycleData[totalCycleCount].tileInfo == null) {
+                throw new PicardException("Could not find tile " + cycleData[totalCycleCount].tileInfo);
```
kshakir: If I'm reading this if+exception message correctly, `"Could not find tile " + cycleData[totalCycleCount].tileInfo` will always translate into `"Could not find tile null"`. Is there an alternative set of variables for the exception message?

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -57,117 +63,189 @@
...
+                    }
+
+                    final InputStream stream = open(bclFile, false, false, false);
+                    int read = stream.read(byteBuffer.array());
 
-                short version = byteBuffer.getShort();
-                int headerSize = byteBuffer.getInt();
+                    //we need to read the first 6 bytes to determine the header size
+                    if (read != INITIAL_HEADER_SIZE) {
+                        close();
```
pshapiro4broad: There's a few places where you call `close()` before returning. It would be easier to use `try/finally` around the call to `readSurfaceTile()` to ensure that `close()` is called no matter how the operation fails.

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -57,117 +63,189 @@
...
+                    //cycleMap is 1 indexed
+                    final ByteBuffer byteBuffer = ByteBuffer.allocate(INITIAL_HEADER_SIZE);
+                    byteBuffer.order(ByteOrder.LITTLE_ENDIAN);
+
+                    final File bclFile = cycleMap.get(i + 1);
+                    if (bclFile == null) {
+                        throw new PicardException("Expected cbcl file for surface " + entry.getKey() + " cycle " + (i + 1) + " but it was not found.");
+                    }
+
+                    final InputStream stream = open(bclFile, false, false, false);
```
pshapiro4broad: This should have its own `try` block to ensure that it's closed on exception.

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -57,117 +63,189 @@
...
+        cachedTile = new byte[cycles][];
+        cachedTilePosition = new int[cycles];
+        for (int i = 1; i <= cycles; i++) {
+            allTiles.put(i, new ArrayList<>());
+        }
+        readSurfaceTile(tileNum, locs, headerOnly);
+        close();
+    }
 
+    private void readSurfaceTile(int tileNum, List<AbstractIlluminaPositionFileReader.PositionInfo> locs,
```
pshapiro4broad: One thing I noticed during yesterday's talk was that this is a very long method, which makes it hard to see what's going on here. It would be easier to understand if it could be split into three or so methods.

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -57,117 +63,189 @@
...
-                TileData[] tileInfo = new TileData[numTiles];
-                for (int j = 0; j < numTiles; j++) {
-                    int tileNum = headerBuffer.getInt();
-                    int numClustersInTile = headerBuffer.getInt();
-                    int uncompressedBlockSize = headerBuffer.getInt();
-                    int compressedBlockSize = headerBuffer.getInt();
-                    tileInfo[j] = new TileData(tileNum, numClustersInTile, uncompressedBlockSize, compressedBlockSize);
+                    boolean pfExcluded = headerBuffer.get() == 1;
+                    //try the next surface if we didn't find the tile
+                    if (tileInfo == null) continue;
```
pshapiro4broad: should have `{` `}` around body of `if`, here and on line 181 below.

src/main/java/picard/illumina/parser/readers/EmpiricalPhasingMetricsOutReader.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public IlluminaPhasingMetrics next() {
+        if (!hasNext()) {
+            throw new NoSuchElementException();
+        }
+        return new IlluminaPhasingMetrics(bbIterator.next());
+    }
+
+    public static class IlluminaPhasingMetrics {
```
pshapiro4broad: static class with final fields can have public final fields, doesn't need getters

src/main/java/picard/illumina/parser/readers/TileMetricsOutReader.java
```diff
@@ -27,30 +27,54 @@
...
+            case 2:
+                headerSize = VERSION_TWO_HEADER_SIZE;
+                recordSize = VERSION_TWO_RECORD_SIZE;
+                break;
+            case 3:
+                headerSize = VERSION_THREE_HEADER_SIZE;
+                recordSize = VERSION_THREE_RECORD_SIZE;
+                break;
+            default:
+                throw new PicardException("Unknown version " + version);
```
pshapiro4broad: Version should be an enum to avoid this issue. you could store the sizes there as well.

src/main/java/picard/illumina/parser/PerTileFileUtil.java
```diff
@@ -22,9 +22,9 @@ public PerTileFileUtil(final String extension, final File base,
         super(true, extension, base, faker, lane, skipEmptyFiles);
         this.fileMap = getTiledFiles(base, matchPattern);
         if (!fileMap.isEmpty()) {
-            this.tiles = Collections.unmodifiableList(new ArrayList<Integer>(this.fileMap.keySet()));
+            this.tiles = new ArrayList<>(this.fileMap.keySet());
```
pshapiro4broad: Setting a field in a base class is not a great way to do this initialization. If it can't be passed to the base class constructor, maybe it should be represented by an abstract getter in the base class, and only stored in the subclasses.

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -51,33 +54,76 @@
  * @author mccowan
  */
 public class TileMetricsUtil {
-    /** The path to the directory containing the tile metrics file relative to the basecalling directory. */
+
+    private static final Log log = Log.getInstance(TileMetricsUtil.class);
+
+    public static String CYCLE_TWENTY_FIVE_NAME = "C25.1";
```
pshapiro4broad: The name should reflect why it's important, e.g. TILE_METRICS_CYCLE_NAME. javadoc would be helpful too.

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -51,33 +54,76 @@
...
+        }
+    }
+
+    public static Collection<Tile> parseTileMetrics(final File tileMetricsOutFile,
+                                                    final Map<Integer, File> phasingMetricsFiles,
+                                                    final ReadStructure readStructure,
+                                                    final ValidationStringency validationStringency)
+            throws FileNotFoundException {
+        TileMetricsOutReader tileMetricsIterator = new TileMetricsOutReader
+                (tileMetricsOutFile, 3);
```
pshapiro4broad: odd line wrap

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -51,33 +54,76 @@
...
+        final Collection<IlluminaTileMetrics> tileMetrics = determineLastValueForLaneTileMetricsCode(tileMetricsIterator);
+        final Map<String, ? extends Collection<IlluminaTileMetrics>> locationToMetricsMap = partitionTileMetricsByLocation(tileMetrics);
+        final Collection<Tile> tiles = new LinkedList<>();
+        for (final Map.Entry<String, ? extends Collection<IlluminaTileMetrics>> entry : locationToMetricsMap.entrySet()) {
```
pshapiro4broad: Can you eliminate the ? here by making partitionTileMetricsByLocation() a generic function? What other types are expected here?

src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
```diff
@@ -92,6 +98,9 @@
     @Option(shortName = "EXT", doc="Append the given file extension to all metric file names (ex. OUTPUT.illumina_lane_metrics.EXT). None if null", optional=true)
     public String FILE_EXTENSION = null;
 
+    @Option(doc = "Boolean the determines if this run is a NovaSeq run or not. (NovaSeq tile metrics files are in cycle 25 directory.", optional = true)
+    public boolean IS_NOVASEQ = false;
```
pshapiro4broad: Is there no way to infer this from the files/directories present on disk? Either to generate this setting or assert that the setting matches the data.

src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
```diff
@@ -135,36 +146,48 @@ public static void main(final String[] args) {
...
+                                                                             final boolean isNovaSeq) {
             final Collection<Tile> tiles;
             try {
-                tiles = TileMetricsUtil.parseTileMetrics(TileMetricsUtil.renderTileMetricsFileFromBasecallingDirectory(illuminaRunDirectory),
-                        readStructure,
-                        validationStringency
-                        );
+                if (isNovaSeq) {
+                    tiles = TileMetricsUtil.parseTileMetrics(
+                            TileMetricsUtil.renderTileMetricsFileFromBasecallingDirectory(illuminaRunDirectory, true),
```
pshapiro4broad: It looks like this boolean argument is the same as isNovaSeq

src/main/java/picard/illumina/LanePhasingMetricsCollector.java
```diff
@@ -60,8 +59,13 @@ public LanePhasingMetricsCollector(final Collection<Tile> laneTiles) {
 
         // Calculate the medians for the collected data
         for (final TileTemplateRead tileTemplateRead : phasingValues.keySet()) {
-            medianPhasingMap.put(tileTemplateRead, medianPercentage(phasingValues.get(tileTemplateRead)));
-            medianPrePhasingMap.put(tileTemplateRead, medianPercentage(prePhasingValues.get(tileTemplateRead)));
+            if (usePercentage) {
+                medianPhasingMap.put(tileTemplateRead, medianPercentage(phasingValues.get(tileTemplateRead)));
```
pshapiro4broad: It might be clearer to make a new method that accepts `usePercentage` e.g. `median(Collection<Float> phaseValues, boolean usePercentage)`. That would avoid some duplicate code here too.

src/test/java/picard/illumina/IlluminaBasecallsToFastqTest.java
```diff
@@ -46,8 +49,10 @@
     private static final File TEST_DATA_DIR = new File("testdata/picard/illumina/25T8B25T/fastq");
     private static final File TEST_DATA_DIR_WITH_4M = new File("testdata/picard/illumina/25T8B25T/fastq_with_4M");
     private static final File TEST_DATA_DIR_WITH_4M4M = new File("testdata/picard/illumina/25T8B25T/fastq_with_4M4M");
+    private static final File TEST_DATA_DIR_WITH_CBCLS = new File("/Users/jcarey/workspace/data/seq/illumina/proc/SL-NVA/170322_A00113_0017_AH23GVDMXX/Data/Intensities/BaseCalls");
```
pshapiro4broad: This needs to be a relative path

src/test/java/picard/illumina/IlluminaBasecallsToFastqTest.java
```diff
@@ -132,18 +137,100 @@ public void testDualBarcodes() throws Exception {
         runStandardTest(1, "dualBarcode.", "barcode_double.params", 2, "25T8B8B25T", DUAL_BASECALLS_DIR, DUAL_TEST_DATA_DIR);
     }
 
+    @Test
+    public void testCbclConvert() throws Exception {
+        //     runNewConverterTest(1, "dualBarcode.", "barcode_double.params", 2, "151T8B8B151T", TEST_DATA_DIR_WITH_CBCLS, DUAL_CBCL_TEST_DATA_DIR);
```
pshapiro4broad: Test not complete

src/test/java/picard/illumina/IlluminaBasecallsToFastqTest.java
```diff
@@ -132,18 +137,100 @@ public void testDualBarcodes() throws Exception {
...
+        try {
+            outputDir.delete();
+            outputDir.mkdir();
+
+            outputDir.deleteOnExit();
+            // Create barcode.params with output files in the temp directory
+            final File libraryParams = new File(outputDir, libraryParamsFile);
+            libraryParams.deleteOnExit();
+            final List<File> outputPrefixes = new ArrayList<File>();
+            final LineReader reader = new BufferedLineReader(new FileInputStream(new File(testDataDir, libraryParamsFile)));
```
pshapiro4broad: use try-with-resources to avoid having to close() later on

src/test/java/picard/illumina/IlluminaBasecallsToSamTest.java
```diff
@@ -252,6 +254,54 @@ public void testDualBarcodes() throws Exception {
...
+        // Create barcode.params with output files in the temp directory
+        final File libraryParams = new File(outputDir, libraryParamsFile);
+        libraryParams.deleteOnExit();
+        final List<File> samFiles = new ArrayList<File>();
+        final LineReader reader = new BufferedLineReader(new FileInputStream(new File(testDataDir, libraryParamsFile)));
+        final PrintWriter writer = new PrintWriter(libraryParams);
+        final String header = reader.readLine();
+        writer.println(header + "\tOUTPUT");
+        while (true) {
```
pshapiro4broad: would be nice to avoid the code duplication in these tests

src/main/java/picard/illumina/parser/NewIlluminaDataProvider.java
jacarey: Agreed but this was a breaking change and it was asked that we don't move it.  We could duplicate the code to a top level class and deprecate this.. but I didn't want to tackle that in this PR.

src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
jacarey: typically we leave that up to the caller which may have more information that we have available. (the normal case is zamboni determining this based on machineType)

alecw:
Hi @jacarey ,

Thanks!  Any idea when a new release will happen that includes this change?  

-Alec

jacarey:
@alecw I just requested it, so I suspect with in a few hours.

Issue #272, opened 2015-08-28
Need to handle reverse-complemented indels in LiftoverVcf


eitanbanks:
Currently LiftoverVcf refuses to liftover indels or mixed records if they fall into a reverse complemented region in the target genome.  However, it is actually possible to liftover these records correctly with some changes to the code.
1. Check that the indel does not span over the boundary of a block in the chain file; if it does span then we should continue to fail to lift it over.
2. Multi-allelic records need to be split up into their component parts and lifted over (and output) as separate records.  This is necessary because indels are given a position which is the position before the event; in the original genome all of the indels in a multi-allelic record start at the same position but once the indels are reverse-complemented this no longer holds true.
3. Indels should be left-aligned if necessary based on the recommendations in the VCF specification. 

@tfenne said that he may one day take a shot at implementing this.

yfarjoun:
Comment regarding 3 above: a mixed record might have snps and indels which would also need to be split (not only multi-indels)

4. Multiple records might end up in the same genomic position and so we should merge them in that case:
10 AGG->AG final G added to inform of the reference)
11 G->C
12 G->A

After liftover 10-12 -> 10-12 (NEGATIVE) should become

10 CCT->CT, TGT 
11 C->G

yfarjoun:
Thinking about this more, we have a problem with genotypes....splitting/merging the variant is one thing, but if there are genotypes, these would need to be split up/merged, and with this come the splitting/merging of fields like PL, which would be quite horrendous (I think).

So, as a first pass, I'll take a crack at doing this for bi-allelic indels that will not require merging with neighboring variants.

yfarjoun:
fixed by #836

PR #836, opened 2017-06-15
Enable LiftOverVcf to liftover bi-allelic indels

yfarjoun:
### Description

- added liftover capability for bi-allelic indels (multiallelic indels, and mixed variants get lifted-over into more than one VC, so it's more complicated.)
- added a function for left-aligning (and normalizing) variants according to the algorithm laid out in:
"Adrian Tan, Gonçalo R. Abecasis and Hyun Min Kang. (2015) Unified Representation of Genetic Variants. Bioinformatics."
- added tests.


yfarjoun:
@eitanbanks would you like to cast a glance on this? @tfenne perhaps you'd like to take a look as well, as you expressed interest in solving the original issue but never got around to it :-)

I am happy to move the core functionality to htsjdk if that seems appropriate.

Fixes #272

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+                final Allele fixedAllele = alleleMap.getOrDefault(allele, allele);
                 fixedAlleles.add(fixedAllele);
             }
             fixedGenotypes.add(new GenotypeBuilder(genotype).alleles(fixedAlleles).make());
         }
         return fixedGenotypes;
     }
-}
+
+    protected static VariantContext leftAlignVariant(final VariantContext vc, final ReferenceSequence referenceSequence) {
```
magicDGS: This method could be useful in HTSJDK too!

yfarjoun: to avoid the long cycle of htsjdk review/release, I've opted to putting the function here...once this PR is accepted, I'm happy to move some of the code over to HTSJDK and remove the function from Picard...does that work? (it's protected here so this will not be a breaking change)

magicDGS: Sounds reasonable for me, the same as I did with the `CreateSequenceDictionary` improvements... Thanks!

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -201,15 +204,16 @@ public static void main(final String[] args) {
...
 
         for (final VariantContext ctx : in) {
             ++total;
             final Interval source = new Interval(ctx.getContig(), ctx.getStart(), ctx.getEnd(), false, ctx.getContig() + ":" + ctx.getStart() + "-" + ctx.getEnd());
             final Interval target = liftOver.liftOver(source, LIFTOVER_MIN_MATCH);
 
-            // if the target is null OR (the target is reverse complemented AND the variant is an indel or mixed), then we cannot lift it over
-            if (target == null || (target.isNegativeStrand() && (ctx.isMixed() || ctx.isIndel()))) {
+            // if the target is null OR (the target is reverse complemented AND the variant is a non-biallelic indel or mixed), then we cannot lift it over
+
```
eitanbanks: extra newline

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -201,15 +204,16 @@ public static void main(final String[] args) {
...
+            if (target == null || (target.isNegativeStrand() && (ctx.isMixed() || ctx.isIndel() && !ctx.isBiallelic()))) {
```
eitanbanks: could you please add parentheses around the final terms -- just for readability so it's clear where the ORs and ANDs apply?

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -218,22 +222,30 @@ public static void main(final String[] args) {
...
 
                 String missingContigMessage = "Encountered a contig, " + target.getContig() + " that is not part of the target reference.";
-                if(WARN_ON_MISSING_CONTIG) {
+                if (WARN_ON_MISSING_CONTIG) {
                     log.warn(missingContigMessage);
                 } else {
                     log.error(missingContigMessage);
                     return EXIT_CODE_WHEN_CONTIG_NOT_IN_REFERENCE;
                 }
+            } else if ( target.isNegativeStrand() && ctx.isIndel() && ctx.isBiallelic()) {
```
eitanbanks: extra space after the parenthesis

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
         return 0;
     }
 
-    protected static GenotypesContext fixGenotypes(final GenotypesContext originals, final Map<Allele, Allele> reverseComplementAlleleMap) {
+    protected static Locatable getFlippedLocation(final VariantContext source, final LiftOver liftOver) {
+        Interval originalLocus = new Interval(source.getContig(), source.getStart() + 1, source.getEnd() + 1);
```
eitanbanks: final

eitanbanks: could you please add a comment as to why you're adding +1 to start and end?  I think this method could use docs since it's not obvious what you're doing.

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+        return liftOver.liftOver(originalLocus);
+    }
+
+    /**
+     * @param source            original variant context
```
eitanbanks: would be good to explain that it's used in a context where the old and new seqs are on different strands

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+     * @param liftOver          the LiftOver object to use for flipping
+     * @param referenceSequence the reference sequence of the target
+     * @return a flipped variant-context.
+     */
+    protected static VariantContext flipIndel(final VariantContext source, final LiftOver liftOver, final ReferenceSequence referenceSequence) {
+        if (!source.isBiallelic()) return null;  //only supporting biallelic indels, for now.
```
eitanbanks: (forever!)

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+
+        final Locatable target = getFlippedLocation(source, liftOver);
+        if (target == null) return null;
+
+        final Map<Allele, Allele> reverseComplementAlleleMap = new HashMap<>(10);
+
+        reverseComplementAlleleMap.clear();
```
eitanbanks: not needed... it was just created in the previous line

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+        final List<Allele> alleles = new ArrayList<>();
+
+        for (final Allele oldAllele : source.getAlleles()) {
+            // target.getStart is 1-based, reference bases are 0-based
+            final String alleleWithoutHead = oldAllele.getBaseString().substring(1, oldAllele.length());
```
eitanbanks: We generally refer (or referred) to this base as the "leading reference base" if that's a better description

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+     * @param referenceSequence the reference sequence of the target
+     * @return a flipped variant-context.
+     */
+    protected static VariantContext flipIndel(final VariantContext source, final LiftOver liftOver, final ReferenceSequence referenceSequence) {
+        if (!source.isBiallelic()) return null;  //only supporting biallelic indels, for now.
+
+        final Locatable target = getFlippedLocation(source, liftOver);
+        if (target == null) return null;
+
+        final Map<Allele, Allele> reverseComplementAlleleMap = new HashMap<>(10);
```
eitanbanks: we know that there will only be 2 entries...

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+
+        reverseComplementAlleleMap.clear();
+        final List<Allele> alleles = new ArrayList<>();
+
+        for (final Allele oldAllele : source.getAlleles()) {
+            // target.getStart is 1-based, reference bases are 0-based
+            final String alleleWithoutHead = oldAllele.getBaseString().substring(1, oldAllele.length());
+            final char refBase = (char) referenceSequence.getBases()[target.getStart() - 1];
+            final Allele fixedAllele = Allele.create(refBase + SequenceUtil.reverseComplement(alleleWithoutHead), oldAllele.isReference());
```
eitanbanks: Question for you: is it worth confirming for deletions that the bases that were deleted in the source are the same ones being deleted in the target?  I.e. if bases "ACC" are deleted in the source then we'd expect their reverse complemented equivalents to be deleted in the target, or else we fail.

I'm not requiring this to be added as part of the review, but I would then want this described in the docs for the tool up top so that it's clear what it does and does not do.

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+            alleles.add(fixedAllele);
+            reverseComplementAlleleMap.put(oldAllele, fixedAllele);
+        }
+
+        VariantContextBuilder builder = new VariantContextBuilder(source.getSource(),
```
eitanbanks: final

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+
+    protected static VariantContext leftAlignVariant(final VariantContext vc, final ReferenceSequence referenceSequence) {
+        // from Adrian Tan, Gonçalo R. Abecasis and Hyun Min Kang. (2015) Unified Representation of Genetic Variants. Bioinformatics.
+
+        boolean changesInAlleles = true;
+
+        int start = vc.getStart();
+        int end = vc.getEnd();
+
+        Map<Allele, byte[]> alleleBasesMap = new HashMap<>();
```
eitanbanks: final

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+        // 1. while changes in alleles do
+        while (changesInAlleles) {
+
+            changesInAlleles = false;
+            // 2. if alleles end with the same nucleotide then
+            if (alleleBasesMap.values().stream()
+                    .collect(Collectors.groupingBy(a -> a[a.length - 1], Collectors.toSet()))
+                    .size() == 1) {
+                // 3. truncate rightmost nucleotide of each allele
+                for (Allele allele : alleleBasesMap.keySet()) {
```
eitanbanks: final Allele

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+                end--;
+                // 4. end if
+            }
+
+            // 5. if there exists an empty allele then
+            if (alleleBasesMap.values().stream()
+                    .map(a -> a.length)
+                    .anyMatch(l -> l == 0)) {
+                // 6. extend alleles 1 nucleotide to the left
+                for (Allele allele : alleleBasesMap.keySet()) {
```
eitanbanks: final Allele

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+                    // the first -1 for zero-base (getBases) versus 1-based (variant position)
+                    // another   -1 to get the base prior to the location of the start of the allele
+                    alleleBasesMap.put(allele, extendOneBaseLeft(alleleBasesMap.get(allele), referenceSequence.getBases()[start - 2]));
+                }
+                changesInAlleles = true;
+                start--;
+
```
eitanbanks: extra newline

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+                // 7. end if
+            }
+        }
+
+        // 8. while leftmost nucleotide of each allele are the same and all alleles have length 2 or more do
```
eitanbanks: 'are' -> 'is'

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+        while (alleleBasesMap.values().stream()
+                .allMatch(a -> a.length >= 2) &&
+
+                alleleBasesMap.values().stream()
+                        .collect(Collectors.groupingBy(a -> a[0], Collectors.toSet()))
+                        .size() == 1
+                ) {
+
+            //9. truncate the leftmost base of the alleles
+            for (Allele allele : alleleBasesMap.keySet()) {
```
eitanbanks: final Allele

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+                alleleBasesMap.put(allele, truncateBase(alleleBasesMap.get(allele), false));
+            }
+            start++;
+        }
+
+        VariantContextBuilder builder = new VariantContextBuilder(vc);
```
eitanbanks: final

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+        builder.start(start);
+        builder.stop(end);
+
+        Map<Allele, Allele> fixedAlleleMap = alleleBasesMap.entrySet().stream()
```
eitanbanks: final

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+                .collect(Collectors.toMap(Map.Entry::getKey, me -> Allele.create(me.getValue(), me.getKey().isReference())));
+        builder.alleles(fixedAlleleMap.values());
+        builder.genotypes(fixGenotypes(vc.getGenotypes(), fixedAlleleMap));
+
+        return builder.make();
+    }
+
+    private static byte[] truncateBase(byte[] allele, boolean truncateRightmost) {
```
eitanbanks: a couple of finals

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+        return Arrays.copyOfRange(allele, truncateRightmost ? 0 : 1, truncateRightmost ? allele.length - 1 : allele.length);
+    }
+
+    private static byte[] extendOneBaseLeft(final byte[] bases, final byte base) {
+
+        byte[] newBases = new byte[bases.length + 1];
```
eitanbanks: yet another final

src/test/java/picard/vcf/LiftoverVcfTest.java
```diff
@@ -159,12 +138,345 @@ public void testWriteOriginalPosition(boolean shouldWriteOriginalPosition) {
...
+        //   position of 13 ^ in result
+        // ...and becomes A->AACG at position 10
+
+        start = CHAIN_SIZE - 13;
+        stop = start;
+        builder.source("test4").stop(stop).start(start).alleles(CollectionUtil.makeList(RefA, ACGT));
+        result_builder.start(10).stop(10).alleles(CollectionUtil.makeList(RefA, AACG));
+        tests.add(new Object[]{builder.make(), reference, result_builder.make()});
+
+//      // outside of chain (due to shifting of anchor point)
```
eitanbanks: extra comment characters here and below

src/test/java/picard/vcf/LiftoverVcfTest.java
```diff
@@ -159,12 +138,345 @@ public void testWriteOriginalPosition(boolean shouldWriteOriginalPosition) {
...
+        return tests.iterator();
+    }
+
+    @Test(dataProvider = "leftAlignAllelesData")
+    public void testLeftAlignVariants(final VariantContext source, final ReferenceSequence reference, final VariantContext result) {
+
+        assertVcAreEqual(leftAlignVariant(source, reference), result);
+    }
+
+    private void assertVcAreEqual(final VariantContext actual, final VariantContext expected) {
```
eitanbanks: why can't you just use Assert.equals(actual, expected)?  That should work.  These tests are already in the VC class I believe.

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -313,21 +323,155 @@ public static void main(final String[] args) {
...
+                alleles);
+
+        builder.id(source.getID());
+        builder.attributes(source.getAttributes());
+
+        builder.genotypes(fixGenotypes(source.getGenotypes(), reverseComplementAlleleMap));
+        builder.filters(source.getFilters());
+        builder.log10PError(source.getLog10PError());
+
+        return leftAlignVariant(builder.make(), referenceSequence);
```
eitanbanks: Thinking about it... why do we need to left align only when going from positive to negative strand?  In theory we should always need to check for the opportunity for left alignment, right?

yfarjoun: If we assume that the input was left aligned, then lifting it over will also be left aligned (on non reverse-strand links)...if your input is bad, then I'm not sure that it's lift-over's job to fix it...

src/test/java/picard/vcf/LiftoverVcfTest.java
yfarjoun: I don't see an implementation of .equals(other) in VariantContext)

src/main/java/picard/vcf/LiftoverVcf.java
eitanbanks: But what if the reference has changed and now the indel can be left aligned even more on the new one?  I guess this would be rare enough that we can ignore it.

src/test/java/picard/vcf/LiftoverVcfTest.java
eitanbanks: How is that possible?  Must be in the GATK code then.
In any case you could definitely make use of VariantContext.hasSameAllelesAs() here.
Or you can be really hacky and just do actual.toString().equals(expected.toString())!

yfarjoun: I'll use hasSameAllelesAs(). thanks.

src/main/java/picard/vcf/LiftoverVcf.java
yfarjoun: I could add a left-align to all indels.

eitanbanks: Up to you.  You can also argue that it's outside the scope of this PR.

yfarjoun:
back to you @eitanbanks. I made a little more changes than you requested, but I think it's for the better...

src/main/java/picard/vcf/LiftoverVcf.java
yfarjoun: done.

eitanbanks:
Looks good.

How annoying would it be to add a few more lines to the test VCFs?  Or do you think you've covered everything with the DataProviders?

yfarjoun:
it seems that not all the failure modes are properly covered...so perhaps I'll add another data-provider that provides various cases that fail liftover....

yfarjoun:
@eitanbanks good enough?

eitanbanks:
Sure.

yfarjoun:
bitmoji

PR #840, opened 2017-06-16
DownsampleSam can optionally yield QualityYieldMetrics.

nh13:
In many cases, I'd like to know the exact # of reads after downsampling, while I am piping the output to other tools that will consume the reads.  It would be nice to have `DownsampleSam` output some metrics, so I repurposed `QualityYieldMetrics` to do so.  @yfarjoun or @tfenne thoughts?

### DIFF
--- a/src/main/java/picard/analysis/CollectQualityYieldMetrics.java
+++ b/src/main/java/picard/analysis/CollectQualityYieldMetrics.java
@@ -50,6 +50,8 @@ import java.io.File;
         programGroup = Metrics.class
 )
 public class CollectQualityYieldMetrics extends SinglePassSamProgram {
+    private QualityYieldMetricsCollector collector = null;
+
     static final String USAGE_SUMMARY = "Collect metrics about reads that pass quality thresholds and Illumina-specific filters.  ";
     static final String USAGE_DETAILS = "This tool evaluates the overall quality of reads within a bam file containing one read group. " +
             "The output indicates the total numbers of bases within a read group that pass a minimum base quality score threshold and " +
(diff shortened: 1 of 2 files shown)


src/main/java/picard/sam/DownsampleSam.java
```diff
@@ -115,10 +133,12 @@ protected int doWork() {
         final SAMFileWriter out = new SAMFileWriterFactory().makeSAMOrBAMWriter(in.getFileHeader(), true, OUTPUT);
         final ProgressLogger progress = new ProgressLogger(log, (int) 1e7, "Wrote");
         final DownsamplingIterator iterator = DownsamplingIteratorFactory.make(in, STRATEGY, PROBABILITY, ACCURACY, RANDOM_SEED);
+        final PublicCollectQualityYieldMetrics metricsCollector = new PublicCollectQualityYieldMetrics();
 
         while (iterator.hasNext()) {
             final SAMRecord rec = iterator.next();
             out.addAlignment(rec);
+            metricsCollector.acceptRead(rec, null);
```
yfarjoun: this should only be done if METRIC_FILE !=null since it needs a more thorough parsing of the bam-file than downsampling...

nh13:
@yfarjoun thanks for the review, do you want to take another look? I had to refactor `CollectQualityYieldMetrics` into a metrics collector since I can't use an instance of `CollectQualityYieldMetrics` in another command line program.

yfarjoun:
that's a bummer... This is now changing core programs....

@ktibbett @jacarey can someone take a look here?

src/main/java/picard/analysis/CollectQualityYieldMetrics.java
```diff
@@ -50,6 +50,8 @@
         programGroup = Metrics.class
 )
 public class CollectQualityYieldMetrics extends SinglePassSamProgram {
+    private QualityYieldMetricsCollector collector = new QualityYieldMetricsCollector();
```
yfarjoun: This is confusing...why not set it to null here, keep the argument defaults as they were and in setup call the right constructor?

src/main/java/picard/analysis/CollectQualityYieldMetrics.java
```diff
@@ -77,74 +79,112 @@
...
-        metrics.TOTAL_READS++;
-        metrics.TOTAL_BASES += length;
+        // The metrics to be accumulated
+        private final QualityYieldMetrics metrics = new QualityYieldMetrics();
 
-        final boolean isPfRead = !rec.getReadFailsVendorQualityCheckFlag();
-        if (isPfRead) {
-            metrics.PF_READS++;
-            metrics.PF_BASES += length;
+        public QualityYieldMetricsCollector() {
```
yfarjoun: do you really need a default constructor?

nh13:
@yfarjoun it was a copy-paste, so the danger is minimal.

src/main/java/picard/analysis/CollectQualityYieldMetrics.java
```diff
@@ -77,74 +79,112 @@
...
-    @Override
-    protected void finish() {
-        final MetricsFile<QualityYieldMetrics, Integer> metricsFile = getMetricsFile();
-        metrics.READ_LENGTH = metrics.TOTAL_READS == 0 ? 0 : (int) (metrics.TOTAL_BASES / metrics.TOTAL_READS);
-        metrics.Q20_EQUIVALENT_YIELD = metrics.Q20_EQUIVALENT_YIELD / 20;
-        metrics.PF_Q20_EQUIVALENT_YIELD = metrics.PF_Q20_EQUIVALENT_YIELD / 20;
+        public void finish() {
+            metrics.READ_LENGTH             = metrics.TOTAL_READS == 0 ? 0 : (int) (metrics.TOTAL_BASES / metrics.TOTAL_READS);
+            metrics.Q20_EQUIVALENT_YIELD    = metrics.Q20_EQUIVALENT_YIELD / 20;
+            metrics.PF_Q20_EQUIVALENT_YIELD = metrics.PF_Q20_EQUIVALENT_YIELD / 20;
```
yfarjoun: I know this is not your doing...but could you correct the javadoc on "PF_Q20_EQUIVALENT_YIELD" (it's missing any mention of PF).

yfarjoun: (what does this metric even mean?!!?!)

src/main/java/picard/analysis/CollectQualityYieldMetrics.java
nh13: I would like the defaults to be stored in one place, so that if you call `CollectQualityYieldMetrics` with no command line options specified, or instantiate a `QualityYieldMetricsCollector`, you get the same defaults.  The other option is to have `static final` variables in `QualityYieldMetricsCollector` that can be referenced by `CollectQualityYieldMetrics`.  Does this make sense what I am trying to achieve?

src/main/java/picard/analysis/CollectQualityYieldMetrics.java
nh13: See my comment above about storing the defaults in one place.   Again, I can make the defaults their own variables that both `CollectQualityYieldMetrics ` and `QualityYieldMetricsCollector` reference.

src/main/java/picard/analysis/CollectQualityYieldMetrics.java
nh13: I'll fix the comment.  My guess is it's trying to estimate effectively how many bases have Q20.  If all the bases are Q40, then you really have effectively twice the number of Q20 bases?  I don't want to remove it since folks may be relying up on it.

nh13:
@yfarjoun I made all your suggested changes, including not having a default constructor (for `QualityYieldMetricsCollector`), since I would like to get this merged ASAP :)

yfarjoun:
@nh13 I understand you are rushed, but My initial comments were just that...comments. 

I think that this change will have to be reviewed by someone else as well. While I am pretty sure that there is no change to the results, I want someone to make sure that this is the pattern we want to use for splitting a CLP from the worker-class. 

Also, I want someone else to check that this is a reasonable way to get what you want...namely the number of reads processed...
Doesn't it make more sense, for your purpose, to emit a small metrics class that has the number of reads emitted? Why go through the trouble of parsing the qual-string when you just want to count the number of reads? Also, I suspect that QualityYieldMetrics will soon have 3 rows (BOTH, FIRST, SECOND) to deal with some sequencer issues on Read2...by tying yourself to this metric, you will have to absorb changes to it down the line.

nh13:
@yfarjoun

(1) I am running `CollectQualityYieldMetrics` after `DownsampleSam` currently, so I wanted the _same_ metric out.  It is useful not to create yet another metric if an existing one suffices. 

(2) I looked at a few collectors already implemented in Picard, and modeled the current one based on them, with some simplifications.  

(3) Regarding having three rows on `QualityYieldMetrics` that's fine too, since if you are thinking of changing that metric (multi-level too perhaps?), then I am going to have to deal with it elsewhere anyhow.  I can always stay on an earlier Picard version if I need to.

@ktibbett or @jacarey can you take a quick peek at this one?

src/main/java/picard/analysis/CollectQualityYieldMetrics.java
```diff
@@ -87,64 +89,98 @@
...
+                quals = tmp;
+            } else {
+                quals = rec.getBaseQualities();
+            }
+
+            // add up quals, and quals >= 20
+            for (final int qual : quals) {
+                metrics.Q20_EQUIVALENT_YIELD += qual;
+                if (qual >= 20) metrics.Q20_BASES++;
+                if (qual >= 30) metrics.Q30_BASES++;
```
jacarey:
```
if(qual >= 30) {
    metrics.Q20_BASES++;
    metrics.Q30_BASES++;
} else if (qual >=20) {
    metrics.Q20_BASES++;
}
```

slightly faster for qualities above 30 not sure if it makes it less readable though

src/main/java/picard/sam/DownsampleSam.java
```diff
@@ -115,10 +122,12 @@ protected int doWork() {
         final SAMFileWriter out = new SAMFileWriterFactory().makeSAMOrBAMWriter(in.getFileHeader(), true, OUTPUT);
         final ProgressLogger progress = new ProgressLogger(log, (int) 1e7, "Wrote");
         final DownsamplingIterator iterator = DownsamplingIteratorFactory.make(in, STRATEGY, PROBABILITY, ACCURACY, RANDOM_SEED);
+        final QualityYieldMetricsCollector metricsCollector = new QualityYieldMetricsCollector(true, false, false);
 
         while (iterator.hasNext()) {
             final SAMRecord rec = iterator.next();
             out.addAlignment(rec);
+            if (METRICS_FILE != null) metricsCollector.acceptRecord(rec, null);
```
jacarey: if check in loop is not great... but I can live with it.

nh13: Agreed, making it of type `Optional<QualityYieldMetricsCollector>` is syntactic sugar :/

nh13:
Thanks @jacarey, merging when tests pass.

PR #843, opened 2017-06-19
DSDEGP-454 Enable the IntelDeflater by default in all invocations of Picard

danxmoran:
### Description

This mirrors what was done for GATK in broadinstitute/gsa-unstable#1572.

Comwell workflow ID in gotc-dev: 7012caa5-ea19-4f99-82f3-a1f2ca200ad6

----

#### Content

#### Review

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,3 @@ report
 jacoco.data
 .gradle
 build
-*.bai
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ report
 jacoco.data
 .gradle
 build
+*.bai
--- a/build.gradle
+++ b/build.gradle
@@ -50,12 +50,21 @@ jacoco {
 final htsjdkVersion = System.getProperty('htsjdk.version', '2.10.1')
 
 dependencies {
-    compile 'com.intel.gkl:gkl:0.5.2'
+    compile('com.intel.gkl:gkl:0.5.2') {
+        exclude module: 'htsjdk'
+    }
     compile 'com.google.guava:guava:15.0'
     compile 'com.github.samtools:htsjdk:' + htsjdkVersion
     //tools dependency for doclet requires sdk devel
     compile(files(((URLClassLoader) ToolProvider.getSystemToolClassLoader()).getURLs()))
     testCompile 'org.testng:testng:6.9.10'
+}
+
+configurations.all {
+    resolutionStrategy {
+        // force the htsjdk version so we don't get a different one transitively
+        force 'com.github.samtools:htsjdk:' + htsjdkVersion
+    }
 }
 
 sourceCompatibility = 1.8
--- a/build.gradle
+++ b/build.gradle
@@ -50,6 +50,7 @@ jacoco {
 final htsjdkVersion = System.getProperty('htsjdk.version', '2.10.1')
 
 dependencies {
+    compile 'com.intel.gkl:gkl:0.5.2'
     compile 'com.google.guava:guava:15.0'
     compile 'com.github.samtools:htsjdk:' + htsjdkVersion
     //tools dependency for doclet requires sdk devel
--- a/build.gradle
+++ b/build.gradle
@@ -58,6 +58,7 @@ dependencies {
     //tools dependency for doclet requires sdk devel
     compile(files(((URLClassLoader) ToolProvider.getSystemToolClassLoader()).getURLs()))
     testCompile 'org.testng:testng:6.9.10'
+    testCompile 'org.apache.commons:commons-lang3:3.6'
 }
 
 configurations.all {
--- a/src/main/java/picard/cmdline/CommandLineProgram.java
+++ b/src/main/java/picard/cmdline/CommandLineProgram.java
@@ -34,7 +34,10 @@ import htsjdk.samtools.metrics.Header;
 import htsjdk.samtools.metrics.MetricBase;
 import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.metrics.StringHeader;
-import htsjdk.samtools.util.*;
+import htsjdk.samtools.util.BlockCompressedOutputStream;
+import htsjdk.samtools.util.BlockGunzipper;
+import htsjdk.samtools.util.IOUtil;
+import htsjdk.samtools.util.Log;
 import htsjdk.variant.variantcontext.writer.Options;
 import htsjdk.variant.variantcontext.writer.VariantContextWriterBuilder;
 
@@ -85,7 +88,8 @@ public abstract class CommandLineProgram {
     public ValidationStringency VALIDATION_STRINGENCY = ValidationStringency.DEFAULT_STRINGENCY;
 
     @Option(doc = "Compression level for all compressed files created (e.g. BAM and GELI).", common=true)
-    public int COMPRESSION_LEVEL = 1; // The default IntelDeflater is slow at levels > 1
+    // The default IntelDeflater is slow at levels > 1
+    public int COMPRESSION_LEVEL = 1;
 
     @Option(doc = "When writing SAM files that need to be sorted, this will specify the number of records stored in RAM before spilling to disk. Increasing this number reduces the number of file handles needed to sort a SAM file, and increases the amount of RAM needed.", optional=true, common=true)
     public Integer MAX_RECORDS_IN_RAM = SAMFileWriterImpl.getDefaultMaxRecordsInRam();
@@ -102,10 +106,10 @@ public abstract class CommandLineProgram {
     @Option(doc="Google Genomics API client_secrets.json file path.", common = true)
     public String GA4GH_CLIENT_SECRETS="client_secrets.json";
 
-    @Option(shortName = "jdk_deflater", doc = "Use the JDK Deflater instead of the IntelDeflater for writing BAMs", common = true)
+    @Option(shortName = "use_jdk_deflater", doc = "Use the JDK Deflater instead of the IntelDeflater for writing BAMs", common = true)
     public Boolean USE_JDK_DEFLATER = false;
 
-    @Option(shortName = "jdk_inflater", doc = "Use the JDK Inflater instead of the IntelInflater for reading BAMs", common = true)
+    @Option(shortName = "use_jdk_inflater", doc = "Use the JDK Inflater instead of the IntelInflater for reading BAMs", common = true)
     public Boolean USE_JDK_INFLATER = false;
     
     private final String standardUsagePreamble = CommandLineParser.getStandardUsagePreamble(getClass());
@@ -209,7 +213,7 @@ public abstract class CommandLineProgram {
                     new Date(), System.getProperty("user.name"), InetAddress.getLocalHost().getHostName(),
                     System.getProperty("os.name"), System.getProperty("os.version"), System.getProperty("os.arch"),
                     System.getProperty("java.vm.name"), System.getProperty("java.runtime.version"),
-                    USE_JDK_DEFLATER ? "JdkDeflater" : "IntelDeflater", USE_JDK_INFLATER ? "JdkInflater" : "IntelInflater",
+                    USE_JDK_DEFLATER ? "Jdk" : "Intel", USE_JDK_INFLATER ? "Jdk" : "Intel",
                     commandLineParser.getVersion());
                 System.err.println(msg);
             }
--- a/src/main/java/picard/cmdline/CommandLineProgram.java
+++ b/src/main/java/picard/cmdline/CommandLineProgram.java
@@ -85,7 +85,7 @@ public abstract class CommandLineProgram {
     public ValidationStringency VALIDATION_STRINGENCY = ValidationStringency.DEFAULT_STRINGENCY;
 
     @Option(doc = "Compression level for all compressed files created (e.g. BAM and GELI).", common=true)
-    public int COMPRESSION_LEVEL = BlockCompressedStreamConstants.DEFAULT_COMPRESSION_LEVEL;
+    public int COMPRESSION_LEVEL = 1; // The default IntelDeflater is slow at levels > 1
 
     @Option(doc = "When writing SAM files that need to be sorted, this will specify the number of records stored in RAM before spilling to disk. Increasing this number reduces the number of file handles needed to sort a SAM file, and increases the amount of RAM needed.", optional=true, common=true)
     public Integer MAX_RECORDS_IN_RAM = SAMFileWriterImpl.getDefaultMaxRecordsInRam();
--- a/src/main/java/picard/cmdline/CommandLineProgram.java
+++ b/src/main/java/picard/cmdline/CommandLineProgram.java
@@ -88,7 +88,8 @@ public abstract class CommandLineProgram {
     public ValidationStringency VALIDATION_STRINGENCY = ValidationStringency.DEFAULT_STRINGENCY;
 
     @Option(doc = "Compression level for all compressed files created (e.g. BAM and GELI).", common=true)
-    // The default IntelDeflater is slow at levels > 1
+    // Reading from a file compressed by the default IntelDeflater at level 1 is ~3x faster than reading from a file
+    // compressed at level 5 (the htsjdk default), and level 1 files are only slightly larger.
     public int COMPRESSION_LEVEL = 1;
 
     @Option(doc = "When writing SAM files that need to be sorted, this will specify the number of records stored in RAM before spilling to disk. Increasing this number reduces the number of file handles needed to sort a SAM file, and increases the amount of RAM needed.", optional=true, common=true)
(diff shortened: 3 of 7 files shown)


.gitignore
```diff
@@ -15,3 +15,4 @@ report
 jacoco.data
 .gradle
 build
+*.bai
```
pshapiro4broad: Why is this necessary? Did a test create a .bai file and not delete it?

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -102,6 +101,12 @@
 
     @Option(doc="Google Genomics API client_secrets.json file path.", common = true)
     public String GA4GH_CLIENT_SECRETS="client_secrets.json";
+
+    @Option(shortName = "jdk_deflater", doc = "Use the JDK Deflater instead of the IntelDeflater for writing BAMs", common = true)
```
pshapiro4broad: I think it would be better to use the same option name as GATK here, e.g. `use_jdk_deflater`, `use_jdk_inflater`.

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -190,12 +204,14 @@ public int instanceMain(final String[] argv) {
...
-                                       " on " + System.getProperty("os.name") + " " + System.getProperty("os.version") +
-                                       " " + System.getProperty("os.arch") + "; " + System.getProperty("java.vm.name") +
-                                       " " + System.getProperty("java.runtime.version") +
-                                       "; Picard version: " + commandLineParser.getVersion());
+                final String msg = String.format(
+                    "[%s] Executing as %s@%s on %s %s %s; %s %s; Deflater: %s; Inflater: %s; Picard version: %s",
+                    new Date(), System.getProperty("user.name"), InetAddress.getLocalHost().getHostName(),
+                    System.getProperty("os.name"), System.getProperty("os.version"), System.getProperty("os.arch"),
+                    System.getProperty("java.vm.name"), System.getProperty("java.runtime.version"),
+                    USE_JDK_DEFLATER ? "JdkDeflater" : "IntelDeflater", USE_JDK_INFLATER ? "JdkInflater" : "IntelInflater",
```
pshapiro4broad: If the prefix text is `Deflater` then I think the value text could be just `Intel` or `JDK`.

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -190,12 +204,14 @@ public int instanceMain(final String[] argv) {
 
             // Output a one liner about who/where and what software/os we're running on
             try {
-            System.err.println("[" + new Date() + "] Executing as " +
-                                       System.getProperty("user.name") + "@" + InetAddress.getLocalHost().getHostName() +
-                                       " on " + System.getProperty("os.name") + " " + System.getProperty("os.version") +
-                                       " " + System.getProperty("os.arch") + "; " + System.getProperty("java.vm.name") +
-                                       " " + System.getProperty("java.runtime.version") +
-                                       "; Picard version: " + commandLineParser.getVersion());
+                final String msg = String.format(
```
pshapiro4broad: Not sure why you changed this to use a variable and `String.format()` instead of using `+`? Do you find this form easier to read?

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -32,10 +34,7 @@
 import htsjdk.samtools.metrics.MetricBase;
 import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.metrics.StringHeader;
-import htsjdk.samtools.util.BlockCompressedOutputStream;
-import htsjdk.samtools.util.BlockCompressedStreamConstants;
-import htsjdk.samtools.util.IOUtil;
-import htsjdk.samtools.util.Log;
+import htsjdk.samtools.util.*;
```
pshapiro4broad: I don't know if we have a policy about this but I think we want to avoid `*` imports in our java code. There is an IDE setting to control this.

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -86,7 +85,7 @@
     public ValidationStringency VALIDATION_STRINGENCY = ValidationStringency.DEFAULT_STRINGENCY;
 
     @Option(doc = "Compression level for all compressed files created (e.g. BAM and GELI).", common=true)
-    public int COMPRESSION_LEVEL = BlockCompressedStreamConstants.DEFAULT_COMPRESSION_LEVEL;
+    public int COMPRESSION_LEVEL = 1; // The default IntelDeflater is slow at levels > 1
```
pshapiro4broad: IMO end of line comments make code harder to read and edit. In most cases, like this, it can just as easily be a comment on the line before, instead of at the end of the line.

.gitignore
danxmoran: Yeah, a couple popped out when I ran tests. Should I move this to my global ignore file instead?

src/main/java/picard/cmdline/CommandLineProgram.java
danxmoran: I do find this one easier to read. It's a bit harder to figure out which values plug into which slots in the log, but it's easier to see the general structure of the message when it's all on one line at the top.

droazen:
@erniebrau and/or @gspowley could one of you have an initial look at this?

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -86,7 +88,8 @@
     public ValidationStringency VALIDATION_STRINGENCY = ValidationStringency.DEFAULT_STRINGENCY;
 
     @Option(doc = "Compression level for all compressed files created (e.g. BAM and GELI).", common=true)
-    public int COMPRESSION_LEVEL = BlockCompressedStreamConstants.DEFAULT_COMPRESSION_LEVEL;
+    // The default IntelDeflater is slow at levels > 1
+    public int COMPRESSION_LEVEL = 1;
```
droazen: This comment is not accurate -- the Intel Deflater is faster than the JDK deflater by 15-45% on all compression levels:

image

The reason why we want to compress at level 1 is because reading from a level 1 compressed file is ~3x faster than reading from a level 5 compressed file (the former default), and only slightly larger:

image

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -102,6 +105,12 @@
 
     @Option(doc="Google Genomics API client_secrets.json file path.", common = true)
     public String GA4GH_CLIENT_SECRETS="client_secrets.json";
+
+    @Option(shortName = "use_jdk_deflater", doc = "Use the JDK Deflater instead of the IntelDeflater for writing BAMs", common = true)
```
droazen: This controls more than just writing BAMs -- it controls anything written via `BlockCompressedOutputStream` (applies to the second argument below as well)

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -190,12 +208,14 @@ public int instanceMain(final String[] argv) {
...
-                                       " on " + System.getProperty("os.name") + " " + System.getProperty("os.version") +
-                                       " " + System.getProperty("os.arch") + "; " + System.getProperty("java.vm.name") +
-                                       " " + System.getProperty("java.runtime.version") +
-                                       "; Picard version: " + commandLineParser.getVersion());
+                final String msg = String.format(
+                    "[%s] Executing as %s@%s on %s %s %s; %s %s; Deflater: %s; Inflater: %s; Picard version: %s",
+                    new Date(), System.getProperty("user.name"), InetAddress.getLocalHost().getHostName(),
+                    System.getProperty("os.name"), System.getProperty("os.version"), System.getProperty("os.arch"),
+                    System.getProperty("java.vm.name"), System.getProperty("java.runtime.version"),
+                    USE_JDK_DEFLATER ? "Jdk" : "Intel", USE_JDK_INFLATER ? "Jdk" : "Intel",
```
droazen: You need to test whether the Intel Deflater/Inflater were actually successfully loaded, and use that result to inform the user which deflater/inflater will be used. What you want to do here is:

```
        final boolean usingIntelDeflater = (BlockCompressedOutputStream.getDefaultDeflaterFactory() instanceof IntelDeflaterFactory && ((IntelDeflaterFactory)BlockCompressedOutputStream.getDefaultDeflaterFactory()).usingIntelDeflater());
        final boolean usingIntelInflater = (BlockGunzipper.getDefaultInflaterFactory() instanceof IntelInflaterFactory && ((IntelInflaterFactory)BlockGunzipper.getDefaultInflaterFactory()).usingIntelInflater());
```

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -190,12 +208,14 @@ public int instanceMain(final String[] argv) {
...
+                    commandLineParser.getVersion());
+                System.err.println(msg);
             }
             catch (Exception e) { /* Unpossible! */ }
         }
```
droazen: This branch needs a unit test. Recommend adding a test proving that the Intel Inflater and Deflater can be loaded successfully.

src/main/java/picard/cmdline/CommandLineProgram.java
droazen: Collapsing imports after a set threshold is standard practice.

build.gradle
```diff
@@ -50,6 +50,7 @@ jacoco {
 final htsjdkVersion = System.getProperty('htsjdk.version', '2.10.1')
 
 dependencies {
+    compile 'com.intel.gkl:gkl:0.5.2'
```
danxmoran: screen shot 2017-07-05 at 9 43 04 am
Confirmed! Is this something that would be good to assert on in our jar-building Gradle tasks?

src/test/java/picard/IntelInflaterDeflaterLoadTest.java
```diff
@@ -0,0 +1,37 @@
+package picard;
```
danxmoran: Here's a test for the successful loading of the Intel inflater / deflater, modeled off of the [GATK tests](https://github.com/broadinstitute/gatk/blob/master/src/test/java/org/broadinstitute/hellbender/IntelInflaterDeflaterIntegrationTest.java).

src/test/java/picard/analysis/CollectMultipleMetricsTest.java
```diff
@@ -335,8 +335,10 @@ void setupBuilder() throws IOException {
         final String flowCellBarcode = "TESTBARCODE";
 
         tempSamFile = File.createTempFile("CollectGcBias", ".bam", TEST_DIR);
+        final File tempSamIndex = new File(tempSamFile.getAbsolutePath().replace("bam", "bai"));
```
danxmoran: This test would successfully clean up after itself when run alone with `-Dtest.single=CollectMultipleMetricsTest` but fail to clean up when run with all tests in the `analysis` package (or all the tests in Picard). Explicitly marking the index file for deletion fixes the problem, but I'm not sure what the root issue is.

src/test/java/picard/analysis/CollectWgsMetricsTest.java
```diff
@@ -136,8 +136,10 @@ void setupBuilder() throws IOException {
 
         //Create Sam Files
         tempSamFile = File.createTempFile("CollectWgsMetrics", ".bam", TEST_DIR);
+        final File tempSamIndex = new File(tempSamFile.getAbsolutePath().replace("bam", "bai"));
```
danxmoran: Same as in `CollectMultipleMetricsTest`.

src/test/java/picard/sam/FilterSamReadsTest.java
```diff
@@ -95,7 +95,7 @@ public void testPairedIntervalFilter(final String intervalFilename, final int ex
         // Build a sam file for testing
         final File inputSam = File.createTempFile("testSam", ".sam", TEST_DIR);
         inputSam.deleteOnExit();
-        final File sortedSamIdx = new File(TEST_DIR, inputSam.getName() + ".idx");
+        final File sortedSamIdx = new File(TEST_DIR, inputSam.getName() + ".bai");
```
danxmoran: This test reliably failed to clean up after itself because of the wrong extension here.

src/main/java/picard/cmdline/CommandLineProgram.java
danxmoran: Added below.

danxmoran:
Ok @droazen, comments addressed!

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -102,6 +106,12 @@
 
     @Option(doc="Google Genomics API client_secrets.json file path.", common = true)
     public String GA4GH_CLIENT_SECRETS="client_secrets.json";
+
+    @Option(shortName = "use_jdk_deflater", doc = "Use the JDK Deflater instead of the Intel Deflater for writing with BlockCompressedOutputStreams", common = true)
+    public Boolean USE_JDK_DEFLATER = false;
+
+    @Option(shortName = "use_jdk_inflater", doc = "Use the JDK Inflater instead of the Intel Inflater for reading with BlockGunzippers", common = true)
+    public Boolean USE_JDK_INFLATER = false;
```
droazen: These argument docs should be targeted at end users rather than developers, and should not mention class names. Eg.,

```
"Use the JDK Deflater instead of the Intel Deflater for writing compressed output"
"Use the JDK Inflater instead of the Intel Inflater for reading compressed input"
```

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -190,12 +209,18 @@ public int instanceMain(final String[] argv) {
...
+                final boolean usingIntelDeflater = (BlockCompressedOutputStream.getDefaultDeflaterFactory() instanceof IntelDeflaterFactory &&
+                        ((IntelDeflaterFactory)BlockCompressedOutputStream.getDefaultDeflaterFactory()).usingIntelDeflater());
+                final boolean usingIntelInflater = (BlockGunzipper.getDefaultInflaterFactory() instanceof IntelInflaterFactory &&
+                        ((IntelInflaterFactory)BlockGunzipper.getDefaultInflaterFactory()).usingIntelInflater());
+                final String msg = String.format(
+                    "[%s] Executing as %s@%s on %s %s %s; %s %s; Deflater: %s; Inflater: %s; Picard version: %s",
+                    new Date(), System.getProperty("user.name"), InetAddress.getLocalHost().getHostName(),
+                    System.getProperty("os.name"), System.getProperty("os.version"), System.getProperty("os.arch"),
+                    System.getProperty("java.vm.name"), System.getProperty("java.runtime.version"),
+                    usingIntelDeflater ? "Jdk" : "Intel", usingIntelInflater ? "Jdk" : "Intel",
```
droazen: These expressions giving the deflater/inflater being used are backwards. Should be:

```
usingIntelDeflater ? "Intel" : "Jdk"
```

and

```
usingIntelInflater ? "Intel" : "Jdk"
```

src/test/java/picard/sam/FilterSamReadsTest.java
droazen: A sam file can't have a `.bai` index though?

src/main/java/picard/cmdline/CommandLineProgram.java
danxmoran: D'oh

src/test/java/picard/sam/FilterSamReadsTest.java
danxmoran: I can't comment on [the line](https://github.com/broadinstitute/picard/pull/843/files#diff-9c09d6174fd9f608424a44e8e09ac593R102) directly, but could it be because the test is using `makeBAMWriter` instead of `makeSAMWriter` below?

src/test/java/picard/sam/FilterSamReadsTest.java
```diff
@@ -99,7 +99,7 @@ public void testPairedIntervalFilter(final String intervalFilename, final int ex
         sortedSamIdx.deleteOnExit();
 
         final SAMFileWriter writer = new SAMFileWriterFactory()
-                .setCreateIndex(true).makeBAMWriter(builder.getHeader(), false, inputSam);
+                .setCreateIndex(true).makeSAMWriter(builder.getHeader(), false, inputSam);
```
danxmoran: This change also stops the test from leaving index files lying around. It also seems to make more sense given the `inputSam` name and extension, but I honestly don't know enough about BAM vs SAM to know if this is the right move. The test passes either way.

droazen: Looks like the right move to me -- the test case explicitly claims to "Build a sam file for testing" (rather than a BAM file)

src/test/java/picard/analysis/CollectMultipleMetricsTest.java
sooheelee: Is the index a valid index? If you run ValidateSamFile on the bam/bai pair, do the files validate?

PR #848, opened 2017-06-27
Add CBCL support to ExtractIlluminaBarcodes and CollectBasecallingMetrics

jacarey:
### Description

This PR adds support for the new basecall converter to extract illumina barcods and collect basecalling metrics. It also removes the basecall extraction and metrics from basecalls to sam and basecalls to fastq to avoid double processing.

----

#### Content

#### Review


src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
```diff
@@ -92,29 +105,31 @@
...
 
-    @Option(doc=ReadStructure.PARAMETER_DOC, shortName="RS")
+    @Option(doc = ReadStructure.PARAMETER_DOC, shortName = "RS")
     public String READ_STRUCTURE;
 
-    @Option(doc="The file to which the collected metrics are written", shortName= StandardOptionDefinitions.OUTPUT_SHORT_NAME, optional = true)
+    @Option(doc = "The file to which the collected metrics are written", shortName = StandardOptionDefinitions.OUTPUT_SHORT_NAME, optional = true)
     public File OUTPUT;
 
+    @Option(doc = "Use the new converter", optional = true)
```
gbggrant: Maybe a little clearer explanation of what the 'new converter' is  and why I should want to use it?

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
```diff
@@ -178,38 +195,94 @@ protected int doWork() {
...
+
+            //locs
+            List<AbstractIlluminaPositionFileReader.PositionInfo> locs = new ArrayList<>();
+            File locsFile = new File(BASECALLS_DIR.getParentFile(), "s.locs");
+            try (LocsFileReader locsFileReader = new LocsFileReader(locsFile)) {
+                while (locsFileReader.hasNext()) {
+                    locs.add(locsFileReader.next());
+                }
+            }
+            IOUtil.assertFileIsReadable(locsFile);
```
gbggrant: Shouldn't you assert it's readable before reading it?  And maybe this should be in a CustomCommandLineValidation method?

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
```diff
@@ -178,38 +195,94 @@ protected int doWork() {
...
+            IOUtil.assertFilesAreReadable(Arrays.asList(filterFiles));
+            Pattern barcodeRegex = Pattern.compile(ParameterizedFileUtil.escapePeriods(
+                    ParameterizedFileUtil.makeLaneTileRegex("_barcode.txt", LANE)));
+            Map<Integer, File> barcodesFiles = new HashMap<>();
+            for (File barcodeFile : getTiledFiles(laneDir, barcodeRegex)) {
+                Matcher tileMatcher = barcodeRegex.matcher(barcodeFile.getName());
+                if (tileMatcher.matches()) {
+                    barcodesFiles.put(Integer.valueOf(tileMatcher.group(1)), barcodeFile);
+                }
+            }
```
gbggrant: assert barcodeFiles are readable?

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
```diff
@@ -178,38 +195,94 @@ protected int doWork() {
...
+        } else {
+            BaseIlluminaDataProvider provider = factory.makeDataProvider();
+
+            while (provider.hasNext()) {
+                final ClusterData cluster = provider.next();
+                addCluster(cluster);
+            }
         }
 
+
```
gbggrant: I don't like this additional line.

src/main/java/picard/illumina/ExtractIlluminaBarcodes.java
```diff
@@ -174,6 +183,9 @@
             "the number available on the machine less NUM_PROCESSORS.")
     public int NUM_PROCESSORS = 1;
 
+    @Option(doc = "Use the new converter", optional = true)
```
gbggrant: Same comment about further description of new functionality here.

src/main/java/picard/illumina/ExtractIlluminaBarcodes.java
```diff
@@ -223,23 +235,78 @@ protected int doWork() {
...
+
+            //locs
+            List<AbstractIlluminaPositionFileReader.PositionInfo> locs = new ArrayList<>();
+            File locsFile = new File(BASECALLS_DIR.getParentFile(), "s.locs");
+            try (LocsFileReader locsFileReader = new LocsFileReader(locsFile)) {
+                while (locsFileReader.hasNext()) {
+                    locs.add(locsFileReader.next());
+                }
+            }
+            IOUtil.assertFileIsReadable(locsFile);
```
gbggrant: Move the assertFileIsReadable up above line 263.

src/main/java/picard/illumina/CheckIlluminaDirectory.java
```diff
@@ -167,7 +167,9 @@ protected int doWork() {
                     filterFileMap.put(fileToTile(filterFile.getName()), filterFile);
                 }
 
-                CbclReader reader = new CbclReader(cbcls, filterFileMap, readStructure.readLengths, tiles.get(0), locs, true);
+                OutputMapping outputMapping = new OutputMapping(readStructure);
+
+                CbclReader reader = new CbclReader(cbcls, filterFileMap, readStructure.readLengths, tiles.get(0), locs, outputMapping.getOutputCycles(), true);
```
nh13: reader/outputMapping can be final I think

src/main/java/picard/illumina/NewIlluminaBasecallsConverter.java
```diff
@@ -74,12 +57,6 @@
...
      * @param includeNonPfReads        If true, will include ALL reads (including those which do not have PF set)
      * @param ignoreUnexpectedBarcodes If true, will ignore reads whose called barcode is not found in barcodeRecordWriterMap,
-     * @param maxNoCalls               The maximum number of no calls to allow in a match.
-     * @param maxMismatches            The maximum number of mismatched calls to allow in a match.
-     * @param minMismatchDelta         The minimum mismatch difference between the best and second best matches.
-     * @param minimumBaseQuality       The minimum base quality to allow for a matching base call.
-     * @param metrics                  The metrics output for barcode metrics.
-     * @param metricsFile              The metrics file to write out the barcode metrics to.
      */
     public NewIlluminaBasecallsConverter(final File basecallsDir, File barcodesDir, final int lane,
```
nh13: I thought this class was going to be renamed in a previous review.  Can you rename it now?

src/main/java/picard/illumina/IlluminaBasecallsToSam.java
```diff
@@ -312,21 +297,14 @@ private void initialize() {
...
+            if (BARCODES_DIR == null) BARCODES_DIR = BASECALLS_DIR;
             basecallsConverter = new NewIlluminaBasecallsConverter<>(BASECALLS_DIR, BARCODES_DIR, LANE, readStructure,
                     barcodeSamWriterMap, true, Math.max(1, MAX_READS_IN_RAM_PER_TILE / numOutputRecords),
                     TMP_DIR, NUM_PROCESSORS,
                     FIRST_TILE, TILE_LIMIT, new QueryNameComparator(),
                     new Codec(numOutputRecords),
                     SAMRecordsForCluster.class, bclQualityEvaluationStrategy,
-                    this.APPLY_EAMSS_FILTER, INCLUDE_NON_PF_READS, IGNORE_UNEXPECTED_BARCODES, MAX_NO_CALLS,
-                    MAX_MISMATCHES, MIN_MISMATCH_DELTA, MINIMUM_QUALITY, metrics, METRICS_FILE);
+                    this.APPLY_EAMSS_FILTER, INCLUDE_NON_PF_READS, IGNORE_UNEXPECTED_BARCODES);
```
nh13: No your doing, but we have `this. APPLY_EAMSS_FILTER` and `INCLUDE_NON_PF_READS`, can we remove `this.`?

src/main/java/picard/illumina/IlluminaBasecallsToFastq.java
```diff
@@ -201,18 +199,6 @@
...
-    @Option(doc = "Maximum mismatches for a barcode to be considered a match.")
-    public int MAX_MISMATCHES = 1;
-
-    @Option(doc = "Minimum difference between number of mismatches in the best and second best barcodes for a barcode to be considered a match.")
-    public int MIN_MISMATCH_DELTA = 1;
-
-    @Option(doc = "Maximum allowable number of no-calls in a barcode read before it is considered unmatchable.")
-    public int MAX_NO_CALLS = 2;
-
-    @Option(doc = "Per-barcode and per-lane metrics written to this file.", shortName = StandardOptionDefinitions.METRICS_FILE_SHORT_NAME, optional = true)
```
nh13: Could this be deprecated first in one release, then removed?  Perhaps remove the functionality and add a custom command line validation to error when it is specified?  Seems like people may be using this.

src/main/java/picard/illumina/NewIlluminaBasecallsConverter.java
```diff
@@ -151,32 +110,35 @@ public NewIlluminaBasecallsConverter(final File basecallsDir, File barcodesDir,
         IOUtil.assertFileIsReadable(locsFile);
         //filter
 
-        Pattern laneTileRegex = Pattern.compile(ParameterizedFileUtil.escapePeriods(
+        Pattern filterRegex = Pattern.compile(ParameterizedFileUtil.escapePeriods(
```
nh13: not yours, but add `final` here.  It's nice to clean it up while you are here.

nh13: Lots more finals here.  IntelliJ will help you.

src/main/java/picard/illumina/NewIlluminaBasecallsConverter.java
```diff
@@ -151,32 +110,35 @@ public NewIlluminaBasecallsConverter(final File basecallsDir, File barcodesDir,
...
+                    ParameterizedFileUtil.makeLaneTileRegex("_barcode.txt(\\.gz|\\.bz2)?", lane)));
+            File[] barcodeTileFiles = getTiledFiles(barcodesDir, barcodeRegex);
+            if (barcodeTileFiles.length != tiles.size()) {
+                throw new PicardException("Barcode files are required for each tile.");
+            }
+            for (File barcodeFile : barcodeTileFiles) {
+                Matcher tileMatcher = barcodeRegex.matcher(barcodeFile.getName());
+                if (tileMatcher.matches()) {
+                    barcodesFiles.put(Integer.valueOf(tileMatcher.group(1)), barcodeFile);
+                }
```
nh13: finals!!!

src/main/java/picard/illumina/NewIlluminaBasecallsConverter.java
```diff
@@ -151,32 +110,35 @@ public NewIlluminaBasecallsConverter(final File basecallsDir, File barcodesDir,
...
-        int index = 0;
-        for (final ReadDescriptor d : readStructure.descriptors) {
-            if (d.type == ReadType.Barcode) {
-                noMatchBarcode[index++] = StringUtil.repeatCharNTimes('N', d.length);
+        if (demultiplex) {
+            Pattern barcodeRegex = Pattern.compile(ParameterizedFileUtil.escapePeriods(
+                    ParameterizedFileUtil.makeLaneTileRegex("_barcode.txt(\\.gz|\\.bz2)?", lane)));
+            File[] barcodeTileFiles = getTiledFiles(barcodesDir, barcodeRegex);
+            if (barcodeTileFiles.length != tiles.size()) {
+                throw new PicardException("Barcode files are required for each tile.");
```
nh13: Can you add in the message how many where found and how many were expected?

src/main/java/picard/illumina/parser/IlluminaDataProviderFactory.java
```diff
@@ -201,7 +205,25 @@ public IlluminaDataProviderFactory(File basecallDirectory, File barcodesDirector
...
+        File laneDir = new File(basecallDirectory, IlluminaFileUtil.longLaneStr(lane));
+
+        List<Integer> tiles = new ArrayList<>();
+        File[] filterFiles = getTiledFiles(laneDir, laneTileRegex);
+        for (File filterFile : filterFiles) {
+            Matcher tileMatcher = laneTileRegex.matcher(filterFile.getName());
+            if (tileMatcher.matches()) {
+                tiles.add(Integer.valueOf(tileMatcher.group(1)));
+            }
+        }
```
nh13: duplicated code (no finals too) from `NewIlluminaBasecallsConverter.java`

src/main/java/picard/illumina/parser/NewIlluminaDataProvider.java
```diff
@@ -12,58 +11,38 @@
...
-        this.noMatchMetric = noMatchMetric;
+        this.reader = new CbclReader(cbcls, filterFileMap, outputMapping.getOutputReadLengths(), tileNum, locs, outputMapping.getOutputCycles(), false);
+        if (barcodeFile != null) {
+            this.barcodeReader = new BarcodeFileReader(barcodeFile);
+        }
     }
 
     @Override
     public void close() {
+        reader.clear();
```
nh13: Why clear when you're just going to close?

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -269,7 +272,8 @@ private void advance() {
             for (int cycle = 0; cycle < outputLengths[read]; cycle++) {
                 CycleData currentCycleData = cycleData[totalCycleCount];
 
-                if (cachedTilePosition[totalCycleCount] >= cachedTile[totalCycleCount].length) {
+                if (cachedTilePosition[totalCycleCount] >= cachedTile[totalCycleCount].length
+                        || cachedTilePosition[totalCycleCount] >= cycleData[totalCycleCount].getTileInfo().getNumClustersInTile()) {
```
nh13: Can you explain this logic?

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -361,7 +364,7 @@ private void cacheTile(int totalCycleCount, TileData tileData, CycleData current
         // Write buffer contents to cached tile array
         // if nonPF reads are included we need to strip them out
         if (!currentCycleData.pfExcluded) {
-            boolean[] filterDatas = cachedFilter.get(tileData.tileNum);
+            List<Boolean> filterDatas = cachedFilter.get(tileData.tileNum);
```
nh13: so man finals

src/main/java/picard/util/ThreadPoolExecutorWithExceptions.java
```diff
@@ -0,0 +1,37 @@
...
+import picard.PicardException;
+
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+public class ThreadPoolExecutorWithExceptions extends ThreadPoolExecutor {
```
nh13: how about some docs?

src/test/java/picard/illumina/parser/readers/CbclReaderTest.java
```diff
@@ -31,9 +31,8 @@ public void testReadValidFile() {
         filters.put(1101, TILE_1101_FILTER);
         LocsFileReader locsFileReader = new LocsFileReader(new File("testdata/picard/illumina/readerTests/s_1_6.locs"));
         List<AbstractIlluminaPositionFileReader.PositionInfo> locs = locsFileReader.toList();
-
-        final CbclReader reader = new CbclReader(Arrays.asList(PASSING_CBCL_C99_1, PASSING_CBCL_C100_1),
-                filters, new int[]{2}, 1101, locs, false);
+        final CbclReader reader = new CbclReader(Arrays.asList(PASSING_CBCL_C1_1, PASSING_CBCL_C2_1),
+                filters, new int[]{2}, 1101, locs, new int[]{1, 2}, false);
```
nh13: How about a test where it fails if no barcode files are found?  How about a test if not all the tiles were found? My apologies if they are already there.

src/main/java/picard/illumina/parser/readers/CbclReader.java
```diff
@@ -406,4 +409,7 @@ public int getHeaderSize() {
         return allTiles;
     }
 
+    public void clear() {
```
nh13: Could this just be `close()`?  I dont' see it being used anywhere else but another `close()` call.

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
nh13: When does this need be used?  It would be nice not to have to know the type of sequencer data you have and it "just works".  I am not sure that's the case anymore.

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
```diff
@@ -273,31 +347,32 @@ public void incrementClusterCount(final Integer tileNumber, final double increme
...
             final Histogram<Integer> laneToPctPfClusterHistogram = tileToPfClusterHistogram.divideByHistogram(tileToClusterHistogram);
             final double meanPctPfClustersPerTile = laneToPctPfClusterHistogram.getMeanBinSize();
-            metrics.MEAN_PCT_PF_CLUSTERS_PER_TILE = (Double.isNaN(meanPctPfClustersPerTile) ?  0 : Double.valueOf(decFormat.format(meanPctPfClustersPerTile * 100)));
+            metrics.MEAN_PCT_PF_CLUSTERS_PER_TILE = (Double.isNaN(meanPctPfClustersPerTile) ? 0 : Double.valueOf(decFormat.format(meanPctPfClustersPerTile * 100)));
             metrics.SD_PCT_PF_CLUSTERS_PER_TILE = Double.valueOf(decFormat.format(laneToPctPfClusterHistogram.getStandardDeviationBinSize(meanPctPfClustersPerTile) * 100));
 
             metrics.TOTAL_CLUSTERS = (long) this.tileToClusterHistogram.getSumOfValues();
-            metrics.PF_CLUSTERS    = (long) this.tileToPfClusterHistogram.getSumOfValues();
-            
+            metrics.PF_CLUSTERS = (long) this.tileToPfClusterHistogram.getSumOfValues();
```
nh13: removing alignments!

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
```diff
@@ -273,31 +347,32 @@ public void incrementClusterCount(final Integer tileNumber, final double increme
...
+
             final ReadStructure readStructure = new ReadStructure(READ_STRUCTURE);
             int templateBaseCountPerCluster = 0;
-            for (int i = 0; i < readStructure.templates.length(); i++) templateBaseCountPerCluster += readStructure.templates.get(i).length;
+            for (int i = 0; i < readStructure.templates.length(); i++)
+                templateBaseCountPerCluster += readStructure.templates.get(i).length;
```
nh13: add `{}`

src/main/java/picard/illumina/ExtractIlluminaBarcodes.java
nh13: Yeah, a big 👎  if you need to intelligently set this based on sequencer type.  These demux tools already hard enough to run.

src/main/java/picard/illumina/ExtractIlluminaBarcodes.java
```diff
@@ -798,16 +915,25 @@ public static BarcodeMatch findBestBarcodeAndUpdateMetrics(final byte[][] readSu
...
-                        if (!SequenceUtil.basesEqual(barcodeBytes[j][i], readSubsequence[j][i])) ++numMismatches;
-                        else if (qualities != null && qualities[j][i] < minimumBaseQuality) ++numMismatches;
+                for (int i = 0; i < barcodeBytes[j].length; ++i) {
+                    if (readSubsequence[j].length <= i) {
+                        break;
+                    }
+                    if (SequenceUtil.isNoCall(readSubsequence[j][i])) {
+                        continue;
+                    }
+                    if (!SequenceUtil.basesEqual(barcodeBytes[j][i], readSubsequence[j][i])) {
```
nh13: so why not just `else if`

src/main/java/picard/illumina/ExtractIlluminaBarcodes.java
```diff
@@ -798,16 +915,25 @@ public static BarcodeMatch findBestBarcodeAndUpdateMetrics(final byte[][] readSu
...
-            // Read sequence and barcode length may not be equal, so we just use the shorter of the two
+
             for (int j = 0; j < barcodeBytes.length; j++) {
-                final int basesToCheck = Math.min(barcodeBytes[j].length, readSubsequence[j].length);
-                for (int i = 0; i < basesToCheck; ++i) {
-                    if (!SequenceUtil.isNoCall(readSubsequence[j][i])) {
-                        if (!SequenceUtil.basesEqual(barcodeBytes[j][i], readSubsequence[j][i])) ++numMismatches;
-                        else if (qualities != null && qualities[j][i] < minimumBaseQuality) ++numMismatches;
+                for (int i = 0; i < barcodeBytes[j].length; ++i) {
+                    if (readSubsequence[j].length <= i) {
```
nh13: why not in the for loop condition?

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
```diff
@@ -178,38 +195,94 @@ protected int doWork() {
...
+
+            if (cbcls.size() == 0) {
+                throw new PicardException("No CBCL files found.");
+            }
+
+            IOUtil.assertFilesAreReadable(cbcls);
+
+            //locs
+            List<AbstractIlluminaPositionFileReader.PositionInfo> locs = new ArrayList<>();
+            File locsFile = new File(BASECALLS_DIR.getParentFile(), "s.locs");
```
pshapiro4broad: We should use a constant for this, e.g. `AbstractIlluminaPositionFileReader.S_LOCS_FILE`

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
```diff
@@ -38,26 +38,39 @@
...
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.SortedMap;
 import java.util.TreeMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static picard.illumina.NewIlluminaBasecallsConverter.getTiledFiles;
```
pshapiro4broad: I think we should avoid static imports like this and instead call the method using `NewIlluminaBasecallsConverter.getTiledFiles()`

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
```diff
@@ -159,17 +174,19 @@ protected int doWork() {
                     barcodeToMetricCounts.put(barcode.toString(), new IlluminaMetricCounts(barcode.toString(), barcodeName, LANE));
                 }
             }
-
-            factory = barcodeToMetricCounts.isEmpty()
-                    ? new IlluminaDataProviderFactory(
+            if (USE_NEW_CONVERTER) {
+                factory = new IlluminaDataProviderFactory(BASECALLS_DIR, BARCODES_DIR, LANE, readStructure, bclQualityEvaluationStrategy);
```
pshapiro4broad: This assigns `factory` but its only use in the `USE_NEW_CONVERTER` case is inside another test for `USE_NEW_CONVERTER`. It seems like it would be simpler to only test `USE_NEW_CONVERTER` once and not use `factory for that case.

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
```diff
@@ -178,38 +195,94 @@ protected int doWork() {
...
+
         onComplete();
         return 0;
     }
 
     /***
      * Process new cluster of Illumina data - increment a running counter of data
      */
-    private void addCluster(final ClusterData cluster) {
+    private synchronized void addCluster(final ClusterData cluster) {
```
pshapiro4broad: IMO any use of `synchronized` should be accompanied by a comment. It's not obvious to me why this block would need to be synchronized. I don't see any global state and the code doesn't appear to be threaded.

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
```diff
@@ -224,7 +297,8 @@ private void onComplete() {
                 counts.addMetricsToFile(file);
                 allLaneCounts.addIlluminaMetricCounts(counts);
             }
-            if (! barcodeToMetricCounts.keySet().contains("")) allLaneCounts.addMetricsToFile(file);  // detect non-indexed case
+            if (!barcodeToMetricCounts.keySet().contains(""))
```
pshapiro4broad: may as well as `{` `}` too

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
pshapiro4broad: IMO this is totally fine, no need to align these.

src/main/java/picard/illumina/NewIlluminaBasecallsConverter.java
jacarey: I'm open to suggestions

src/main/java/picard/illumina/ExtractIlluminaBarcodes.java
```diff
@@ -223,23 +235,78 @@ protected int doWork() {
...
+        final List<PerTileBarcodeExtractor> extractors = new ArrayList<>(factory.getAvailableTiles().size());
+
+        if (USE_NEW_CONVERTER) {
+            File laneDir = new File(BASECALLS_DIR, IlluminaFileUtil.longLaneStr(LANE));
+
+            File[] cycleDirs = IOUtil.getFilesMatchingRegexp(laneDir, IlluminaFileUtil.CYCLE_SUBDIRECTORY_PATTERN);
+
+            //CBCLs
+            List<File> cbcls = new ArrayList<>();
+            Arrays.asList(cycleDirs)
```
pshapiro4broad: Some refactoring needed here to avoid code duplication. E.g. the code on line 261 could be moved into a static method in `AbstractIlluminaPositionFileReader`.

src/main/java/picard/illumina/parser/readers/CbclReader.java
jacarey: This is for the case where you have a odd number of PF reads despite the fact that you have to have a even number since there are 2 reads per byte. Perhaps I could do away with the first check in the or.. this may be a bit belt and suspenders.

src/main/java/picard/illumina/parser/NewIlluminaDataProvider.java
jacarey: clear could be part of the close but we need to clear out the cache since the thread executor jobs hold reference to these readers.

src/main/java/picard/illumina/IlluminaBasecallsToFastq.java
jacarey: Basecalls to fastq only created metrics for one release while we were testing and there are lots of guts inside that aren't necessary now that extract barcodes can do all of this.. I'd be hesitant for it to stick around at all and i doubt anyone has used it in that time (1 week or so)

src/main/java/picard/illumina/ExtractIlluminaBarcodes.java
jacarey: The rational here was that this new convert could (and likely will) support other sequencer types and replace the outdated internal machinery for data providers. That said this was left as an explicit argument because I would consider it experimental. I'm happy to change it to be all encompassing and do things under the covers.

src/main/java/picard/illumina/ExtractIlluminaBarcodes.java
pshapiro4broad: It looks like you could merge it into a single `if` test (after you move the first `if` into the loop test) since the only two outcomes are either "increment `numMismatches`" or "do nothing"

src/main/java/picard/illumina/parser/readers/CbclReader.java
jacarey: I have my local variables should be final inspection off.

src/main/java/picard/illumina/NewIlluminaBasecallsConverter.java
nh13: You're asking the wrong person about naming software and classes.

src/main/java/picard/illumina/parser/readers/CbclReader.java
nh13: If I run a NovaSeq, do I hvae to set this to true, or does it work out of the box?

src/main/java/picard/illumina/ExtractIlluminaBarcodes.java
jacarey: if with continue is slightly faster in that if the initial condition is met the second doesn't even need to be evaluated.

I thought this wouldn't make much of a difference (since I figured the compiler would optimize this down to the same bytecode.. However, several rounds of profiling showed a significant difference. I'd be happy to share my timings if you like.

src/main/java/picard/illumina/IlluminaBasecallsToFastq.java
nh13: 👍

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
```diff
@@ -38,26 +38,39 @@
...
-        - *  Requires a Lane and an input file of Barcodes to expect.
-        - *  Outputs metrics:
-        - *    *  Mean Clusters Per Tile
-        - *    *  Standard Deviation of Clusters Per Tile
-        - *    *  Mean Pf Clusters Per Tile
-        - *    *  Standard Deviation of Pf Clusters Per Tile
-        - *    *  Mean Percentage of Pf Clusters Per Tile
-        - *    *  Standard Deviation of Percentage of Pf Clusters Per Tile
-        - */
+ - *  A Command line tool to collect Illumina Basecalling metrics for a sequencing run
```
jacarey: great question... not sure to be honest.. i think Intellij just reformatted this.. but i'll fix it

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
jacarey: factory.makeProvider is used outside of that in the else.

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
jacarey: Actually not needed.. I was testing with a parellell stream but that proved unfruitful

src/main/java/picard/illumina/parser/readers/CbclReader.java
jacarey: I assume you are talking about USE_NEW_CONVERTER even though this comment is down here.Currently the old converter does not support cbcls. So yes you would have to set it to true.  I'm happy to have it detect this case (presence of cbcls), however once we can use the new converter with say HiSeqXs we might have to revisit this.

jacarey:
@nh13 I removed USE_NEW_CONVERTER in lieu of a method to check for cbcls so it should "just work" now.

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
pshapiro4broad: It looks to me that you could merge this `if` with the one below, since the only intervening line doesn't rely on `factory`. Once that's done, you can change the new converter line to: 
```
setupNewDataProvider(new IlluminaDataProviderFactory(BASECALLS_DIR, BARCODES_DIR, LANE, readStructure, bclQualityEvaluationStrategy));
```

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
```diff
@@ -178,22 +196,82 @@ protected int doWork() {
...
+                ParameterizedFileUtil.makeLaneTileRegex(".filter", LANE)));
+        final File[] filterFiles = NewIlluminaBasecallsConverter.getTiledFiles(laneDir, laneTileRegex);
+
+        IOUtil.assertFilesAreReadable(Arrays.asList(filterFiles));
+        final Pattern barcodeRegex = Pattern.compile(ParameterizedFileUtil.escapePeriods(
+                ParameterizedFileUtil.makeLaneTileRegex("_barcode.txt", LANE)));
+
+
+        final Map<Integer, File> barcodesFiles = new HashMap<>();
+        IOUtil.assertFilesAreReadable(new ArrayList<>(barcodesFiles.values()));
```
pshapiro4broad: `barcodesFiles` is empty here. you could put this assert inside the `if` on line 261.

src/main/java/picard/illumina/CollectIlluminaBasecallingMetrics.java
```diff
@@ -178,22 +196,82 @@ protected int doWork() {
                         IlluminaDataType.PF,
                         IlluminaDataType.Position,
                         IlluminaDataType.Barcodes);
+            }
         }
 
         unmatched_barcode = StringUtil.repeatCharNTimes('N', barcodeLength);
```
pshapiro4broad: This should be `unmatchedBarcode`.

sooheelee:
Hey @jacarey, what do you think of this. When deprecating arguments, leave them in the code and note that they are deprecated and what has replaced the functionality, if anything. After a while, after some number of releases, then remove the deprecated argument altogether. 

We do this sort of transitive deprecation in GATK. This manner of transitioning is less jarring for our users. I just got a user question today asking what happened to the `USE_NEW_CONVERTER` option. I haven't had a chance to clean up the release notes for 2.9.5, let alone 2.10.0, due to higher priority GATK work. In the meanwhile,  users are left searching the repo for what has happened.

jacarey:
@sooheelee There were a few reasons we didn't bother deprecating this option. 

1) It was marked experimental. 
2) It was only present in one release. 
3) It was only relevant for people using NovaSeqs. 
4) The upgrade path is automatic. Just remove the flag and it automatically determines if it needs the new converter or not. 

Rather than have people ignore a deprecation warning and potentially have more users using the option we opted to remove it.

PR #853, opened 2017-07-05
Add option to CollectWgsMetrics to output theoretical sensitivity at various allele fractions

fleharty:
### Description

This adds the THEORETICAL_SENSITIVITY_OUTPUT option to CollectWgsMetrics.  This option allows a user to specify a file that will contain the theoretical sensitivities at various allele fractions.

This augments the current TheoreticalHetSensitivity so that we can now calculate theoretical sensitivities for non-germline samples.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/analysis/CollectRawWgsMetrics.java
+++ b/src/main/java/picard/analysis/CollectRawWgsMetrics.java
@@ -30,6 +30,8 @@ import picard.cmdline.CommandLineProgramProperties;
 import picard.cmdline.Option;
 import picard.cmdline.programgroups.Metrics;
 
+import java.io.File;
+
 import static picard.cmdline.StandardOptionDefinitions.MINIMUM_MAPPING_QUALITY_SHORT_NAME;
 
 /**
--- a/src/main/java/picard/analysis/CollectWgsMetrics.java
+++ b/src/main/java/picard/analysis/CollectWgsMetrics.java
@@ -118,6 +118,9 @@ static final String USAGE_DETAILS = "<p>This tool collects metrics about the fra
     @Option(doc="Sample Size used for Theoretical Het Sensitivity sampling. Default is 10000.", optional = true)
     public int SAMPLE_SIZE=10000;
 
+    @Option(doc="Output for Theoretical Sensitivity metrics.  Default is null.", optional = true)
+    public File THEORETICAL_SENSITIVITY_OUTPUT;
+
     @Option(doc = "If true, fast algorithm is used.")
     public boolean USE_FAST_ALGORITHM = false;
 
@@ -133,7 +136,7 @@ static final String USAGE_DETAILS = "<p>This tool collects metrics about the fra
     private SAMFileHeader header = null;
 
     private final Log log = Log.getInstance(CollectWgsMetrics.class);
-    private static final double LOG_ODDS_THRESHOLD = 3.0;
+    private static final double LOG_ODDS_THRESHOLD = 3;
 
     /** Metrics for evaluating the performance of whole genome sequencing experiments. */
     public static class WgsMetrics extends MergeableMetricBase {
@@ -466,6 +469,31 @@ static final String USAGE_DETAILS = "<p>This tool collects metrics about the fra
         final MetricsFile<WgsMetrics, Integer> out = getMetricsFile();
         processor.addToMetricsFile(out, INCLUDE_BQ_HISTOGRAM, dupeFilter, mapqFilter, pairFilter);
         out.write(OUTPUT);
+
+        if(THEORETICAL_SENSITIVITY_OUTPUT != null) {
+            final double[] depthDoubleArray = TheoreticalSensitivity.normalizeHistogram(collector.getUnfilteredDepthHistogram());
+            final double[] baseQDoubleArray = TheoreticalSensitivity.normalizeHistogram(collector.getUnfilteredBaseQHistogram());
+
+            collector.getUnfilteredBaseQHistogram();
+            collector.getUnfilteredDepthHistogram();
+
+            TheoreticalSensitivityMetrics theoreticalSensitivityMetrics = new TheoreticalSensitivityMetrics();
+            int theoreticalHetSensitivitySampleSize = 10000;
+
+            double logOddsThreshold = 6.2; // This threshold is used because it is the value used for MuTect2.
+            theoreticalSensitivityMetrics.SENSITIVITY_AT_0_1 = TheoreticalSensitivity.theoreticalSensitivity(depthDoubleArray, baseQDoubleArray, theoreticalHetSensitivitySampleSize, logOddsThreshold, 0.001);
+            theoreticalSensitivityMetrics.SENSITIVITY_AT_0_5 = TheoreticalSensitivity.theoreticalSensitivity(depthDoubleArray, baseQDoubleArray, theoreticalHetSensitivitySampleSize, logOddsThreshold, 0.005);
+            theoreticalSensitivityMetrics.SENSITIVITY_AT_01 = TheoreticalSensitivity.theoreticalSensitivity(depthDoubleArray, baseQDoubleArray, theoreticalHetSensitivitySampleSize, logOddsThreshold, 0.01);
+            theoreticalSensitivityMetrics.SENSITIVITY_AT_02 = TheoreticalSensitivity.theoreticalSensitivity(depthDoubleArray, baseQDoubleArray, theoreticalHetSensitivitySampleSize, logOddsThreshold, 0.02);
+            theoreticalSensitivityMetrics.SENSITIVITY_AT_05 = TheoreticalSensitivity.theoreticalSensitivity(depthDoubleArray, baseQDoubleArray, theoreticalHetSensitivitySampleSize, logOddsThreshold, 0.05);
+            theoreticalSensitivityMetrics.SENSITIVITY_AT_10 = TheoreticalSensitivity.theoreticalSensitivity(depthDoubleArray, baseQDoubleArray, theoreticalHetSensitivitySampleSize, logOddsThreshold, 0.10);
+            theoreticalSensitivityMetrics.SENSITIVITY_AT_30 = TheoreticalSensitivity.theoreticalSensitivity(depthDoubleArray, baseQDoubleArray, theoreticalHetSensitivitySampleSize, logOddsThreshold, 0.30);
+            theoreticalSensitivityMetrics.SENSITIVITY_AT_50 = TheoreticalSensitivity.theoreticalSensitivity(depthDoubleArray, baseQDoubleArray, theoreticalHetSensitivitySampleSize, logOddsThreshold, 0.50);
+
+            final MetricsFile<TheoreticalSensitivityMetrics, Double> tsOut = getMetricsFile();
+            tsOut.addMetric(theoreticalSensitivityMetrics);
+            tsOut.write(THEORETICAL_SENSITIVITY_OUTPUT);
+        }
 
         return 0;
     }
--- a/src/main/java/picard/analysis/TheoreticalSensitivity.java
+++ b/src/main/java/picard/analysis/TheoreticalSensitivity.java
@@ -41,14 +41,15 @@ public class TheoreticalSensitivity {
 
     private static final Log log = Log.getInstance(TheoreticalSensitivity.class);
     private static final int SAMPLING_MAX = 600; //prevent 'infinite' loops
-    private static final int MAX_CONSIDERED_DEPTH = 1000; //no point in looking any deeper than this, otherwise GC overhead is too high.
+    private static final int MAX_CONSIDERED_DEPTH = 10000; //no point in looking any deeper than this, otherwise GC overhead is too high.
+    private static int randomSeed = 51;
 
     /**
-     * @param depthDistribution the probability of depth n is depthDistribution[n] for n = 0, 1. . . N - 1
+     * @param depthDistribution   the probability of depth n is depthDistribution[n] for n = 0, 1. . . N - 1
      * @param qualityDistribution the probability of quality q is qualityDistribution[q] for q = 0, 1. . . Q
-     * @param sampleSize sample size is the number of random sums of quality scores for each m
-     * @param logOddsThreshold is the log_10 of the likelihood ratio required to call a SNP,
-     * for example 5 if the variant likelihood must be 10^5 times greater
+     * @param sampleSize          sample size is the number of random sums of quality scores for each m
+     * @param logOddsThreshold    is the log_10 of the likelihood ratio required to call a SNP,
+     *                            for example 5 if the variant likelihood must be 10^5 times greater
      */
     public static double hetSNPSensitivity(final double[] depthDistribution, final double[] qualityDistribution,
                                            final int sampleSize, final double logOddsThreshold) {
@@ -56,12 +57,12 @@ public class TheoreticalSensitivity {
     }
 
     /**
-     * @param depthDistribution the probability of depth n is depthDistribution[n] for n = 0, 1. . . N - 1
+     * @param depthDistribution   the probability of depth n is depthDistribution[n] for n = 0, 1. . . N - 1
      * @param qualityDistribution the probability of quality q is qualityDistribution[q] for q = 0, 1. . . Q
-     * @param sampleSize sample size is the number of random sums of quality scores for each m
-     * @param logOddsThreshold is the log_10 of the likelihood ratio required to call a SNP,
-     * for example 5 if the variant likelihood must be 10^5 times greater.
-     * @param withLogging true to output log messages, false otherwise.
+     * @param sampleSize          sample size is the number of random sums of quality scores for each m
+     * @param logOddsThreshold    is the log_10 of the likelihood ratio required to call a SNP,
+     *                            for example 5 if the variant likelihood must be 10^5 times greater.
+     * @param withLogging         true to output log messages, false otherwise.
      */
     public static double hetSNPSensitivity(final double[] depthDistribution, final double[] qualityDistribution,
                                            final int sampleSize, final double logOddsThreshold, final boolean withLogging) {
@@ -143,7 +144,7 @@ public class TheoreticalSensitivity {
         private Random rng;
 
         RouletteWheel(final double[] weights) {
-            rng = new Random(51);
+            rng = new Random(randomSeed);
             N = weights.length;
 
             probabilities = new ArrayList<>();
@@ -203,5 +204,137 @@ public class TheoreticalSensitivity {
             }
         }
         return normalizedHistogram;
+    }
+
+    /**
+     * Determines if a variant would be called under the particular conditions of a given total depth, alt depth,
+     * average base qualities, allele fraction of variant and log odds threshold necessary to exceed to call variant.
+     * @param totalDepth Depth at the site to be called, both alt and ref.
+     * @param altDepth Number of alt bases at this site.
+     * @param averageQuality Average Phred-scaled quality of bases
+     * @param alleleFraction Allele fraction we are attempting to detect
+     * @param logOddsThreshold Log odds threshold necessary to exceed for variant to be called
+     * @return
+     */
+    public static boolean isCalled(int totalDepth, int altDepth, double averageQuality, double alleleFraction, double logOddsThreshold) {
+        double threshold;
+        double sumOfQualities = altDepth * averageQuality;
+        threshold = 10.0 * (altDepth * Math.log10(1.0 / alleleFraction) + (totalDepth - altDepth) * Math.log10(1.0 / (1.0 - alleleFraction)) + logOddsThreshold);
+
+        return sumOfQualities > threshold;
+    }
+
+    /**
+     * Draw from a binomial distribution.
+     * @param trials Number of trials to perform
+     * @param p Probability of individual success
+     * @param uniformRNG Random number generator to use for making draw
+     * @return Number of total successes
+     */
+    public static int binomialDraw(final int trials, final double p, final Random uniformRNG) {
+        if (p > 1.0 || p < 0) {
+            throw new PicardException("Probabilities should be between 0 and 1, found value " + p + ".");
+        }
+
+        int successes = 0;
+        for (int i = 0; i < trials; i++) {
+            if (uniformRNG.nextDouble() < p) {
+                successes++;
+            }
+        }
+        return successes;
+    }
+
+    /**
+     * Calculates the theoretical sensitivity with a given Phred-scaled quality score distribution at a constant
+     * depth.
+     * @param depth Depth to compute sensitivity at
+     * @param qualityDistribution Phred-scaled quality score distribution
+     * @param logOddsThreshold Log odd threshold necessary to exceed for variant to be called
+     * @param sampleSize sampleSize is the total number of simulations to run
+     * @param alleleFraction the allele fraction to evaluate sensitivity at
+     * @param randomSeed random number seed to use for random number generator
+     * @return
+     */
+    public static double sensitivityAtConstantDepth(final int depth, final double[] qualityDistribution, final double logOddsThreshold, final int sampleSize, final double alleleFraction, final int randomSeed) {
+        final RouletteWheel qualityRW = new RouletteWheel(trimDistribution(qualityDistribution));
+        final Random uniformRNG = new Random(randomSeed);
+
+        int altDepth = 0;
+        int calledVariants = 0;
+        for (int k = 0; k < sampleSize; k++) {
+            altDepth = binomialDraw(depth, alleleFraction, uniformRNG);
+
+            int sumOfQualities = 0;
+            for (int i = 0; i < altDepth; i++) {
+                sumOfQualities += qualityRW.draw();
+            }
+            if (isCalled(depth, altDepth, (double) sumOfQualities / (double) altDepth, alleleFraction, logOddsThreshold)) {
+                calledVariants++;
+            }
+        }
+        return (double) calledVariants / sampleSize;
+    }
+
+    /**
+     * Calculates the theoretical sensitivity with a given Phred-scaled quality score distribution at a constant
+     * depth.
+     * @param depth Depth to compute sensitivity at
+     * @param qualityDistribution Phred-scaled quality score distribution
+     * @param logOddsThreshold Log odds threshold necessary to exceed for variant to be called
+     * @param sampleSize the total number of simulations to run
+     * @param alleleFraction the allele fraction to evaluate sensitivity at
+     * @return
+     */
+    public static double sensitivityAtConstantDepth(final int depth, final double[] qualityDistribution, final double logOddsThreshold, final int sampleSize, final double alleleFraction) {
+        return sensitivityAtConstantDepth(depth, qualityDistribution, logOddsThreshold, sampleSize, alleleFraction, randomSeed);
+    }
+
+    /**
+     * Calculates the theoretical sensitivity with a given Phred-scaled quality score distribution and depth
+     * distribution.
+     * @param depthDistribution Depth distribution to compute theoretical sensitivity over
+     * @param qualityDistribution Phred-scaled quality score distribution
+     * @param sampleSize the total number of simulations to run
+     * @param logOddsThreshold Log odds threshold necessary to exceed for variant to be called
+     * @param alleleFraction the allele fraction to evaluate sensitivity at
+     * @return
+     */
+    public static double theoreticalSensitivity(final double[] depthDistribution, final double[] qualityDistribution,
+                                                final int sampleSize, final double logOddsThreshold, final double alleleFraction) {
+        double sensitivity = 0.0;
+        for (int k = 0; k < depthDistribution.length; k++) {
+            if(k % 10 == 0) {
+                log.info("Calculting sensitivity at depth " + k + " of " + depthDistribution.length);
+            }
+            sensitivity += sensitivityAtConstantDepth(k, qualityDistribution, logOddsThreshold, sampleSize, alleleFraction) * depthDistribution[k];
+        }
+        return sensitivity;
+    }
+
+    /**
+     * Removes trailing zeros in a distribution.  The purpose of this function is to prevent other
+     * functions from evaluating in regions where the distribution has zero probability.
+     * @param distribution Distribution of base qualities
+     * @return Distribution of base qualities removing any trailing zeros
+     */
+    public static double[] trimDistribution(final double[] distribution) {
+        int endOfDistribution = 0;
+
+        // Locate the index of the distribution where all the values remaining at
+        // larger indices are zero.
+        for(endOfDistribution = distribution.length-1;endOfDistribution >= 0;endOfDistribution--) {
+            if(distribution[endOfDistribution] != 0) {
+                break;
+            }
+        }
+
+        // Remove trailing zeros.
+        final double[] trimmedDistribution = new double[endOfDistribution+1];
+        for(int i = 0;i <= endOfDistribution;i++) {
+            trimmedDistribution[i] = distribution[i];
+        }
+
+        return trimmedDistribution;
     }
 }
--- a/src/main/java/picard/analysis/TheoreticalSensitivityMetrics.java
+++ b/src/main/java/picard/analysis/TheoreticalSensitivityMetrics.java
@@ -1,0 +1,33 @@
+package picard.analysis;
+
+import htsjdk.samtools.metrics.MetricBase;
+
+/**
+ * Created by fleharty on 6/30/17.
+ */
+public class TheoreticalSensitivityMetrics extends MetricBase {
+    /** Theoretical sensitivity at 0.1% allele fraction */
+    public double SENSITIVITY_AT_0_1 = 0.0;
+    /** Theoretical sensitivity at 0.1% allele fraction */
+
+    /** Theoretical sensitivity at 0.5% allele fraction */
+    public double SENSITIVITY_AT_0_5 = 0.0;
+
+    /** Theoretical sensitivity at 1% allele fraction */
+    public double SENSITIVITY_AT_01 = 0.0;
+
+    /** Theoretical sensitivity at 2% allele fraction */
+    public double SENSITIVITY_AT_02 = 0.0;
+
+    /** Theoretical sensitivity at 5% allele fraction */
+    public double SENSITIVITY_AT_05 = 0.0;
+
+    /** Theoretical sensitivity at 10% allele fraction */
+    public double SENSITIVITY_AT_10 = 0.0;
+
+    /** Theoretical sensitivity at 30% allele fraction */
+    public double SENSITIVITY_AT_30 = 0.0;
+
+    /** Theoretical sensitivity at 50% allele fraction */
+    public double SENSITIVITY_AT_50 = 0.0;
+}
--- a/src/test/java/picard/analysis/TheoreticalSensitivityTest.java
+++ b/src/test/java/picard/analysis/TheoreticalSensitivityTest.java
@@ -207,7 +207,7 @@ public class TheoreticalSensitivityTest {
     }
 
     @Test(dataProvider = "hetSensDataProvider")
-    public void testHetSensTargeted(final double expected, final File metricsFile) throws Exception{
+    public void testHetSensTargeted(final double expected, final File metricsFile) throws Exception {
         final double tolerance = 0.000_000_01;
 
         final MetricsFile Metrics = new MetricsFile();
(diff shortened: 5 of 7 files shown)


yfarjoun:
given that most of the somatic data is still exomes and not genomes, would it make sense to do this for CollectHsMetrics while you're at it?

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -467,6 +470,31 @@ protected int doWork() {
         processor.addToMetricsFile(out, INCLUDE_BQ_HISTOGRAM, dupeFilter, mapqFilter, pairFilter);
         out.write(OUTPUT);
 
+        if(THEORETICAL_SENSITIVITY_OUTPUT != null) {
+            final double[] depthDoubleArray = TheoreticalSensitivity.normalizeHistogram(collector.getUnfilteredDepthHistogram());
+            final double[] baseQDoubleArray = TheoreticalSensitivity.normalizeHistogram(collector.getUnfilteredBaseQHistogram());
+
+            collector.getUnfilteredBaseQHistogram();
```
yfarjoun: what do these 2 lines do?

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -467,6 +470,31 @@ protected int doWork() {
...
+            collector.getUnfilteredDepthHistogram();
+
+            TheoreticalSensitivityMetrics theoreticalSensitivityMetrics = new TheoreticalSensitivityMetrics();
```
yfarjoun: final

src/main/java/picard/analysis/TheoreticalSensitivityMetrics.java
```diff
@@ -0,0 +1,33 @@
+package picard.analysis;
+
+import htsjdk.samtools.metrics.MetricBase;
+
+/**
+ * Created by fleharty on 6/30/17.
+ */
+public class TheoreticalSensitivityMetrics extends MetricBase {
```
yfarjoun: wouldn't it be better to just emit a histogram with the values you want? it would be much easier to extend to arbitrary values. You could then also have the list of desired values as an input parameter to the CLP...

yfarjoun:
nm. you are doing that! :-)

src/main/java/picard/analysis/CollectWgsMetrics.java
fleharty: Removed these lines.  Fixed.

src/main/java/picard/analysis/CollectWgsMetrics.java
fleharty: fixed

fleharty:
@yfarjoun I'd be more than happy to present on the math at the next Palantir meeting.

src/main/java/picard/analysis/TheoreticalSensitivityMetrics.java
fleharty: @yfarjoun I'm not entirely sure I understand what you mean by histogram.
As I understand it, the histogram stores key-value pairs where the value is a non-negative integer.  I agree with you that we should make this tool so that we can emit sensitivities for any list of allele fractions the user is interested in.

fleharty: What I'm trying to say is, I don't see how to use a histogram to store a list of allele fractions with their corresponding sensitivities.

src/test/java/picard/analysis/TheoreticalSensitivityTest.java
```diff
@@ -216,13 +216,117 @@ public void testHetSensTargeted(final double expected, final File metricsFile) t
...
+    @DataProvider(name = "TheoreticalSensitivityConstantDepthDataProvider")
+    public Object[][] fractionalAlleleSensDataProvider() {
+        final File wgsMetricsFile = new File(TEST_DIR, "test_Solexa-332667.wgs_metrics");
+        final File targetedMetricsFile = new File(TEST_DIR, "test_25103070136.targeted_pcr_metrics");
+
+        //These magic numbers come from a separate implementation of the code in R.
+        return new Object[][]{
+                // expected sensitivity, metrics file, allele fraction, constant depth, sample size.
+                {1.00, wgsMetricsFile, .5, 30, 10000},
+                {1.00, wgsMetricsFile, .5, 30, 10000},
```
fleharty: Note to self, this is a duplicated test.

src/test/java/picard/analysis/TheoreticalSensitivityTest.java
```diff
@@ -216,13 +216,117 @@ public void testHetSensTargeted(final double expected, final File metricsFile) t
...
+                {0.78, targetedMetricsFile, .1, 30, 10000},
+                {0.26, targetedMetricsFile, 0.1, 10, 10000}
```
fleharty: I need to add a test for high depth and low allele fraction.  That was actually the original purpose of this test...

Also, I need to make sure my claim about the magic numbers coming from R is true.

src/test/java/picard/analysis/TheoreticalSensitivityTest.java
```diff
@@ -216,13 +216,117 @@ public void testHetSensTargeted(final double expected, final File metricsFile) t
...
+            Assert.assertEquals(result, expected, tolerance);
+        }
+    }
+
+    @DataProvider(name = "TheoreticalSensitivityDataProvider")
+    public Object[][] arbFracSensDataProvider() {
+        final File wgsMetricsFile = new File(TEST_DIR, "test_Solexa-332667.wgs_metrics");
+
+        // This test acts primarily as an intergration test.  The sample size of 100
+        // is not quite large enough to converge properly
```
fleharty: Mention that I use a small sample size so as to keep the test run time short.

src/test/java/picard/analysis/TheoreticalSensitivityTest.java
```diff
@@ -216,13 +216,117 @@ public void testHetSensTargeted(final double expected, final File metricsFile) t
...
+        };
+    }
+
+    @Test(dataProvider = "TheoreticalSensitivityDataProvider")
+    public void testSensitivity(final double expected, final File metricsFile, final double alleleFraction, final int sampleSize) throws Exception {
+        // This tests Theoretical Sensitivity using distributions on both base quality scores
+        // and the depth histogram.
+
+        // We use a pretty forgiving tolerance here because for these tests
+        // we are not using large enough sample sizes to converge.
```
fleharty: Again, mention that the reason here is to keep tests from running too long.

src/test/java/picard/analysis/TheoreticalSensitivityTest.java
```diff
@@ -216,13 +216,117 @@ public void testHetSensTargeted(final double expected, final File metricsFile) t
...
+                // the values converge.
+                {wgsMetricsFile, 0.01, 300},
+                {targetedMetricsFile, 0.01, 50}
+        };
+    }
+
+    @Test (dataProvider = "equivalanceHetVsArbitrary")
+    public void testHetVsArbitrary(final File metricsFile, final double tolerance, final int sampleSize) throws Exception {
+        // This test compares Theoretical Sensitivity for arbitrary allele fractions with the theoretical het sensitivity
+        // model.  Since allele fraction of 0.5 is equivalent to a het, these should provide the same answer.
```
fleharty: Mention that there are minor differences in the two models, but they shouldn't matter here.  They might matter for low allele fraction, but not for het variants.

src/main/java/picard/analysis/TheoreticalSensitivityMetrics.java
yfarjoun: The "key" to the histogram is of any Comparable type. Integer is Comparable, so that works, but you can use Double as well. You could use a String even! You can see an example for having a double in the histogram here: 
picard/sam/markduplicates/util/AbstractMarkDuplicatesCommandLineProgram.java:162

fleharty: fixes

src/test/java/picard/analysis/TheoreticalSensitivityTest.java
fleharty: fixed.

src/test/java/picard/analysis/TheoreticalSensitivityTest.java
fleharty: fixed.

fleharty:
@yfarjoun 
Thanks for the preliminary review.  
- added histogram that contains a list of allele fractions.
- added option for user specified allele fractions
- fixed issue with random seed from SamRecordSetBuilder

Back to you.

src/main/java/picard/analysis/directed/CollectTargetedMetrics.java
```diff
@@ -98,6 +102,11 @@ protected abstract COLLECTOR makeCollector(final Set<MetricAccumulationLevel> ac
     @Option(doc="Sample Size used for Theoretical Het Sensitivity sampling. Default is 10000.", optional = true)
     public int SAMPLE_SIZE=10000;
 
+    @Option(doc="Output for Theoretical Sensitivity metrics.  Default is null.", optional = true)
```
yfarjoun: specify that this is for the values specified in  the `ALLELE_FRACTION` argument

src/main/java/picard/analysis/directed/CollectTargetedMetrics.java
```diff
@@ -98,6 +102,11 @@ protected abstract COLLECTOR makeCollector(final Set<MetricAccumulationLevel> ac
...
+    public File THEORETICAL_SENSITIVITY_OUTPUT;
+
+    @Option(doc="Allele fraction to run theoretical sensitivity on.", optional = true)
+    public List<Double> ALLELE_FRACTION = new LinkedList<>(Arrays.asList(0.001, 0.005, 0.01, 0.02, 0.05, 0.1, 0.2, 0.3, 0.5));
```
yfarjoun: Arrays.asList() already returns a List<> so why do you wrap it in a LinkedList?

src/main/java/picard/analysis/TheoreticalSensitivityMetrics.java
```diff
@@ -0,0 +1,14 @@
...
+    /** Theoretical Het Sensitivity for log-odss of 6.2 **/
```
yfarjoun: in comment: not `Het`.

yfarjoun: Do we need the fixed metrics for anything? Seems like a red-herring.

src/main/java/picard/analysis/directed/CollectTargetedMetrics.java
fleharty: List<> will throw an UnsupportedOperation if  you later try to use .put().

fleharty: The .put() is used in the command line parsing, so I can't easily wrap it later.

yfarjoun: got it. thanks for that. so please wrap in an ArrayList instead.

src/main/java/picard/analysis/TheoreticalSensitivityMetrics.java
fleharty: By fixed metrics, I assume you mean the TheoreticalHet sensitivities for different log odds.

I had a discussion with @davidbenjamin about the log odds threshold, and I think we need to rethink the threshold we use for TheoreticalHetSensitivity.  This was just a nice way to include THS with a log odds threshold of 6.2.

I'll remove these two because we get THS with log odds of 3.0 in the other metrics file.  And we get TS with log odds of 6.2 for 0.5 allele fraction.

src/main/java/picard/analysis/CollectRawWgsMetrics.java
```diff
@@ -30,6 +30,8 @@
 import picard.cmdline.Option;
 import picard.cmdline.programgroups.Metrics;
 
+import java.io.File;
```
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,162 @@ public int draw() {
...
+            final double[] depthDoubleArray = TheoreticalSensitivity.normalizeHistogram(depthHistogram);
+            final double[] baseQDoubleArray = TheoreticalSensitivity.normalizeHistogram(baseQHistogram);
+
+            final TheoreticalSensitivityMetrics theoreticalSensitivityMetrics = new TheoreticalSensitivityMetrics();
+
+            theoreticalSensitivityMetrics.HET_SENSITIVITY_LOD3_0 = TheoreticalSensitivity.hetSNPSensitivity(depthDoubleArray, baseQDoubleArray, sampleSize, 3.0);
+            theoreticalSensitivityMetrics.HET_SENSITIVITY_LOD6_2 = TheoreticalSensitivity.hetSNPSensitivity(depthDoubleArray, baseQDoubleArray, sampleSize, 6.2);
+            double logOddsThreshold = 6.2; // This threshold is used because it is the value used for MuTect2.
+
+            Histogram<Double> sensitivityHistogram = new Histogram<>();
```
fleharty: A yes, good point.
fixed

src/main/java/picard/analysis/directed/CollectTargetedMetrics.java
fleharty: fixed

src/main/java/picard/analysis/directed/CollectTargetedMetrics.java
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivityMetrics.java
fleharty: Also, I was thinking about saving the 20/20/20 metric in this file.  It is a type of TheoreticalSensitivity, so it seems reasonable to me to put it there.  What do you think?

yfarjoun:
Let's talk about that offline.

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -467,6 +471,10 @@ protected int doWork() {
         processor.addToMetricsFile(out, INCLUDE_BQ_HISTOGRAM, dupeFilter, mapqFilter, pairFilter);
         out.write(OUTPUT);
 
+        if(THEORETICAL_SENSITIVITY_OUTPUT != null) {
```
fleharty: fixed

fleharty:
@yfarjoun After removing the TheoreticalHetSensitivity for LOD=3 and LOD=6.2
there are no fields in the metrics file (other than the histogram).  It appears that the file MetricsFile.read() function in HTSJDK isn't set up to deal with metric files that only contain a histogram, but no other fields.

There are two solutions that I see to this.
1. Fix HTSJDK MetricsFile.read() to deal with the case where a metric file contains no fields other than the histogram.
2. Add a dummy field, to be removed after HTSJDK is fixed.

yfarjoun:
OK. I would say add a dummy field.

fleharty:
@yfarjoun Thanks again for the review, back to you.

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -41,27 +40,28 @@
 
     private static final Log log = Log.getInstance(TheoreticalSensitivity.class);
     private static final int SAMPLING_MAX = 600; //prevent 'infinite' loops
-    private static final int MAX_CONSIDERED_DEPTH = 1000; //no point in looking any deeper than this, otherwise GC overhead is too high.
+    private static final int MAX_CONSIDERED_DEPTH = 10000; //no point in looking any deeper than this, otherwise GC overhead is too high.
```
yfarjoun: what if someone wants super low AF sensitivity? will the code still be able to estimate it? should this be a parameter rather than a static final?

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -143,7 +143,7 @@ We use an O(1) stochastic acceptance algorithm -- see Physica A, Volume 391, Pag
         private Random rng;
 
         RouletteWheel(final double[] weights) {
-            rng = new Random(51);
+            rng = new Random(randomSeed);
```
yfarjoun: why do you need to control the random seed?

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+    }
+
+    /**
+     * Draw from a binomial distribution.
+     * @param trials Number of trials to perform
+     * @param p Probability of individual success
+     * @param uniformRNG Random number generator to use for making draw
+     * @return Number of total successes
+     */
+    public static int binomialDraw(final int trials, final double p, final Random uniformRNG) {
```
yfarjoun: why not use apache.Commons.BinominalDistribution?

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+     * @param totalDepth Depth at the site to be called, both alt and ref.
+     * @param altDepth Number of alt bases at this site.
+     * @param averageQuality Average Phred-scaled quality of bases
+     * @param alleleFraction Allele fraction we are attempting to detect
+     * @param logOddsThreshold Log odds threshold necessary to exceed for variant to be called
+     * @return
+     */
+    public static boolean isCalled(int totalDepth, int altDepth, double averageQuality, double alleleFraction, double logOddsThreshold) {
+        double threshold;
+        double sumOfQualities = altDepth * averageQuality;
```
yfarjoun: is there a reason you want to pass in averageQuality and not sumOfQualities? that seems to be the original value and also the one you need...so why divide and then multiply?

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+        threshold = 10.0 * (altDepth * Math.log10(1.0 / alleleFraction) + (totalDepth - altDepth) * Math.log10(1.0 / (1.0 - alleleFraction)) + logOddsThreshold);
```
yfarjoun: instead of the 1/x you can just put a `-` in front of the log10

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+     * @param alleleFraction the allele fraction to evaluate sensitivity at
+     * @return
+     */
+    public static double theoreticalSensitivity(final double[] depthDistribution, final double[] qualityDistribution,
+                                                final int sampleSize, final double logOddsThreshold, final double alleleFraction) {
+        if (alleleFraction > 1.0 || alleleFraction < 0.0) {
+            throw new PicardException("Allele fractions must be between 0 and 1.");
+        }
+        double sensitivity = 0.0;
+        for (int k = 0; k < depthDistribution.length; k++) {
```
yfarjoun: if `k` is depth, then call it `depth`, not `k`.

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+        for (int i = 0;i <= endOfDistribution;i++) {
+            trimmedDistribution[i] = distribution[i];
+        }
+
+        return trimmedDistribution;
+    }
+
+    /**
+     * This is a utility function
+     * @param theoreticalSensitivityOutput File to save to results ot theoretical sensitivity.
```
yfarjoun: re-read/re-write javadoc

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+     * @param tsOut MetricsFile object to save results of theoretical sensitivity to.
+     * @param sampleSize Number of samples to take for each depth.
```
yfarjoun: the word `sample` is used in two different meaning in this javadoc...needs clarification

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+     * @param depthHistogram Histogram of depth distribution for sample.
+     * @param baseQHistogram Histogram of Phred-scaled quality scores.
+     * @param alleleFractions Allele fractions
```
yfarjoun: alleleFractions needs more detail on what the parameter is.

src/main/java/picard/analysis/TheoreticalSensitivityMetrics.java
```diff
@@ -0,0 +1,11 @@
...
+
+import htsjdk.samtools.metrics.MetricBase;
+
+/**
+ * Created by fleharty on 6/30/17.
+ */
+public class TheoreticalSensitivityMetrics extends MetricBase {
+    /** This is a dummy value that allows for this to be read properly from a file. */
+    public int DUMMY_FIELD = 0;
+}
```
yfarjoun: another newline in the end to make github happy

src/main/java/picard/analysis/TheoreticalSensitivity.java
yfarjoun: Given that you want to go to high depth, it would make sense to bin the depth distribution in a smart manner (that gives higher fidelity for small depths) and iterate over that binned distribution.

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+     * @param randomSeed random number seed to use for random number generator
+     * @return
+     */
+    public static double sensitivityAtConstantDepth(final int depth, final double[] qualityDistribution, final double logOddsThreshold, final int sampleSize, final double alleleFraction, final int randomSeed) {
+        final RouletteWheel qualityRW = new RouletteWheel(trimDistribution(qualityDistribution));
+        final Random uniformRNG = new Random(randomSeed);
+
+        int altDepth = 0;
+        int calledVariants = 0;
+        for (int k = 0; k < sampleSize; k++) {
```
yfarjoun: using some properties of the qDist, (mean and stdev) you should be able to figure out at what point (altDepth+alleleFraction) you can ignore the details of the distribution and use the meanQuality directly to approximate sumOfQualities

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+     * functions from evaluating in regions where the distribution has zero probability.
+     * @param distribution Distribution of base qualities
+     * @return Distribution of base qualities removing any trailing zeros
+     */
+    public static double[] trimDistribution(final double[] distribution) {
+        int endOfDistribution = 0;
+
+        // Locate the index of the distribution where all the values remaining at
+        // larger indices are zero.
+        for (endOfDistribution = distribution.length-1;endOfDistribution >= 0;endOfDistribution--) {
```
yfarjoun: spaces around `;` and `-`

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+            if (distribution[endOfDistribution] != 0) {
+                break;
+            }
+        }
+
+        // Remove trailing zeros.
+        final double[] trimmedDistribution = new double[endOfDistribution+1];
+        for (int i = 0;i <= endOfDistribution;i++) {
```
yfarjoun: spaces

src/test/java/picard/analysis/TheoreticalSensitivityTest.java
```diff
@@ -216,13 +216,118 @@ public void testHetSensTargeted(final double expected, final File metricsFile) t
...
+
+    @DataProvider(name = "TheoreticalSensitivityDataProvider")
+    public Object[][] arbFracSensDataProvider() {
+        final File wgsMetricsFile = new File(TEST_DIR, "test_Solexa-332667.wgs_metrics");
+
+        // This test acts primarily as an intergration test.  The sample size of 100
+        // is not quite large enough to converge properly, but is used for the purpose of
+        // keeping the compute time of the tests short.
+        return new Object[][] {
+                {0.90, wgsMetricsFile,       .5, 100},
```
yfarjoun: why so much space here?

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: This was put in as one way to test convergence in one of the tests.

src/main/java/picard/analysis/TheoreticalSensitivityMetrics.java
fleharty: fixed

fleharty:
@davidbenjamin Thanks for agreeing to review this.
If you would like for me to rebase or anything to make it easier for you to review let me know.

Also, I anticipate an objection from you on the issue of how it converges, it's super slow.  I would like to speed it up, and if you have thoughts on how best to do that, I would be open to making those changes.

fleharty:
@davidbenjamin 
Before you review, I'll address Yossi's comments, and then I'll send it back to you.

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: We discussed using a log distribution on depth for each sensitivityAtConstantDepth.  I think this is a sensible thing to do, but I think there are other (hopefully even better) ways to speed this up that I will discuss with @davidbenjamin

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+     * Determines if a variant would be called under the particular conditions of a given total depth, alt depth,
+     * average base qualities, allele fraction of variant and log odds threshold necessary to exceed to call variant.
+     * @param totalDepth Depth at the site to be called, both alt and ref.
+     * @param altDepth Number of alt bases at this site.
+     * @param averageQuality Average Phred-scaled quality of bases
+     * @param alleleFraction Allele fraction we are attempting to detect
+     * @param logOddsThreshold Log odds threshold necessary to exceed for variant to be called
+     * @return
+     */
+    public static boolean isCalled(int totalDepth, int altDepth, double averageQuality, double alleleFraction, double logOddsThreshold) {
```
davidbenjamin: What, in words, is the criterion for `isCalled`?  I'm asking because `allele fraction` here seems possibly different from the command line argument allele fraction.  The latter, I think, is the true allele fraction, which the former can't be because only the alt and total depths, which are subject to sampling error, are observed.

I could imagine various ways to define theoretical sensitivity with the allele fraction in this method being:

* an exact value.  `isCalled` is true if the likelihood ratio between a) het with exactly this allele fraction and b) hom ref  is sufficiently large.
* a minimum. `isCalled` is true if the likelihood of being a het with *at least* this allele fraction is sufficiently large.
* irrelevant. `isCalled` could use a criterion that marginalizes out the latent true allele fraction and does model comparison between a) hom ref and b) int_0^1 likelihood(af) P(af) d af.  This is what Mutect now does.
* irrelevant. `isCalled` plugs in af = alt / total in calculating the likelihood.  This is what Mutect used to do.

I'm not saying any one is more right than the others, (well, I guess that between the last two my preference is clear) but I want to know which one the code implements.

src/main/java/picard/analysis/TheoreticalSensitivity.java
davidbenjamin: Agreed.

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+        }
+        return successes;
+    }
+
+    /**
+     * Calculates the theoretical sensitivity with a given Phred-scaled quality score distribution at a constant
+     * depth.
+     * @param depth Depth to compute sensitivity at
+     * @param qualityDistribution Phred-scaled quality score distribution
+     * @param logOddsThreshold Log odd threshold necessary to exceed for variant to be called
```
davidbenjamin: Log odd*s*

davidbenjamin: And same as above about "necessary to exceed"

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+     * @param sampleSize sampleSize is the total number of simulations to run
+     * @param alleleFraction the allele fraction to evaluate sensitivity at
+     * @param randomSeed random number seed to use for random number generator
+     * @return
+     */
+    public static double sensitivityAtConstantDepth(final int depth, final double[] qualityDistribution, final double logOddsThreshold, final int sampleSize, final double alleleFraction, final int randomSeed) {
+        final RouletteWheel qualityRW = new RouletteWheel(trimDistribution(qualityDistribution));
+        final Random uniformRNG = new Random(randomSeed);
```
davidbenjamin: Utils.getRandomGenerator

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+
+        int altDepth = 0;
```
davidbenjamin: This variable is local to the `for` loop and should not be declared outside it.

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+        int calledVariants = 0;
+        for (int k = 0; k < sampleSize; k++) {
+            altDepth = binomialDraw(depth, alleleFraction, uniformRNG);
```
davidbenjamin: `final`

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+
+            int sumOfQualities = 0;
```
davidbenjamin: `final int sumOfQualities = IntStream.range(0, altDepth).map(n -> qualityRW.draw()).sum()`

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+     * @param depthDistribution Depth distribution to compute theoretical sensitivity over
+     * @param qualityDistribution Phred-scaled quality score distribution
+     * @param sampleSize the total number of simulations to run
+     * @param logOddsThreshold Log odds threshold necessary to exceed for variant to be called
+     * @param alleleFraction the allele fraction to evaluate sensitivity at
+     * @return
+     */
+    public static double theoreticalSensitivity(final double[] depthDistribution, final double[] qualityDistribution,
+                                                final int sampleSize, final double logOddsThreshold, final double alleleFraction) {
+        if (alleleFraction > 1.0 || alleleFraction < 0.0) {
```
davidbenjamin: `Utils.validateArg()`

src/main/java/picard/analysis/TheoreticalSensitivity.java
davidbenjamin: Let's discuss these. . .

src/main/java/picard/analysis/TheoreticalSensitivity.java
davidbenjamin: Use `Arrays.copyOfRange()`

src/main/java/picard/analysis/TheoreticalSensitivity.java
davidbenjamin: Elsewhere you use `simulations`, which would seem to work here, too.

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+     * @param theoreticalSensitivityOutput File to save to results ot theoretical sensitivity.
+     * @param tsOut MetricsFile object to save results of theoretical sensitivity to.
+     * @param sampleSize Number of samples to take for each depth.
+     * @param depthHistogram Histogram of depth distribution for sample.
+     * @param baseQHistogram Histogram of Phred-scaled quality scores.
+     * @param alleleFractions Allele fractions
+     */
+    public static void writeOutput(final File theoreticalSensitivityOutput, final MetricsFile<TheoreticalSensitivityMetrics, Double> tsOut, final int sampleSize,
+                                   final Histogram depthHistogram, final Histogram baseQHistogram, final List<Double> alleleFractions) {
+        if (theoreticalSensitivityOutput != null) {
```
davidbenjamin: To reduce nesting, use 
```java
if (theoreticalSensitivityOutput == null) {
    return;
}
```

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+            final double logOddsThreshold = 6.2; // This threshold is used because it is the value used for MuTect2.
```
davidbenjamin: But (see above) is the criterion for making a call the same as Mutect's?

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+            final double[] depthDoubleArray = TheoreticalSensitivity.normalizeHistogram(depthHistogram);
```
davidbenjamin: `depthDistribution` or `depthPdf` would be a more descriptive name.

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+            final double[] baseQDoubleArray = TheoreticalSensitivity.normalizeHistogram(baseQHistogram);
```
davidbenjamin: Likewise.  `baseQualityDistribution`.

src/main/java/picard/analysis/TheoreticalSensitivity.java
davidbenjamin: Is `normalizeHistogram` any different from `MathUtils.normalizeFromRealSpace`?

src/test/java/picard/analysis/TheoreticalSensitivityTest.java
```diff
@@ -216,13 +216,118 @@ public void testHetSensTargeted(final double expected, final File metricsFile) t
...
+        final Histogram qualityHistogram = histograms.get(1);
+
+        final double[] qualityDistribution = TheoreticalSensitivity.normalizeHistogram(qualityHistogram);
+        final double[] depthDistribution = TheoreticalSensitivity.normalizeHistogram(depthHistogram);
+
+        final double resultFromTS = TheoreticalSensitivity.theoreticalSensitivity(depthDistribution, qualityDistribution, sampleSize, 3, 0.5);
+        final double resultFromTHS = TheoreticalSensitivity.hetSNPSensitivity(depthDistribution, qualityDistribution, sampleSize, 3);
+
+        Assert.assertEquals(resultFromTHS, resultFromTS, tolerance);
+    }
```
davidbenjamin: I would like to see one unit test with a back-of-the-envelope result.

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: I suspect you are referring to the Utils in GATK4.  Unfortunately this is Picard :(

davidbenjamin: Whoops, I thought this was the picard package within the GATK.

src/main/java/picard/analysis/TheoreticalSensitivity.java
davidbenjamin: If Picard has such a thing.

src/main/java/picard/analysis/TheoreticalSensitivity.java
davidbenjamin: If Picard has such a thing.

yfarjoun:
this PR needs a rebase and please respond to review comments....

fleharty:
@yfarjoun Noted.  The algorithm as it is is too slow.  I need to make a few improvements to speed it up.

yfarjoun:
bump!

src/main/java/picard/analysis/TheoreticalSensitivityMetrics.java
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: That's true, but this would be an additional approximation to speed things up when the runtime isn't very long now.  I think if we find speed to be an issue I'd be happy to revisit this.

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: fixed

src/test/java/picard/analysis/TheoreticalSensitivityTest.java
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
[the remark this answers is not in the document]
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -204,4 +204,174 @@ public int draw() {
...
+     * average base qualities, allele fraction of variant and log odds threshold necessary to exceed to call variant.
+     * @param totalDepth Depth at the site to be called, both alt and ref.
+     * @param altDepth Number of alt bases at this site.
+     * @param averageQuality Average Phred-scaled quality of bases
+     * @param alleleFraction Allele fraction we are attempting to detect
+     * @param logOddsThreshold Log odds threshold necessary to exceed for variant to be called
+     * @return
+     */
+    public static boolean isCalled(int totalDepth, int altDepth, double averageQuality, double alleleFraction, double logOddsThreshold) {
+        double threshold;
```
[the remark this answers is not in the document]
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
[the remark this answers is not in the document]
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: Honestly, I don't know, and I don't know to what you are referring to "see above"

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: fixed.

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: fixed

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: I can't find this function normalizeFromRealSpace.

PR #860, opened 2017-07-10
- oops broke the old option EXPECT_ALL_READ_GROUPS_TO_MATCH. Sorry.

yfarjoun:
In a previous change I mistakenly renamed option EXPECT_ALL_READ_GROUPS_TO_MATCH. This PR restores that option.

### DIFF
--- a/src/main/java/picard/fingerprint/CrosscheckReadGroupFingerprints.java
+++ b/src/main/java/picard/fingerprint/CrosscheckReadGroupFingerprints.java
@@ -60,6 +60,11 @@ public class CrosscheckReadGroupFingerprints extends CrosscheckFingerprints {
             "and print out a library x library matrix with LOD scores.")
     public boolean CROSSCHECK_LIBRARIES = false;
 
+    @Option(doc="Expect all read groups' fingerprints to match, irrespective of their sample names.  By default (with this value set to " +
+            "false), read groups with different sample names are expected to mismatch, and those with the same sample name are expected " +
+            "to match.", mutex = {"EXPECT_ALL_GROUPS_TO_MATCH"})
+    public boolean EXPECT_ALL_READ_GROUPS_TO_MATCH = false;
+
     @Override
     protected String[] customCommandLineValidation() {
         final List<String> errors = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


src/main/java/picard/fingerprint/CrosscheckReadGroupFingerprints.java
```diff
@@ -86,6 +91,8 @@
     @Override
     protected int doWork() {
 
+        if (EXPECT_ALL_READ_GROUPS_TO_MATCH) EXPECT_ALL_GROUPS_TO_MATCH = EXPECT_ALL_READ_GROUPS_TO_MATCH;
```
pshapiro4broad: Since this is java it should have `{` `}` around the body.

All the other options that migrated to the base class kept their original names. Was the original name incorrect? Or is it now no longer specific to read groups?

src/main/java/picard/fingerprint/CrosscheckReadGroupFingerprints.java
```diff
@@ -60,6 +60,11 @@
             "and print out a library x library matrix with LOD scores.")
     public boolean CROSSCHECK_LIBRARIES = false;
 
+    @Option(doc="Expect all read groups' fingerprints to match, irrespective of their sample names.  By default (with this value set to " +
+            "false), read groups with different sample names are expected to mismatch, and those with the same sample name are expected " +
+            "to match.", mutex = {"EXPECT_ALL_GROUPS_TO_MATCH"})
+    public boolean EXPECT_ALL_READ_GROUPS_TO_MATCH = false;
+
     @Override
     protected String[] customCommandLineValidation() {
```
pshapiro4broad: Is there a reason why this method wasn't also moved into `CrosscheckFingerprints`?

src/main/java/picard/fingerprint/CrosscheckReadGroupFingerprints.java
ronlevine: Should there be a test?

yfarjoun: I'm a little lothe to adding a test for a deprecated class...

yfarjoun: @pshapiro4broad it was originally correct, but once you allow "rollup" or "check_by" things that are not read-groups the logic changes...and so the name should reflect that.

yfarjoun:
I'm assuming the 👍 means I can merge once I responded to your comments and tests pass @pshapiro4broad .

pshapiro4broad:
> I'm assuming the 👍 means I can merge once I responded to your comments and tests pass @pshapiro4broad .

Yes,  that is what I meant by that

PR #862, opened 2017-07-10
Better error message when dictionaries differ in GatherVcfs.

nh13:


src/main/java/picard/vcf/GatherVcfs.java
```diff
@@ -106,7 +106,13 @@ private static void assertSameSamplesAndValidOrdering(final List<File> inputFile
 
         for (final File f : inputFiles) {
             final VCFFileReader in = new VCFFileReader(f, false);
-            dict.assertSameDictionary(in.getFileHeader().getSequenceDictionary());
+            try {
+                dict.assertSameDictionary(in.getFileHeader().getSequenceDictionary());
+            } catch (final AssertionError e) {
+                log.error("File #1: " + inputFiles.get(0));
```
pshapiro4broad: Wouldn't it be be better to create a new exception and add the desired text there? E.g.
```
throw new AssertionError(String.format("Dictionary '%s' didn't match dictionary '%s'" inputFiles.get(0), f), e);
```
It also doesn't seem correct to use `AssertionError` in non-test code but that is being done in a few different places, and would be more work to fix.

yfarjoun: I disagree with @pshapiro4broad as assertSameDictionary has information that we want to see (how the dictionaries differ) so this seems to be a reasonable way of getting both the information about the dictionaries and the files from whence the came. 👍

nh13: Happy to merge when I get a 👍  from @pshapiro4broad

pshapiro4broad: I'm not objecting to `assertSameDirectory()`, just its use of `AssertionError`. `AssertionError` in Java is intended to be used only if something 'impossible' happens. It's not supposed to be used for programs to report a validation error, which is what this is. But I can see that there is precedent for its use this way in the picard codebase.

What about my first comment, regarding throwing a more detailed message, instead of putting the file names in the error log? That would match the code below that uses `throw new IllegalArgumentException`. Using logs plus throw means that the file names might not come out next to the exception message.

All that being said, if neither of these points is compelling you have my 👍 to merge.

tfenne: Just for argument's sake ... the documentation for `AssertionError` says "Thrown to indicate that an assertion has failed.".  It says nothing about program bugs, or unreachable/impossible situations.

pshapiro4broad: I agree, the javadoc is not very helpful. There is more info in the Oracle doc [Programming With Assertions](https://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html) but it's a lot to read. The Guava doc [ConditionalFailuresExplained](https://github.com/google/guava/wiki/ConditionalFailuresExplained) is a good summary of the different exception types and when they should be used. Of course these are all just suggested conventions, since any programmer can do anything they like in their code base.

yfarjoun:
I think that the discussion has come to an end, and this can be merged! :shipit:

gbggrant:
@nh13 bump

PR #864, opened 2017-07-11
- fixed bug in Liftover pertaining to indels that straddle two "links…

### LINKED ISSUE #859: Updates to LiftoverVcf break on indel that span two chain blocks
thanks to the user who reported this error. 

The resulting error message looks like this:

```
Exception in thread "main" java.lang.IllegalStateException: BUG: GenomeLoc 1:1634405-1634409 has a size == 5 but the variation reference allele has length 2 this = [VC Unknown @ 1:1634405-1634409 Q. of type=INDEL
        at htsjdk.variant.variantcontext.VariantContext.validateStop(VariantContext.java:1327)
        at htsjdk.variant.variantcontext.VariantContext.validate(VariantContext.java:1294)
        at htsjdk.variant.variantcontext.VariantContext.<init>(VariantContext.java:401)
        at htsjdk.variant.variantcontext.VariantContextBuilder.make(VariantContextBuilder.java:494)
        at htsjdk.variant.variantcontext.VariantContextBuilder.make(VariantContextBuilder.java:488)
        at picard.vcf.LiftoverVcf.liftSnp(LiftoverVcf.java:359)
        at picard.vcf.LiftoverVcf.doWork(LiftoverVcf.java:245)
        at picard.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:205)
        at picard.cmdline.PicardCommandLine.instanceMain(PicardCommandLine.java:94)
        at edu.mit.broad.picard.cmdline.PicardPrivateCommandLine.main(PicardPrivateCommandLine.java:24)
```

I know what the problem is. I think I fixed it already, but I'm trying to build a test that covers this case.

yfarjoun:
fixes #859


src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -101,17 +101,24 @@
...
-     * Filter name to use when a target is lifted over, but the reference allele doens't match the new reference.
+     * Filter name to use when a target is lifted over, but the reference allele doesn't match the new reference.
      */
     public static final String FILTER_MISMATCHING_REF_ALLELE = "MismatchedRefAllele";
 
     /**
+     * Filter name to use when an indel cannot be lifted over since it straddles two links which means
+     * that it is unclear what are the right alleles to be used.
+     */
+    public static final String FILTER_INDEL_STRADDLES_TWO_LINKS = "IndelStraddlesMultipleLinks";
```
eitanbanks: My only major issue with this PR is that I haven't got a clue what this means.  What on earth is an indel that straddles multiple links?!  I'm thinking that it means that it overlaps multiple intervals from the chain file?  If so, I think that "links" is the wrong term here; better just to use intervals.  So I would say "IndelStraddlesMultipleChainIntervals"

eitanbanks: Or better yet: "IndelSpansMultipleChainIntervals"

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -210,16 +219,24 @@ protected int doWork() {
             ++total;
             final Interval source = new Interval(ctx.getContig(), ctx.getStart(), ctx.getEnd(), false, ctx.getContig() + ":" + ctx.getStart() + "-" + ctx.getEnd());
             final Interval target = liftOver.liftOver(source, LIFTOVER_MIN_MATCH);
+
+            if (target == null) {
+                rejectVariant(ctx, FILTER_NO_TARGET);
+                continue;
+            }
+            if (ctx.getReference().length() != target.length()){
```
eitanbanks: Is it always the case that this means that the indel spans multiple chain intervals?  I'm not sure I see why that is.  Could you please add a comment to the code explaining what you are testing for here?

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -210,16 +219,24 @@ protected int doWork() {
...
             // if the target is null OR (the target is reverse complemented AND the variant is a non-biallelic indel or mixed), then we cannot lift it over
-            if (target == null || (target.isNegativeStrand() && (ctx.isMixed() || (ctx.isIndel() && !ctx.isBiallelic())))) {
-                final String reason = (target == null) ? FILTER_NO_TARGET : FILTER_CANNOT_LIFTOVER_INDEL;
-                rejects.add(new VariantContextBuilder(ctx).filter(reason).make());
-                failedLiftover++;
+            if (target.isNegativeStrand() && (ctx.isMixed() || ctx.isIndel() && !ctx.isBiallelic())) {
+                rejectVariant(ctx, FILTER_CANNOT_LIFTOVER_INDEL);
+
             } else if (!refSeqs.containsKey(target.getContig())) {
-                rejects.add(new VariantContextBuilder(ctx).filter(FILTER_NO_TARGET).make());
```
eitanbanks: Just as a side point, maybe it's worth emitting a warning here?  Shouldn't this only happen when the chain file is bad?

src/main/java/picard/vcf/LiftoverVcf.java
yfarjoun: ok. intervals in a chain. sure.

src/main/java/picard/vcf/LiftoverVcf.java
yfarjoun: no, this is legit. it happens all the time when part of the source reference isn't represented in the target reference or when a variant overlaps the "deleted" part of an interval in a chain

PR #868, opened 2017-07-12
- yet another reverse-compatibility issue in CrosscheckReadgroupFingerprints

yfarjoun:
I got the logic of when to emit a matrix and when to emit a metric-file reversed (relative to the old way) this PR fixes that.


yfarjoun:
@gbggrant the tests were written with the wrong model in mind: that crosscheck by library and sample creates a metricFile where crosscheck by read-group creates a matrix, in reality the roles are reversed and so the tests need to be re-written. I'm happy to do that, but I don't want to do it in this PR, due to the time pressure.

gbggrant:
@yfarjoun ok - can you make a Jira to add the test back?

👍

PR #870, opened 2017-07-13
Fixed tests that were removed from CrosscheckReadgroupFingerprints

yfarjoun:
compare with previous version to see actual changes.

### DIFF
--- a/src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
+++ b/src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
@@ -21,37 +21,36 @@ import static picard.fingerprint.FingerprintIdDetails.multipleValuesString;
  */
 public class CrosscheckReadGroupFingerprintsTest {
 
-    private final static File TEST_DIR = new File("testdata/picard/fingerprint/");
-    private final static File HAPLOTYPE_MAP = new File(TEST_DIR, "Homo_sapiens_assembly19.haplotype_database.subset.txt");
+    private static final File TEST_DIR = new File("testdata/picard/fingerprint/");
+    private static final File HAPLOTYPE_MAP = new File(TEST_DIR, "Homo_sapiens_assembly19.haplotype_database.subset.txt");
 
-    static private final File NA12891_r1_sam = new File(TEST_DIR, "NA12891.over.fingerprints.r1.sam");
-    static private final File NA12891_r2_sam = new File(TEST_DIR, "NA12891.over.fingerprints.r2.sam");
+    private static final File NA12891_r1_sam = new File(TEST_DIR, "NA12891.over.fingerprints.r1.sam");
+    private static final File NA12891_r2_sam = new File(TEST_DIR, "NA12891.over.fingerprints.r2.sam");
 
     //this is a copy of a previous one, but with a different sample name
-    static private final File NA12891_named_NA12892_r1_sam = new File(TEST_DIR, "NA12891_named_NA12892.over.fingerprints.r1.sam");
+    private static final File NA12891_named_NA12892_r1_sam = new File(TEST_DIR, "NA12891_named_NA12892.over.fingerprints.r1.sam");
 
-    static private final File NA12892_r1_sam = new File(TEST_DIR, "NA12892.over.fingerprints.r1.sam");
-    static private final File NA12892_r2_sam = new File(TEST_DIR, "NA12892.over.fingerprints.r2.sam");
+    private static final File NA12892_r1_sam = new File(TEST_DIR, "NA12892.over.fingerprints.r1.sam");
+    private static final File NA12892_r2_sam = new File(TEST_DIR, "NA12892.over.fingerprints.r2.sam");
 
-    static private File NA12891_r1, NA12891_r2, NA12891_named_NA12892_r1, NA12892_r1, NA12892_r2;
+    private static File NA12891_r1, NA12891_r2, NA12891_named_NA12892_r1, NA12892_r1, NA12892_r2;
 
-    static private final int NA12891_r1_RGs = 27;
-    static private final int NA12891_r2_RGs = 26;
-    static private final int NA12892_r1_RGs = 25;
-    static private final int NA12892_r2_RGs = 26;
+    private static final int NA12891_r1_RGs = 27;
+    private static final int NA12891_r2_RGs = 26;
+    private static final int NA12892_r1_RGs = 25;
+    private static final int NA12892_r2_RGs = 26;
 
-    private static final Map<CrosscheckMetric.DataType, List<String>> lookupMap = new HashMap<>(4);
+    private final Map<CrosscheckMetric.DataType, List<String>> lookupMap = new HashMap<>(4);
 
     @BeforeTest
-    static public void setup() throws IOException {
+    public void setup() throws IOException {
         NA12891_r1 = SamTestUtils.createIndexedBam(NA12891_r1_sam, NA12891_r1_sam);
         NA12891_r2 = SamTestUtils.createIndexedBam(NA12891_r2_sam, NA12891_r2_sam);
         NA12891_named_NA12892_r1 = SamTestUtils.createIndexedBam(NA12891_named_NA12892_r1_sam, NA12891_named_NA12892_r1_sam);
         NA12892_r1 = SamTestUtils.createIndexedBam(NA12892_r1_sam, NA12892_r1_sam);
         NA12892_r2 = SamTestUtils.createIndexedBam(NA12892_r2_sam, NA12892_r2_sam);
 
-        lookupMap.put(CrosscheckMetric.DataType.FILE, new ArrayList<>());
-        lookupMap.get(CrosscheckMetric.DataType.FILE).addAll(Arrays.asList("LEFT_FILE", "RIGHT_FILE"));
+        lookupMap.put(CrosscheckMetric.DataType.FILE, new ArrayList<>(Arrays.asList("LEFT_FILE", "RIGHT_FILE")));
 
         lookupMap.put(CrosscheckMetric.DataType.SAMPLE, new ArrayList<>());
         lookupMap.get(CrosscheckMetric.DataType.SAMPLE).addAll(Arrays.asList("LEFT_SAMPLE", "RIGHT_SAMPLE"));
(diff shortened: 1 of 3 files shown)


src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
```diff
@@ -0,0 +1,312 @@
...
+import java.util.stream.Stream;
+
+import static picard.fingerprint.FingerprintIdDetails.multipleValuesString;
+
+/**
+ * Tests for CrosscheckReadGroupFingerprints
+ */
+public class CrosscheckReadGroupFingerprintsTest {
+
+    private final static File TEST_DIR = new File("testdata/picard/fingerprint/");
```
pshapiro4broad: Should be consistent here with the ordering of modifiers. I think the preferred order is `private final static`. There is an IDE setting to highlight this and fix it for you.

src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
```diff
@@ -0,0 +1,312 @@
...
+    static private final File NA12892_r2_sam = new File(TEST_DIR, "NA12892.over.fingerprints.r2.sam");
+
+    static private File NA12891_r1, NA12891_r2, NA12891_named_NA12892_r1, NA12892_r1, NA12892_r2;
+
+    static private final int NA12891_r1_RGs = 27;
+    static private final int NA12891_r2_RGs = 26;
+    static private final int NA12892_r1_RGs = 25;
+    static private final int NA12892_r2_RGs = 26;
+
+    private static final Map<CrosscheckMetric.DataType, List<String>> lookupMap = new HashMap<>(4);
```
pshapiro4broad: Should this really be static? It doesn't look it needs to be.

src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
```diff
@@ -0,0 +1,312 @@
...
+
+    @BeforeTest
+    static public void setup() throws IOException {
```
pshapiro4broad: `static` unnecessary here, you only need that on `@BeforeClass` and `@AfterClass`. The test class will have been instantiated before `@BeforeTest` is called.

src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
```diff
@@ -0,0 +1,312 @@
...
+        NA12891_r1 = SamTestUtils.createIndexedBam(NA12891_r1_sam, NA12891_r1_sam);
+        NA12891_r2 = SamTestUtils.createIndexedBam(NA12891_r2_sam, NA12891_r2_sam);
+        NA12891_named_NA12892_r1 = SamTestUtils.createIndexedBam(NA12891_named_NA12892_r1_sam, NA12891_named_NA12892_r1_sam);
+        NA12892_r1 = SamTestUtils.createIndexedBam(NA12892_r1_sam, NA12892_r1_sam);
+        NA12892_r2 = SamTestUtils.createIndexedBam(NA12892_r2_sam, NA12892_r2_sam);
+
+        lookupMap.put(CrosscheckMetric.DataType.FILE, new ArrayList<>());
+        lookupMap.get(CrosscheckMetric.DataType.FILE).addAll(Arrays.asList("LEFT_FILE", "RIGHT_FILE"));
```
pshapiro4broad: You don't need to do this in two steps, you could write
```
lookupMap.put(CrosscheckMetric.DataType.FILE, new ArrayList<>(Arrays.asList("LEFT_FILE", "RIGHT_FILE")));
```

If you don't need the list to be modifiable then you can skip the `new ArrayList<>()` constructor call.

src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
```diff
@@ -0,0 +1,312 @@
...
+    private void doTest(final String[] args, final File metrics, final int expectedRetVal, final int expectedNMetrics, final CrosscheckMetric.DataType expectedType, final boolean expectAllMatch) throws IOException {
+
+        final CrosscheckReadGroupFingerprints crossChecker = new CrosscheckReadGroupFingerprints();
+        Assert.assertEquals(crossChecker.instanceMain(args), expectedRetVal);
+
+        final MetricsFile<CrosscheckMetric, Comparable<?>> metricsOutput = new MetricsFile<>();
+        metricsOutput.read(new FileReader(metrics));
+
+        Assert.assertFalse(metricsOutput.getMetrics().stream()
+                .anyMatch(m -> m.DATA_TYPE != expectedType));
```
pshapiro4broad: IMO it would be clearer to write this using `assertTrue(...noneMatch(...))`

src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
```diff
@@ -0,0 +1,312 @@
...
+        Assert.assertEquals(metricsOutput.getMetrics().size(), expectedNMetrics);
+
+        // at the readgroup level things aren't always conclusive...
+        if (!metricsOutput.getMetrics().isEmpty() && expectedType != CrosscheckMetric.DataType.READGROUP) {
+            Assert.assertTrue(metricsOutput.getMetrics().stream()
+                    .anyMatch(m -> m.RESULT != CrosscheckMetric.FingerprintResult.INCONCLUSIVE));
+        }
+
+        //check that fields that should have an actual value, indeed do
+        for(final String fieldName : lookupMap.get(expectedType)) {
```
pshapiro4broad: missing space after `for`

src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
```diff
@@ -0,0 +1,312 @@
...
+            try {
+                final Field field = CrosscheckMetric.class.getField(fieldName);
+                Assert.assertTrue(metricsOutput.getMetrics().stream().allMatch(m -> {
+                    try {
+                        return field.get(m) != multipleValuesString && field.get(m) != null;
+                    } catch (IllegalAccessException e) {
+                        e.printStackTrace();
+                        return false;
```
pshapiro4broad: Since we know this is going to hit the assert above and throw, why not let the exception get thrown in this case? That would avoid the need to print the stack trace here. Same comment applies to line 240 below.

src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
```diff
@@ -0,0 +1,312 @@
...
+        Assert.assertTrue(metrics.canRead());
+
+        try (Stream<String> lines = Files.lines(metrics.toPath(), Charset.defaultCharset())) {
+            long numOfLines = lines.count() - 1; //subtract one to account for the column header
+            Assert.assertEquals(numOfLines, expectedNMetrics);
+        }
+
+        if (expectAllMatch) {
+            try (Stream<String> lines = Files.lines(metrics.toPath(), Charset.defaultCharset())) {
+                lines.skip(1).forEach(s->{
```
pshapiro4broad: missing spaces around `->` operator

src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
```diff
@@ -0,0 +1,312 @@
...
+                    final List<String> strings = Arrays.asList(s.split("\t"));
+                    strings.subList(1,strings.size()).forEach(str->Assert.assertTrue(Double.parseDouble(str)>0D, "expected positive value, found: " + str));
```
pshapiro4broad: Missing many spaces here, easier to just reformat in the IDE (command-option-L)

src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
yfarjoun: According to https://stackoverflow.com/questions/41177031/throwing-exception-inside-the-java-8-stream-foreach this cannot be done since the signature for the function in "allMatch" doesn't throw...

src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
```diff
@@ -0,0 +1,312 @@
...
+                Assert.assertTrue(metricsOutput.getMetrics().stream().allMatch(m -> {
+                    try {
+                        return field.get(m) != multipleValuesString && field.get(m) != null;
+                    } catch (IllegalAccessException e) {
+                        e.printStackTrace();
+                        return false;
+                    }
+                }));
+
+            } catch (NoSuchFieldException e) {
```
yfarjoun: this one I removed.

yfarjoun:
@pshapiro4broad back for review.

src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
pshapiro4broad: You could catch & rethrow a `RuntimeException` to handle that. But it looks like what you want to do here is fail as soon as any metric value is not correct. E.g.
```
for (final String fieldName: lookupMap.get(expectedType)) {
    final Field field = CrosscheckMetric.class.getField(fieldName);
    for (CrosscheckMetric metric : metricsOutput.getMetrics()) {
        Assert.assertNotNull(field.get(metric));
        Assert.assertNotEquals(field.get(metric), multipleValuesString);
    }
}
```
This will give you more precise information about the failure when it occurs as well.

src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
```diff
@@ -0,0 +1,312 @@
...
+import picard.vcf.SamTestUtils;
+
+import java.io.*;
+import java.lang.reflect.Field;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.util.*;
+import java.util.stream.Stream;
+
+import static picard.fingerprint.FingerprintIdDetails.multipleValuesString;
```
pshapiro4broad: static imports make code harder to read and should be avoided, IMO

yfarjoun:
I'm trying to convert this to a data-provider....so meta.

yfarjoun:
done.

src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
```diff
@@ -0,0 +1,300 @@
...
+
+/**
+ * Tests for CrosscheckReadGroupFingerprints
+ */
+public class CrosscheckReadGroupFingerprintsTest {
```
pshapiro4broad: Would be nice to avoid code duplication with `CrosscheckFingerprintsTest` if possible

yfarjoun: I kept it as is on purpose...the idea is that CrosscheckReadgroupFingerprints is deprecated, and so I revived the original CrosscheckReadgroupFingerprints tests as proof that the refactoring didn't change anything....once the deprecation becomes a removal, there will be no code-duplication...does that make sense?

pshapiro4broad: OK

PR #880, opened 2017-07-25
Revert default compression level to 5.

### LINKED ISSUE #879: New default compression level 1 is a bigger change than advertised
I expect that a lot of testing was done internally at Broad on the intel deflator/inflator but I'm having a hard time understanding the assertions that:

1. The intel deflator produces BAMs at compression=1 only "slightly larger" than the old default of 5
2. That reading BAMs compressed at level 1 is 3x faster than reading BAMs compressed at level 5.

The following test is 100% anecdotal, but highlights that the above is not universal.  I took a single BAM file that represents ~7000X coverage of a 100 gene panel, and used the latest release JAR of picard to run `SamFormatConverter` and emit BAMs at the default (`1`) compression level and again at `5`, both using the intel deflator.  The BAM at compression level 1 came out at 3,152,719,064 bytes; the one at level `5` came out at 1,811,428,073 bytes.  I.e. going from `5` to `1` produced a 74% increase in file size.

I then tried running `CollectInsertSizeMetrics`, thinking that is a Picard tool that decompresses the entire BAM but then does very little beyond that (just reads the isize field).  Running on the level `1` BAM was _slower_ than running on the level `5` BAM by about 13% (1.11 minutes vs. 0.98 minutes).  The test was run on a Mac with a local SSD; I would imagine results to be significantly worse if using any kind of network storage.  I'm wondering if the observed "3x faster" reading is just the amount of time spent in the decompression code and doesn't account for the time to read the bytes from disk?

The change in the default compression level should also, IMHO, have triggered a major version bump.  It's non-backwards compatible in that the results produced are significantly different, meanwhile this change is easy to miss since it won't cause any programs to outright fail (unless they run out of disk space).

How about changing the default back to `5` until this is better understood?

danxmoran:
### Description

Fixes #879.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/cmdline/CommandLineProgram.java
+++ b/src/main/java/picard/cmdline/CommandLineProgram.java
@@ -88,9 +88,7 @@ public abstract class CommandLineProgram {
     public ValidationStringency VALIDATION_STRINGENCY = ValidationStringency.DEFAULT_STRINGENCY;
 
     @Option(doc = "Compression level for all compressed files created (e.g. BAM and GELI).", common=true)
-    // Reading from a file compressed by the default IntelDeflater at level 1 is ~3x faster than reading from a file
-    // compressed at level 5 (the htsjdk default), and level 1 files are only slightly larger.
-    public int COMPRESSION_LEVEL = 1;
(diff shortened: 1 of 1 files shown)


ktibbett:
👍

PR #882, opened 2017-07-27
Add CLP for calculating Uq

jsotobroad:


src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -560,6 +560,22 @@ public static void fixNmMdAndUq(final SAMRecord record, final ReferenceSequenceF
             record.setAttribute(SAMTag.NM.name(), SequenceUtil.calculateSamNmTag(record, referenceBases, 0, isBisulfiteSequence));
         }
         if (record.getBaseQualities() != SAMRecord.NULL_QUALS) {
```
pshapiro4broad: This `if` is unnecessary since it's tested again inside `calculateUq()`

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -560,6 +560,22 @@ public static void fixNmMdAndUq(final SAMRecord record, final ReferenceSequenceF
...
+    /** Calculates and sets the UQ tag from the record and the reference
+     *
+     * @param record the record to be fixed
+     * @param refSeqWalker a ReferenceSequenceWalker that will be used to traverse the reference
+     * @param isBisulfiteSequence a flag indicating whether the sequence came from bisulfite-sequencing.
+     *
+     * No return value, modifies the provided record.
+     */
+    public static void calculateUq(final SAMRecord record, final ReferenceSequenceFileWalker refSeqWalker, final boolean isBisulfiteSequence) {
+        final byte[] referenceBases = refSeqWalker.get(refSeqWalker.getSequenceDictionary().getSequenceIndex(record.getReferenceName())).getBases();
```
pshapiro4broad: This should be moved inside the `if` to avoid unnecessary work when the `if` is false.

src/main/java/picard/sam/CalculateUqTag.java
```diff
@@ -0,0 +1,79 @@
...
+    protected int doWork() {
+        validateInputs();
+
+        final SamReader reader = createSamReader();
+
+        final SAMFileWriter writer = createSamFileWriter(reader, log);
+
+        final ReferenceSequenceFileWalker refSeq = new ReferenceSequenceFileWalker(REFERENCE_SEQUENCE);
+
+        StreamSupport.stream(reader.spliterator(),false)
```
pshapiro4broad: Need spaces here, after `,` and around `->`, etc. May be easiest to just reformat in the IDE.

src/main/java/picard/sam/CalculateUqTag.java
```diff
@@ -0,0 +1,79 @@
...
+            "<pre>" +
+            "java -jar picard.jar CalculateUqTag \\<br />" +
+            "      I=sorted.bam \\<br />" +
+            "      O=UQtagged.bam \\<br />"+
+            "</pre>" +
+            "<hr />";
+
+    private final Log log = Log.getInstance(CalculateUqTag.class);
+
+    public static void main(final String[] argv) { new CalculateUqTag().instanceMainWithExit(argv); }
```
pshapiro4broad: format methods using multiple lines
```
public static void main(final String[] argv) {
    new CalculateUqTag().instanceMainWithExit(argv);
}
```

src/main/java/picard/sam/SetNmMdAndUqTags.java
```diff
@@ -110,4 +104,25 @@ protected int doWork() {
         writer.close();
         return 0;
     }
+
+    protected void validateInputs(){
```
pshapiro4broad: need space after `)` for all methods

src/test/java/picard/sam/SetNmMdAndUqTagsTest.java
```diff
@@ -30,16 +30,32 @@ public void TestValidSort(final File input, final File reference) throws IOExcep
 
         sort(input, sortOutput);
         fixFile(sortOutput, fixOutput, reference);
-        validate(fixOutput,validateOutput, reference);
+        validate(fixOutput,validateOutput, reference, false);
     }
 
-    private void validate(final File input, final File output, final File reference) {
+    @Test(dataProvider = "filesToFix")
+    public void TestUQValidSort(final File input, final File reference) throws IOException {
```
pshapiro4broad: methods start with a lowercase letter, e.g. `testUQValidSort()`

src/test/java/picard/sam/SetNmMdAndUqTagsTest.java
```diff
@@ -67,4 +83,15 @@ private void fixFile(final File input, final File output, final File reference)
         SetNmMdAndUqTags setNmMdAndUqTags = new SetNmMdAndUqTags();
         Assert.assertEquals(setNmMdAndUqTags.instanceMain(args), 0, "Fix did not succeed");
     }
+
+    private void calcUQ(final File input, final File output, final File reference) throws IOException {
+
+        final String[] args = new String[] {
```
pshapiro4broad: `new String[]` unnecessary here

src/test/java/picard/sam/SetNmMdAndUqTagsTest.java
```diff
@@ -67,4 +83,15 @@ private void fixFile(final File input, final File output, final File reference)
...
+                "INPUT="+input,
```
pshapiro4broad: need spaces around `+`

src/test/java/picard/sam/SetNmMdAndUqTagsTest.java
```diff
@@ -30,16 +30,32 @@ public void TestValidSort(final File input, final File reference) throws IOExcep
...
+        sortOutput.deleteOnExit();
+        final File fixOutput = File.createTempFile("Fix", ".bam");
+        fixOutput.deleteOnExit();
+        final File validateOutput = File.createTempFile("Sort", ".validation_report");
+        validateOutput.deleteOnExit();
+
+        sort(input, sortOutput);
+        calcUQ(sortOutput, fixOutput, reference);
+        //ignore warnings because a bam with no NM tag throws warnings
+        validate(fixOutput,validateOutput, reference, true);
```
pshapiro4broad: missing space after `,`

src/test/java/picard/sam/SetNmMdAndUqTagsTest.java
```diff
@@ -67,4 +83,15 @@ private void fixFile(final File input, final File output, final File reference)
...
+
+    private void calcUQ(final File input, final File output, final File reference) throws IOException {
+
+        final String[] args = new String[] {
+                "INPUT="+input,
+                "OUTPUT="+output,
+                "REFERENCE_SEQUENCE="+reference };
+
+        CalculateUqTag calculateUqTag = new CalculateUqTag();
+        Assert.assertEquals(calculateUqTag.instanceMain(args), 0, "Fix did not succeed");
```
pshapiro4broad: IMO assert text like this doesn't add value. In this case the text doesn't seem to match the operation; even if it did, it doesn't tell us much about the actual failure.

src/main/java/picard/sam/CalculateUqTag.java
```diff
@@ -0,0 +1,79 @@
...
+        validateInputs();
+
+        final SamReader reader = createSamReader();
+
+        final SAMFileWriter writer = createSamFileWriter(reader, log);
+
+        final ReferenceSequenceFileWalker refSeq = new ReferenceSequenceFileWalker(REFERENCE_SEQUENCE);
+
+        StreamSupport.stream(reader.spliterator(),false)
+                .peek(rec->{if(!rec.getReadUnmappedFlag()) AbstractAlignmentMerger.calculateUq(rec, refSeq, IS_BISULFITE_SEQUENCE);})
```
pshapiro4broad: Need to format `if` on different lines and use `{` `}` around the body.

src/main/java/picard/sam/CalculateUqTag.java
jsotobroad: intellij tricked me!

src/main/java/picard/sam/CalculateUqTag.java
jsotobroad: I'm not sure if I agree with this, I think it just adds characters that you don't need and makes it harder to read. I did change some of the spacing though

jsotobroad:
so bad at formatting..... 😞

src/main/java/picard/sam/CalculateUqTag.java
pshapiro4broad: Yeah intellij folds the text so it looks like it's a single line in the editor. sneaky.

src/main/java/picard/sam/SetNmMdAndUqTags.java
```diff
@@ -96,21 +96,21 @@ protected int doWork() {
 
         final ReferenceSequenceFileWalker refSeq = new ReferenceSequenceFileWalker(REFERENCE_SEQUENCE);
 
-        StreamSupport.stream(reader.spliterator(),false)
-                .peek(rec->{if(!rec.getReadUnmappedFlag()) AbstractAlignmentMerger.fixNmMdAndUq(rec, refSeq, IS_BISULFITE_SEQUENCE);})
+        StreamSupport.stream(reader.spliterator(), false)
+                .peek(rec -> {if (!rec.getReadUnmappedFlag()) AbstractAlignmentMerger.fixNmMdAndUq(rec, refSeq, IS_BISULFITE_SEQUENCE);})
```
pshapiro4broad: Everything in `SetNmMdAndUqTags.doWork()` and `CalculateUqTag.doWork()` could be refactored to a single method that accepts a `AbstractAlignmentMerger.fixNmMdAndUq()` or `AbstractAlignmentMerger.calculateUq()` method since they both have the same signature.

nh13:
Can you have the CLP named consistently with other similar tools?  Why can't you just use `SetNmMdAndUqTags`?

jsotobroad:
Because its called `SetNmM`AND`UqTags`.  I didnt think it would make sense for a clp named that to only do one of them.  I can change it to `SetUqTag` (it was named that originally fyi :P)

nh13:
My two cents is that having a tool for every variation of setting a tag is a bit overkill. Could you modify the original tool to have "tags to set" or something like that? Can other folks chime in as while the tool is valid this feels like a slippery slope to having the kitchen sink in Picard.

jsotobroad:
I actually don't mind bring it back into the original CLP at all.  I was debating back and forth for a while about what makes the most sense between basically your idea and what I have in this PR.

vdauwera:
I would favor going for a single multifunctional tool rather than many one-trick tools. Keeping the number of tools down helps keep maintenance and support burden down too.

Issue #902, opened 2017-08-09
Speeding up MarkDuplicates by not modifying per read PG tag


jsotobroad:
We want to speed up MarkDuplicates by not having it write out a PG tag for every read in the bam file.  If we no longer are writing out this tag then on the 2nd pass we can turn off the `EAGERLY_DECODE` option [here](https://github.com/broadinstitute/picard/blob/master/src/main/java/picard/sam/markduplicates/util/AbstractMarkDuplicatesCommandLineProgram.java#L210) (changes only turn this off for 2nd pass) when reading the SAM file and this gives a ~16% speedup on WGS data.  This is due to the fact that if we no longer are modifying the PG tag then all we do with the record on the 2nd pass is look at the RG tag for every read so we save time by not decoding the whole record which makes writing the file out is much faster.

Some runtimes before and after the changes.  Runtime is in minutes

Still writing PG tag with `EAGERLY_DECODE` enabled on both passes 
![image](https://user-images.githubusercontent.com/13023616/29133770-88910294-7d02-11e7-8315-31cfc179e201.png)

Not Writing PG tag with `EAGERLY_DECODE` on second pass disabled
![image](https://user-images.githubusercontent.com/13023616/29133854-cba451bc-7d02-11e7-8e05-67d194d0d650.png)

These were all run on HDD,  you do see further gains with SSD but thats outside the realm of the proposed changes.

Some questions to answer is if this is an acceptable change in general to MD (and MBA as far as writing PG tags to every read).  If this is an acceptable change, should this be the default mode for these programs in the future.

vdauwera:
I've always wondered why we tagged every read in this way, considering we don't do it with any other tools that modify reads (afaik). I would support getting rid of it as long as we keep a PG tag in the header to tell us the data has been deduped.

yfarjoun:
I'm for it. :+1:

Also would be nice if

1. MBA would leave it's own PG record (not only that of the aligner) so that the arguments of that swiss-army-knife of sequencing tools can be recorded somewhere (not so easy)
2. PG records are chained using the PP tag
3. MBA also drops the PG tag.

magicDGS:
This can be closed after #907, no?

PR #907, opened 2017-08-14
Add PG tag to every read option

jsotobroad:
Adding the option to not write out PG tags for MarkDuplicates and MergeBamAlignment.  Default behavior is kept the same as it was before but you can get speed increases if you set them appropriately.

### DIFF
--- a/src/main/java/picard/sam/AbstractAlignmentMerger.java
+++ b/src/main/java/picard/sam/AbstractAlignmentMerger.java
@@ -91,7 +91,7 @@ public abstract class AbstractAlignmentMerger {
     private boolean addMateCigar = false;
     private boolean unmapContaminantReads = false;
     private UnmappingReadStrategy unmappingReadsStrategy = UnmappingReadStrategy.DO_NOT_CHANGE;
-    private boolean addPGTagToReads = false;
+    private boolean addPGTagToReads = true;
 
 
     private final SamRecordFilter alignmentFilter = new SamRecordFilter() {
--- a/src/main/java/picard/sam/AbstractAlignmentMerger.java
+++ b/src/main/java/picard/sam/AbstractAlignmentMerger.java
@@ -26,7 +26,6 @@ package picard.sam;
 import htsjdk.samtools.*;
 import htsjdk.samtools.filter.FilteringSamIterator;
 import htsjdk.samtools.filter.SamRecordFilter;
-import htsjdk.samtools.reference.ReferenceSequence;
 import htsjdk.samtools.reference.ReferenceSequenceFileWalker;
 import htsjdk.samtools.SAMFileHeader.SortOrder;
 import htsjdk.samtools.util.*;
@@ -125,14 +124,22 @@ public abstract class AbstractAlignmentMerger {
 
         /** Adds a record to the sink. */
         void add(final SAMRecord rec) {
-            if (writer != null) writer.addAlignment(rec);
-            if (sorter != null) sorter.add(rec);
+            if (writer != null) {
+                writer.addAlignment(rec);
+            }
+            if (sorter != null) {
+                sorter.add(rec);
+            }
         }
 
         /** Closes the underlying resource. */
         void close() {
-            if (this.writer != null) this.writer.close();
-            if (this.sorter != null) this.sorter.doneAdding();
+            if (this.writer != null) {
+                this.writer.close();
+            }
+            if (this.sorter != null) {
+                this.sorter.doneAdding();
+            }
         }
     }
 
@@ -315,7 +322,10 @@ public abstract class AbstractAlignmentMerger {
         this.maxRecordsInRam = maxRecordsInRam;
     }
 
-    /** Allows the caller to decide whether or not to write the PG tag to reads when applicable */
+    /**
+     * Set addPGTagToReads. If true, the PG will be added to reads when applicable. If false, the PG tag will not be added.
+     * Default is true
+     */
     public void setAddPGTagToReads(final boolean addPGTagToReads) {
         this.addPGTagToReads = addPGTagToReads;
     }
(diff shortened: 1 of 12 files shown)


src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -314,6 +315,11 @@ public void setMaxRecordsInRam(final int maxRecordsInRam) {
         this.maxRecordsInRam = maxRecordsInRam;
     }
 
+    /** Allows the caller to decide whether or not to write the PG tag to reads when applicable */
```
pshapiro4broad: Where would be an appropriate place to document the default value of this property? Maybe here?

pshapiro4broad: For a docstring, I think it's better to describe behavior of the property, e.g. `Set addPGTagToReads. If true, the PG will be added to reads when ... If false, the PG tag will not be added.`

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -354,7 +357,7 @@ protected int doWork() {
                     rec.getReadName().equals(opticalDuplicateQueryName) ||
                     recordInFileIndex == nextOpticalDuplicateIndex;
 
-            rec.setAttribute(DUPLICATE_TYPE_TAG, null);
+            if (CLEAR_DT) rec.setAttribute(DUPLICATE_TYPE_TAG, null);
```
pshapiro4broad: I realize a lot of code here isn't written this way, but I think we should migrate to the Java coding standards in the picard code base. That means even a one-line `if` like this is written on two lines and has `{` `}` around its body.

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -392,7 +395,7 @@ protected int doWork() {
             if (this.REMOVE_DUPLICATES            && rec.getDuplicateReadFlag()) continue;
             if (this.REMOVE_SEQUENCING_DUPLICATES && isOpticalDuplicate)         continue;
 
-            if (PROGRAM_RECORD_ID != null)  rec.setAttribute(SAMTag.PG.name(), chainedPgIds.get(rec.getStringAttribute(SAMTag.PG.name())));
+            if (PROGRAM_RECORD_ID != null && ADD_PG_TAG_TO_READS) rec.setAttribute(SAMTag.PG.name(), chainedPgIds.get(rec.getStringAttribute(SAMTag.PG.name())));
```
pshapiro4broad: This `if` should also be written on two lines with `{ }`

src/main/java/picard/sam/util/ReadOutputCommandLineProgram.java
```diff
@@ -0,0 +1,14 @@
+package picard.sam.util;
+
+import org.broadinstitute.barclay.argparser.Argument;
+import picard.cmdline.CommandLineProgram;
+
+/**
+ * Abstract class that holds parameters and methods common to classes that want to add PG tags to reads in SAM/BAM files
+ *
+ */
+public abstract class ReadOutputCommandLineProgram extends CommandLineProgram {
```
pshapiro4broad: Is `ReadOutput` the best name for this class? Would a typical picard developer expect to see the `ADD_PG_TAG` argument in it? Or maybe there's no good name for this concept?

src/main/java/picard/sam/markduplicates/util/AbstractOpticalDuplicateFinderCommandLineProgram.java
```diff
@@ -27,14 +27,15 @@
...
+import picard.sam.util.ReadOutputCommandLineProgram;
 
 /**
  * Abstract class that holds parameters and methods common to classes that optical duplicate detection.  We put them here so that
  * the explanation about how read names are parsed is in once place
  *
  * @author Tim Fennell
  */
-public abstract class AbstractOpticalDuplicateFinderCommandLineProgram extends CommandLineProgram {
+public abstract class AbstractOpticalDuplicateFinderCommandLineProgram extends ReadOutputCommandLineProgram {
```
pshapiro4broad: This provides a new command line argument for all subclasses. Is that option applicable for all subclasses? Or are only some subclasses affected by this argument?

src/main/java/picard/sam/util/ReadOutputCommandLineProgram.java
jsotobroad: I wasn't sure what would be a good name either.  This is something I discussed with @jacarey as well

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
jsotobroad: 👍 I'll try to clean up wherever I see this if its something we want to move towards, all for it!

src/main/java/picard/sam/markduplicates/util/AbstractOpticalDuplicateFinderCommandLineProgram.java
jsotobroad: hmm `EstimateLibraryComplexity` inherits this but doesn't use it.  I cant see an obvious way to get only things that would use to inherit it but I'll see what I can do

RoelKluin:
Thank you for this improvement, however, there are several tools that could also benefit from this feature. For instance, MarkDuplicatesWithMateCigar could be improved similarly. Also, can not writing the PG tag to every read be the default please? Besides speed improvement, I observe 25% larger files after markduplicates, and this might be the culprit. Please reopen.

yfarjoun:
hi @RoelKluin,

you might want to either comment on the original issue (#902) or open a new issue. commenting on a merged PR will not have much effect otherwise. 

Alternatively, you are most welcome to submit a PR of your own, if you'd like!

Cheers,

Yossi.

PR #909, opened 2017-08-14
Add documentation creation and update (pushed to gh-pages)

jacarey:
Added document annotations.
Generate docs and upload to gh-pages in their respective gradle tasks.

#### Content

#### Review

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -360,3 +360,32 @@ uploadArchives {
         System.out.println("Uploading version $version")
     }
 }
+
+ext.htmlDir = new File("build/docs/html")
+
+//update static web docs
+task copyJavadoc(dependsOn: 'javadoc', type: Copy) {
+    from 'build/docs/javadoc'
+    into "$htmlDir/javadoc"
+}
+
+task copyPicardDoc(dependsOn: 'picardDoc', type: Copy){
+    from 'build/docs/picarddoc'
+    into "$htmlDir/_includes"
+}
+
+task updateGhPages(dependsOn: ['copyJavadoc', 'copyPicardDoc']){
+  outputs.dir htmlDir
+}
+
+updateGhPages.finalizedBy publishGhPages
+
+githubPages {
+  repoUri = 'git@github.com:broadinstitute/picard.git'
+  targetBranch = 'gh-pages'
+  deleteExistingFiles = false
+  pages {
+    from htmlDir
+    into '.'
+  }
+}
--- a/build.gradle
+++ b/build.gradle
@@ -371,7 +371,7 @@ task copyJavadoc(dependsOn: 'javadoc', type: Copy) {
 
 task copyPicardDoc(dependsOn: 'picardDoc', type: Copy){
     from 'build/docs/picarddoc'
-    into "$htmlDir/_includes"
+    into "$htmlDir/picarddoc"
 }
 
 task updateGhPages(dependsOn: ['copyJavadoc', 'copyPicardDoc']){
@@ -386,6 +386,6 @@ githubPages {
   deleteExistingFiles = false
   pages {
     from htmlDir
-    into '.'
+    into 'newdocs'
   }
 }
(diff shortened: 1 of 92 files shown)


build.gradle
```diff
@@ -360,3 +360,32 @@ uploadArchives {
...
+    from 'build/docs/javadoc'
+    into "$htmlDir/javadoc"
+}
+
+task copyPicardDoc(dependsOn: 'picardDoc', type: Copy){
+    from 'build/docs/picarddoc'
+    into "$htmlDir/picarddoc"
+}
+
+task updateGhPages(dependsOn: ['copyJavadoc', 'copyPicardDoc']){
```
pshapiro4broad: Would it be too much to spell out `github` here? I didn't know what `gh` was until I got to line 383.

build.gradle
```diff
@@ -360,3 +360,32 @@ uploadArchives {
...
+  outputs.dir htmlDir
+}
+
+updateGhPages.finalizedBy publishGhPages
```
pshapiro4broad: Where is `publishGhPages` defined?

pshapiro4broad:
Can you provide a link to the docs built using this script?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Mon Oct 24 16:22:03 EDT 2016
+#Mon Aug 14 16:27:51 EDT 2017
```
pshapiro4broad: Is this date updated by a script? Am not sure why we need a timestamp here.

src/main/java/picard/analysis/CollectBaseDistributionByCycle.java
```diff
@@ -38,6 +38,7 @@
 
 import htsjdk.samtools.util.StringUtil;
 import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.barclay.help.DocumentedFeature;
 import picard.PicardException;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
```
pshapiro4broad: Imports don't appear to be sorted in many of these files. This isn't the worst thing but it would be good to have them in canonical order to avoid unnecessary churn in the future.

vdauwera:
The annotation requirements are related to the new parser and doc system, Barclay. We have yet to do a pass to add the Barclay annotations to all the tools we want to document. Within Picard itself we don't currently have documentable things that are not both clp and documented feature, but we could in future like we do in GATK (and the two are designed to be compatible).

gradle/wrapper/gradle-wrapper.properties
jacarey: yeah updated by the wrapper

build.gradle
jacarey: in the sbt plugin https://github.com/ajoberstar/gradle-git

build.gradle
jacarey: I think it is because the branch is gh-pages.. but no reason we couldn't.

jacarey:
@pshapiro4broad Link to the generated docs https://github.com/broadinstitute/picard/tree/gh-pages/newdocs

also there as some CLPs that are deprecated that we aren't annotating.. but in general for picard we will document all clps

pshapiro4broad:
Sorry, my non-knowledge about github docs is getting in the way here. Can you provide a link to the new docs files that will render in my browser? I've never used gh-docs before so this is all new to me.

cmnbroad:
@jacarey @vdauwera I'm not sure what the plan for this is, but the documentation generated by this process is for the Barclay, posix-like syntax, which is not currently the Picard default.

vdauwera:
@cmnbroad can you remind me whether it's possible to emit the Picard style syntax in the docs? I completely forgot that rather important detail.

cmnbroad:
@vdauwera No, the whole doc system is build on the new parser/syntax. I think most of the difference is in the syntax change ("--INPUT" vs. "INPUT=") though. If thats the extent of it, it might be possible to use a FreeMarker macro that does the appropriate transform.

vdauwera:
Ah right, that does ring a bell. It is possible for us to build a Picard jar with new-style syntax, right? Even without doing the actual integration with GATK4?

cmnbroad:
@vdauwera Yes, the gradle build has a picardBarclayShadowJar task to build an uber-jar called picardBarclay.jar. We're not currently packaging/distributing it in the .zip/.tar though.

PR #929, opened 2017-09-18
LiftoverVcf Error

bbimber:
### Description

I hit the following error when running LiftoverVcf on a VCF file.  I assume this issue is due to the fact that while picard tries to update the alleles in genotypes for reverse complementation, it does not update the alleles in VariantContextBuilder to reflect any of those changes.  This PR applies similar logic to the set of Ref/Alt alleles as already happens for the genotype alleles.

The stacktrace for the error i found is below:

INFO    2017-09-18 13:10:34     LiftoverVcf     Loading up the target reference genome.
INFO    2017-09-18 13:10:51     LiftoverVcf     Lifting variants over and sorting.
INFO    2017-09-18 13:10:52     LiftOver        Interval chr01:32029-32050 failed to match chain 2 because intersection length 19 < minMatchSize 20.9 (0.8636364 < 0.95)
[Mon Sep 18 13:10:53 PDT 2017] picard.vcf.LiftoverVcf done. Elapsed time: 0.71 minutes.
srun: error: exanode-2-6: task 0: Exited with exit code 1
Runtime.totalMemory()=98784247808
To get help, see http://broadinstitute.github.io/picard/index.html#GettingHelp
Exception in thread "main" java.lang.IllegalStateException: Allele in genotype A* not in the variant context [T*, C]
        at htsjdk.variant.variantcontext.VariantContext.validateGenotypes(VariantContext.java:1360)
        at htsjdk.variant.variantcontext.VariantContext.validate(VariantContext.java:1298)
        at htsjdk.variant.variantcontext.VariantContext.<init>(VariantContext.java:401)
        at htsjdk.variant.variantcontext.VariantContextBuilder.make(VariantContextBuilder.java:494)
        at htsjdk.variant.variantcontext.VariantContextBuilder.make(VariantContextBuilder.java:488)
        at picard.vcf.LiftoverVcf.tryToAddVariant(LiftoverVcf.java:369)
        at picard.vcf.LiftoverVcf.doWork(LiftoverVcf.java:287)
        at picard.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:268)
        at picard.cmdline.PicardCommandLine.instanceMain(PicardCommandLine.java:98)
        at picard.cmdline.PicardCommandLine.main(PicardCommandLine.java:108)

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/vcf/LiftoverVcf.java
+++ b/src/main/java/picard/vcf/LiftoverVcf.java
@@ -255,7 +255,10 @@ public class LiftoverVcf extends CommandLineProgram {
             final ReferenceSequence refSeq;
 
             // if the target is null OR (the target is reverse complemented AND the variant is a non-biallelic indel or mixed), then we cannot lift it over
-            if (!refSeqs.containsKey(target.getContig())) {
+            if (target.isNegativeStrand() && (ctx.isMixed() || ctx.isIndel() && !ctx.isBiallelic())) {
+                rejectVariant(ctx, FILTER_CANNOT_LIFTOVER_INDEL);
+
+            } else if (!refSeqs.containsKey(target.getContig())) {
                 rejectVariant(ctx, FILTER_NO_TARGET);
 
                 final String missingContigMessage = "Encountered a contig, " + target.getContig() + " that is not part of the target reference.";
@@ -265,23 +268,19 @@ public class LiftoverVcf extends CommandLineProgram {
                     log.error(missingContigMessage);
                     return EXIT_CODE_WHEN_CONTIG_NOT_IN_REFERENCE;
                 }
-            } else if (target.isNegativeStrand()) {
-                if (ctx.isMixed() || (ctx.isIndel() && !ctx.isBiallelic())){
-                    rejectVariant(ctx, FILTER_CANNOT_LIFTOVER_INDEL);
-                }
-                else {
-                    refSeq = refSeqs.get(target.getContig());
-                    //flipping negative strand intervals:
-                    final VariantContext flippedVC = flipVC(ctx, refSeq, target);
-                    if (flippedVC == null) {
-                        throw new IllegalArgumentException("Unexpectedly found null VC. This should have not happened.");
-                    } else {
-                        tryToAddVariant(flippedVC, refSeq, ctx);
-                    }
+            } else if (target.isNegativeStrand() && ctx.isIndel() && ctx.isBiallelic()) {
+                refSeq = refSeqs.get(target.getContig());
+                //flipping indels:
+
+                final VariantContext flippedIndel = flipIndel(ctx, refSeq, target, WRITE_ORIGINAL_POSITION);
+                if (flippedIndel == null) {
+                    throw new IllegalArgumentException("Unexpectedly found null VC. This should have not happened.");
+                } else {
+                    tryToAddVariant(flippedIndel, refSeq, ctx);
                 }
             } else {
                 refSeq = refSeqs.get(target.getContig());
-                final VariantContext liftedVariant = liftSimpleVariant(ctx, target);
+                final VariantContext liftedVariant = liftSimpleVariant(ctx, target, WRITE_ORIGINAL_POSITION);
 
                 tryToAddVariant(liftedVariant, refSeq, ctx);
             }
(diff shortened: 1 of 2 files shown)


bbimber:
Update: I looked closer at the original LiftoverVcf code and I think there were actually some big problems with existing behavior.  These are:

1) It tried to reverse complement indels if the liftover is not he negative strand, but does not do this for simple SNVs.
2) I dont understand the intent of the essentially global reverseComplementAlleleMap.  Perhaps the intention was to avoid re-calculating reverse complements; however, I think the effect of storing those results across all positions, without any accounting for if a position is actually reverse complemented, is that it could be relatively easy to inappropriately convert genotypes in fixGenotypes.  

This updated pull request should feed all indels and SNVs lifted to the negative strand through one codepath to do the reverse complement, and do any fixing of alleles and genotypes in one place.  We no longer need to do this in tryToAddVariant().

I apparently cant see the TravisCI results, so I cant tell what any errors there might be.

yfarjoun:
Hi @bbimber could you give an example of what variant doesn't lift properly? 
I agree that there seems to be some problems with the implementation, but I suspect that you might have missed a few subtle points. 

For example, SNPs (forward and reverse) are lifted over here: picard/vcf/LiftoverVcf.java:285

Regarding the reverse-complementing map, I agree that there's something funny going on, and I'm looking into it, but it would be good to increase our arsenal of tests. perhaps you could first add a test (disabled) which currently fails to liftover?

yfarjoun:
I now see that the tests are not covering lifting-over VCs that have genotypes...which is where you must be having trouble. I'll add a few tests and submit a PR.

yfarjoun:
I wrote a test that I thought would cover your use-case and it works...it seems (from the line numbers in your stack-trace that you are using an old version. Could you try again with the current version and provide a stack-trace if that fails again?

Since it is still unclear what the problem is (if any) I am closing this PR. please open an issue to discuss the problems you are seeing.

bbimber:
I apologize for the series of 5 commits and comments here, and I was also incorrect in my comment in the middle of this thread about the current implementation not covering reverse complementation of simple SNVs (which I edited a few hours before you rely, though your reply might have been based on it).  Yes, the HTSJDK portion of the stack is out of date (though not picard); however, it's pretty easy to see what code is being hit in VariantContext.  

The core of the problem is that while reverse complemented indels update both the alleles and genotypes, liftSimpleVariant() will reverse complement the alleles and update these.  fixGenotypes() is called downstream; however, liftSimpleVariant() provides no information that could be used here.  

This PR (including the 5 commits.  again, unclear when you saw all), fixes that, and also does a few other code clean-ups for there is consistent transfer of information from the source VC to the lifted one for both indels and SNVs.  

It is worth writing tests, and I agree that the current tests dont catch this because a) most dont include genotypes which are required to hit this, and b) most unit tests wouldnt hit this since they dont cover the hand-off from liftSimpleVariant() to tryToAddVariant().  

I am happy to do that, but before I do, is there any process to port changes here to GATK4?  If not, is this a bad use of time and should I put in a PR on GATK's tool?

yfarjoun:
Gatk will not have a separate implementation of lift over.

Can you look into my previous that puts in tests and see if it solves your
issue?

bbimber:
Isnt GATK4 replacing both GATK3 and Picard?  I think LiftoverVcf is in fact ported (and out of date) in GATK4's repo:

https://github.com/broadinstitute/gatk/blob/dd7e19a58fece8d165f5f8d2d17f88ad3ddf2666/src/main/java/org/broadinstitute/hellbender/tools/picard/vcf/LiftOverVcf.java

There must be some process in place to handle changes in picard and porting to GATK4, yes?  Porting changes within LiftoverVcf shouldnt be that big a deal since the classes are similar, but since the tests are more distinct it seems like it's worth just doing that direct on GATK4, rather than in picard.

I put a couple comments on your other PR.  That does less to consolidate between the SNV and indel codepaths, but should fix the issue that prompted this.

cmnbroad:
@bimber The GATK versions of LiftOverVcf and other Picard tools will soon be removed and GATK will delegate to the actual Picard code (via the Picard jar), so there is no need to update the GATK version of LiftOver.

bbimber:
I realize we're getting off topic, but how does this fit into the idea of GATK4 unifying GATK and picard's code bases?  this has been a big part of the messaging around GATK4.  

Will GATK4 and Picard each depend on some kind of core project which includes the lower level code, and each project produce a separate JAR with implementations of the various tools?

vdauwera:
GATK4 will take in Picard as a library and expose the tools alongside the GATK tools. Over time some tools may migrate over to the GATK codebase if there is a functional advantage to do so. We had initially envisioned a complete merge of the codebases but that raised some concerns, mostly related to codebase complexity and consequences for the Picard developer community. So we've opted for this less dramatic library strategy to minimize disruption for now.

PR #931, opened 2017-09-19
Added code that tests all the @DataProviders in the package

yfarjoun:
And makes sure that the do not error out. This will avoid scenarios where due to some change a data-provider fails to complete and the tests are silently skipped.

Found two cases of methods that were "marked" as data-providers, but were not, so I removed their annotation.

----

#### Content

#### Review

### DIFF
--- a/src/test/java/picard/TestDataProviders.java
+++ b/src/test/java/picard/TestDataProviders.java
@@ -3,88 +3,33 @@ package picard;
 import org.testng.Assert;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
+import picard.cmdline.ClassFinder;
 
-import java.io.File;
-import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.net.URL;
-import java.util.ArrayList;
-import java.util.Enumeration;
-import java.util.List;
 
 /**
- * Created by farjoun on 9/19/17.
+ * This test is a mechanism to check that none of the data-providers fail to run.
+ * It is needed because in the case that a data-provider fails (for some reason, perhaps a change in some other code
+ * causes it to throw an exception) the tests that rely on it will be sliently skipped.
+ * The only mention of this will be in test logs but since we normally avoid reading these logs and rely on the
+ * exit code, it will look like all the tests have passed.
+ *
+ * @author Yossi Farjoun
  */
 public class TestDataProviders {
 
-
-    /**
-     * Taken from: https://stackoverflow.com/a/862130/360496
-     *
-     * Scans all classes accessible from the context class loader which belong
-     * to the given package and subpackages.
-     *
-     * @param packageName The base package
-     * @return The classes
-     * @throws ClassNotFoundException
-     * @throws IOException
-     */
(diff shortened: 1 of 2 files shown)


src/test/java/picard/TestDataProviders.java
```diff
@@ -0,0 +1,96 @@
...
+                classes.addAll(findClasses(file, packageName + "." + file.getName()));
+            } else if (file.getName().endsWith(".class")) {
+                classes.add(Class.forName(packageName + '.' + file.getName().substring(0, file.getName().length() - 6)));
+            }
+        }
+        return classes;
+    }
+
+    @Test
+    public void testAllDataProviders() throws URISyntaxException, IOException, ClassNotFoundException, IllegalAccessException, InstantiationException, InvocationTargetException {
```
pshapiro4broad: What is the purpose of this test? It doesn't seem like a test at all. Maybe if you asserted that the signatures of the provider methods are correct, that would make it a useful test.

src/test/java/picard/TestDataProviders.java
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * Recursive method used to find all classes in a given directory and
+     * subdirs.
+     *
+     * @param directory   The base directory
+     * @param packageName The package name for classes found inside the base directory
+     * @return The classes
+     * @throws ClassNotFoundException
+     */
+    private List<Class> findClasses(File directory, String packageName) throws ClassNotFoundException {
```
pshapiro4broad: Can this be done using `ClassFinder` to avoid code duplication?

yfarjoun:
It tests that all the providers do not throw an exception...

pshapiro4broad:
> It tests that all the providers do not throw an exception

If a provider throws an exception, and is used by a test, then running all the tests would find that, wouldn't it?

yfarjoun:
Nope. if  provider throws an exception that test is "skipped". (try it!)

pshapiro4broad:
> if  provider throws an exception that test is "skipped".

I believe you. That's pretty bad. :(

src/test/java/picard/TestDataProviders.java
yfarjoun: oooh...thanks!

yfarjoun:
done. @pshapiro4broad back to you.

yfarjoun:
oh, you approved already, so I'll merge once tests pass.

lbergelson:
@yfarjoun This is a good idea, we should do something like this in GATK and htsjdk as well.

yfarjoun:
thanks! I'm happy to put it into htsjdk, you can do gatk...OK?

pshapiro4broad:
I wonder if JUnit has these kinds of problems. JUnit was bad for a while and TestNG picked up the slack; these days JUnit is getting better and better and TestNG seems to have stagnated.

magicDGS:
@pshapiro4broad - as far as I know, there is no concept of `@DataProvider` in JUnit, only parametrized test classes. There is a project for that (https://github.com/TNG/junit-dataprovider), but I've never use it...

pshapiro4broad:
I know JUnit 5 has an experimental feature [@ParameterizedTest](http://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests) which appears to be equivalent. Using the parameterized test classes would also work but that would require a lot of work to port existing tests.

magicDGS:
I didn't know about that experimental feature. Sounds really good, but it is true that porting the tests to JUnit5 will be really painful...

Issue #589, opened 2016-07-07
MergeBamAlignment creates invalid BAMs in the presence of secondary alignments


tfenne:
I just came across a very odd case, where a colleague was running `bwa mem` with the flags setup so that short supplementary reads were being output as _secondary_/_not primary_ alignments, and were getting hard clipped.  However, when running the resulting file through `MergeBamAlignment` the output of MergeBamAlignment was an invalid BAM file containing two primary records for each of R1 and R2.

I've reduced the problem down to a single read-pair unmapped sam file and bwa produced sam file, and also created variants with the read marked as supplementary and as secondary but soft-clipped instead of hard clipped.  The problem exhibits with both the soft and hard clipped secondary reads, but works fine with the secondary read.

My recommendation to my colleague is to change how `bwa` is being run to not output split reads as secondary, which will work around this, and so I'm not inclined to try and track this down and fix it myself.

Files are attached to reproduce the problem.  The should work against any `HG19` reference that has the core chromosomes and `chrM` in it.  Commands for merging are as follows (`picard` is just an alias ):

```
picard MergeBamAlignment UNMAPPED=unmapped.sam ALIGNED=bwa.hard_clipped_secondary.sam    R=hg19/hg19.fasta MAX_GAPS=-1 SO=queryname MC=true OUTPUT=merged.hard_clipped_secondary.sam   
picard MergeBamAlignment UNMAPPED=unmapped.sam ALIGNED=bwa.hard_clipped_supplemental.sam R=hg19/hg19.fasta MAX_GAPS=-1 SO=queryname MC=true OUTPUT=merged.hard_clipped_supplemental.sam
picard MergeBamAlignment UNMAPPED=unmapped.sam ALIGNED=bwa.soft_clipped_secondary.sam    R=hg19/hg19.fasta MAX_GAPS=-1 SO=queryname MC=true OUTPUT=merged.soft_clipped_secondary.sam   
```

[sam_files_to_recreate_issue.zip](https://github.com/broadinstitute/picard/files/352416/sam_files_to_recreate_issue.zip)

PR #932, opened 2017-09-19
fix: MergeBamAlignment creates invalid BAMs in the presence of second…

DenisVerkhoturov:
### Description
The problem originates in [`HitsForInserts#coordinateByHitIndex`](../blob/9a7ba13f2fad45d8bf48816c2bc926a3821243ca/src/main/java/picard/sam/HitsForInsert.java#L196).  This method makes decision whether read is paired or not on assumption that all the records have HI-tag but ignores other criteria such as **RNEXT** (*Ref. name of the mate/next read*) and **PNEXT** (*Position of the mate/next read*) [SAM](https://samtools.github.io/hts-specs/SAMv1.pdf) fields. So paired reads are determined as two unrelated ones in the case when HI-tag is absent.

----

## Implementation
* Added new method (`coorditanteByMate`) for correlating pair-aware alignments. It does not require presence of HI-tags and performs work based on the data contained in fields: **RNAME**, **RNEXT**, **POS**, **PNEXT**.
* Added `MergeBamAlignmentTest#testPairedReadsAreProcessedProperlyInAbsenceOfHitIndexTags` running [MergeBamAlignment](src/main/java/picard/sam/MergeBamAlignment.java) with SAM file contained pair-aware alignments without HI-tag. 
* Added `testCoordinateByMateMethodWorksProperly` for the new method.
* Changed `BestMapqPrimaryAlignmentSelectionStrategy`. It uses implemented method `coordinateByMate` instead of `coordinateByHitIndex` now. We kept the old method due to backward compatibility.

## Performance
All the benchmarks were measured by [JMH](http://openjdk.java.net/projects/code-tools/jmh/) (**Mode** - thrpt, **Cnt** - 100, **Units** - ops/s)
<table style="text-align: center;">
    <tr>
        <th colspan="2">Benchmark</th>
        <th colspan="2">100 paired alignments</th>
        <th colspan="2">500 paired alignments</th>
        <th colspan="2">10_000 paired alignments</th>
    </tr>
    <tr>
        <th>Method</th>
        <th>Case</th>
        <th>Score</th>
        <th>Error</th>
        <th>Score</th>
        <th>Error</th>
        <th>Score</th>
        <th>Error</th>
    </tr>
    <tr>
        <td rowspan="3">coordinateByHitIndex</td>
        <td>HI-tag and mate is absent</td>
        <td>272_334.682</td>
        <td>± 3852.273</td>
        <td>27_056.185</td>
        <td>± 543.235</td>
        <td>95.384</td>
        <td>± 0.354</td>
    </tr>
    <tr>
        <td>HI-tag is absent</td>
        <td>282_456.059</td>
        <td>± 5726.669</td>
        <td>35_603.756</td>
        <td>± 954.916</td>
        <td>696.495</td>
        <td>± 7.419</td>
    </tr>
    <tr>
        <td>HI-tag is present</td>
        <td>280_571.088</td>
        <td>± 1900.595</td>
        <td>27_734.340</td>
        <td>± 914.012</td>
        <td>95.101</td>
        <td>± 0.935</td>
    </tr>
    <tr>
        <td rowspan="3">coordinateByMate</td>
        <td>HI-tag and mate is absent</td>
        <td>180_912.877</td>
        <td>± 1983.565</td>
        <td>17_965.070</td>
        <td>± 242.398</td>
        <td>55.060</td>
        <td>± 2.967</td>
    </tr>
    <tr>
        <td>HI-tag is absent</td>
        <td>174_695.474</td>
        <td>± 1786.275</td>
        <td>17_139.180</td>
        <td>± 128.948</td>
        <td>51.079</td>
        <td>± 0.542</td>
    </tr>
    <tr>
        <td>HI-tag is present</td>
        <td>185_531.494</td>
        <td>± 1975.027</td>
        <td>18_120.429</td>
        <td>± 157.350</td>
        <td>57.640</td>
        <td>± 3.658</td>
    </tr>
</table>
New method is 1.7 times slower but this caused by using required fields that allow rearranging alignments in any situations. And the only situation `coordinateByIndex` is much faster is when we have no **Hi-tag** and pair-aware alignments are being rearranged incorrectly (case described in issue).

### Resolves
[MergeBamAlignment creates invalid BAMs in the presence of secondary alignments](https://github.com/broadinstitute/picard/issues/589)

#### Content

#### Review


src/main/java/picard/sam/HitsForInsert.java
```diff
@@ -194,45 +195,100 @@ public void setPrimaryAlignment(final int primaryAlignmentIndex) {
...
-            } else if (rightHi != null) {
-                firstOfPairOrFragment.add(i, null);
-            } else {
-                // Both alignments do not have HI, so push down the ones on the right.
-                // Right is arbitrarily picked to push down.
-                secondOfPair.add(i, null);
+            final Integer hiTagValueOfFirst = firstOfPairOrFragment.get(i).getIntegerAttribute(SAMTag.HI.name());
+            final Integer hiTagValueOfSecond = secondOfPair.get(i).getIntegerAttribute(SAMTag.HI.name());
+
+            if (nonNull(hiTagValueOfFirst) && nonNull(hiTagValueOfSecond)) {
```
pshapiro4broad: What is the rationale for using these functions instead of comparing against `null` directly? My understanding is that these methods are provided so we have convenient lambda forms for streaming operations, etc. Using them inline like this doesn't look like an improvement to me.

src/main/java/picard/sam/HitsForInsert.java
```diff
@@ -27,9 +27,12 @@
 import htsjdk.samtools.SAMTag;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Objects;
+
+import static java.util.Objects.isNull;
```
pshapiro4broad: In general static imports should be avoided as they make code harder to read.

src/main/java/picard/sam/HitsForInsert.java
```diff
@@ -194,45 +195,100 @@ public void setPrimaryAlignment(final int primaryAlignmentIndex) {
...
-                firstOfPairOrFragment.add(i, null);
-            } else {
-                // Both alignments do not have HI, so push down the ones on the right.
-                // Right is arbitrarily picked to push down.
-                secondOfPair.add(i, null);
+            final Integer hiTagValueOfFirst = firstOfPairOrFragment.get(i).getIntegerAttribute(SAMTag.HI.name());
+            final Integer hiTagValueOfSecond = secondOfPair.get(i).getIntegerAttribute(SAMTag.HI.name());
+
+            if (nonNull(hiTagValueOfFirst) && nonNull(hiTagValueOfSecond)) {
+                if (hiTagValueOfFirst < hiTagValueOfSecond) secondOfPair.add(i, null);
```
pshapiro4broad: The original code didn't do this, but the new code should use `{` `}` around all if/else body statements, even if they're a single line.

src/main/java/picard/sam/HitsForInsert.java
```diff
@@ -194,45 +195,100 @@ public void setPrimaryAlignment(final int primaryAlignmentIndex) {
...
+    }
+
+    /**
+     * @return true if records belong to the same pairwise alignment
+     */
+    private static boolean arePair(SAMRecord first, SAMRecord second) {
+        return nonNull(first) && nonNull(second)
+                && first.getMateAlignmentStart() == second.getAlignmentStart()
+                && first.getAlignmentStart() == second.getMateAlignmentStart()
+                && Objects.equals(first.getReferenceName(), second.getMateReferenceName())
```
pshapiro4broad: Why use `Object.equals()` here? is it possible that one of these is `null`?

src/main/java/picard/sam/HitsForInsert.java
```diff
@@ -194,45 +195,100 @@ public void setPrimaryAlignment(final int primaryAlignmentIndex) {
...
+        secondOfPair.clear();
+        secondOfPair.addAll(newSecondOfPair);
+
+        renumberHitIndex();
+    }
+
+    /**
+     * @return true if records belong to the same pairwise alignment
+     */
+    private static boolean arePair(SAMRecord first, SAMRecord second) {
```
pshapiro4broad: This looks like it should be a non-static method on `SAMRecord`, as it only deals with fields in that object and takes that object as its argument.

src/main/java/picard/sam/HitsForInsert.java
DenisVerkhoturov: I prefer to use functional style instead of procedural even if it could decrease performance a little bit because it usually increases code readability. Especially for functions of Java standard library. I could revert all cases when I use static import if it is the concern. So is it critical?

src/main/java/picard/sam/HitsForInsert.java
DenisVerkhoturov: `SAMRecord` is part of `htsjdk` and this was the only constraint I haven't done it. Could I just make this function a `SamRecord#arePair` method and make associated pull-request to [`htsjdk`](https://github.com/samtools/htsjdk)?

src/main/java/picard/sam/HitsForInsert.java
DenisVerkhoturov: Since both `SamRecod#mReferenceName` and `SamRecord#mMateReferenceName` [are nullable](https://github.com/samtools/htsjdk/blob/934fb62e8b95071787bad57ee49244adcb655e6b/src/main/java/htsjdk/samtools/SAMRecord.java#L571) we need to ensure that they are non-null or simply delegate it to `Objects.equals` to avoid `NullPointerException`. This is a theoretical case that may hardly ever occur in runtime. So, do we need to prevent this case or can I unwrap this check?

src/main/java/picard/sam/HitsForInsert.java
DenisVerkhoturov: Done [here](https://github.com/broadinstitute/picard/pull/932/commits/e418168597725fdc18640ecc687348efcc2ee169).

src/main/java/picard/sam/HitsForInsert.java
pshapiro4broad: Not sure what you mean about functional vs procedural... one form is an expression consisting of a function call, the other is a binary expression using the`!=` operator. Both are equally functional and procedural, as I understand these terms.

Readability is a subjective argument; to me, using `!= null` is more readable. Given that it requires a static import to avoid making the code much wordier than the original, I don't think it's worth the change.

src/main/java/picard/sam/HitsForInsert.java
pshapiro4broad: It's OK, I was wondering if this was added to fix a known problem or was just precautionary.

Looking at `SamRecord` I see that it's common to check these fields for `null` so it seems wise to do it here too.

It would be great if picard had nullable constraints in the code base! That would be a huge PR though.

src/main/java/picard/sam/HitsForInsert.java
DenisVerkhoturov: Done [here](https://github.com/broadinstitute/picard/pull/932/commits/96bc12a053504430c0ae1c1aafcea8667924648a).

src/main/java/picard/sam/HitsForInsert.java
DenisVerkhoturov: Replaced by operators as was discussed [here](https://github.com/broadinstitute/picard/pull/932#discussion_r139697092).

src/main/java/picard/sam/HitsForInsert.java
DenisVerkhoturov: `Optional` can be rather helpful here and it would be just great if we could use [`Try`](https://github.com/vavr-io/vavr/blob/master/vavr/src/main/java/io/vavr/control/Try.java) and [`Lazy` ](https://github.com/vavr-io/vavr/blob/master/vavr/src/main/java/io/vavr/Lazy.java) types from [vavr](https://github.com/vavr-io/vavr) to reduce service-code and to clarify Picard logic!=D

src/main/java/picard/sam/HitsForInsert.java
DenisVerkhoturov: Could you please review [associated request](https://github.com/samtools/htsjdk/pull/994) to `htsjdk`?

src/main/java/picard/sam/HitsForInsert.java
DenisVerkhoturov: Speaking about operators vs functions, I have always thought that since Java doesn't allow to make partial application of an operator it is convenient to use functions in such cases. Let's imagine we have `equals` function that takes two objects and checks their equality. So `isNull` function is just a partial application of `equals` function with fixed `null` as one of the arguments. Then `nonNull` is negated `isNull`. So that way, for the sake of consistency it is easier to use functions instead of operators in Java not only in streams but everywhere in mine opinion. Is it a little bit excessively? :smile:

pshapiro4broad: It sounds like you have a more enlightened view of functional programming in Java than I do. :)

Thanks for the explanation, I now understand the distinction you're making between expressions and functions, and why it's more consistent to use functions everywhere. I can't say I'm convinced but I'll definitely be giving it more thought.

yfarjoun:
@DenisVerkhoturov there are many "esthetic" changes mixed in with substantive changes. Could you please rebase and separate into two commits, one that is purely esthetic and one that is substantive? it will make it much easier to understand what you are doing.

DenisVerkhoturov:
@yfarjoun Sure, done=)

src/main/java/picard/sam/HitsForInsert.java
```diff
@@ -235,42 +239,81 @@ public void coordinateByHitIndex() {
...
     }
 
+    /**
+     * This method lines up alignments in firstOfPairOrFragment and secondOfPair lists so that the first and the second
+     * records of each pair have the same index.
+     */
+    void coordinateByMate() {
+        final List<SAMRecord> newFirstOfPairOrFragment = new ArrayList<>();
+        final List<SAMRecord> newSecondOfPair = new ArrayList<>();
+
```
tbl3rd: Looks like this file preferred ++i to i++.

src/test/java/picard/sam/MergeBamAlignmentTest.java
```diff
@@ -1907,5 +1907,84 @@ public void testHeaderFromMappedBreaks(final String filename) throws IOException
...
+            final SAMRecord second = new SAMRecord(header);
+
+            second.setReferenceName("chrm2");
+            second.setAlignmentStart(15);
+            second.setMateReferenceName(SAMRecord.NO_ALIGNMENT_REFERENCE_NAME);
+            second.setAlignmentStart(SAMRecord.NO_ALIGNMENT_START);
+
+            hitsForInsert.addFirstOfPairOrFragment(second);
+        }
+
```
tbl3rd: Is 3 some function of HitsForInsert::numHits?

src/main/java/picard/sam/HitsForInsert.java
```diff
@@ -217,16 +219,18 @@ public void coordinateByHitIndex() {
                 secondOfPair.add(i, null);
             }
         }
+    }
 
-        // Now renumber any correlated alignments, and remove hit index if no correlated read.
```
tbl3rd: Thank you for breaking this out and cleaning up hi, and so on.

DenisVerkhoturov: Thank you for your feedback=)

src/main/java/picard/sam/HitsForInsert.java
DenisVerkhoturov: As you might see ([here](https://github.com/broadinstitute/picard/pull/932/files#diff-fb83641c8b09326f5028047ee001cda6R153) for instance), there was a usage of a postfix increment in this file before. So I just followed this example.
All increment operators in this file changed to prefix form.

src/main/java/picard/sam/HitsForInsert.java
```diff
@@ -235,42 +239,81 @@ public void coordinateByHitIndex() {
...
+        firstOfPairOrFragment.clear();
+        firstOfPairOrFragment.addAll(newFirstOfPairOrFragment);
+        secondOfPair.clear();
+        secondOfPair.addAll(newSecondOfPair);
+
+        renumberHitIndex();
+    }
+
+    /**
+     * Identifies weather records present pairwise alignment or not.
```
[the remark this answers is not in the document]
DenisVerkhoturov: Thank you, fixed [here](https://github.com/broadinstitute/picard/pull/932/commits/e4c8c0d42aae2e02b70dec7f630f41edce6af860).

src/test/java/picard/sam/MergeBamAlignmentTest.java
DenisVerkhoturov: Thank you, fixed [here](https://github.com/broadinstitute/picard/pull/932/commits/e4c8c0d42aae2e02b70dec7f630f41edce6af860).

PR #934, opened 2017-09-19
Dont revert genotypes after fixing

bbimber:
@yfarjoun Because your earlier PR isnt committed I'm opening the merge to your branch, but if you want to merge yours and then make a separate PR to master that's fine.

This does two things: 1) tryToAddVariant() shouldnt stomp over the fixed genotypes using those from the source VC, 2) adds genotypes to the test VCF, which gives coverage over this scenario.  prior to this change, that updated VCF alone would trigger this error.

### Description

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/vcf/LiftoverVcf.java
+++ b/src/main/java/picard/vcf/LiftoverVcf.java
@@ -331,7 +331,6 @@ public class LiftoverVcf extends CommandLineProgram {
 
         final VariantContextBuilder builder = new VariantContextBuilder(vc);
 
-        builder.genotypes(source.getGenotypes());
         builder.filters(source.getFilters());
         builder.log10PError(source.getLog10PError());
         builder.attributes(source.getAttributes());
@@ -438,7 +437,6 @@ public class LiftoverVcf extends CommandLineProgram {
 
         final Map<Allele, Allele> reverseComplementAlleleMap = new HashMap<>(2);
 
-        reverseComplementAlleleMap.clear();
         final List<Allele> alleles = new ArrayList<>();
 
         for (final Allele oldAllele : source.getAlleles()) {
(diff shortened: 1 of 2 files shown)


src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -331,7 +331,6 @@ private void tryToAddVariant(final VariantContext vc, final ReferenceSequence re
 
         final VariantContextBuilder builder = new VariantContextBuilder(vc);
 
-        builder.genotypes(source.getGenotypes());
```
yfarjoun: when will the genotypes get added then?

bbimber:
LiftoverVcf line 410 or 464.  

When the new VariantContextBuilder is created on line 332, it will copy the Genotypes from vc, which is the VCB we made with the lifted coordinates.  By calling genotypes() with source.getGenotypes(), we actually revert all the upstream work to fixGenotypes() and do the opposite of what we want.  Tests are passing...see the augmented VCF in this PR and run tests with or without that line.

bbimber:
In addition: when running the modified (or original) LiftoverVcf on a VCF with a lot of samples (~350) I regularly get GCOutOfMemoryErrors.  I strongly suspect this is due to the code constantly creating and re-creating entire VariantContext and Genotypes objects with each step.  I have a local version of LiftoverVcf that rewrites this code to do two key things:

1) While lifting the VC, pass around VariantContextBuilder and only actually create the final VariantContext once per site, instead of 2-3 times per site like now.

2) We shouldnt need to update Genotypes with every modification.  I believe this gets really expensive when you have a lot of samples.  Rather than track allele changes as a map of Allele->Allele, we only really need the list of original genotypes and list of replaced ones.  If we have this information, we can get away with a single step where we replace old/new genotypes based on index.  That saves a lot of work.

If you're willing to look at the more substantial change, this is a big performance boost and the only way I can get this tool to work on a VCF with lots of samples.

yfarjoun:
I'm wiling to look at performance boost and bug-fixes...but not PRs that do
both at the same time. :-)

bbimber:
Well good b/c that's not what's in this PR - maybe we can get this bugfix closed out?

yfarjoun:
@bbimber I merged this into my branch. Looking forward to your performance improvement PR!

PR #940, opened 2017-09-28
Adapt CollectRrbsMetrics to SinglePassSamProgram

tcibinan:
### Description

```CollectRrbsMetrics``` extends from ```CommandLineProgram```. In order to optimize code and make it more readable ```SinglePassSamProgram``` was added.

### Implementation
- **CRM** extends from **SPSP** and can be used by ```CollectMultipleMetrics```;
- Add tests for running **CRM** by `CollectMultipleMetrics`;
- Add more tests for **CRM**;
- Argument ```OUTPUT``` in **SPSP** was made optional *(it could be a point for a discussion)* in case where **CRM** runs with argument ```METRICS_FILE_PREFIX``` instead of argument ```OUTPUT``` and 2 other related arguments ```CHART_OUTPUT``` and ```SUMMARY_OUTPUT```:

```
java -jar picard.jar CollectRrbsMetrics
R=reference_sequence.fasta
I=input.bam
M=basename_for_metrics_output_files
```

----

#### Content

#### Review

### CODE REVIEW

File: src/main/java/picard/analysis/CollectMultipleMetrics.java [hunk 1]
```diff
@@ -103,6 +104,29 @@ SinglePassSamProgram makeInstance(final String outbase, final String outext, fin
     }
 
     public static enum Program implements ProgramInterface {
+        CollectRrbsMetrics {
+            @Override
+            public SinglePassSamProgram makeInstance(final String outbase, final String outext, final File input, final File reference, final Set<MetricAccumulationLevel> metricAccumulationLevel, final File dbSnp, final File intervals) {
+                final CollectRrbsMetrics program = new CollectRrbsMetrics();
+                program.OUTPUT = new File(outbase + ".rrbc.detail_metrics" + outext);
```
yfarjoun: rrbc->rrbs (x3)

src/main/java/picard/analysis/CollectRrbsMetrics.java
```diff
@@ -82,7 +76,18 @@
...
             "observed for each site.</p>" +
 
-            "<h4>Usage example:</h4>" +
+            "<p>It is possible to launch CollectRrbsMetrics tool by setting all output files or by indicating a base name for metrics files only. See examples:</p>" +
+
+            "<h4>Usage example 1:</h4>" +
+            "<pre>" +
+            "java -jar picard.jar CollectRrbsMetrics \\<br />" +
+            "      I=input.bam \\<br />" +
+            "      O=detail_output.bam \\<br />" +
```
yfarjoun: O=detail_output.txt, not .bam

src/main/java/picard/analysis/CollectRrbsMetrics.java
```diff
@@ -82,7 +76,18 @@
...
+            "      CHART=chart_output.pdf \\<br />" +
+            "      S=summary_output.bam \\<br />" +
```
yfarjoun: .txt not .bam

src/main/java/picard/analysis/SinglePassSamProgram.java
```diff
@@ -56,7 +56,7 @@
     @Argument(shortName = StandardOptionDefinitions.INPUT_SHORT_NAME, doc = "Input SAM or BAM file.")
     public File INPUT;
 
-    @Argument(shortName = "O", doc = "File to write the output to.")
+    @Argument(shortName = "O", doc = "File to write the output to.", optional = true)
```
yfarjoun: why is this optional?

yfarjoun:
why the change to specifying the outputs explicitly? what's the benefit there?

src/main/java/picard/analysis/SinglePassSamProgram.java
tcibinan: `CollectRrbsMetrics` didn't have `OUTPUT` argument (only `METRICS_FILE_PREFIX`). But now it extends from `SinglePassSamProgram` and you can't run the metrics without **mandatory  argument** `OUTPUT`. 

So `OUTPUT` was made optional in order to be able to run `CollectRrbsMetrics` without it.

Also we can use that excessive `OUTPUT` argument with two more to specify output files explicitly. It can be useful and it is a way to use this unnecessary argument.

tcibinan:
Please check the comment above.

src/main/java/picard/analysis/SinglePassSamProgram.java
yfarjoun: ok.

src/main/java/picard/analysis/CollectMultipleMetrics.java
tcibinan: Done.

src/main/java/picard/analysis/CollectRrbsMetrics.java
tcibinan: Done.

src/main/java/picard/analysis/CollectRrbsMetrics.java
tcibinan: Done.

src/main/java/picard/analysis/SinglePassSamProgram.java
yfarjoun: thinking about this more, this isn't "OK". this changes the behavior of all the other programs that extend from SinglePassSamProgram.

can you find a solution that doens't change the behavior/API for the other programs?

src/test/java/picard/metrics/CollectRrbsMetricsTest.java
```diff
@@ -1,102 +0,0 @@
-/*
```
yfarjoun: why is this test removed? we should still be testing CollectRrbsMetrics directly, not only via collect multiple metrics.

src/main/java/picard/analysis/SinglePassSamProgram.java
yfarjoun: I owuld almost prefer to simply rename METRICS_FILE_PREFIX -> "OUTPUT" and to do away with the specific metric/chart/summary output variables....the only problem is that it would create an API change...

@cnorman, perhaps you can suggest solutions for this...given your knowledge of the parsing system.

yfarjoun:
please respond to the review here if this is still interesting.

src/main/java/picard/analysis/SinglePassSamProgram.java
yfarjoun: I suggest that you implement it like CollectGcBiasMetrics is implemented, i.e. OUTPUT is for the detailed_metrics, SUMMARY_OUTPUT is for summary_metrics and CHART_OUTPUT for the pdf. You can use METRICS_FILE_PREFIX as Mutex, or remove it and change the API for this program (but not for all of the Single-Pass-Sam programs)

kbergin:
Hi @tcibinan - Would you be available to complete this PR?

tcibinan:
Hi! It's been a long time since I created the pull request. I'll try to find time to update the code to the latest changes and add requested changes but I can't be certain when. So if there is somebody who have enough courage to finish the request I'll be very pleased. Otherwise, I'll complete it one day.

yfarjoun:
@tcibinan I'm wondering if this PR is needed now with the changes I made to CMM...perhaps you could re-tool CollectRrbsMetrics to only use the regular paramters from CMM and take additional parameter via EXTRA_ARGUMENTS?

kbergin:
Hey @tcibinan ! Going through and looking at open PRs and issues in Picard and wanted to check in if you had seen Yossi's comment, or had any thoughts about the future of this PR?

PR #941, opened 2017-09-28
Converted the tests of @DataProviders to using....a @DataProvider!

yfarjoun:
### Description
Improves the Test of DataProviders to use a @DataProvider which will show all the tests separately so that in case of failures it will not abort after the first failure.

----

#### Content

#### Review

### DIFF
--- a/src/test/java/picard/TestDataProviders.java
+++ b/src/test/java/picard/TestDataProviders.java
@@ -1,6 +1,7 @@
 package picard;
 
 import org.testng.Assert;
+import org.testng.annotations.BeforeSuite;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 import picard.cmdline.ClassFinder;
@@ -39,6 +40,12 @@ public class TestDataProviders {
         }
     }
 
+    @Test
+    public void IndependentTestOfDataProviderTest() throws IllegalAccessException, InvocationTargetException, InstantiationException {
+        testAllDataProvidersdata();
+    }
+
+
     @DataProvider(name = "DataprovidersThatDontTestThemselves")
     public Iterator<Object[]> testAllDataProvidersdata() throws IllegalAccessException, InstantiationException, InvocationTargetException {
         int i = 0;
@@ -66,6 +73,15 @@ public class TestDataProviders {
         final Class clazz = container.clazz;
         System.err.println("Method: " + method + " Class: " + clazz.getName());
 
-        method.invoke(clazz.newInstance());
+        Object instance = clazz.newInstance();
+
+        // Some tests assume that the @BeforeSuite methods will be called before the @DataProviders
+        for (final Method otherMethod : clazz.getMethods()) {
+            if (otherMethod.isAnnotationPresent(BeforeSuite.class)) {
+                otherMethod.invoke(instance);
+            }
+        }
+
+        method.invoke(instance);
     }
 }
--- a/src/test/java/picard/TestDataProviders.java
+++ b/src/test/java/picard/TestDataProviders.java
@@ -1,18 +1,18 @@
 package picard;
 
 import org.testng.Assert;
-import org.testng.annotations.BeforeSuite;
-import org.testng.annotations.DataProvider;
-import org.testng.annotations.NoInjection;
+import org.testng.annotations.*;
 import org.testng.annotations.Test;
+import org.testng.collections.Sets;
 import picard.cmdline.ClassFinder;
+import picard.util.TestNGUtil;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
+import java.util.*;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 /**
  * This test is a mechanism to check that none of the data-providers fail to run.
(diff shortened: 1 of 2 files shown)


src/test/java/picard/TestDataProviders.java
```diff
@@ -21,21 +26,46 @@
 public class TestDataProviders {
 
     @Test
-    public void testAllDataProviders() throws IllegalAccessException, InstantiationException, InvocationTargetException {
-        int i = 0;
+    public void IndependentTestOfDataProviderTest() throws IllegalAccessException, InvocationTargetException, InstantiationException {
```
pshapiro4broad: When writing tests it's OK (IMO) to use `throws Exception` since we don't to bother updating our tests every time the list of thrown exceptions changes.

src/test/java/picard/TestDataProviders.java
```diff
@@ -21,21 +26,46 @@
...
-                    method.invoke(testClass.newInstance());
-                    i++;
+                    data.add(new Object[]{method, testClass});
                 }
             }
         }
-        Assert.assertNotSame(i,0);
-        System.err.println("Found: "+ i + " @DataProviders.");
+        Assert.assertTrue(data.size() > 1);
+        Assert.assertEquals(data.stream().filter(c -> ((Method) c[0]).getName().equals("testAllDataProvidersdata")).count(), 1);
```
pshapiro4broad: You filter here but that doesn't affect the data returned below. Not sure if that was your intent or not.

src/test/java/picard/TestDataProviders.java
```diff
@@ -21,21 +26,46 @@
...
-                    System.err.println("Method: " + testClass.getName() + ":" + method.getName());
-                    method.invoke(testClass.newInstance());
-                    i++;
+                    data.add(new Object[]{method, testClass});
                 }
             }
         }
-        Assert.assertNotSame(i,0);
-        System.err.println("Found: "+ i + " @DataProviders.");
+        Assert.assertTrue(data.size() > 1);
```
pshapiro4broad: I think in general you don't want to put asserts inside a dataprovider. If you want to do this, you should do it inside `IndependentTestOfDataProviderTest()` above.

src/test/java/picard/TestDataProviders.java
pshapiro4broad: method names should begin with lowercase letter

src/test/java/picard/TestDataProviders.java
```diff
@@ -21,21 +26,46 @@
...
+        Assert.assertEquals(data.stream().filter(c -> ((Method) c[0]).getName().equals("testAllDataProvidersdata")).count(), 1);
+
+        return data.iterator();
+    }
+
+    // @NoInjection annotations required according to this test:
+    // https://github.com/cbeust/testng/blob/master/src/test/java/test/inject/NoInjectionTest.java
+    @Test(dataProvider = "DataprovidersThatDontTestThemselves")
```
pshapiro4broad: Why write this using a data provider? Was there an issue with this being written as a test without a data provider?

src/test/java/picard/TestDataProviders.java
yfarjoun: do you prefer it with throws Exception?

src/test/java/picard/TestDataProviders.java
yfarjoun: that was the intent, but I can move it to the independent test as you suggested earlier.

src/test/java/picard/TestDataProviders.java
yfarjoun: more meta...

I wanted to be sure that the test itself is working...so this is how it tests itself.

src/test/java/picard/TestDataProviders.java
pshapiro4broad: Yes, for test methods only

src/test/java/picard/TestDataProviders.java
pshapiro4broad: Personally I would wait until the test fails in some way, and then add a check for that failure (such as if you weren't getting any dataproviders, you'd add an assert like the one you added on line 48).

You could spend a lot of time testing your test code but IMO it's not where I'd spend my time, especially when it comes to testing for failures that have not occurred yet.

yfarjoun: not sure what you are referring to....but untested code has bugs, by definition. since the small testing framework I built for testing the DataProviders is code...I wanted to be sure that it works...thus a test. In similar additions in htsjdk I found many DataProviders that were silently failing....so I don't think that this is pre-mature.

src/test/java/picard/TestDataProviders.java
```diff
@@ -21,21 +26,63 @@
...
+
+        Assert.assertTrue(data.hasNext(), "Found no data from testAllDataProvidersdata. Something is wrong.");
+
+        Assert.assertEquals( StreamSupport.stream(Spliterators.spliteratorUnknownSize(data, 0), false)
+                        .filter(c -> ((Method) c[0]).getName().equals("testAllDataProvidersdata")).count(), 1,
+                "getDataProviders didn't find testAllDataProvidersdata, which is in this class. Something is wrong.");
+    }
+
+    @DataProvider(name = "DataprovidersThatDontTestThemselves")
+    public Iterator<Object[]> testAllDataProvidersdata() throws Exception {
```
pshapiro4broad: I think capitalization should be `testAllDataProvidersData`

src/test/java/picard/TestDataProviders.java
```diff
@@ -21,21 +26,63 @@
...
+
+    // @NoInjection annotations required according to this test:
+    // https://github.com/cbeust/testng/blob/master/src/test/java/test/inject/NoInjectionTest.java
+    @Test(dataProvider = "DataprovidersThatDontTestThemselves")
+    public void testDataProviderswithDP(@NoInjection final Method method, final Class clazz) throws
+            IllegalAccessException, InstantiationException {
+
+        Object instance = clazz.newInstance();
+
+        Set<Method> methodSet = Sets.newHashSet();
```
pshapiro4broad: This seems like an unnecessary use of `Sets` as you can just call `new HashSet<>()` here

src/test/java/picard/TestDataProviders.java
```diff
@@ -21,21 +26,63 @@
...
+        methodSet.addAll(Arrays.asList(clazz.getDeclaredMethods()));
+        methodSet.addAll(Arrays.asList(clazz.getMethods()));
```
pshapiro4broad: As with data provider methods, I don't think you want to invoke inherited before suite methods.

src/test/java/picard/TestDataProviders.java
```diff
@@ -21,21 +26,63 @@
...
+
+        // Some tests assume that the @BeforeSuite methods will be called before the @DataProviders
```
pshapiro4broad: This seems like a problem. This can (will always?) result in `BeforeSuite` code being called more than once for a test suite. That could be a problem, as BeforeSuite code can assume it will only be called once.

yfarjoun: OK. I'll see if I can leave out classes that have a @BeforeSuite and not call it.

src/test/java/picard/TestDataProviders.java
yfarjoun: done.

src/test/java/picard/TestDataProviders.java
yfarjoun: done.

src/test/java/picard/TestDataProviders.java
yfarjoun: done.

src/test/java/picard/TestDataProviders.java
yfarjoun: not sure what is meant here.

src/test/java/picard/util/TestNGUtil.java
```diff
@@ -93,4 +94,33 @@ public static void assertEqualDoubleArrays(final double[] lhs, final double[] rh
...
     }
+
+    /** A Method that returns all the Methods that are annotated with @DataProvider
+     * in a given package. Should be moved to htsjdk and used from there
+     *
+     * @param pakkage the package under hwich to look for classes and methods
+     * @return an iterator to collection of Object[]'s consisting of {Method method, Class clazz} pair.
+     * where method has the @DataProviderAnnotation and is a member of clazz.
+     */
+    public static Iterator<Object[]> getDataProviders(final String pakkage) {
```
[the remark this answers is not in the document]
yfarjoun: done.

src/test/java/picard/util/TestNGUtil.java
```diff
@@ -93,4 +94,33 @@ public static void assertEqualDoubleArrays(final double[] lhs, final double[] rh
...
+        List<Object[]> data = new ArrayList<>();
+        final ClassFinder classFinder = new ClassFinder();
+        classFinder.find(pakkage, Object.class);
+
+        for (final Class<?> testClass : classFinder.getClasses()) {
+            if (Modifier.isAbstract(testClass.getModifiers()) || Modifier.isInterface(testClass.getModifiers()))
+                continue;
+            Set<Method> methodSet = Sets.newHashSet();
+            methodSet.addAll(Arrays.asList(testClass.getDeclaredMethods()));
+            methodSet.addAll(Arrays.asList(testClass.getMethods()));
```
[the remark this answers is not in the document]
yfarjoun: I was thinking that I might want it to be run twice, so that it is run for each of the classes that inherits (e.g. MarkDuplicates testing framework)....what do you think?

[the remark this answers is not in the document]
pshapiro4broad: OK, I guess that makes sense, I guess the instances are different so if the method isn't static it will have a different context when run in a derived class vs the base class. (I do wonder if most of our data providers can be declared as static, and don't even need an instance to run. In the case of `MarkDuplicatesTest` the data providers could be declared static and don't depend on any inherited state, so this test is invoking them 5x when it only needs to invoke them 1x.)

src/test/java/picard/TestDataProviders.java
pshapiro4broad: Sorry, the comment was left out of order.. I was thinking that, just like data provider methods, using `getMethods()` will result in `BeforeSuite` being executed twice if it's inherited. Most likely we don't have any cases of `@BeforeSuite` being ended up as an inherited method.

src/test/java/picard/util/TestNGUtil.java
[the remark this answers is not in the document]
yfarjoun: OTOH, they would be invokes several times by TestNG...so it can't be that bad that they get invoked a few more times here...can it?

src/test/java/picard/util/TestNGUtil.java
```diff
@@ -93,4 +94,33 @@ public static void assertEqualDoubleArrays(final double[] lhs, final double[] rh
             Assert.assertTrue(compareDoubleWithAccuracy(lhs[i], rhs[i], accuracy), "Arrays disagree at position " + i + ":  " + lhs[i] + " vs. " + rhs[i] + ". ");
         }
     }
+
+    /** A Method that returns all the Methods that are annotated with @DataProvider
+     * in a given package. Should be moved to htsjdk and used from there
+     *
+     * @param pakkage the package under hwich to look for classes and methods
```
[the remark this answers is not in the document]
yfarjoun: done.

PR #947, opened 2017-10-03
LiftoverVcf performance improvements

bbimber:
This was motivated by the fact that I found I was unable to successfully run LiftoverVcf on a relatively small VCF (in terms of positions), that had about 300 samples in it.  I frequently hit GCOutOfMemory errors, even when giving 96GB of RAM.  If I strip out genotypes from this VCF, it can lift, so I suspect that the large number of samples are a contributing factor.  I looked over the code and saw two things:

1) It nearly always creates 2 (or more) VariantContextBuilder/VariantContexts per position.  In this proposal, we pass around a VariantContextBuilder, and only create the new VC once.

2) If the VC gets reverse complemented, we need to update alleles and genotypes.  However, the current code does this way more frequently than necessary.  Each time we need to swap alleles, the current code essentially maintains a map of Old Allele -> New Alelle, and updates with each step that alters alleles.  This happens once for simple inversions and twice for indels.  Because the set of alleles is an ordered list, we really only need to maintain the original allele list and the final list.  the index of the original allele can be used to map to the updated one, and this only needs to be done once. 

3) Because I hope to use pieces of this code outside of the LiftoverVcf tool, I refactored most of the logic to lift a VariantContext to a new interval into a separate LiftoverUtils class.  Most methods remain private.  It tries to better split the complicated logic behind lifting one VariantContext from LiftoverVcf itself.

----

#### Content

#### Review


yfarjoun:
Thanks for this @bbimber. Can you also positively state that this new version was able to lift-over your 300-sample vcf?

bbimber:
Yes, it was the only way i could get that to happen.  I've run this version on the original 300 sample VCF that sparked this issue and it worked considerably faster as well.

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,255 @@
+package picard.util;
```
yfarjoun: needs license

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,255 @@
...
+
+        builder.start(start);
+        builder.stop(end);
+
+        final Map<Allele, Allele> fixedAlleleMap = alleleBasesMap.entrySet().stream()
+                .collect(Collectors.toMap(Map.Entry::getKey, me -> Allele.create(me.getValue(), me.getKey().isReference())));
+
+        //retain original order:
+        List<Allele> fixedAlleles = alleles.stream().map(a -> fixedAlleleMap.get(a)).collect(Collectors.toList());
+        //Collections.reverse(fixedAlleles);
```
yfarjoun: why this comment?

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,255 @@
...
+        return fixedGenotypes;
+    }
+
+    /**
+     *    Normalizes and left aligns a {@link VariantContext}.
+     *
+     *    Based on Adrian Tan, Gonçalo R. Abecasis and Hyun Min Kang. (2015)
+     *    Unified Representation of Genetic Variants. Bioinformatics.
+     *
+     * @return a new {@link VariantContext} which represents the same variation as vc but has
```
yfarjoun: correct javaDoc

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,255 @@
...
+     * been normalized and left-aligned
+     */
+    protected static void leftAlignVariant(VariantContextBuilder builder, int start, int end, final List<Allele> alleles, final ReferenceSequence referenceSequence) {
```
yfarjoun: Several comments here...

- would it be good to return the original builder as well so that one could use in functional form?
- Please make a big note about side-effect (changing the builder) in the javaDoc.
- Please convert all parameters to final
- do you need start and end as parameters? why not pull them out of builder?
- similarly for the alleles...why not pull them out of the builder?

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,255 @@
...
+     * If this interval is in the opposite orientation, all alleles and genotypes will be reverse complemented
+     * and indels will be left-aligned.  Currently this is only able to invert biallelic indels, and null will be
+     * returned for any unsupported VC.
+     * @param source The VariantContext to lift
+     * @param target The target interval
+     * @param refSeq The reference sequence, which should match the target interval
+     * @param writeOriginalPosition If true, INFO field annotations will be added to store the original position and contig
+     * @return The lifted VariantContext.  This will be null if the input VariantContext could not be lifted.
+     */
+    public static VariantContext liftVariant(VariantContext source, Interval target, ReferenceSequence refSeq, boolean writeOriginalPosition){
```
yfarjoun: finals

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,255 @@
...
+        return builder.make();
+    }
+
+    protected static VariantContextBuilder liftSimpleVariantContext(VariantContext source, Interval target){
+        if (target == null || source.getReference().length() != target.length()) {
+            return null;
+        }
+
+        // Build the new variant context
+        final VariantContextBuilder builder = new VariantContextBuilder(source);
```
yfarjoun: Comment that this will copy over the genotypes and all the annotations, and filters

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,255 @@
...
+        if (source.isIndel()){
+            leftAlignVariant(vcb, start, stop, vcb.getAlleles(), refSeq);
+        }
+
+        vcb.genotypes(fixGenotypes(source.getGenotypes(), origAlleles, vcb.getAlleles()));
+
+        return vcb;
+    }
+
+    private static List<Allele> reverseComplementAlleles(List<Allele> originalAlleles, Interval target, ReferenceSequence refSeq, boolean isBiAllelicIndel, boolean addToStart){
```
yfarjoun: finals

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,255 @@
...
+        final List<Allele> alleles = new ArrayList<>();
+
+        for (final Allele oldAllele : originalAlleles) {
+            alleles.add(LiftoverUtils.reverseComplement(oldAllele, target, refSeq, isBiAllelicIndel, addToStart));
+        }
+
+        return alleles;
+    }
+
+    private static Allele reverseComplement(Allele oldAllele, Interval target, ReferenceSequence referenceSequence, boolean isBiAllelicIndel, boolean addToStart){
```
yfarjoun: finals

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,255 @@
...
+            if (!addToStart) alleleBuilder.append((char) referenceSequence.getBases()[target.getEnd() - 1]);
+
+            return Allele.create(alleleBuilder.toString(), oldAllele.isReference());
+        }
+        else {
+            return Allele.create(SequenceUtil.reverseComplement(oldAllele.getBaseString()), oldAllele.isReference());
+        }
+    }
+
+    protected static GenotypesContext fixGenotypes(final GenotypesContext originals, List<Allele> originalAlleles, List<Allele> newAlleles) {
```
yfarjoun: finals

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,255 @@
...
+        //not currently supported
+        if (source.isIndel() && !source.isBiallelic()){
+            return null;
+        }
+
+        List<Allele> origAlleles = new ArrayList<>(source.getAlleles());
+        VariantContextBuilder vcb = new VariantContextBuilder(source);
+        vcb.chr(target.getContig());
+
+        final boolean addToStart = source.isIndel() && target.getStart() > 1;
```
yfarjoun: now that this boolean is going to be used over 3 functions, we might want to document what it is better....

Please note that when an indel is reverse-complemented the allele needs to be shifted by one base to the left. However, this cannot be done if the indel ends up being at the beginning of the contig after the RC. So, this boolean's job is to protect against such a, admitably rare, event.

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,255 @@
...
+        for (final Allele oldAllele : originalAlleles) {
+            alleles.add(LiftoverUtils.reverseComplement(oldAllele, target, refSeq, isBiAllelicIndel, addToStart));
+        }
+
+        return alleles;
+    }
+
+    private static Allele reverseComplement(Allele oldAllele, Interval target, ReferenceSequence referenceSequence, boolean isBiAllelicIndel, boolean addToStart){
+
+        if (oldAllele.isSymbolic()){
```
yfarjoun: what about no-calls ?

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    protected static GenotypesContext fixGenotypes(final GenotypesContext originals, List<Allele> originalAlleles, List<Allele> newAlleles) {
+        // optimization: if nothing needs to be fixed then don't bother
+        if (originalAlleles.equals(newAlleles)) {
+            return originals;
+        }
+
+        final GenotypesContext fixedGenotypes = GenotypesContext.create(originals.size());
+        for (final Genotype genotype : originals) {
```
yfarjoun: This is where the map (from old allele to new allele) would help. But you removed it. now you need to mess around with indexes and stuff. Consider the case of many genotypes...would it not be better to have a map?

The alleles in the genotype must be part of the VC's alleles, so the checking (for symbolic but not no-call).

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,255 @@
+package picard.util;
+
+import htsjdk.samtools.reference.ReferenceSequence;
```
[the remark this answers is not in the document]
bbimber: ok, though i think most of these pre-date my changes.  

do you happen to know if I can get intellij to auto-correct any of this kind of stuff?

src/main/java/picard/util/LiftoverUtils.java
bbimber: ok

src/main/java/picard/util/LiftoverUtils.java
bbimber: ok.  i see many of these classes are missing licences?  SortVcf being an example, and LiftoverVcf (pre-dated my PR), also seems to lack one?

src/main/java/picard/util/LiftoverUtils.java
bbimber: i thought this was handled upstream, but i will add a check here too

src/main/java/picard/util/LiftoverUtils.java
bbimber: ok

src/main/java/picard/util/LiftoverUtils.java
yfarjoun: Please add to your new files in this PR. I'll open an issue to add licenses to the ones that miss it. thanks.

src/main/java/picard/util/LiftoverUtils.java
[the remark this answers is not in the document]
yfarjoun: not sure about intelliJ....

src/main/java/picard/util/LiftoverUtils.java
bbimber: does builder have these getters?  it would have been nice if it did...

i'll add those notes in javadoc.  

as to returning the builder: this is basically a private method.  it's protected only so tests can access it.  in theory maybe that might be useful, but because of the deficiencies in VariantContextBuilder's get methods, i'm not sure what this by itself gains you...

src/main/java/picard/util/LiftoverUtils.java
bbimber: ok

src/main/java/picard/util/LiftoverUtils.java
bbimber: ok

src/main/java/picard/util/LiftoverUtils.java
bbimber: ok

src/main/java/picard/util/LiftoverUtils.java
bbimber: The reason the map was awkward (or at least the original usage of it was), if because translating genotypes wasnt a one-step process.  You need to do the initial reverse complement, and then maybe left align, etc.  We ended up with multiple maps, updating the genotypes multiple times (the thing i wanted to avoid).  It's entirely possible this approach could have been made to work, but it seemed messy when I way working on it.

The reference and alt alleles are in fact an ordered list.  Genotypes (at least in the raw file), just point to allele index, and dont actually 'know' what that allele means.  If I could someone prevent any parsing of genotypes at all until after we fixed up the alleles that would be even better.  Dealing with updating alleles with genotypes at every change is an unnecessary pain.  

I changed this to make a one-time map in this method based on originalAlleles and newAlleles, instead of using indexOf() each time we iterate all the genotypes.

src/main/java/picard/util/LiftoverUtils.java
bbimber: removed

yfarjoun:
hmmm. that rebase didn't work well...please fix.

src/main/java/picard/util/LiftoverUtils.java
yfarjoun: yes...would be nice to add these getters....didn't realize they are missing...

ok. about not returning.

src/main/java/picard/util/LiftoverUtils.java
yfarjoun: oh and finals here too.

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,296 @@
+package picard.util;
+
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015 The Broad Institute
```
yfarjoun: year 2017

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,296 @@
...
+            if (addToStart) {
+                alleleBuilder.append((char) referenceSequence.getBases()[target.getStart() - 2]);
+            }
+            alleleBuilder.append(SequenceUtil.reverseComplement(oldAllele.getBaseString().substring(1, oldAllele.length())));
+            if (!addToStart) {
+                alleleBuilder.append((char) referenceSequence.getBases()[target.getEnd() - 1]);
+            }
+
+            return Allele.create(alleleBuilder.toString(), oldAllele.isReference());
+        }
```
yfarjoun: } else {

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -0,0 +1,296 @@
...
+
+        final GenotypesContext fixedGenotypes = GenotypesContext.create(originals.size());
+        for (final Genotype genotype : originals) {
+            final List<Allele> fixedAlleles = new ArrayList<>();
+            for (final Allele allele : genotype.getAlleles()) {
+                if (allele.isSymbolic() || allele.isNoCall()) {
+                    fixedAlleles.add(allele);
+                }
+                else {
+                    Allele newAllele = alleleMap.get(allele);
```
yfarjoun: no need for this check

src/main/java/picard/util/LiftoverUtils.java
bbimber: I get that, I'm saying this allele-swapping process would be made easier if that's not how it worked.  If VariantContext could do lazy parsing of the genotypes (i.e. so they remain 0/1) until after we swap around the ref/alt alleles then this translation wouldnt matter one bit.  In any case, that's not what it does.

In the updated PR I done a one-time loop to make this old/new allele map, and translate alleles for each genotype based on this.

src/main/java/picard/util/LiftoverUtils.java
bbimber: ok

src/main/java/picard/util/LiftoverUtils.java
bbimber: ok, though you guys should really get some kind of build step or tool to do these across the project...

src/main/java/picard/util/LiftoverUtils.java
bbimber: symbolic probably not, but for NoCall I think we do, right?

src/main/java/picard/util/LiftoverUtils.java
bbimber: ok

bbimber:
I think this covers requested changes and fixes the rebase.

src/main/java/picard/util/LiftoverUtils.java
yfarjoun: right, the NoCall is allowed to be in the genotype, while missing from the VC, but not the symbolic...so please remove the check for the symbolic allele as it is confusing.

bbimber: ok, removed it

yfarjoun:
It seems that symbolic and no-call genotypes are not tested...do you mind adding a test so that we know this case is correctly done?

yfarjoun:
Thanks!

sooheelee:
Since you are plumbing the code actively, may I ask a question? This relates to a user report at 
<https://gatkforums.broadinstitute.org/gatk/discussion/10753/picard-liftover-from-build-38-to-37-fails-on-alternate-contigs#latest>.

When LiftoverVcf encounters intervals that the CHAIN does not map to the target reference, why don't these variants go to the REJECT file, instead of erring the run?

bbimber:
Well I didnt write the original tool, so I cant speak to the original motivation.

However, the error seems like a reasonable response.  If you generate a chain file from Genome X to Genome Y, how it is valid for this chain file to map to a chromosome that doesnt exist in Y?  I could imagine there are situations where the user drops unplaced contigs in the genome they use, but the chain file was generated on a complete genome; however, I'd argue those are not actually the same genome.  It would also seem that if your chain file specifies source/target chromosomes with invalid names that it's equally likely you simply have the wrong chain file.  Or perhaps you have a chain file for the UCSC genome and you are using the Ensembl one.  Honestly i think the way the whole field deals so loosely w/ genome definitions and contig names is kind of a mess (maybe in need of a stronger file format?), but that's not the point here.

I could maybe see allowing a command line option to control the behavior when unexpected contigs are encountered.  It's possible that'll give you valid results in some cases, but I'd still find it suspect and if it was added it might be worth putting some kind of 'unsafe operation' warning on it, like GATK does for some things.

Issue #955, opened 2017-10-12
INVALID_FLAG_FIRST_OF_PAIR error for paired reads


zkamvar:
### Affected tool(s)
_ValidateSamFile_, _MarkDuplicates_

### Affected version(s)

version 2.9.0

### Description 

I've logged a description of this problem at https://github.com/zkamvar/read-processing/issues/11

In short, I'm receiving the error `INVALID_FLAG_FIRST_OF_PAIR` on a bam file consisting of merged reads from paired and unpaired data when I attempt to run `MarkDuplicates`. The error is applied to two reads that appear to be paired (they have the 80 and 128 flags, respectively). This happened for 3 out of 55 samples. The strange part is that this has no problem when I attempt to run this using only paired-end data. I apologize for the bug report if this is a non-issue, I just couldn't find anything about this elsewhere.

Below is the error log of my run.

<details>
<summary>picard error report</summary>

```
[Thu Oct 12 11:42:43 CDT 2017] picard.sam.markduplicates.MarkDuplicates MAX_FILE_HANDLES_FOR_READ_ENDS_MAP=1000 INPUT=[BAMS/SS.11.16_fixed.bam] OUTPUT=BAMS/SS.11.16_dupmrk.bam METRICS_FILE=BAMS/SS.11.16_marked_dup_metrics.txt ASSUME_SORTED=true    MAX_SEQUENCES_FOR_DISK_READ_ENDS_MAP=50000 SORTING_COLLECTION_SIZE_RATIO=0.25 REMOVE_SEQUENCING_DUPLICATES=false TAGGING_POLICY=DontTag REMOVE_DUPLICATES=false DUPLICATE_SCORING_STRATEGY=SUM_OF_BASE_QUALITIES PROGRAM_RECORD_ID=MarkDuplicates PROGRAM_GROUP_NAME=MarkDuplicates READ_NAME_REGEX=<optimized capture of last three ':' separated fields as numeric values> OPTICAL_DUPLICATE_PIXEL_DISTANCE=100 VERBOSITY=INFO QUIET=false VALIDATION_STRINGENCY=STRICT COMPRESSION_LEVEL=5 MAX_RECORDS_IN_RAM=500000 CREATE_INDEX=false CREATE_MD5_FILE=false GA4GH_CLIENT_SECRETS=client_secrets.json
[...]

PR #956, opened 2017-10-12
Website: add catch for invalid flags in explain-flags.html

zkamvar:
### Description

I have added a catch in explain-flags.html that will send up a warning **Invalid flag: read not paired** in the case of invalid flag values like 128. I ran into this in issue #955, where I had visited the explain flags page, but had no indication that the flag was invalid. 

#### Content

#### Review

### DIFF
--- a/explain-flags.html
+++ b/explain-flags.html
@@ -20,17 +20,16 @@
     function explainFlags() {
         var flagValue = parseInt(document.getElementById('tb').value); //returns 0 or NaN if can't parse
         var summary  = ""; 
-        var badFlags = ""; // Stores the flags that don't make sense with unpaired reads
+        var badFlags = []; // Stores the flags that don't make sense with unpaired reads
         var asterix  = ""; // indicator for the user which flags are invalid with unpaired reads
         var pairedRead = lstFlags[0][1] & flagValue;
         for(var i = 0; i < lstFlags.length; i++) {
             var checkbox = document.getElementById('cb' + i)
             if(lstFlags[i][1] & flagValue) {
+                asterix = " (0x" + lstFlags[i][1].toString(16) + ")";
                 if (!pairedRead && (lstFlags[i][1] & 235)) {
-                    badFlags += " " + lstFlags[i][1];
-                    asterix = "<b>*</b>";
-                } else {
-                    asterix = "";
+                    badFlags.push(" 0x" + lstFlags[i][1].toString(16));
+                    asterix += "<b>*</b>";
                 }
                 summary  += " &nbsp; &nbsp; " + lstFlags[i][0] + asterix + "<br>";
                 checkbox.checked = true;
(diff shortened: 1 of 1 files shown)


explain-flags.html
```diff
@@ -20,10 +20,15 @@
     function explainFlags() {
         var flagValue = parseInt(document.getElementById('tb').value); //returns 0 or NaN if can't parse
         var summary = "";
+        var pairedRead = lstFlags[0][1] & flagValue;
         for(var i = 0; i < lstFlags.length; i++) {
             var checkbox = document.getElementById('cb' + i)
             if(lstFlags[i][1] & flagValue) {
-                summary  += " &nbsp; &nbsp; " + lstFlags[i][0] + "<br>";
+                if (!pairedRead && i < 8){
```
yfarjoun: What's special about 8? it looks as though you are saying that if a read is not paired than none of the first 8 flags can be set. the table below shows where this is true:

flag | not valid without 0x1
-----|-----
|0x1 | yes (but impossible)
0x2 | yes
0x4 | no
0x8 | yes
0x10 | no
0x20 | yes
0x40 | yes
0x80 | yes

In the Sam spec this is described by the line 
>  If 0x1 is unset, no assumptions can be made about 0x2, 0x8, 0x20, 0x40 and 0x80.

which Picard took to mean "invalid"

the "OR" of these flags is 234 (decimal) so you can use that in the code (or use binary 0b11101011)

```
...
if (!pairedRead && (lstFlag[i][1] & 235)) {
...
}

File: explain-flags.html [hunk 2]
```diff
@@ -20,10 +20,15 @@
...
+                    summary = " &nbsp; &nbsp; <b>Invalid flag: read not paired</b><br>";
```
yfarjoun: even though it isn't valid, I would still like to be able to see the flag...

yfarjoun: instead of hiding the summary in the event of an invalid flag, could you add a note to the end explaining ?
perhaps a summary of issue:
```
- Flag XXX cannot be set when read is not paired"
- ....
```

explain-flags.html
zkamvar: This is why I shouldn't code when I'm tired...

I looked at the `switchToMate()` function and saw:

https://github.com/broadinstitute/picard/blob/56f200e7fbb332bc9de788f8021495b4d739a241/explain-flags.html#L48

I thought that those must be all the flags assuming 0x1 😅  I can fix this, no problem!

explain-flags.html
zkamvar: I agree, this is a much better solution.

vdauwera:
Can you show some screenshots/staging pages of what the output looks like?

zkamvar:
Two things happen when a flag assuming paired reads is triggered when the read is does not have the `0x1` flag:

1. In the summary, the offending flags are marked with a "\*". 
2. A warning pops up showing which (integer) flags are cannot be set.

Here's what the warning looks like if the flag is set to 80:

image

No warning is generated if it is set to 81:

image

All the possible flags that are invalid when `0x1` is unset:

image

Please let me know if there's anything that I can do to make the output clearer.

yfarjoun:
@vdauwera you can see it in action by viewing the file, asking for the "raw" version, saving that as a html and opening it in your browser....

zkamvar:
I just realized, there's also htmlpreview! http://htmlpreview.github.io/?https://github.com/zkamvar/picard/blob/fix-explain/explain-flags.html

It won't be pretty because of the Jekyll markup, but it will be functional.

yfarjoun:
- Please output the flags in hex and add `0x` in front so it's clearer.
- Add commas when there are more than one invalid flag.
- Add the flags' hex value after their description in the list. Like so:
`read unmapped (0x2)`

zkamvar:
@yfarjoun I have updated the script with the requested features.

yfarjoun:
I think that this final version looks good. @vdauwera ? what do you say?

vdauwera:
Looks good to me, thanks

PR #964, opened 2017-10-25
Tool to split unmapped SAM/BAM

meganshand:
### Description

This adds a simple tool to take a large unmapped bam and split it into many smaller unmapped bams while retaining the sort order. This is motivated by wanting to align a large file in parallel on multiple machines at the same time.

----

#### Content

#### Review


src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+        if (!firstSeenMates.isEmpty()) {
+            SAMUtils.processValidationError(new SAMValidationError(SAMValidationError.Type.MATE_NOT_FOUND,
+                    "Found " + firstSeenMates.size() + " unpaired mates", null), VALIDATION_STRINGENCY);
+        }
+        return 0;
+    }
+
+    private SAMFileWriter[] generateWriters(final SAMFileWriterFactory factory, final SAMFileHeader header, int splitToNFiles) {
+        int digits = String.valueOf(splitToNFiles).length();
+        final DecimalFormat fileNameFormatter = new DecimalFormat(String.format("%0" + digits + "d", 0));
```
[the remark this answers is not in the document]
yfarjoun: or maybe take the prefix as an argument with default to "shard_" ?

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+
+        if (TOTAL_READS_IN_INPUT < 1) {
+            return new String[]{
+                    "Cannot set TOTAL_READS_IN_INPUT to a number less than 1."
+            };
+        }
+
+        return null;
+    }
+}
```
yfarjoun: extra NL to make github happy...also in the test file.

src/test/java/picard/sam/SAMSplitterTest.java
```diff
@@ -0,0 +1,108 @@
...
+                "INPUT=" + PAIRED_FILE.getAbsolutePath(),
+                "TOTAL_READS_IN_INPUT=10",
+                "SPLIT_TO_N_TEMPLATES=3",
+                "OUTPUT_DIR=" + tmpDir
+        };
+
+        runPicardCommandLine(args);
+
+        final File out1 = new File(tmpDir + "1.bam");
+        final File out2 = new File(tmpDir + "2.bam");
```
yfarjoun: please add deleteOnExit to the outputs

src/test/java/picard/sam/SAMSplitterTest.java
```diff
@@ -0,0 +1,108 @@
...
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Tests for SAMSplitter
+ */
+public class SAMSplitterTest extends CommandLineProgramTest {
+    private static final String TEST_DATA_DIR = "testdata/picard/sam/bam2fastq";
+    private static final File PAIRED_FILE = new File(TEST_DATA_DIR + "/paired/ok/sorted-pair.sam");
+    private static final ValidateSamTester VALIDATE_SAM_TESTER = new ValidateSamTester();
```
yfarjoun: shouldn't be static

src/test/java/picard/sam/SAMSplitterTest.java
```diff
@@ -0,0 +1,108 @@
...
+        final SamReader inReader = SamReaderFactory.makeDefault().open(PAIRED_FILE);
+        VALIDATE_SAM_TESTER.assertSamValid(out);
+        Assert.assertEquals(inReader.iterator().toList().size(), outReader.iterator().toList().size());
+    }
+
+    @Test(expectedExceptions = PicardException.class)
+    public void testZeroOutput() throws IOException {
+        final String tmpDir = IOUtil.getDefaultTmpDir().getAbsolutePath() + "/";
+
+        final String [] args = new String[]{
```
yfarjoun: why does this break?

src/test/java/picard/sam/SAMSplitterTest.java
```diff
@@ -0,0 +1,108 @@
...
+                "INPUT=" + PAIRED_FILE.getAbsolutePath(),
+                "TOTAL_READS_IN_INPUT=10",
+                "SPLIT_TO_N_TEMPLATES=5",
+                "OUTPUT_DIR=" + tmpDir
+        };
+
+        runPicardCommandLine(args);
+
+        final File out = new File(tmpDir + "1.bam");
```
yfarjoun: delete output file on exit

src/test/java/picard/sam/SAMSplitterTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+        final File out1 = new File(tmpDir + "1.bam");
+        final File out2 = new File(tmpDir + "2.bam");
+        final SamReader reader1 = SamReaderFactory.makeDefault().open(out1);
+        final SamReader reader2 = SamReaderFactory.makeDefault().open(out2);
+
+        Assert.assertEquals(reader1.getFileHeader(), reader2.getFileHeader());
+        VALIDATE_SAM_TESTER.assertSamValid(out1);
+        VALIDATE_SAM_TESTER.assertSamValid(out2);
+        Assert.assertEquals(reader1.iterator().toList().size(), 6);
```
yfarjoun: We have guava in picard, so you can use 
```
com.google.common.collect.Iterators.size(iterator) 
```
instead of converting to a list (also next line and further below)

src/test/java/picard/sam/SAMSplitterTest.java
```diff
@@ -0,0 +1,108 @@
...
+        };
+
+        runPicardCommandLine(args);
+
+        final File out1 = new File(tmpDir + "1.bam");
+        final File out2 = new File(tmpDir + "2.bam");
+        final SamReader reader1 = SamReaderFactory.makeDefault().open(out1);
+        final SamReader reader2 = SamReaderFactory.makeDefault().open(out2);
+
+        Assert.assertEquals(reader1.getFileHeader(), reader2.getFileHeader());
```
yfarjoun: also assert that the new header is equal to the original header

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+
+import java.io.File;
+import java.text.DecimalFormat;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * <p/>
+ * Splits the input queryname sorted or querygrouped SAM/BAM file and writes it into
```
nh13: `querygrouped` -> `query-grouped`

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+ * within each output BAM and if the BAMs are concatenated in order (output files are named
+ * numerically) the order of the reads will match the original BAM. As this tool is
+ * intended for unmapped input files, all secondary and supplementary reads will be removed.
+ */
+@CommandLineProgramProperties(
+        summary = SAMSplitter.USAGE_SUMMARY + SAMSplitter.USAGE_DETAILS,
+        oneLineSummary = SAMSplitter.USAGE_SUMMARY,
+        programGroup = SamOrBam.class)
+@DocumentedFeature
+public class SAMSplitter extends CommandLineProgram {
```
nh13: 1. `SAM` -> `Sam` everywhere.  See how other tools are named please.
2. Can we rename this to `SplitBamBySize`?  This is more in line with the names of other tools.  Avoid verbs, use nouns.

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+    public File INPUT;
+
+    @Argument(shortName = "N", doc = "Split to have approximately N templates per output file.")
+    public int SPLIT_TO_N_TEMPLATES;
+
+    @Argument(shortName = "TOTAL_READS", doc = "Total number of reads in the input file.")
+    public int TOTAL_READS_IN_INPUT;
+
+    @Argument(shortName = "ODIR", doc = "Directory in which to output the split BAM files.")
+    public File OUTPUT_DIR;
```
nh13: Consider just using `OUTPUT` and short name `O`.

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+            "     I=paired_unmapped_input.bam \\<br />" +
+            "     OUTPUT_DIR=out_dir \\ <br />" +
+            "     TOTAL_READS_IN_INPUT=800000000" +
+            "     SPLIT_TO_N_TEMPLATES=24000000" +
+            "</pre>" +
+            "<hr />";
+    @Argument(doc = "Input SAM/BAM file to split", shortName = StandardOptionDefinitions.INPUT_SHORT_NAME)
+    public File INPUT;
+
+    @Argument(shortName = "N", doc = "Split to have approximately N templates per output file.")
```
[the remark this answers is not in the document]
nh13: I would argue that we should have `total_input_templates` not `total_input_reads`, since then we can just increment a "template" counter each time we see the primary alignment of either a fragment or the first end of a pair.  This keeps reads from the same template together, and also gets rid of the annoying `PAIRED_INPUT` argument (what if we have a mix).

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+
+    @Argument(shortName = "TOTAL_READS", doc = "Total number of reads in the input file.")
+    public int TOTAL_READS_IN_INPUT;
+
+    @Argument(shortName = "ODIR", doc = "Directory in which to output the split BAM files.")
+    public File OUTPUT_DIR;
+
+    @Argument(shortName = "PAIR", doc = "If true the input bam contains paired reads and " +
+            "each template is assumed to contain two reads (rather than one for unpaired inputs).")
+    public boolean PAIRED_INPUT = true;
```
nh13: Can you find a way not to have this?  Count by templates? Examine the first few records to see if they are paired end (I'd prefer just to say templates).

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+            "</pre>" +
+            "<hr />";
+    @Argument(doc = "Input SAM/BAM file to split", shortName = StandardOptionDefinitions.INPUT_SHORT_NAME)
+    public File INPUT;
+
+    @Argument(shortName = "N", doc = "Split to have approximately N templates per output file.")
+    public int SPLIT_TO_N_TEMPLATES;
+
+    @Argument(shortName = "TOTAL_READS", doc = "Total number of reads in the input file.")
+    public int TOTAL_READS_IN_INPUT;
```
nh13: templates not reads?

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+        final ProgressLogger progress = new ProgressLogger(log);
+        for (final SAMRecord currentRecord : reader) {
+            if (currentRecord.isSecondaryOrSupplementary())
+                continue;
+
+            int writerIndex = templateIndex / templatesPerFile;
+            if (PAIRED_INPUT) {
+                final String currentReadName = currentRecord.getReadName();
+                final SAMRecord firstRecord = firstSeenMates.remove(currentReadName);
+                if (firstRecord == null) {
```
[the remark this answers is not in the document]
nh13: Again, I would argue that `recordsWritten` is updated only when we see the primary alignment of either a fragment or a the first read of a pair (template)

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+                    firstSeenMates.put(currentReadName, currentRecord);
+                } else {
+                    final SAMRecord read1 =
+                            currentRecord.getFirstOfPairFlag() ? currentRecord : firstRecord;
+                    final SAMRecord read2 =
+                            currentRecord.getFirstOfPairFlag() ? firstRecord : currentRecord;
```
nh13: One line?

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+
+            int writerIndex = templateIndex / templatesPerFile;
+            if (PAIRED_INPUT) {
+                final String currentReadName = currentRecord.getReadName();
+                final SAMRecord firstRecord = firstSeenMates.remove(currentReadName);
+                if (firstRecord == null) {
+                    firstSeenMates.put(currentReadName, currentRecord);
+                } else {
+                    final SAMRecord read1 =
+                            currentRecord.getFirstOfPairFlag() ? currentRecord : firstRecord;
```
nh13: One line?

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+                writers[writerIndex].addAlignment(currentRecord);
+                templateIndex++;
+            }
+
+            progress.record(currentRecord);
+        }
+
+        for (final SAMFileWriter w : writers) {
+            w.close();
+        }
```
nh13: I don't know if we prefer using foreach in Picard? @yfarjoun ?

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+    protected int doWork() {
+        IOUtil.assertFileIsReadable(INPUT);
+        final SamReader reader = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(INPUT);
+        final SAMFileHeader header = reader.getFileHeader();
+        if (header.getSortOrder() == SAMFileHeader.SortOrder.coordinate) {
+            throw new PicardException("Cannot split coordinate sorted bam. Only queryname sorted (or querygrouped) " +
+                    "input will ensure that templates are not split into multiple files.");
+        }
+
+        final Map<String, SAMRecord> firstSeenMates = new HashMap<>();
```
nh13: `firstSeenMates` is not an intuitive variable name.  Perhaps `readPairCache`? 

Also, I don't see why you need this if you assume the input is queryname or query-grouped?

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+        final SAMFileWriterFactory factory = new SAMFileWriterFactory();
+        double unitConverter = PAIRED_INPUT ? 2.0 : 1.0;
```
nh13: final

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+
+        int splitToNFiles = (int) Math.round((TOTAL_READS_IN_INPUT / unitConverter) / SPLIT_TO_N_TEMPLATES);
+        if (splitToNFiles < 1) {
+            throw new PicardException("Splitting " + TOTAL_READS_IN_INPUT + " reads into " + SPLIT_TO_N_TEMPLATES +
+                    " templates per file would result in fewer than 1 output file.");
+        }
+
+        int templatesPerFile = (int) Math.ceil((TOTAL_READS_IN_INPUT / (double) splitToNFiles) / unitConverter);
+        SAMFileWriter[] writers = generateWriters(factory, header, splitToNFiles);
```
nh13: I stopped looking at this further as I think you need another pass based on @yfarjoun and my suggestions.

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        if (!firstSeenMates.isEmpty()) {
+            SAMUtils.processValidationError(new SAMValidationError(SAMValidationError.Type.MATE_NOT_FOUND,
+                    "Found " + firstSeenMates.size() + " unpaired mates", null), VALIDATION_STRINGENCY);
+        }
+        return 0;
+    }
+
+    private SAMFileWriter[] generateWriters(final SAMFileWriterFactory factory, final SAMFileHeader header, int splitToNFiles) {
```
nh13: `splitToNFiles` can be final

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+        int digits = String.valueOf(splitToNFiles).length();
```
nh13: more finals

src/main/java/picard/sam/SAMSplitter.java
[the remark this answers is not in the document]
nh13: I'd prefer a default prefix and an option to override.

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+        final DecimalFormat fileNameFormatter = new DecimalFormat(String.format("%0" + digits + "d", 0));
+        int fileIndex = 1;
+        SAMFileWriter[] writers = new SAMFileWriter[splitToNFiles];
+        for (int i = 0; i < splitToNFiles; i++) {
+            writers[i] = factory.makeSAMOrBAMWriter(header, true, new File(OUTPUT_DIR.getAbsolutePath() + "/" +
```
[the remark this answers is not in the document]
nh13: Or use `Paths.get`.  It's best not to assume a `/` directory separator.

src/test/java/picard/sam/SAMSplitterTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Tests for SAMSplitter
+ */
+public class SAMSplitterTest extends CommandLineProgramTest {
+    private static final String TEST_DATA_DIR = "testdata/picard/sam/bam2fastq";
+    private static final File PAIRED_FILE = new File(TEST_DATA_DIR + "/paired/ok/sorted-pair.sam");
+    private static final ValidateSamTester VALIDATE_SAM_TESTER = new ValidateSamTester();
+
+    public String getCommandLineProgramName() {
```
nh13: I don't see this used anywhere.  If it overrides, use `@Override`.

src/test/java/picard/sam/SAMSplitterTest.java
```diff
@@ -0,0 +1,108 @@
...
+import picard.sam.testers.ValidateSamTester;
+
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Tests for SAMSplitter
+ */
+public class SAMSplitterTest extends CommandLineProgramTest {
+    private static final String TEST_DATA_DIR = "testdata/picard/sam/bam2fastq";
```
nh13: Not the biggest fan of using other unrelated classes' test data.  But ok.

src/test/java/picard/sam/SAMSplitterTest.java
```diff
@@ -0,0 +1,108 @@
...
+    private static final File PAIRED_FILE = new File(TEST_DATA_DIR + "/paired/ok/sorted-pair.sam");
+    private static final ValidateSamTester VALIDATE_SAM_TESTER = new ValidateSamTester();
+
+    public String getCommandLineProgramName() {
+        return SAMSplitter.class.getSimpleName();
+    }
+
+    @Test
+    public void testOkFile() throws IOException {
+        final String tmpDir = IOUtil.getDefaultTmpDir().getAbsolutePath() + "/";
```
nh13: Don't add the "/"

src/test/java/picard/sam/SAMSplitterTest.java
```diff
@@ -0,0 +1,108 @@
...
+        final File out1 = new File(tmpDir + "1.bam");
+        final File out2 = new File(tmpDir + "2.bam");
+        final SamReader reader1 = SamReaderFactory.makeDefault().open(out1);
+        final SamReader reader2 = SamReaderFactory.makeDefault().open(out2);
+
+        Assert.assertEquals(reader1.getFileHeader(), reader2.getFileHeader());
+        VALIDATE_SAM_TESTER.assertSamValid(out1);
+        VALIDATE_SAM_TESTER.assertSamValid(out2);
+        Assert.assertEquals(reader1.iterator().toList().size(), 6);
+        Assert.assertEquals(reader2.iterator().toList().size(), 4);
```
nh13: I would argue that you care about order (see the description of the tool).  Can you validate the order of the query names are the same in the input and outputs?

src/test/java/picard/sam/SAMSplitterTest.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Test
+    public void testOneOutput() throws IOException {
+        final String tmpDir = IOUtil.getDefaultTmpDir().getAbsolutePath() + "/";
```
nh13: slash

src/test/java/picard/sam/SAMSplitterTest.java
```diff
@@ -0,0 +1,108 @@
...
+        final File out = new File(tmpDir + "1.bam");
+        final SamReader outReader = SamReaderFactory.makeDefault().open(out);
+        final SamReader inReader = SamReaderFactory.makeDefault().open(PAIRED_FILE);
+        VALIDATE_SAM_TESTER.assertSamValid(out);
+        Assert.assertEquals(inReader.iterator().toList().size(), outReader.iterator().toList().size());
+    }
+
+    @Test(expectedExceptions = PicardException.class)
+    public void testZeroOutput() throws IOException {
+        final String tmpDir = IOUtil.getDefaultTmpDir().getAbsolutePath() + "/";
```
nh13: slash

src/main/java/picard/sam/SAMSplitter.java
```diff
@@ -0,0 +1,183 @@
...
+        if (header.getSortOrder() == SAMFileHeader.SortOrder.coordinate) {
+            throw new PicardException("Cannot split coordinate sorted bam. Only queryname sorted (or querygrouped) " +
+                    "input will ensure that templates are not split into multiple files.");
+        }
+
+        final Map<String, SAMRecord> firstSeenMates = new HashMap<>();
+        final SAMFileWriterFactory factory = new SAMFileWriterFactory();
+        double unitConverter = PAIRED_INPUT ? 2.0 : 1.0;
+
+        int splitToNFiles = (int) Math.round((TOTAL_READS_IN_INPUT / unitConverter) / SPLIT_TO_N_TEMPLATES);
```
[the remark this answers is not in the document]
meganshand: done

src/main/java/picard/sam/SAMSplitter.java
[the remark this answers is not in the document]
meganshand: done

src/main/java/picard/sam/SAMSplitter.java
meganshand: done

src/main/java/picard/sam/SAMSplitter.java
meganshand: done

src/main/java/picard/sam/SAMSplitter.java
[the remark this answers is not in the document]
meganshand: I added an argument with a default of "shard"

src/main/java/picard/sam/SAMSplitter.java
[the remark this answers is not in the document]
meganshand: done

src/test/java/picard/sam/SAMSplitterTest.java
meganshand: I copied the file into a new directory.

src/test/java/picard/sam/SAMSplitterTest.java
meganshand: fixed

src/test/java/picard/sam/SAMSplitterTest.java
meganshand: It's an abstract method in `CommandLineProgramTest`. It looks like most test classes use it without `@Override`.

src/test/java/picard/sam/SAMSplitterTest.java
meganshand: done

src/test/java/picard/sam/SAMSplitterTest.java
meganshand: done

src/test/java/picard/sam/SAMSplitterTest.java
meganshand: The problem is that the version in the header could be different. Is there a way to compare the header excluding versions? For now I can just change the input header, but it that brittle?

src/test/java/picard/sam/SAMSplitterTest.java
meganshand: done

src/test/java/picard/sam/SAMSplitterTest.java
meganshand: done

src/test/java/picard/sam/SAMSplitterTest.java
meganshand: done

src/test/java/picard/sam/SAMSplitterTest.java
meganshand: done

src/test/java/picard/sam/SAMSplitterTest.java
meganshand: It shouldn't, I fixed it.

yfarjoun:
@megan, I think that @nh13  idea is that each shard there will be many consecutive blocks, and since the concern around records that span blocks and mates that span blocks will only be localized to the boundary between the blocks. If the number of blocks per shard is >>1 this can result in significant savings. 

Now, while I think that this idea is worth thinking about, there are a few obstacles that @nh13 might be making light of....the major one is that we normally have no index for an unmapped bam. So one cannot "jump" to the right place. Without an index it isn't clear (to me) how to figure out where is the beginning (or end) of the record that spans a block.

I suspect that enhancements to this tool, and hints regarding how to overcome the technical difficulties would be appreciated....but in the meantime, we need to have a tool that works...even if not the most efficient.

src/test/java/picard/sam/SAMSplitterTest.java
yfarjoun: hmmm. is it possible to make the output bams the same version as the inputs? afterall, splitting a bam shouldn't change its version...if this is a thing, please make sure it's tested...

src/main/java/picard/sam/SAMSplitter.java
yfarjoun: I've seen other tools use strange things other thatn OUTPUT, but I agree with @nh13 that it's easier to assume that OUTPUT will define the output somehow...

src/main/java/picard/sam/SplitSamBySize.java
```diff
@@ -0,0 +1,152 @@
...
+            writers[writerIndex].addAlignment(currentRecord);
+            lastReadName = currentRecord.getReadName();
+            recordsWritten++;
+            progress.record(currentRecord);
+        }
+
+        for (final SAMFileWriter w : writers) {
+            w.close();
+        }
+
```
yfarjoun: extra nl

src/main/java/picard/sam/SplitSamBySize.java
```diff
@@ -0,0 +1,152 @@
...
+        return 0;
+    }
+
+    private SAMFileWriter[] generateWriters(final SAMFileWriterFactory factory, final SAMFileHeader header, final int splitToNFiles) {
+        final int digits = String.valueOf(splitToNFiles).length();
+        final DecimalFormat fileNameFormatter = new DecimalFormat(OUT_PREFIX + "_" + String.format("%0" + digits + "d", 0) + BamFileIoUtils.BAM_FILE_EXTENSION);
+        int fileIndex = 1;
+        SAMFileWriter[] writers = new SAMFileWriter[splitToNFiles];
```
yfarjoun: final

src/main/java/picard/sam/SplitSamBySize.java
```diff
@@ -0,0 +1,152 @@
...
+        for (int i = 0; i < splitToNFiles; i++) {
+            writers[i] = factory.makeSAMOrBAMWriter(header, true, new File(OUTPUT_DIR, fileNameFormatter.format(fileIndex++)));
+        }
+        return writers;
+    }
+
+    protected String[] customCommandLineValidation() {
+        if (TOTAL_READS_IN_INPUT < 1) {
+            return new String[]{
+                    "Cannot set TOTAL_READS_IN_INPUT to a number less than 1."
```
yfarjoun: add a "found ..." to the end of the message

src/main/java/picard/sam/SplitSamBySize.java
```diff
@@ -0,0 +1,152 @@
...
+        final SamReader reader = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(INPUT);
+        final SAMFileHeader header = reader.getFileHeader();
+        if (header.getSortOrder() == SAMFileHeader.SortOrder.coordinate) {
+            log.warn("Splitting a coordinate sorted bam may result in invalid bams " +
+                    "that do not always contain each read's mate in the same bam.");
+        }
+
+        final SAMFileWriterFactory factory = new SAMFileWriterFactory();
+
+        int splitToNFiles = SPLIT_TO_N_FILES != 0 ? SPLIT_TO_N_FILES : (int) Math.ceil(TOTAL_READS_IN_INPUT / SPLIT_TO_N_READS);
```
yfarjoun: you need to "promote" either the numerator or denominator to double so that the division isn't truncated. 

My guess is that if you currently have TOTAL_READS_IN_INPUT=11 and SPLIT_TO_N_READS=6 you'll get one file, but your code implies that you would like it to be 2.

src/main/java/picard/sam/SplitSamBySize.java
```diff
@@ -0,0 +1,152 @@
...
+    @Argument(shortName = "N_FILES", doc = "Split to N files.", optional = true, mutex = {"SPLIT_TO_N_READS"})
+    public int SPLIT_TO_N_FILES;
+
+    @Argument(shortName = "TOTAL_READS", doc = "Total number of reads in the input file.")
+    public int TOTAL_READS_IN_INPUT;
+
+    @Argument(shortName = "O", doc = "Directory in which to output the split BAM files.")
+    public File OUTPUT_DIR;
+
+    @Argument(shortName = "OUT_PREFIX", doc = "Prefix to name output files.")
```
yfarjoun: make the doc more explicit..."Output files will be named <OUT_PREFIX>_N.bam, where N enumerates the output file."

src/main/java/picard/sam/SplitSamBySize.java
```diff
@@ -0,0 +1,152 @@
...
+            mutex = {"SPLIT_TO_N_FILES"})
+    public int SPLIT_TO_N_READS;
+
+    @Argument(shortName = "N_FILES", doc = "Split to N files.", optional = true, mutex = {"SPLIT_TO_N_READS"})
+    public int SPLIT_TO_N_FILES;
+
+    @Argument(shortName = "TOTAL_READS", doc = "Total number of reads in the input file.")
+    public int TOTAL_READS_IN_INPUT;
+
+    @Argument(shortName = "O", doc = "Directory in which to output the split BAM files.")
```
yfarjoun: There should be a StandardOptionDefinitions.OUTPUT

src/main/java/picard/sam/SplitSamBySize.java
```diff
@@ -0,0 +1,152 @@
...
+        for (final SAMRecord currentRecord : reader) {
+            if (recordsWritten >= readsPerFile && !lastReadName.equals(currentRecord.getReadName())) {
+                writerIndex++;
+                recordsWritten = 0;
+            }
+            writers[writerIndex].addAlignment(currentRecord);
+            lastReadName = currentRecord.getReadName();
+            recordsWritten++;
+            progress.record(currentRecord);
+        }
```
yfarjoun: Would you like to emit a warning if the number of reads found differs from the number of reads predicted?

src/main/java/picard/sam/SplitSamBySize.java
```diff
@@ -0,0 +1,152 @@
...
+        if (header.getSortOrder() == SAMFileHeader.SortOrder.coordinate) {
+            log.warn("Splitting a coordinate sorted bam may result in invalid bams " +
+                    "that do not always contain each read's mate in the same bam.");
+        }
+
+        final SAMFileWriterFactory factory = new SAMFileWriterFactory();
+
+        int splitToNFiles = SPLIT_TO_N_FILES != 0 ? SPLIT_TO_N_FILES : (int) Math.ceil(TOTAL_READS_IN_INPUT / SPLIT_TO_N_READS);
+        int readsPerFile = (int) Math.ceil(TOTAL_READS_IN_INPUT / (double) splitToNFiles);
+        SAMFileWriter[] writers = generateWriters(factory, header, splitToNFiles);
```
yfarjoun: What's the advantage of setting up the writers in advance? As far as I can see the only actual use of knowing how many writers you'll have is that you are able to put the "correct" amount of zeros in the output filename....but even that doesn't require that you create the writers in advance. If on the other hand you generate the writers on the fly, the tool would be able to by used without knowledge of the size of the input...

lbergelson:
@meganshand Hadoop-bam splits bam files by diving at random into a block and trying to identify read starts.  They use it for this exact problem.  It's super sketchy and gross, but empirically it works.  We use it in spark to split bams into blocks of approximately equal file size.

lbergelson:
We can also generate a new type of index called a splitting index that records safe split points.  If you generate that at the same time that you generate the bai file it's an inexpensive add-on.  (We have a tool that does that.)

lbergelson:
On further reflection, you're looking at unmapped bams, so you won't have generated any index.  Scratch that previous comment.

meganshand:
@lbergelson That's quite the solution! If it works then yes this tool could definitely be split by block, but at the moment I think I will save that for future work.

lbergelson:
@meganshand That seems entirely reasonable.

src/main/java/picard/sam/SplitSamBySize.java
```diff
@@ -0,0 +1,152 @@
...
+        int fileIndex = 1;
+        SAMFileWriter[] writers = new SAMFileWriter[splitToNFiles];
+        for (int i = 0; i < splitToNFiles; i++) {
+            writers[i] = factory.makeSAMOrBAMWriter(header, true, new File(OUTPUT_DIR, fileNameFormatter.format(fileIndex++)));
+        }
+        return writers;
+    }
+
+    protected String[] customCommandLineValidation() {
+        if (TOTAL_READS_IN_INPUT < 1) {
```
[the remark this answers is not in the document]
yfarjoun: not sure what happens if someone gives TOTAL_READS_IN_INPUT=0 on the commandline...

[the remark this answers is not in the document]
yfarjoun: or rather SPLIT_TO_N_FILES=0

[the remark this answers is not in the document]
meganshand: Ok, I tried that and it should now be caught by `SPLIT_TO_N_FILES <= 1 && SPLIT_TO_N_READS <= 1`.

src/test/java/picard/sam/SplitSamBySizeTest.java
```diff
@@ -71,8 +68,13 @@ public void testOkFile() throws IOException {
         final SAMRecordIterator inputIter = input.iterator();
         final SamReader reader1 = SamReaderFactory.makeDefault().open(out1);
         final SamReader reader2 = SamReaderFactory.makeDefault().open(out2);
+        final SAMFileHeader inputHeader = input.getFileHeader();
+        final SAMFileHeader reader1Header = reader1.getFileHeader();
 
-        Assert.assertEquals(input.getFileHeader(), reader1.getFileHeader());
+        // The output BAM versions might not match the input header version if the input is outdated.
+        inputHeader.setAttribute(SAMFileHeader.VERSION_TAG, reader1Header.getVersion());
```
yfarjoun: When the version changes can you emit a warning?

meganshand: done

src/main/java/picard/sam/SplitSamBySize.java
```diff
@@ -0,0 +1,162 @@
...
+ * multiple BAM files, each with an approximately equal number of reads. This will retain the sort order
+ * within each output BAM and if the BAMs are concatenated in order (output files are named
+ * numerically) the order of the reads will match the original BAM.
+ */
+@CommandLineProgramProperties(
+        summary = SplitSamBySize.USAGE_SUMMARY + SplitSamBySize.USAGE_DETAILS,
+        oneLineSummary = SplitSamBySize.USAGE_SUMMARY,
+        programGroup = SamOrBam.class)
+@DocumentedFeature
+public class SplitSamBySize extends CommandLineProgram {
```
nh13: `SplitSamBySize` -> size of what?  Perhaps `SplitSamByNumberOfReads` or is that too verbose?

src/main/java/picard/sam/SplitSamBySize.java
```diff
@@ -0,0 +1,162 @@
...
+    public File INPUT;
+
+    @Argument(shortName = "N_READS", doc = "Split to have approximately N reads per output file.", mutex = {"SPLIT_TO_N_FILES"})
+    public int SPLIT_TO_N_READS;
+
+    @Argument(shortName = "N_FILES", doc = "Split to N files.", mutex = {"SPLIT_TO_N_READS"})
+    public int SPLIT_TO_N_FILES;
+
+    @Argument(shortName = "TOTAL_READS", doc = "Total number of reads in the input file.")
+    public int TOTAL_READS_IN_INPUT;
```
nh13: I am not sure why this option is needed when `SPLIT_TO_N_READS` is given.  Can't we just output the first `SPLIT_TO_N_READS` reads, then create a new writer?

src/main/java/picard/sam/SplitSamBySize.java
```diff
@@ -0,0 +1,162 @@
...
+            writers[writerIndex].addAlignment(currentRecord);
+            lastReadName = currentRecord.getReadName();
+            readsWritten++;
+            progress.record(currentRecord);
+        }
+
+        if (progress.getCount() != TOTAL_READS_IN_INPUT) {
+            log.warn(String.format("The TOTAL_READS_IN_INPUT (%s) provided does not match the reads found in the " +
+                            "input file (%s).", TOTAL_READS_IN_INPUT, progress.getCount())
+                   );
```
nh13: I think this is an indent error.

src/main/java/picard/sam/SplitSamBySize.java
meganshand: The problem I'm trying to avoid when I don't have `TOTAL_READS_IN_INPUT` is that there could be a remainder number of reads in the final file that ranges from none to `SPLIT_TO_N_READS`. Since the use case I was thinking of was to split for alignment, I want to avoid aligning a file with too small a number of reads in it. One potential solution to this would be to combine the last two files, but then you run the risk of having one file that's double the size of the others, which also isn't ideal.

nh13: I see, I just think knowing the total # of input reads ahead of time is onerous or heavy-handed for users.   Why not just have re-write the last two files if the last one will have too few reads?

meganshand: Do you mean, to only combine the last two files if the last file has two few reads and otherwise leave them as two files? One potential issue there is that I have to define what is too few reads. I suppose I could make that a user specified value? 

Another option could be to have this tool optionally calculate the `TOTAL_READS_IN_INPUT` itself, but then wouldn't I either have to traverse the BAM twice or have an index file? Is there a quick way to get the total number of reads in Picard?

meganshand: Ok, I think I have a new solution. Instead of splitting the file to keep number of reads approximately equivalent in each file, I can split by on disk file size so that the number of bytes in the output files will be approximately equivalent. Obviously I'll still keep query-grouped reads together. This eliminates needing to know the number of reads in the input since it's very easy to get the size of the file.

src/main/java/picard/sam/SplitSamBySize.java
```diff
@@ -87,7 +84,15 @@
     protected int doWork() {
         IOUtil.assertFileIsReadable(INPUT);
         IOUtil.assertDirectoryIsWritable(OUTPUT);
-        final SamReader reader = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(INPUT);
+        SeekablePathStream inStream;
+        try {
+            inStream = new SeekablePathStream(INPUT.toPath());
+        } catch (IOException e) {
+            e.printStackTrace();
```
yfarjoun: add informative log..."Failed to open stream around file xxx" etc.

src/main/java/picard/sam/SplitSamBySize.java
```diff
@@ -101,29 +106,52 @@ protected int doWork() {
...
-                   );
+            // We finished the block, write it out
+            if (currentBlockEnd != bufferedBlockEnd || !iterator.hasNext()) {
+                int idxInBlock = 0;
+                final long bufferedBlockSize = bufferedBlockEnd - bufferedBlockStart;
+                for (final SAMRecord record : blockRecords) {
+                    // Estimate the current position in bytes by subdividing the buffered block by the number of records in it
+                    final long currentPosition = bufferedBlockStart + ((idxInBlock * bufferedBlockSize) / blockRecords.size());
+                    if (currentPosition >= inputBytesPerFile * (writerIndex + 1) && !lastReadName.equals(record.getReadName())) {
+                        writerIndex++;
```
yfarjoun: if your estimation is wrong, would you get more than N_FILES files? or possibly less? it would be good to protect against this somehow (not change the writer if it's already N_FILES-1 and also emit an warning if the actual number of files with data differs from the requested number. )

yfarjoun: I have to say that I'm having a tough time following the logic used here. can you explain how it works, in a comment perhaps..

meganshand:
It seems that splitting by byte size is overly complicated for what this tool is meant for. In order to avoid the user having to know the number of reads in their input file I've made it an optional argument, and if it's not provided we traverse the input twice, the first time to get that total count. This should provide some flexibility. I also changed when we open the writers so we don't have to open them all before beginning. Thanks for taking a look at this @yfarjoun @nh13!

src/main/java/picard/sam/SplitSamByNumberOfReads.java
```diff
@@ -0,0 +1,168 @@
...
+            currentWriter.addAlignment(currentRecord);
+            lastReadName = currentRecord.getReadName();
+            readsWritten++;
+            progress.record(currentRecord);
+        }
+
+        if (progress.getCount() != totalReads) {
+            log.warn(String.format("The totalReads (%s) provided does not match the reads found in the " +
+                    "input file (%s). Files may not be split evenly or number of files may not " +
+                    "match what was requested.", totalReads, progress.getCount())
```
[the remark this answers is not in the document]
yfarjoun: maybe also report the number of reads that the other files have (on average)

src/main/java/picard/sam/SplitSamByNumberOfReads.java
```diff
@@ -0,0 +1,168 @@
...
+            log.warn("Splitting a coordinate sorted bam may result in invalid bams " +
+                    "that do not always contain each read's mate in the same bam.");
+        }
+        if (!header.getVersion().equals(SAMFileHeader.CURRENT_VERSION)) {
+            log.warn(String.format("Input file's version is %s, but the current SAM format version is %s. Outputs will be written " +
+                    "with current version.", header.getVersion(), SAMFileHeader.CURRENT_VERSION));
+        }
+
+        final ProgressLogger firstPassProgress = new ProgressLogger(log, 1000000, "Counted");
+        if (TOTAL_READS_IN_INPUT == 0) {
```
yfarjoun: Can you check that input file is a regular file to avoid eating-up a stream?

src/main/java/picard/sam/SplitSamByNumberOfReads.java
```diff
@@ -0,0 +1,168 @@
...
+            "     I=paired_unmapped_input.bam \\<br />" +
+            "     OUTPUT=out_dir \\ <br />" +
+            "     TOTAL_READS_IN_INPUT=800000000 \\ <br />" +
+            "     SPLIT_TO_N_READS=48000000" +
+            "</pre>" +
+            "<hr />";
+    @Argument(doc = "Input SAM/BAM file to split", shortName = StandardOptionDefinitions.INPUT_SHORT_NAME)
+    public File INPUT;
+
+    @Argument(shortName = "N_READS", doc = "Split to have approximately N reads per output file.", mutex = {"SPLIT_TO_N_FILES"})
```
yfarjoun: explain in what way the actual value will differ from the argument.

src/main/java/picard/sam/SplitSamByNumberOfReads.java
meganshand: I don't know how to check if a file is a stream directly, but I put an error and non-zero return code before the first pass iteration begins. It should still fail quickly with a more informative error message.

src/main/java/picard/sam/SplitSamByNumberOfReads.java
```diff
@@ -69,7 +70,9 @@
     @Argument(doc = "Input SAM/BAM file to split", shortName = StandardOptionDefinitions.INPUT_SHORT_NAME)
     public File INPUT;
 
-    @Argument(shortName = "N_READS", doc = "Split to have approximately N reads per output file.", mutex = {"SPLIT_TO_N_FILES"})
+    @Argument(shortName = "N_READS", doc = "Split to have approximately N reads per output file. The actual number of " +
+            "reads per output file will vary by no more than the number of output files times one minus the maximum " +
```
yfarjoun: I think you swapped "times" and "minus one" leading to a description seems wrong.

src/main/java/picard/sam/SplitSamByNumberOfReads.java
```diff
@@ -106,11 +110,17 @@ protected int doWork() {
         final ProgressLogger firstPassProgress = new ProgressLogger(log, 1000000, "Counted");
         if (TOTAL_READS_IN_INPUT == 0) {
             final SamReader firstPass = readerFactory.referenceSequence(REFERENCE_SEQUENCE).open(INPUT);
+            final SAMRecordIterator firstPassIterator = firstPass.iterator();
+            if (readerIterator.hasNext() && !firstPassIterator.hasNext()) {
```
yfarjoun: hmmm. according to the javadoc here: htsjdk.samtools.BAMFileReader#getIterator() only one iterator is allowed to exist at a given time, so I'm not sure how this is working....I think that to do it properly, you need to open another reader to the same file.

meganshand:
@yfarjoun Tests are passing again, back to you!

src/main/java/picard/sam/SplitSamByNumberOfReads.java
```diff
@@ -0,0 +1,185 @@
...
+ */
+@CommandLineProgramProperties(
+        summary = SplitSamByNumberOfReads.USAGE_SUMMARY + SplitSamByNumberOfReads.USAGE_DETAILS,
+        oneLineSummary = SplitSamByNumberOfReads.USAGE_SUMMARY,
+        programGroup = SamOrBam.class)
+@DocumentedFeature
+public class SplitSamByNumberOfReads extends CommandLineProgram {
+    static final String USAGE_SUMMARY = "Splits a SAM or BAM file to multiple BAMs.";
+    static final String USAGE_DETAILS = "This tool splits the input query-grouped SAM/BAM file into multiple BAM files " +
+            "while maintaining the sort order. This can be used to split a large unmapped BAM in order to parallelize alignment." +
```
yfarjoun: space after `.` in string.

src/main/java/picard/sam/SplitSamByNumberOfReads.java
```diff
@@ -89,9 +95,14 @@
 
     protected int doWork() {
         IOUtil.assertFileIsReadable(INPUT);
+        if (TOTAL_READS_IN_INPUT == 0 && !Files.isRegularFile(INPUT.toPath())) {
+            log.error("INPUT is not a regular file. If TOTAL_READS_IN_INPUT is not supplied, INPUT cannot be a stream.");
```
yfarjoun: print the value of INPUT here as well

src/main/java/picard/sam/SplitSamByNumberOfReads.java
```diff
@@ -106,11 +117,13 @@ protected int doWork() {
         final ProgressLogger firstPassProgress = new ProgressLogger(log, 1000000, "Counted");
         if (TOTAL_READS_IN_INPUT == 0) {
             final SamReader firstPass = readerFactory.referenceSequence(REFERENCE_SEQUENCE).open(INPUT);
+            final SAMRecordIterator firstPassIterator = firstPass.iterator();
```
yfarjoun: sorry, I didn't realiz that you had two readers open already and my comments made you move from a for-loop to a while...I think that the original for loop is nicer.... :-/

src/main/java/picard/sam/SplitSamByNumberOfReads.java
```diff
@@ -106,11 +117,13 @@ protected int doWork() {
         final ProgressLogger firstPassProgress = new ProgressLogger(log, 1000000, "Counted");
         if (TOTAL_READS_IN_INPUT == 0) {
             final SamReader firstPass = readerFactory.referenceSequence(REFERENCE_SEQUENCE).open(INPUT);
```
yfarjoun: rename firstPass -> firstPassReader

src/main/java/picard/sam/SplitSamByNumberOfReads.java
```diff
@@ -137,15 +151,18 @@ protected int doWork() {
             readsWritten++;
             progress.record(currentRecord);
         }
+        currentWriter.close();
+        CloserUtil.close(reader);
 
         if (progress.getCount() != totalReads) {
             log.warn(String.format("The totalReads (%s) provided does not match the reads found in the " +
```
yfarjoun: shouldn't many of the `%s` in the format string be %d ?

src/test/java/picard/sam/SplitSamByNumberOfReadsTest.java
```diff
@@ -0,0 +1,257 @@
...
+
+        Assert.assertEquals(reader1.getFileHeader(), reader2.getFileHeader());
+        VALIDATE_SAM_TESTER.assertSamValid(out1);
+        VALIDATE_SAM_TESTER.assertSamValid(out2);
+
+        compareInputWithOutputs(reader1, reader2, inputIter, 6);
+    }
+
+    @Test
+    public void testStreamWithoutTotalReads() throws IOException, NoSuchFieldException, IllegalAccessException {
```
yfarjoun: no longer throws NoSuchFieldException

src/test/java/picard/sam/SplitSamByNumberOfReadsTest.java
```diff
@@ -0,0 +1,257 @@
...
+import org.testng.annotations.Test;
+import picard.cmdline.CommandLineProgramTest;
+import picard.sam.testers.ValidateSamTester;
+import com.google.common.collect.Iterators;
+
+import java.io.File;
+import java.io.FileDescriptor;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.lang.reflect.Field;
```
yfarjoun: optimize imports

yfarjoun:
hopefully final round of nit-picks!

meganshand:
@yfarjoun Thanks, I think I got all of your comments.

src/main/java/picard/sam/SplitSamByNumberOfReads.java
```diff
@@ -155,10 +148,10 @@ protected int doWork() {
         CloserUtil.close(reader);
 
         if (progress.getCount() != totalReads) {
-            log.warn(String.format("The totalReads (%s) provided does not match the reads found in the " +
-                            "input file (%s). Files may not be split evenly or number of files may not " +
-                            "match what was requested. There were %s files generated each with around %s " +
-                            "reads except the last file which contained %s reads.",
+            log.warn(String.format("The totalReads (%d) provided does not match the reads found in the " +
+                            "input file (%d). Files may not be split evenly or number of files may not " +
```
yfarjoun: input file is a string, not number...

yfarjoun:
one final-final comment and you've got my 👍 

@nh13 can you re-review  as you are currently listed as requesting a change.

src/main/java/picard/sam/SplitSamByNumberOfReads.java
meganshand: Sorry, that's printing the number of reads in the input file.

yfarjoun: I read too quickly.... 
:+1:

yfarjoun:
merging after not hearing from @nh13. please open an issue if there's a problem with this implementation.

PR #971, opened 2017-11-08
Enable LiftOverVcf to emit a variant when the reference base has changed 

yfarjoun:
### Description

Enables LiftOverVcf to process variants where the reference changed, swapping the reference and the alternate alleles. 

This has been implemented for bi-allelic SNPs only. 
----

#### Content

#### Review


src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -35,6 +37,17 @@
...
+    /**
+     * Attribute used to store the fact that the alt and ref alleles of the variant have been swapped, while all the INFO annotations have not.
+     */
+
+    public static final Collection<String> DEFAULT_TAGS_TO_REVERSE = Arrays.asList("AF");
+    public static final Collection<String> DEFAULT_TAGS_TO_DROP = Arrays.asList("MAX_AF");
+
+    public static final String SWAPPED_ALLELES = "SwappedAlleles";
+
+    public static Log log = Log.getInstance(LiftoverUtils.class);
```
pshapiro4broad: can be `final`

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -192,6 +207,81 @@ protected static GenotypesContext fixGenotypes(final GenotypesContext originals,
     }
 
     /**
+     * method to swap the reference and alt alleles of a bi-allelic, SNP
+     *
+     * @param vc
```
pshapiro4broad: Provide doc text for `@param` and `@return` or leave them out of the javadoc

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -192,6 +207,81 @@ protected static GenotypesContext fixGenotypes(final GenotypesContext originals,
...
+        swappedBuilder.attribute(SWAPPED_ALLELES, true);
+
+        // You need to get the baseString in order to forget who's reference and who's variant
+        swappedBuilder.alleles(Arrays.asList(vc.getAlleles().get(1).getBaseString(), vc.getAlleles().get(0).getBaseString()));
+
+
+        final Map<Allele, Allele> alleleMap = new HashMap<>();
+        for (int idx = 0; idx < vc.getAlleles().size(); idx++) {
+            // The index of the allele changed from 1 to 0 and vice versa, but the bases remain the same.
+            alleleMap.put(vc.getAlleles().get(idx), swappedBuilder.getAlleles().get(1 - idx));
```
pshapiro4broad: IMO a loop here is less clear than writing out the two `put()` calls you'd otherwise have.

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -192,6 +207,81 @@ protected static GenotypesContext fixGenotypes(final GenotypesContext originals,
...
+            swappedGenotypes.add(builder.make());
+        }
+        swappedBuilder.genotypes(swappedGenotypes);
+
+        for (String key : vc.getAttributes().keySet()) {
+            if (annotationsToDrop.contains(key)) {
+                swappedBuilder.rmAttribute(key);
+            }
+
+            if (annotationsToReverse.contains(key)) {
```
pshapiro4broad: `else if` here? Seems like you can't both drop and reverse an annotation.

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -192,6 +207,81 @@ protected static GenotypesContext fixGenotypes(final GenotypesContext originals,
...
+                final double afLikeAttribute = vc.getAttributeAsDouble(key, -1);
+
+                if (afLikeAttribute == -1) {
+                    log.warn("Trying to reverse attribute " + key +
+                            "but found value that parsable as a double (" + vc.getAttribute(key, "") +
```
pshapiro4broad: found value that _isn't_ parsable ?

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -192,6 +207,81 @@ protected static GenotypesContext fixGenotypes(final GenotypesContext originals,
...
+                            ") in variant " + vc + "results might be wrong.");
+                }
+
+                if (afLikeAttribute < 0 || afLikeAttribute > 1) {
+                    log.warn("Trying to reverse attribute " + key +
+                            "but found value that isn't between 0 and 1 (" + afLikeAttribute + ") in variant " + vc + "results might be wrong.");
```
pshapiro4broad: Both messages seem to be missing a `:` between the error message and the incorrect values. E.g. I'd expect to see "Trying to reverse attribute X but found value that isn't between 0 and 1: (X) in variant V. Results might be wrong."

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -192,6 +207,81 @@ protected static GenotypesContext fixGenotypes(final GenotypesContext originals,
...
+                swappedBuilder.rmAttribute(key);
+            }
+
+            if (annotationsToReverse.contains(key)) {
+                final double afLikeAttribute = vc.getAttributeAsDouble(key, -1);
+
+                if (afLikeAttribute == -1) {
+                    log.warn("Trying to reverse attribute " + key +
+                            "but found value that parsable as a double (" + vc.getAttribute(key, "") +
+                            ") in variant " + vc + "results might be wrong.");
```
pshapiro4broad: Missing a space before `results`, also looks like a typo, extra `might be wrong.` here and below.

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -209,6 +299,16 @@ protected static void leftAlignVariant(final VariantContextBuilder builder, fina
         int theStart = start;
         int theEnd = end;
 
+        final String refString = StringUtil.bytesToString(referenceSequence.getBases(), start-1, end - start + 1);
+        //make sure that referenceAllele matches reference
+        final Allele refAllele = alleles.stream().filter(Allele::isReference).findAny().orElse(null);
+        if (refAllele == null) throw new RuntimeException("How did that happen? No reference allele?");
```
pshapiro4broad: A shorter way to write this is to use `orElseThrow()` instead of `orElse()`. e.g. `alleles.stream().filter(Allele::isReference).findAny().orElseThrow(new RuntimeException("How did that happen? No reference allele?"))`

src/test/java/picard/util/LiftoverVcfTest.java
```diff
@@ -392,6 +465,77 @@ public void testFlipIndel(final VariantContext source, final ReferenceSequence r
...
+
+        builder.genotypes(genotypeBuilder.make());
+        result_builder.genotypes(resultGenotypeBuilder.make());
+
+        tests.add(new Object[]{builder.make(), reference, result_builder.make()});
+        return tests.iterator();
+    }
+
+
+        @Test(dataProvider = "snpWithChangedRef")
```
pshapiro4broad: fix indent, this line and next

src/test/java/picard/util/LiftoverVcfTest.java
```diff
@@ -474,8 +619,8 @@ public void testFlipIndel(final VariantContext source, final ReferenceSequence r
 
         // AA/A in initial polyA repeat -> CA/C at the beginning
         result_builder.start(1).stop(2).alleles(CollectionUtil.makeList(RefCA, C));
-        builder.source("test3");
-        for (start = 2; start <= 11; start++) {
+        for (start = 2; start <= 10; start++) {
+            builder.source("test3-"+ start);
```
pshapiro4broad: should have a space between the string and the `+`

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -209,6 +301,16 @@ protected static void leftAlignVariant(final VariantContextBuilder builder, fina
         int theStart = start;
         int theEnd = end;
 
+        final String refString = StringUtil.bytesToString(referenceSequence.getBases(), start-1, end - start + 1);
+        //make sure that referenceAllele matches reference
+        final Allele refAllele = alleles.stream().filter(Allele::isReference).findAny().orElseThrow(()->new RuntimeException("How did that happen? No reference allele?"));
+
```
eitanbanks: Do we really need this check here?  It seems more appropriate to put this check where the ref allele was created.

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -161,13 +164,14 @@
      */
     public static final String ATTEMPTED_LOCUS = "AttemptedLocus";
 
+
```
eitanbanks: extra space added

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -226,6 +230,12 @@ protected int doWork() {
         if (WRITE_ORIGINAL_POSITION) {
             for (final VCFInfoHeaderLine line : ATTRS) outHeader.addMetaDataLine(line);
         }
+
+        outHeader.addMetaDataLine(new VCFInfoHeaderLine(LiftoverUtils.SWAPPED_ALLELES, 0, VCFHeaderLineType.Flag,
+                "The REF and the ALT alleles have been swapped due to changes in the reference. " +
+                "NOTA BENE: No INFO field annotations reflect this swap, and in the genotypes, " +
```
eitanbanks: Well, TAGS_TO_REVERSE do reflect this swap...

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -362,6 +373,13 @@ private void tryToAddVariant(final VariantContext vc, final ReferenceSequence re
                 final String refString = StringUtil.bytesToString(ref, vc.getStart() - 1, vc.getEnd() - vc.getStart() + 1);
 
                 if (!refString.equalsIgnoreCase(allele.getBaseString())) {
+                    // consider that the ref and the alt may have been swapped in a simple biallelic SNP
+                    if (vc.getAlleles().size() == 2 &&
+                            vc.getAlleles().stream().map(Allele::length).allMatch(l -> l == 1) &&
```
eitanbanks: This is weird.  Why not use isBiallelic() and isSNP()?

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -362,6 +373,13 @@ private void tryToAddVariant(final VariantContext vc, final ReferenceSequence re
...
+                            refString.equalsIgnoreCase(vc.getAlleles().stream().filter(Allele::isNonReference).findFirst().get().getBaseString())) {
+                        sorter.add(LiftoverUtils.swapRefAlt(vc,TAGS_TO_REVERSE, TAGS_TO_DROP));
```
eitanbanks: I feel like there should be one more piece of logic here, but will give you the final say in overriding this.
If the original AF was 1.0 then this isn't really an allele swap; rather it's a reference fix and we should just remove the record (since no samples are polymorphic in the new reference).

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -192,6 +207,83 @@ protected static GenotypesContext fixGenotypes(final GenotypesContext originals,
...
+        if (!vc.isBiallelic() || !vc.isSNP() ) {
+            throw new IllegalArgumentException("swapRefAlt can only process biallelic, SNPS, found " + vc.toString() );
+        }
+
+        final VariantContextBuilder swappedBuilder = new VariantContextBuilder(vc);
+
+        swappedBuilder.attribute(SWAPPED_ALLELES, true);
+
+        // You need to get the baseString in order to forget who's reference and who's variant
```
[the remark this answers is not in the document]
yfarjoun: I tried explaining why I put the getBaseString in the first place...but clearly my comment is unclear...I'll reword.

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -192,6 +207,83 @@ protected static GenotypesContext fixGenotypes(final GenotypesContext originals,
...
+            final GenotypeBuilder builder = new GenotypeBuilder(genotype).alleles(swappedAlleles);
+            if (genotype.hasAD() && genotype.getAD().length == 2) {
+                final int[] origAD = genotype.getAD();
+                builder.AD(new int[]{origAD[1], origAD[0]});
+            }
+
+            //Flip PL
+            if (genotype.hasPL() && genotype.getPL().length == 3) {
+                final int[] origPL = genotype.getPL();
+                builder.PL(new int[]{origPL[2], origPL[1], origPL[0]});
```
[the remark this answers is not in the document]
yfarjoun: using Commons version.

src/main/java/picard/util/LiftoverUtils.java
yfarjoun: hmmm. But that really doesn't fit with the flow of the original algorithm as the actual check against the reference happens at the end, before trying to add the VC...if you feel strongly about this, let's talk more in person.

src/main/java/picard/vcf/LiftoverVcf.java
yfarjoun: right....fixing.

src/main/java/picard/vcf/LiftoverVcf.java
yfarjoun: So, you would like the VC to be dropped if it turns out to be monomorphic after the allele swapping? My sense is that a monomorphic variant is easily removed post-hoc but that it could be useful for some to see them before being removed...

src/main/java/picard/vcf/LiftoverVcf.java
yfarjoun: because I was being too technical and forgot about these utility functions....

src/main/java/picard/vcf/LiftoverVcf.java
eitanbanks: Okay, so maybe we should keep it in but filter it?  Filter could be e.g. ReferenceFix.

src/main/java/picard/util/LiftoverUtils.java
eitanbanks: I don't feel strongly... it's just weird.  As a reviewer of the code it just looks out of place.
Can we move it up to the method that calls this?

yfarjoun:
Merging this as it was approved by @eitanbanks who only asked for minimal changes (which were done)

PR #972, opened 2017-11-09
Increase test coverage

yfarjoun:
### Description

Lots of new tests for various uncovered parts of the picard code-base. 
----

#### Content

#### Review


yfarjoun:
- fixed some bugs (discovered while testing) in CollectSequencingArtifactsMetrics and in ConvertArtifactsToOxoG.
- removed java bug workaround for a bug that was fixed in java7

yfarjoun:
@meganshand  is out, so I need someone else to review this....any takers?

src/test/java/picard/util/IntervalListToolsTest.java
```diff
@@ -0,0 +1,185 @@
...
+                {IntervalListTools.Action.SUBTRACT, totalBasesInDict - 60, 2 + totalContigsInDict},
+                {IntervalListTools.Action.SYMDIFF, totalBasesInDict - 61, 3 + totalContigsInDict},
+                {IntervalListTools.Action.OVERLAPS, totalBasesInDict - 150, 2 + totalContigsInDict},
+        };
+    }
+
+    @Test(dataProvider = "actionAndTotalBasesWithInvertData")
+    public void testActionsWithInvert(final IntervalListTools.Action action, final long bases, final int intervals) throws IOException {
+        final IntervalList il = tester(action, true);
+        Assert.assertEquals(il.getBaseCount(), bases, "unexpcted number of bases found.");
```
gbggrant: 'unexpected'

src/test/java/picard/util/IntervalListToolsTest.java
```diff
@@ -0,0 +1,185 @@
...
+        Assert.assertEquals(il.getIntervals().size(), intervals, "unexpcted number of intervals found.");
```
gbggrant: and here

src/test/java/picard/util/IntervalListToolsTest.java
```diff
@@ -0,0 +1,185 @@
...
+                {IntervalListTools.Action.SUBTRACT, 60, 2},
+                {IntervalListTools.Action.SYMDIFF, 61, 3},
+                {IntervalListTools.Action.OVERLAPS, 150, 2},
+        };
+    }
+
+    @Test(dataProvider = "actionAndTotalBasesData")
+    public void testActions(final IntervalListTools.Action action, final long bases, final int intervals) throws IOException {
+        final IntervalList il = tester(action);
+        Assert.assertEquals(il.getBaseCount(), bases, "unexpcted number of bases found.");
```
gbggrant: "unexpected" (and wrong in line below)

src/main/java/picard/TestStringUtilSplit.java
```diff
@@ -6,20 +6,21 @@
...
+public class TestStringUtilSplit {
     private static final String TEXT = "C0A69ACXX111213:6:1101:10000:144257\t83\t5\t128984606\t60\t76M\t=\t128984542\t-140\tAGTGTTAGAACTTCCTCCCCAAAGCATATACTTCAGTGGCAAGCTGTCCTGGATGAAGGTATGACCAACCAGATCA\t@FFFEECC>EFHBJIGIFGIEIJJJIHED<IEHIGIIJIIIGJIGJIIIIIJGGCJIIGIHHHBHGHFFDFFFC@@\tXT:A:U\tNM:i:0\tSM:i:37\tAM:i:37\tX0:i:1\tX1:i:0\tXM:i:0\tXO:i:0\tXG:i:0\tMD:Z:76";
 
     public static void main(String[] args) {
-        new Test().run();
+        new TestStringUtilSplit().run();
     }
 
-    public void run() {
+    public double run() {
```
pshapiro4broad: Why does this return a value? It doesn't appear to be used anywhere.

src/main/java/picard/TestStringUtilSplit.java
```diff
@@ -6,20 +6,21 @@
...
         final int ITERATIONS = 1000000;
         final String[] fields = new String[10000];
         final StopWatch watch = new StopWatch();
-
+        final long stringUtilTime;
         watch.start();
         for (int i=0; i<ITERATIONS; ++i) {
```
pshapiro4broad: Should have spaces around operators, e.g. `(int i = 0; i < ITERATIONS; ++i)`. Using command-option-L in intellij should fix them all.

src/main/java/picard/TestStringUtilSplit.java
```diff
@@ -6,20 +6,21 @@
...
             if (StringUtil.split(TEXT, fields, '\t') > 100) {
```
pshapiro4broad: Why is this test necessary? It looks like it will always be false, and including it adds noise to the benchmark time.

src/main/java/picard/TestStringUtilSplit.java
```diff
@@ -6,20 +6,21 @@
 import java.util.StringTokenizer;
 
 /**
- *
+ * A class that can be used to check which is faster, StringUtil.split or StringTokenizer.
+ * Currently, StringUtil is faster.
  */
-public class Test {
+public class TestStringUtilSplit {
```
pshapiro4broad: If this is a Test then it should be in the test source tree.

src/main/java/picard/TestStringUtilSplit.java
```diff
@@ -38,6 +42,8 @@ public void run() {
         }
         watch.stop();
         System.out.println("StringTokenizer took " + watch.getElapsedTime());
+        tokenizerTime = watch.getElapsedTime();
+        return (tokenizerTime + 0D) / stringUtilTime;
     }
 
     public int split(final String s, final String[] tokens, final String token) {
```
pshapiro4broad: `new StringTokenizer(s, token, false)` is the same as `new StringTokenizer(s, token)`

src/main/java/picard/analysis/CollectOxoGMetrics.java
```diff
@@ -313,7 +301,7 @@ protected int doWork() {
 
             // Skip sites at the end of chromosomes 
             final byte[] bases = refWalker.get(info.getSequenceIndex()).getBases();
-            if (pos < 3 || pos > bases.length - 3) continue;
+            if (pos <= CONTEXT_SIZE|| pos > bases.length - CONTEXT_SIZE) continue;
```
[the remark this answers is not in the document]
yfarjoun: yeah...I think it was wrong before! we only need to make sure that we are CONTEXT_SIZE bases away from the edges of the reference. a fixed value of 3 is incorrect in 2 ways: 

- it consistently avoids the second and second-to last bases in a reference (normally not a problem due to the N-masking in the ends)
- it would result in OOB exception if one runs it with CONTEXT_SIZE=5 (again, normally not a problem due to the lack of reads that manage to align so close to the edge of a contig)

src/test/java/picard/util/IntervalListToolsTest.java
yfarjoun: thanks.

src/test/java/picard/util/IntervalListToolsTest.java
yfarjoun: thanks.

src/main/java/picard/TestStringUtilSplit.java
yfarjoun: In one of the commits, I had it run in a test, that would affirm that using StringUtil is faster than StringTokenizer, but that test kept failing in travis (and passing on my computer...) so I removed the test....

src/main/java/picard/TestStringUtilSplit.java
yfarjoun: I agree. It's not being run.

src/main/java/picard/TestStringUtilSplit.java
yfarjoun: Leaving it in enables us to run in from within the main jar. given that it isn't being run during the tests, I think that's a better option.

src/main/java/picard/TestStringUtilSplit.java
yfarjoun: (not my code...) TIL

src/main/java/picard/TestStringUtilSplit.java
```diff
@@ -38,6 +42,8 @@ public void run() {
         }
         watch.stop();
         System.out.println("StringTokenizer took " + watch.getElapsedTime());
+        tokenizerTime = watch.getElapsedTime();
+        return (tokenizerTime + 0D) / stringUtilTime;
```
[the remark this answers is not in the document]
yfarjoun: ok

yfarjoun:
@gbggrant @pshapiro4broad

I had to rebase and it was messy...so I squashed all the initial commits, and kept two:
One commit that had all the changes that you already reviewed, (plus the changes needed for resolving the conflict) and one commit that addressed your comments.

yfarjoun:
argh. I need to rebase again...could I get a fast-ish review once that's done?

yfarjoun:
bleh. that was not fun...

had to rebase over the documentation changes...

re-review please?

src/main/java/picard/TestStringUtilSplit.java
pshapiro4broad: Don't we have a test .jar? If not, we should be able to generate one as part of the build.

If this is only run as part of a test, why does it need to be included in the main .jar? I'm not sure why this is being committed to the picard repo in the first place; it looks like test code that was run once to show that X was faster than Y. Maybe the test here is an assert that this speed difference continues to be true?

In general, do we have a way to annotate classes that are public and part of the main .jar but are not intended to be used (e.g. not part of the picard API)?

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -253,38 +253,68 @@
             IntervalList act(final List<IntervalList> list, final List<IntervalList> __) {
                 return IntervalList.concatenate(list);
             }
+
+            @Override
+            public boolean takesSecondInput() {
+                return false;
```
pshapiro4broad: If this is a boolean value that all enum instances must provide it would be simpler to model it as a boolean field in the enum, instead of a method.

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -253,38 +253,68 @@
...
+            }
         },
         INTERSECT("The sorted and merged set of all loci that are contained in all of the INPUTs.") {
-            @Override
-            IntervalList act(final List<IntervalList> list, final List<IntervalList> __) {
+            @Override IntervalList act(final List<IntervalList> list, final List<IntervalList> __) {
```
pshapiro4broad: `@Override` should be on its own line, e.g. the same as the rest of this file is formatted.

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -253,38 +253,68 @@
...
+
+            @Override
+            public boolean takesSecondInput() {
+                return false;
+            }
         },
         UNION("Like CONCATENATE but with UNIQUE and SORT implied, the result being the set-wise union of all INPUTS, " +
                 "with overlapping and abutting intervals merged into one.") {
             @Override
             IntervalList act(final List<IntervalList> list, final List<IntervalList> __) {
```
pshapiro4broad: I don't think `__` is normally used in java names (although I don't think it's completely disallowed as it is in C/C++). `unused` would be OK if we don't want to use `list2` here.

src/main/java/picard/vcf/GatherVcfs.java
```diff
@@ -61,32 +61,38 @@ public GatherVcfs() {
...
+            if (areAllBlockCompressed(INPUT) && areAllBlockCompressed(CollectionUtil.makeList(OUTPUT))) {
+                log.info("Gathering by copying gzip blocks. Will not be able to validate position non-overlap of files.");
+                if (CREATE_INDEX)
+                    log.warn("Index creation not currently supported when gathering block compressed VCFs.");
+                gatherWithBlockCopying(INPUT, OUTPUT);
+            } else {
+                log.info("Gathering by conventional means.");
+                gatherConventionally(sequenceDictionary, CREATE_INDEX, INPUT, OUTPUT);
+            }
+        } catch (IllegalArgumentException e) {
```
pshapiro4broad: Why is important to catch this runtime exception when we throw another runtime exception, `PicardException`, above? I'm not familiar with how `doWork()` works; are some errors OK to catch and `return 1;` and others should crash the program?

src/main/java/picard/vcf/GatherVcfs.java
```diff
@@ -61,32 +61,38 @@ public GatherVcfs() {
...
-            gatherWithBlockCopying(INPUT, OUTPUT);
-        }
-        else {
-            log.info("Gathering by conventional means.");
-            gatherConventionally(sequenceDictionary, CREATE_INDEX, INPUT, OUTPUT);
+        try {
+            assertSameSamplesAndValidOrdering(INPUT);
+            if (areAllBlockCompressed(INPUT) && areAllBlockCompressed(CollectionUtil.makeList(OUTPUT))) {
+                log.info("Gathering by copying gzip blocks. Will not be able to validate position non-overlap of files.");
+                if (CREATE_INDEX)
```
pshapiro4broad: All `if`s should have a `{` `}` block.

src/main/java/picard/vcf/GatherVcfs.java
```diff
@@ -61,32 +61,38 @@ public GatherVcfs() {
...
-            log.info("Gathering by copying gzip blocks. Will not be able to validate position non-overlap of files.");
-            if (CREATE_INDEX) log.warn("Index creation not currently supported when gathering block compressed VCFs.");
-            gatherWithBlockCopying(INPUT, OUTPUT);
-        }
-        else {
-            log.info("Gathering by conventional means.");
-            gatherConventionally(sequenceDictionary, CREATE_INDEX, INPUT, OUTPUT);
+        try {
+            assertSameSamplesAndValidOrdering(INPUT);
+            if (areAllBlockCompressed(INPUT) && areAllBlockCompressed(CollectionUtil.makeList(OUTPUT))) {
```
pshapiro4broad: Using `Collections.singletonList(OUTPUT)` would be more efficient

src/main/java/picard/vcf/GatherVcfs.java
```diff
@@ -211,7 +221,8 @@ private static void gatherWithBlockCopying(final List<File> vcfs, final File out
 
                 // a) It's good to check that the end of the file is valid and b) we need to know if there's a terminator block and not copy it
                 final BlockCompressedInputStream.FileTermination term = BlockCompressedInputStream.checkTermination(f);
-                if (term == BlockCompressedInputStream.FileTermination.DEFECTIVE) throw new PicardException(f.getAbsolutePath() + " does not have a valid GZIP block at the end of the file.");
+                if (term == BlockCompressedInputStream.FileTermination.DEFECTIVE)
```
pshapiro4broad: It's nice to have these formatted across lines but now it's more obvious that they're missing `{` `}`'s

src/main/java/picard/util/IntervalListTools.java
yfarjoun: good point!

src/main/java/picard/vcf/GatherVcfs.java
yfarjoun: no...this is silliness on my part.

yfarjoun:
Thanks for the comments @pshapiro4broad 

fixed it!

gbggrant:
@pshapiro4broad bump

src/main/java/picard/TestStringUtilSplit.java
pshapiro4broad: If the return value is unused (and undocumented) it would be better to return `void`.

src/main/java/picard/analysis/CollectOxoGMetrics.java
```diff
@@ -346,7 +334,7 @@ protected int doWork() {
         for (final List<Calculator> calcs : calculators.values()) {
             for (final Calculator calc : calcs) {
                 final CpcgMetrics m = calc.finish();
-                m.SAMPLE_ALIAS = StringUtil.join(",", new ArrayList<String>(samples));
+                m.SAMPLE_ALIAS = StringUtil.join(",", new ArrayList<>(samples));
```
pshapiro4broad: Would be better to write as `String.join(",", samples);`

src/main/java/picard/TestStringUtilSplit.java
```diff
@@ -6,51 +6,53 @@
...
+ * A class that can be used to check which is faster, StringUtil.split or StringTokenizer.
+ * Currently, StringUtil is faster.
  */
-public class Test {
+public class TestStringUtilSplit {
     private static final String TEXT = "C0A69ACXX111213:6:1101:10000:144257\t83\t5\t128984606\t60\t76M\t=\t128984542\t-140\tAGTGTTAGAACTTCCTCCCCAAAGCATATACTTCAGTGGCAAGCTGTCCTGGATGAAGGTATGACCAACCAGATCA\t@FFFEECC>EFHBJIGIFGIEIJJJIHED<IEHIGIIJIIIGJIGJIIIIIJGGCJIIGIHHHBHGHFFDFFFC@@\tXT:A:U\tNM:i:0\tSM:i:37\tAM:i:37\tX0:i:1\tX1:i:0\tXM:i:0\tXO:i:0\tXG:i:0\tMD:Z:76";
 
     public static void main(String[] args) {
-        new Test().run();
+        new TestStringUtilSplit().run();
```
pshapiro4broad: All methods here should be declared static and can be called directly without instantiating the class.

src/main/java/picard/vcf/GatherVcfs.java
yfarjoun: in reading this again, the part that I care about is deleting what might be partially created vcf. given how the gathering works the truncated output might be valid...so it's important to delete it as the user might not be checking return codes.

src/main/java/picard/TestStringUtilSplit.java
yfarjoun: I'm reverting my changes as they pertain to this class...someone else can deal with it.

yfarjoun:
I reverted the changes to Test.java.

PR #979, opened 2017-11-17
ExtractIlluminaBarcodes should fail when sample barcode lengths mismatch

nh13:
`ExtractIlluminaBarcodes` should fail when the # of sample barcodes does not match between the read structure and the barcodes file.

### DIFF
--- a/src/test/java/picard/illumina/ExtractIlluminaBarcodesTest.java
+++ b/src/test/java/picard/illumina/ExtractIlluminaBarcodesTest.java
@@ -30,6 +30,7 @@ import org.testng.annotations.AfterTest;
 import org.testng.annotations.BeforeTest;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
+import picard.PicardException;
 import picard.cmdline.CommandLineProgramTest;
 import picard.illumina.parser.BaseIlluminaDataProvider;
 import picard.illumina.parser.ClusterData;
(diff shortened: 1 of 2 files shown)


nh13:
The issue for me occurs whent the sample barcodes are length `N` but the `RunInfo.xml` was length greater than `N`, and so the _unmatched_ barcode in the metrics file does not match the length of the sample barcodes.  I could change also solve my problem by changing how the _unmatched_ barcode is created (look at the barcodes) file, but still have a check to make sure the sample barcodes are shorter than what the read structure implies.

nh13:
@yfarjoun I could see this change being an issue for the Broad Institute, where we may not want to use all the bases in index read, or even sample barcodes inline in R1/R2?

fleharty:
@gbggrant Thanks!  I will take a look at this tomorrow.

nh13:
@fleharty the behavior may actually be OK, given we may not want to use all the index bases for sample barcodes. @tfenne if I recall correctly, sometimes this can be the case?

src/test/java/picard/illumina/ExtractIlluminaBarcodesTest.java
```diff
@@ -334,4 +335,20 @@ private void testParsing(final IlluminaDataProviderFactory factory, final ReadSt
         retval.read(new FileReader(metricsFile));
         return retval;
     }
+
+    @Test(expectedExceptions=PicardException.class)
+    public void testBarcodeLengthMismatch() throws Exception {
+        final File metricsFile = File.createTempFile("dual.", ".metrics");
+        metricsFile.deleteOnExit();
+
+        final String[] args = new String[]{
```
fleharty: Space after []

tfenne:
@nh13 Certainly in the past it has been useful to be able to do things like pool a set of samples, some of which have 6bp barcodes and some of which have 8bp barcodes, then do a 6bp _or_ 8bp barcode read on the instrument and have it all work out on the software side.  I.e. a mix of 6 and 8bp barcodes in the BARCODE_FILE, compatible with either a 6bp barcode read performed, or an 8bp barcode read performed.

fleharty:
@nh13 Is this ready to go?  I think we can ignore the pullapprove thing.

nh13:
@fleharty could you please re-read @tfenne's comment? It may be a good idea to get @ktibbett's eyes on this too!

fleharty:
@nh13 Thanks for reminding me about @tfenne comment.  I agree that it would be a good idea to get @ktibbett comments.

ktibbett:
@nh13 , could we have this behavior triggered by an additional (boolean) parameter that defaults to false to preserve current behavior?  It's helpful to be able to tolerate mismatches in some circumstances (as when, for example, a run fails after the first barcode read,but we still want to use the first read).

yfarjoun:
@nh13 this is your court.

nh13:
Will take a look, thanks!

yfarjoun:
bump

kbergin:
Bump @nh13

Issue #451, opened 2016-02-18
[Bug] When removing duplicates, MarkDuplicates does not remove unmapped records whose mate is a duplicate


sooheelee:
When the mapped mate of a singly mapping pair is marked as duplicate, its unmapped mate is not marked as duplicate. This is expected behavior as MarkDuplicates ignores unmapped reads from consideration.

If, however, I add 'REMOVE_DUPLICATES'=true to the MarkDuplicates command, the singly mapped duplicates are removed but the unmapped mates remain in the resulting file. This is undesirable behavior and these unmapped mates should also be removed.
- This bug applies to MarkDuplicates and likely MarkDuplicatesWithMateCigar as well.
- This bug pertains to singly mapping pairs as defined in [this dictionary entry](http://gatkforums.broadinstitute.org/gatk/discussion/6976/dictionary-mate-unmapped-records?new=1). I tested this and [Tutorial#6747](http://gatkforums.broadinstitute.org/gatk/discussion/6747) explains its [example data](https://drive.google.com/open?id=0BzI1CyccGsZiWURLdUdfRjVQazg) that you can use to recapitulate the bug.
- This bug could also potentially apply to secondary records whose primary mappings are marked as duplicate. I did not test this scenerio but recommend this be addressed as well.
- This bug could also potentially apply to any tool that considers and removes singly mapping reads.

In `6747_snippet_piped_markduplicates.bam`, in which duplicates are marked according to [Tutorial#6747](http://gatkforums.broadinstitute.org/gatk/discussion/6747) using MarkDuplicates, I have 255 singly mapping reads that are also duplicate (SAM flag `1033`) and no unmapped reads marked as duplicate (SAM flag `1029`). 

```
-bash-3.2$ samtools view -c -f 1033 6747_snippet_piped_markduplicates.bam
255
-bash-3.2$ samtools view -c -f 1029 6747_snippet_piped_markduplicates.bam
0
```

Using the following command, I get a list of read names for duplicate singly mapping records:

```
-bash-3.2$ samtools view -f 1033 6747_snippet_piped_markduplicates.bam | cut -f1 | sort | uniq
H0164ALXX140820:2:1101:1944:34606
H0164ALXX140820:2:1101:20356:2293
. [truncated]
H0164ALXX140820:2:2224:31389:66479
H0164ALXX140820:2:2224:31825:66180
```

[...]

PR #1018, opened 2017-12-07
Fix secondary, supplementary and unmapped duplicate reads removing

NataliaLeznikova:
**Description**

Initial issue:https://github.com/broadinstitute/picard/issues/451

According to an issue, "when the mapped mate of a singly mapping pair is marked as duplicate, its unmapped mate is not marked as duplicate. This is expected behaviour as MarkDuplicates ignores unmapped reads from consideration.
If, however,'REMOVE_DUPLICATES'=true, the singly mapped duplicates are removed but the unmapped mates remain in the resulting file. This is undesirable behaviour and these unmapped mates should also be removed."

**Implementation**

We added new collection 'nonProperlyMappedPairDuplicatesNames'. This collection is needed only if 'REMOVE DUPLICATES = true': it executes an additional pass through "doWork()" and collects names of secondary, supplementary and unmapped reads that are duplicates. At the final stage of writing new file it allows to check wether a secondary, supplementary or unmapped read is in duplicate set or not, and remove it, if it is.

We also added integration tests on the main functionality of MarkDuplicates and examined all metrics which are used by MarkDuplicates (DuplicationMetrics). Test cases are:

- marking duplicates with/without removing them (including optical ones);
- marking duplicates in files containing barcodes(tags BARCODE_TAG, READ_ONE_BARCODE_TAG, READ_TWO_BARCODE_TAG) with/without removing duplicates;
- files which include secondary or supplementary alignment duplicate reads;
- PicardException on the case: "This program requires input that are either coordinate or query sorted. Found unsorted".

**Performance**

We measured performance by benchmarking method runPicardCommandLine() with different arguments: different SAM/BAM files and checked both how program works with and without removing duplicates.

Comparing old and new implementations of MD (REMOVE_DUPLICATES = false)

|             MarkDuplicatesBenchmark                     |  Cnt  | Score (new impl.)| Score (old impl.) |Percent| Units| 
|:--------------------------------------------------------|:-----:|:----------------:|:-----------------:|:-----:|:----:|
| benchmark(big_dupes.sam) - 1000 reads                   |  50   |  1.364 ± 0.084   |   1.198 ± 0.080   | +14%  |ops/ms|

Comparing old and new implementations of MD (REMOVE_DUPLICATES = true)

|             MarkDuplicatesBenchmark                        |  Cnt  | Score (new impl.)| Score (old impl.) |Percent| Units| 
|:----------------------------------------------------------|:-----:|:----------------:|:-----------------:|:-----:|:----:|
| benchmark(6747_snippet_markduplicates.bam) - 276000 reads |  50   |  0.131 ± 0.014   |   0.110 ± 0.007   |  +19% |ops/ms|
| benchmark(big_dupes.sam) - 1000 reads                     |  50   |  1.049 ± 0.100   |   0.880 ± 0.063   |  +19% |ops/ms|


src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -935,5 +955,42 @@ public int compare(final ReadEndsForMarkDuplicates lhs, final ReadEndsForMarkDup
...
 
+        final CloseableIterator<SAMRecord> iteratorForRemoveDuplicates = headerAndIteratorForRemoveDuplicates.iterator;
+        while (iteratorForRemoveDuplicates.hasNext()) {
+            final SAMRecord rec = iteratorForRemoveDuplicates.next();
+
+            final boolean nextDuplicateIndexIsNecessary = recordInFileIndex > nextDuplicateIndex &&
+                    (sortOrder == SAMFileHeader.SortOrder.coordinate || !rec.getReadName().equals(duplicateQueryName));
+
+            if (nextDuplicateIndexIsNecessary) {
+                nextDuplicateIndex = (duplicateIndexes.hasNext() ? duplicateIndexes.next() : NO_SUCH_INDEX);
```
tlangs: Input variables shouldn't be mutated. A new variable should be introduced to hold this value.

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -656,6 +670,8 @@ private ReadEndsForMarkDuplicates buildReadEnds(final SAMFileHeader header, fina
      */
     private void generateDuplicateIndexes(final boolean useBarcodes, final boolean indexOpticalDuplicates) {
         final int entryOverhead;
+
+
```
tlangs: Unnecessary whitespace

src/test/java/picard/sam/markduplicates/MarkDuplicatesExceptionTest.java
```diff
@@ -0,0 +1,47 @@
...
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+import picard.PicardException;
+import picard.cmdline.CommandLineProgramTest;
+
+import java.io.File;
+import java.util.*;
+
+public class MarkDuplicatesExceptionTest extends CommandLineProgramTest {
```
tlangs: The tests in this class could go in `MarkDuplicatesIntegrationTests`

src/test/java/picard/sam/markduplicates/MarkDuplicatesIntegrationTests.java
```diff
@@ -0,0 +1,378 @@
...
+        args.add("OUTPUT=" + output.getAbsolutePath());
+        final File metrics = new File(outputDir, TEST_BASE_NAME + ".integration_metrics");
+        args.add("METRICS_FILE=" + metrics.getAbsolutePath());
+
+        Assert.assertEquals(runPicardCommandLine(args), 0);
+        examineDuplicationMetrics(input, output, metrics, false);
+    }
+
+    @Test(dataProvider = "MarkDuplicatesIntegrationTest")
+    public void markDuplicatesWithRemovingDuplicatesIntegrationTest(final String input) {
```
tlangs: `markDuplicatesWithRemovingDuplicatesIntegrationTest` and `secondarySupplementaryUnmappedWithRemovingDuplicatesIntegrationTest` are exactly the same. The contents of the test should be pulled out into its own test and the two separate data providers should be one line methods that call the abstracted test.

yfarjoun:
I'm not happy with the fact that the implementation includes a **third** traversal over the input file. I think that the stated result (removing duplicate for reads whose mate is unmapped) without resorting to a third pass.

here are my thoughts:

1. When a read is marked as a duplicate, if it's mate is unmapped, store it's name in a DiskBackedSortingCollection<String>. 
2. When writing the resulting file to disk, filter out reads that are unmapped && whose name is in the collection.

Thoughts?

NataliaLeznikova:
Hello, @yfarjoun!

Thank you for your answer. 

As far as we have understood, we don't have information about reads' mates at the time of the second traversal. We suppose that it's impossible to find out whether a duplicate read has a mate or not as mates can be located at the different parts of the file, so that is why we are offering this implementation. 

Please, correct us if we are wrong.

yfarjoun:
I think that unmapped reads whos mates are marked as duplicates, should also be marked as duplicates....if this isn't happening now, I think it's a bug...I'll open an issue about it.

NataliaLeznikova:
Hello, @yfarjoun!

We definitely agree with you that there is no sense in diving into mark duplicates algorithm until we are not sure if it's actually a bug. We measured the performance on 17.35GB bam and figured out that after the 'fixed' branch worked twice longer than master one.

So there is one thing that seems to be obvious - the third traversal through file leads to performance decrease. We will be waiting for your decision on this point.

yfarjoun:
no one objected to my suggestion that we close this issue...so I'm following through.

PR #1029, opened 2017-12-13
Add DEDUPLICATE_RECORDS option to RevertSam.

nh13:
### Description

In some cases, the same record may be found multiple times in a BAM
file.  This option allows the user to remove the duplicate record,
whereas the default behavior is to discard the records (and mates if
present).

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/sam/RevertSam.java
+++ b/src/main/java/picard/sam/RevertSam.java
@@ -62,10 +62,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.text.DecimalFormat;
 import java.text.NumberFormat;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 /**
  * Reverts a SAM file by optionally restoring original quality scores and by removing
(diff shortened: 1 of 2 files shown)


src/main/java/picard/sam/RevertSam.java
```diff
@@ -164,6 +167,10 @@
             "the program will exit with an Exception instead of exiting cleanly. Output BAM will still be valid.")
     public double MAX_DISCARD_FRACTION = 0.01;
 
+    @Argument(doc = "If SANITIZE=true discard duplicate records.  Duplicate records will have the same values for all field" +
+            "including tags.")
```
tfenne: I think you probably need to add a note here that "deduplicate" doesn't have anything do with the usual "is duplicate flag" definition of the word.  Maybe we can find a better word to use instead?

src/main/java/picard/sam/RevertSam.java
```diff
@@ -374,6 +383,21 @@ public void revertSamRecord(final SAMRecord rec) {
                     }
                 }
 
+                // Remove records that have the SAM SAMString (*** SLOW ***)
+                if (DEDUPLICATE_RECORDS) {
+                    final Iterator<SAMRecord> iter = recs.iterator();
+                    final Set<String> samStrings = new HashSet<>();
```
tfenne: 1. I'm unclear why this is better than just requiring that there be only a single unpaired/R1/R2 in the collection.  That's tested below, so even if they are not identical, it'll still fail if there are, e.g. two R1s in the collection.
2. I'd be a little concerned that if the extended attributes are in a different order they might generate different strings while otherwise being identical.  Is `SAMRecord.equals()` not trustworthy for some reason?

nh13:
@tfenne thanks for the quick review.  I implemented your suggestion to keep only the _first_ read for R1/R2/unpaired  respectively.  I improved the error message a little too.

src/main/java/picard/sam/RevertSam.java
```diff
@@ -374,24 +378,46 @@ public void revertSamRecord(final SAMRecord rec) {
...
-                    }
+                // If we have paired reads, then check that there is exactly one first of pair and one second of pair.
+                // Otherwise, check that we have only one unpaired read.
+                if (firsts > 0 || seconds > 0) { // if we have any paired reads
+                    if (firsts != 1 || seconds != 1) { // if we do not have exactly one R1 and one R2
+                        if (KEEP_FIRST_DUPLICATE && firsts >= 1 && seconds >= 1) { // if we have at least one R1 and one R2, we can discard all but the first encountered
+                            discarded += recs.size() - 2;
+                            recs = Arrays.asList(firstRecord, secondRecord);
+                        }  else {
+                            log.debug("Discarding " + recs.size() + " reads with name " + recs.get(0).getReadName() + " because  we found " + firsts + " R1s " + seconds + " R2s and " + unpaired + " unpaired reads.");
```
tfenne: I would just use commas here instead of `+` so that all the string concatenation is only done if the log level is debug.

PR #1031, opened 2017-12-14
Bug fix: CollectWgsMetricsWithNonZeroCoverage fails to produce a plot

nh13:
when there is zero coverage.

Also fixed up a few warnings.

### DIFF
--- a/src/main/resources/picard/analysis/wgsHistogram.R
+++ b/src/main/resources/picard/analysis/wgsHistogram.R
@@ -1,5 +1,5 @@
 ##
-# Copyright (c) 2016, Nils Homer
+# Copyright (c) 2016-2017, Nils Homer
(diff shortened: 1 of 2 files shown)


nh13:
@yfarjoun or @lbergelson, care to take a look?  This happens when we have zero coverage data (yikes!).

nh13:
@yfarjoun good idea, I added in a test to make sure the pdf is created.   I squashed the commits to be ready for the merge.

yfarjoun:
Thanks. in the future, please refrain from squashing. We can easily squash in github and it only makes reviewing more difficult. (in this case, it's really not a problem since the changes are so minimal...)

yfarjoun:
Looks great! thanks for the fix and test!

👍

PR #1034, opened 2017-12-21
[Do Not Merge] Add filterbytag filter to FilterSamReads

jsotobroad:
Option to filter by include or exclude tag values added to FilterSamReads.  The motivation for this is to allow someone to select all reads from a cells using the CB or CR tag from a bam but could in general be used to select any reads based on value(s) of a tag.

#### Content

#### Review


jsotobroad:
This is currently waiting for a htsjdk release to get the latest TagFilter changes (https://github.com/samtools/htsjdk/commit/335cd0622bacee9936b5ee923cb9eb6649a0b5e0).  It is currently using a local version htsjdk hash

src/main/java/picard/sam/FilterSamReads.java
```diff
@@ -122,17 +124,28 @@ public String toString() {
...
+
+    @Argument(doc = "The tag to select from input SAM/BAM",
+            optional = true,
+            shortName = "T")
+    public String TAG;
+
+    @Argument(doc = "List of tag values to extract from Input SAM/BAM",
+            shortName = "TV",
+            optional = true)
+    public List<String> TAG_VALUE;
```
nh13: I am curious about:
1. Why can't the `JS` argument be used here instead of adding arguments?  For your use case, this would be easier than adding a PR
2. How are array-value tags handled?

jsotobroad: I will have a new PR in the near future and will answer your questions there.

PR #1035, opened 2017-12-22
MarkDuplicates refactoring + readers closing (preparation for parallel implementation)

nkarulin:
# Motivation
The proposed changes can be mainly considered as prerequisites to the `ParallelMarkDuplicates` pull request (https://github.com/broadinstitute/picard/pull/1036), essentially preparing `MarkDuplicates` to be inherited by the parallel implementation.
Some of the changes (like some methods decomposition and moving useBarcode local variable to the field) and test extension are just MarkDuplicates improvements.

# Changes in the original MarkDuplicates class

md_refactoring

1. SortingCollection fields (`pairSort`, `fragSort`, `duplicateIndexes`, `opticalDuplicateIndexes`, `representativeReadIndicesForDuplicates`) decapsulation (private -> package private).
2. `buildSortedReadEndLists` and `generateDuplicateIndexes` methods decapsulation (private -> package private).
3. `useBarcodes` is a package-private field (not local variable). So, `useBarcodes` argument of `buildSortedReadEndLists` and `generateDuplicateIndexes` methods are removed.
4. Decomposition of the output file writing and metric calculation from `doWork()` method into `writeFileAndCalculateMetric()` package-private method.
5. local variable `ReadEndsForMarkDuplicatesMap` from `buildSortedReadEndLists()` method is a field and is renamed to `pairedWithNoMate`
6. Decomposition of paired ends combination in `buildSortedReadEndLists()` method into `combinePairEnds()` package-private method
7. Calling of `doneAdding()` method of `pairSort` and `fragSort` SortingCollections is moved from the end of `buildSortedReadEndLists()` method to the new `freeUpMemory()` method.
8. Decomposition of `maxInMemory` duplicate indexes calculation into `maxInMemoryForDuplicateIndexes()` package-private method.

Also `firstOfPair` boolean field is introduced in the `ReadEndsForMarkDuplicates` class with appropriate codec changes.

## Resource closing
In the `openInputs` method, which returns iterator over inputs 1+ `SamReader` objects are created but never closed. That was not crucial for standard use case, 2 or 4 unclosed file descriptors usually lead to no problems.

But when we have extended `MarkDuplicates` tests, `Too many open Files` exception appeared on tests launch.

We propose to keep the only set of readers and close them in the end of `doWork` method of all `AbstractMarkDuplicates` children.

# Tests extension
1.  Many tests were parameterized and some new tests were added to test paired read ends in cases, when those ends are in the different chromosomes. This multi-chromosome testing is crucial for parallel implementation, where there is multi-chromosome reads merging part. But also they are fine as an extension of the current tests set, so they places in the current PR with MarkDuplicates refactoring.
2.  New tagging duplicates sets tests
3.  New tagging policy tests


src/main/java/picard/sam/markduplicates/util/ReadEndsForMarkDuplicates.java
```diff
@@ -37,7 +37,7 @@
        - int: read1ReferenceIndex, read1Coordinate, read2ReferenceIndex, read2Coordinate, duplicateSetSize
        - long: read1IndexInFile, read2IndexInFile
      */
-    protected static final int SIZE_OF = (1 * 1) + (5 * 2) + (5 * 4) + (8 * 2) + 1
+    protected static final int SIZE_OF = (1 * 1) + (5 * 2) + (5 * 4) + (8 * 2) + 1 + 1
```
als364: Explain the new extra byte (from firstOfPair) in the comment above.

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -418,11 +432,11 @@ protected int doWork() {
         reportMemoryStats("Before output close");
         out.close();
         reportMemoryStats("After output close");
+    }
 
-        // Write out the metrics
-        finalizeAndWriteMetrics(libraryIdGenerator);
-
-        return 0;
+    // To be overrode in parallel implementation
```
tbl3rd: overrode -=> overridden

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -418,11 +432,11 @@ protected int doWork() {
...
+    CloseableIterator<RepresentativeReadIndexer> representativeReadIndicesIterator() {
+        return this.representativeReadIndicesForDuplicates.iterator();
```
tbl3rd: Why this. here? You delete it elsewhere.

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -525,80 +539,91 @@ private void buildSortedReadEndLists(final boolean useBarcodes) {
...
-                            pairedEnds.read1ReferenceIndex = matesRefIndex;
-                            pairedEnds.read1Coordinate = matesCoordinate;
-                            pairedEnds.read1IndexInFile = indexForRead;
-                            pairedEnds.orientation = ReadEnds.getOrientationByte(rec.getReadNegativeStrandFlag(),
-                                    pairedEnds.orientation == ReadEnds.R);
-                        }
-
-                        pairedEnds.score += DuplicateScoringStrategy.computeDuplicateScore(rec, this.DUPLICATE_SCORING_STRATEGY);
-                        this.pairSort.add(pairedEnds);
+                        this.pairSort.add(combinePairEnds(pairedEnds, fragmentEnd));
```
tbl3rd: Good idea.

src/main/java/picard/sam/markduplicates/util/ReadEndsForMarkDuplicates.java
```diff
@@ -37,7 +37,7 @@
...
             + 8 + // last 8 == reference overhead
             13; // This is determined experimentally with JProfiler
```
tbl3rd: That size doesn't look aligned (before or now), but maybe that doesn't matter.
Shouldn't (5 * 2) be (6 * 2) according to the comment?
Shouldn't (8 * 2) be (2 * 8) for consistency?
And why the +1s?
Maybe just determine the whole thing with the profiler.

src/test/java/picard/sam/testers/SamFileTester.java
```diff
@@ -1,3 +1,27 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
 package picard.sam.testers;
```
tbl3rd: All good. Thanks.

src/main/java/picard/sam/markduplicates/util/ReadEndsForMarkDuplicates.java
nkarulin: I have profiled the class with `java.lang.instrument.Instrumentation`. It shows that an implementation-specific (Java HotSpot(TM) 64-Bit Server) approximation of the amount of storage consumed by the `ReadEndsForMarkDuplicates` object equals to __72 bytes__.
Adding new boolean field has no influence on the approximated value.
Current value of `SIZE_OF` is __70__. So they are pretty equal.

After recounting this could looks like this:
```java
    // What do we need to store you ask?  Well, we need to store:
    protected static final int SIZE_OF =
              (2 * 1)   // - byte: orientation, orientationForOpticalDuplicates
            + (6 * 2)   // - short: libraryId, readGroup, tile, x, y, score
            + (5 * 4)   // - int: read1ReferenceIndex, read1Coordinate, read2ReferenceIndex, read2Coordinate, duplicateSetSize
            + (2 * 8)   // - long: read1IndexInFile, read2IndexInFile
            + 1         // - boolean: firstOfPair
            + 8         // last 8 == reference overhead
            + 13;       // This is determined experimentally with JProfiler
```
Which is equal to __72__. 
Is this OK, or should we just paste 72 from profiling?

src/main/java/picard/sam/markduplicates/util/ReadEndsForMarkDuplicates.java
nkarulin: @als364, thanks for the review!
Please have a look at my comment about `SIZE_OF` value: https://github.com/broadinstitute/picard/pull/1035#discussion_r165039319.

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -182,11 +182,14 @@
...
-    private SortingLongCollection opticalDuplicateIndexes;
-    private SortingCollection<RepresentativeReadIndexer> representativeReadIndicesForDuplicates;
+    SortingCollection<ReadEndsForMarkDuplicates> pairSort;
+    SortingCollection<ReadEndsForMarkDuplicates> fragSort;
+    SortingLongCollection duplicateIndexes;
+    SortingLongCollection opticalDuplicateIndexes;
+    SortingCollection<RepresentativeReadIndexer> representativeReadIndicesForDuplicates;
+
+    DiskBasedReadEndsForMarkDuplicatesMap pairedWithNoMate;
+    boolean useBarcodes;
```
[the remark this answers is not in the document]
nkarulin: @tbl3rd, thanks for the review!
Yes, those fields must be accessible in parallel implementation. For example for `useBarcode` field here: https://github.com/broadinstitute/picard/pull/1036/files#diff-afce78182525d79877825df5962e642fR340

src/main/java/picard/sam/markduplicates/util/ReadEndsForMarkDuplicates.java
tbl3rd: That makes sense to me.  Thanks.

tbl3rd: You can do that or summarize how to tell when SIZE_OF goes wrong and how to fix it,
but either way, thank you for straightening that out for us.

src/main/java/picard/sam/markduplicates/util/ReadEndsForMarkDuplicates.java
als364: That comment makes sense, thanks!

yfarjoun:
We have a Spark-based implementation in GATK4.0, so I'm inclined to close this so that we do not have to maintain 2 different parallel implementations of MD. Thanks for the work.

PR #1036, opened 2017-12-22
Parallel MarkDuplicates implementation

nkarulin:
**!**  Includes the [MarkDuplicates refactoring PR](https://github.com/broadinstitute/picard/pull/1035) in picard-tools. Current PR is separated for convenience.

**!**  Depends on [SortingCollection: freeUpMemory util method PR](https://github.com/samtools/htsjdk/pull/1061) in htsjdk

# Motivation
NGS data processing performance is more and more crucial factor in many research and clinic fields.
More and more NGS tools demonstrate good and even excellent scalability while running on multi-core computers.
Now it’s time to introduce some scalability to MarkDuplicates as well.
# MarkDuplicates parallel implementation description
## General Purpose of MarkDuplicates parallelization and slice problem
The main concept of MarkDuplicates parallelization is to run several MarkDuplicates instances on several peaces of input file in parallel and then merge the results. The only problem is in case when paired reads are separated by a slice. MarkDuplicates will not mark those paired reads. 
slice

Since there could be pair ends even on the different chromosomes, there are no saved slices we could know before processing. For example [NA12878.mapped.ILLUMINA.bwa.CEU.exome.20121211.bam](ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/phase3/data/NA12878/exome_alignment/NA12878.mapped.ILLUMINA.bwa.CEU.exome.20121211.bam) has 5793 duplicate pairs which ends are on the different chromosomes.

## Structure
Parallel implementation is introduced as separate class inheriting original `MarkDuplicates` and composes `MarkDuplicatesInstance`s. `MarkDuplicatesInstance` also inherits original MarkDuplicates, it has some specific behavior but marking duplicates logic is fully inherited from the original version.
There is some refactoring in the original MarkDuplicates class which is for convenience described in the [separate pull request](https://github.com/broadinstitute/picard/pull/1035).
Also small util `ArgsPreparer` class added in `picard.sam.markduplicates.util` package.

md_class_idagram__1_

### High-level example  
ParallelMarkDuplicates forks 3 MarkDuplicateInstances and waits until they finished building SortingCollections of ReadEnds. Then It merges pairs between those two parts and lets them generate duplicate indexes and write their part of the file. Finally, ParallelMarkDuplicates gather BAM  files into one and merge end write metric file. 

seq2

### More detailed example
md_sequencial_diagram2

# Functional Features description
## 1. Pair ends merging algorithm
Algorithm of pair ends combining is optimized to minimize count of Loading in RAM and Unloading to disk of sequence maps in `CoordinateSortedPairInfoMap` (`DiskBasedReadEndsForMarkDuplicatesMap`).
After finishing `buildingSortedReadEnds` stage we have **N** of `MarkDuplicatesInstances`. Each instance is processing serial list of sequences (could be of size 1): from `leftmostSeqIndex` to `rightmostSeqIndex`. Merging has linear complexity.

```java
// mds - list of MarkDuplicatesInstances
for (int mdIndex = 0; mdIndex < mds.size() - 2 /*skip unmapped*/; mdIndex++) {
    MarkDuplicatesInstance md1 = mds.get(mdIndex);
    for (int seqId = md1.leftmostSeqIndex; seqId <= md1.rightmostSeqIndex; seqId++) {
        mergeAllPairsFromMDInstanceOfSequence(mdIndex, seqId);
    }
}
```
Example (where there are only 2 instances) of `mergeAllPairsFromMDInstanceOfSequence` method is illustrated below on the picture.

merging_pair_of_mdinstance

### Pair ends combining
After we have found corresponding read ends which are in the different parallel processing pieces, we combine them into two pairedEnds by the same method original MarkDuplicates does but with the corrected `inFileIndice`. 

endscombination
So both MDInstances will define duplicates considering assembled pairedEnd and results will be equals to the original MarkDuplicates. But each of those out-of-piece ends will not be written by the MDInstance (corresponding to `InFileIndex`). 

## 2. Correct optical duplicates handling
`MarkDuplicatesInstance` uses `OpticalDuplicateFinder` with modified `findOpticalDuplicates()` method. `findOpticalDuplicates()` after processing list of ends returns array of flags: which of them are optical duplicates. The modification drops flags for those ends which are not from the current piece, so they will not be count as optical duplicate (to avoid counting them twice). But still we set the end's field `isOpticalDuplicate` to `true` so they will be added to `opticalDuplicateIndexes` SortingCollection.

On the `doneAddingDuplicateIndexes` call special delimiter index (-1) is added to the `duplicateIndexes` SortingLongCollection. Before piece writing MDInstance will skip all negative indices till the delimiter.  
duplicateindexes

```java
while (duplicateIndexes.hasNext()) {
    if (duplicateIndexes.next() == DELIMITER_DUPLICATE_INDEX) {
        break;
    }
}
```

## 3. Correct duplicate sets tagging
On the `doneAddingDuplicateIndexes` call if `TAG_DUPLICATE_SET_MEMBERS` is `true` special `delimiterReadIndexer` is added to the `representativeReadIndicesForDuplicates`.

`MarkDuplicatesInstance` returns modified iterator over `RepresentativeReadIndexer`s which skips all representativeReadIndices till `delimiterReadIndexer` and corrects `representativeReadIndexInFile` by incrementing it's value by `recordsInFileBefore` of the current `MarkDuplicatesInstance`. Original value means index in the current part file, corrected value is an index in the result file. So `DI` tag will be correct.

# Additional necessary changes
## Keeping the original list of program arguments in CommandLineProgram
Keeping the original list of program arguments is needed to pass it from *master* MarkDuplicates (`ParallelMarkDuplicates`) to *slave* MarkDuplicates (`MarkDuplicatesInstance`). There is `commandLine` string field in `CommandLineProgram` class, but it is really tricky to parse it back to the original strings array (Array values should be handled, some default values are invalid command line arguments like `READ_NAME_REGEX=<optimized capture of last three ':' separated fields as numeric values>`). 

Alternatively all ~40 MarkDuplicates argument fields could be set manually after MarkDuplicates instancing.

## Public iterator for `DiskBasedReadEndsForMarkDuplicatesMap`
`iterator()` public method was added added to `DiskBasedReadEndsForMarkDuplicatesMap` class. The method return `CloseableIterator` over `Map.Entry<String, ReadEndsForMarkDuplicates>` of encapsulated `pairInfoMap`. That is needed for pair ends combining algorithm.

# Benchmarks 
### 8CPU machine: 10Gb BAM file (33% duplicates)
This example shows that heap size greater then needed downgrades the performance both for original and parallel MD.

Records by chromosome map:

screenshot from 2017-12-11 17-45-11

Max speed up ~**3.3x**

### 8CPU machine: 1,5Gb BAM file
This example shows, that there could be parallelisation limit due to low chromosome variety.

aura
screenshot from 2017-12-11 18-02-37

Max speed up ~**2.8x**

### 8CPU machine: 5,5Gb BAM file, barcodes are enabled
BAM has very steady records distribution between chromosomes (each has 2-5 million of records). 
screenshot from 2017-12-11 18-22-21

Max speed up ~**3.7x**

### AWS c4.8xlarge: 17Gb BAM file, 4G heap
Lunching 17Gb exome ([NA12878.mapped.ILLUMINA.bwa.CEU.exome.20121211.bam](ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/phase3/data/NA12878/exome_alignment/NA12878.mapped.ILLUMINA.bwa.CEU.exome.20121211.bam)) in the AWS instance shows good scalability.

17gbmap
awsparallelmd

Max speed up ~**5.9x**

# Testing
There are several new ParallelMarkDuplicates tests which inherits `MarkDuplicatesTest` with new `ParallelMarkDuplicatesTester`. `ParallelMarkDuplicatesTester` represents generated sam files as indexed bam files, so `ParallelMarkDuplicates` could process it.
*  `ParallelMarkDuplicatesTest`
*  `ParallelMarkDuplicateWithMissingBarcodeTest`
*  `ParallelMarkDuplicateWithMissingReadOneBarcodeTest`
*  `ParallelMarkDuplicateWithMissingSampleBarcodeTest`

Also there was integration tests suit, which contains of:  
* Simple MD run scenario
* Simple MD run scenario with barcodes enabled
* CLEAR_DT option
* COMMENT option
* READ_NAME_REGEX disabling
* REMOVE_DUPLICATES=TRUE scenario
* TAGGING_POLICY=DontTag/OpticalOnly/All scenario
* TAG_DUPLICATE_SET_MEMBERS=true scenario

There were lunches on different files of different sizes and duplicates cases, launches on different threads number. Result BAM/SAM were tested to be fully equals (md5sum) to the original MarkDuplicates output.


nh13:
Very exciting!

> The only problem is in case when paired reads are separated by a slice. MarkDuplicates will not mark those paired reads.
My apologies if I am missing something, does this implementation mark the all paired end reads, even those that are "separated by a slice"?  Or more concretely, does this mark the same reads as the non-parallel implementation?

nkarulin:
@nh13, yes, as it is said in Testing section, the output of parallel implementation is fully equal to the original MarkDuplicates.
It is true for reads tagging (TAGGING_POLICY, TAG_DUPLICATE_SET_MEMBERS) and all other scenarios.  Output metrics also do not differ.

yfarjoun:
This is indeed quite exciting. The only thing that dampens my excitement is that there is a Spark enabled implementation of MarkDuplicates in GATK4. It It would be a shame to support so many different implementations of MarkDuplicates. 

@lbergelson Can you share timing results from Spark mark duplicates? It would be good to see if the overhead is similar or of of the implementations can learn a trick or two from the other implementation and decide on which one to keep.

magicDGS:
I vote for having the different implementations if they provide the same results. For example, there are some people that are not familiar with Spark or that they prefer to use Picard over GATK (due to the size of the jar file), and thus a parallel implementation will be useful. In addition, having different algorithm for parallelize is a good way to being able to experiment with improvements in any of them - discarding this change because the performance of the Spark version is better avoids further development of this algorithm to improve speed even more in a non-spark system....

lbergelson:
@yfarjoun I'm happy to share timing information as soon as we have anything reliable.  We're doing very active development on it right now, but we still don't know exactly how the final form is going to look / perform.   I wouldn't block this implementation because of the upcoming spark one.  If this one works that's great.  Once we have the spark one we can compare.  It's very possible that a hand tuned threading framework for MarkDuplicates could outperform sparks more general approach in many circumstances.  If we can get the spark version to scale well across multiple node that will be an obvious advantage for it, since I assume this can't make use of multiple nodes in an easy way.

yfarjoun:
After much delay, I'm inclined to close this PR as we have a Spark-based implementation of MD in GATK.

exander77:
@yfarjoun I don't get it, this PR is not merged and WAPR does not use Spark version: https://github.com/broadinstitute/warp

PR #1046, opened 2018-01-02
Pull out some of the metrics being fixed in HsMetricsCollector into a separate method

eitanbanks:
### Description

This minor change is needed so that a class that extends HsMetricsCollector can update these metrics differently.  Nothing is really changed in this PR.

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/analysis/directed/HsMetricCollector.java
+++ b/src/main/java/picard/analysis/directed/HsMetricCollector.java
@@ -86,10 +86,7 @@ public class HsMetricCollector extends TargetMetricsCollector<HsMetrics> {
                 new String[]{"BAIT_SET",  "BAIT_TERRITORY",  "ON_BAIT_BASES",  "NEAR_BAIT_BASES",  "OFF_BAIT_BASES",  "PCT_OFF_BAIT",  "ON_BAIT_VS_SELECTED",  "MEAN_BAIT_COVERAGE"}
         );
 
-        hsMetrics.BAIT_DESIGN_EFFICIENCY = (double) hsMetrics.TARGET_TERRITORY / (double) hsMetrics.BAIT_TERRITORY;
-        hsMetrics.PCT_USABLE_BASES_ON_BAIT   = hsMetrics.ON_BAIT_BASES   / (double) targetMetrics.PF_BASES;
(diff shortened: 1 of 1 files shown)


src/main/java/picard/analysis/directed/HsMetricCollector.java
```diff
@@ -101,6 +98,21 @@ public HsMetrics convertMetric(final TargetMetrics targetMetrics) {
...
 
+    /**
+     * Updates several of the metrics in the provided Metrics object so that they better reflect the particular target strategy
+     *
+     * @param hsMetrics the metrics object to fix
+     * @param bases the total number of bases
+     * @param selectedPairs the total number of selected pairs
+     * @param selectedUniquePairs the number of selected pairs that are unique
+     */
+    protected void updateMetrics(final HsMetrics hsMetrics, final long bases, final long selectedPairs, final long selectedUniquePairs) {
```
pshapiro4broad: This looks like it could be `static`, unless we expect users to override this method? If it can be static, I would also consider making it a `HsMetrics` member function, e.g. 
```
public void updateMetrics(final long bases, final long selectedPairs, final long selectedUniquePairs) {
    ...
```

eitanbanks: Yeah, that's the whole point.  I'm overriding it (in a picard-private class that extends this one).

src/main/java/picard/analysis/directed/HsMetricCollector.java
```diff
@@ -109,7 +121,7 @@ public HsMetrics convertMetric(final TargetMetrics targetMetrics) {
      * @return the hs penalty - a multiplier that tells if you want, e.g. 20X coverage, then you will
      *         need to produce this many PF aligned bases per target bases in your design!
      */
-    private double calculateHsPenalty(final Long librarySize, final TargetMetrics targetMetrics, final int coverageGoal) {
+    protected double calculateHsPenalty(final Long librarySize, final TargetMetrics targetMetrics, final int coverageGoal) {
```
pshapiro4broad: If you widen the visibility of this method, it would be good to fix the javadoc so it describes all the params.

eitanbanks: ok, will do

PR #1049, opened 2018-01-02
DSDEGP-1626: Only check expected output cycles

jacarey:
DSDEGP-1626

This PR fixes the bug described here: https://gatkforums.broadinstitute.org/gatk/discussion/comment/42610

This ensures that cbcls are only checked for output cycles.
----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/illumina/CheckIlluminaDirectory.java
+++ b/src/main/java/picard/illumina/CheckIlluminaDirectory.java
@@ -121,10 +121,10 @@ public class CheckIlluminaDirectory extends CommandLineProgram {
         final List<Integer> failingLanes = new ArrayList<>();
         int totalFailures = 0;
 
-        final int[] expectedCycles = new OutputMapping(readStructure).getOutputCycles();
+        final OutputMapping outputMapping = new OutputMapping(readStructure);
         log.info("Checking lanes(" + StringUtil.join(",", LANES) + " in basecalls directory (" + BASECALLS_DIR
                 .getAbsolutePath() + ")\n");
-        log.info("Expected cycles: " + StringUtil.intValuesToString(expectedCycles));
+        log.info("Expected cycles: " + StringUtil.intValuesToString(outputMapping.getOutputCycles()));
 
         for (final Integer lane : LANES) {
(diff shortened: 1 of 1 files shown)


src/main/java/picard/illumina/CheckIlluminaDirectory.java
```diff
@@ -168,9 +168,8 @@ protected int doWork() {
                     filterFileMap.put(fileToTile(filterFile.getName()), filterFile);
                 }
 
-                final OutputMapping outputMapping = new OutputMapping(readStructure);
-
-                final CbclReader reader = new CbclReader(cbcls, filterFileMap, readStructure.readLengths, tiles.get(0), locs, outputMapping.getOutputCycles(), true);
+                final CbclReader reader = new CbclReader(cbcls, filterFileMap, outputMapping.getOutputReadLengths(),
```
pshapiro4broad: It looks like the significant change here is replacing `readStructure.readLengths` with `outputMapping.getOutputReadLengths()`. Why did you remove the variable `expectedCycles` which was caching `outputMapping.getOutputCycles()` as well?

jacarey:
Adding test data that would make this tests possible is .... complicated.

src/main/java/picard/illumina/CheckIlluminaDirectory.java
jacarey: I'm caching the full outputmapping now. It is one extra get call now. Not sure its worth caching that as well.

pshapiro4broad: Well, it's not a simple getter (it calls `ReadStructure.getCycles()`), but I'll leave it up to you if it's worth caching or not.

alecw:
Hi @sooheelee , I would think you could use any run folder, and simply replace the "natural" read structure with something unnatural but compatible, e.g. make the last template read 1 base shorter, and add a 1 base skip at the end.

sooheelee:
Thanks for the clarification @alecw. I thought perhaps it was easiest to get some data from the user but what you say clarifies that it would be just as easy to modify existing data.

jacarey:
I have tested it with real data. The issue is setting this up to be run properly with testdata for a unit test.

PR #1079, opened 2018-01-16
Add filter by tag option to FilterSamReads

jsotobroad:
This gives people a easier way to filter reads by tag values.  Currently this only works for `String` tag values.  The single cell community heavily uses tags in bams and this addition provides a simple way to select useful things (cell barcode, molecular barcodes) from a bam.  The original use case for this is a way to subset a bam into specific cells using the cell barcode tag.

#### Content

#### Review


src/test/java/picard/sam/FilterSamReadsTest.java
```diff
@@ -114,18 +114,32 @@ public void testReadFilters(final FilterSamReads.Filter filterType, final List<S
...
+                {FilterSamReads.Filter.includeTagValues, "JAVASCRIPT_FILE"},
+                {FilterSamReads.Filter.excludeTagValues, "JAVASCRIPT_FILE"},
+
+                {FilterSamReads.Filter.excludeReadList, "TAG"},
+                {FilterSamReads.Filter.includeReadList, "TAG"},
+                {FilterSamReads.Filter.excludeAligned, "TAG"},
+                {FilterSamReads.Filter.includeAligned, "TAG"},
+                {FilterSamReads.Filter.includePairedIntervals, "TAG"},
+                {FilterSamReads.Filter.excludeAligned, "TAG"},
+                {FilterSamReads.Filter.includeAligned, "TAG"}
```
yfarjoun: also TAG  but not TAG_VALUE is bad..right?

yfarjoun: and vice versa...

src/main/java/picard/sam/FilterSamReads.java
```diff
@@ -205,7 +218,7 @@ public String getHelpDoc() {
     @Argument(
             doc = "Create <OUTPUT>.reads file containing names of reads from INPUT and OUTPUT (for debugging purposes.)",
             optional = true)
-    public boolean WRITE_READS_FILES = true;
+    public boolean WRITE_READS_FILES = false;
```
yfarjoun: I have no idea why this defaults to true...and I agree with you that it should default to false...however, that is a change to the default behaviour, and thus an api change...I don't think that it's appropriate to put it into this PR...

Please open an issue  and/or a separate PR to address this. (and remove it from this PR.)

nh13:
@jsotobroad I don't see any response to the comment I made in #1034, namely, why can't users just use `JS` argument?

src/main/java/picard/sam/FilterSamReads.java
jsotobroad: will do.  I think i changed this because it was being annoying when i was testing and just forgot i made the change

jsotobroad:
Yes @nh13 sorry I totally forgot sorry.  The main reason I want this added functionality is really just the user experience.  I totally agree that the javascript option includes this functionality plus a lot more but I think its somewhat scary for some of our users to come up with a javascript file that accomplishes what they want.  Being able to say to someone who wants to filter out cells for instance, "hey just use this tool and provide values to these two command line options" is much more convenient / user friendly than asking someone who probably doesn't much if any programming experience to "come up with a javascript file that is given access to this record object that you don't really know it got there and use that object's members to construct the filter you want".

nh13:
@jsotobroad I totally understand your points, then my suggestion would be to strongly consider creating a separate tool for this purpose, as now I will get confused which options to use in this tool when looking at the help message.  Why not a simple tool called `FilterSamByTag`?  I think that having a new tool has to be weighed against the proliferation of tools within Picard. Also, I totally understand having easy options for folks who don't have programming experience, and in this case I agree, but we have to be careful to not to over-simplify.

src/main/java/picard/sam/FilterSamReads.java
```diff
@@ -332,19 +355,20 @@ protected int doWork() {
         checkInputs(Arrays.asList(Filter.includeReadList, Filter.excludeReadList), READ_LIST_FILE, "READ_LIST_FILE").ifPresent(errors::add);
         checkInputs(Collections.singletonList(Filter.includePairedIntervals), INTERVAL_LIST, "INTERVAL_LIST").ifPresent(errors::add);
         checkInputs(Collections.singletonList(Filter.includeJavascript), JAVASCRIPT_FILE, "JAVASCRIPT_FILE").ifPresent(errors::add);
+        checkInputs(Arrays.asList(Filter.includeTagValues, Filter.excludeTagValues), TAG, "TAG").ifPresent(errors::add);
```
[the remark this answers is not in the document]
jsotobroad: so passing in no values `TAG_VALUE` is technically valid.  You wont filter out any reads but it also won't fail because it seems the way the command line parser deals with `Array` input types is that by default it is an empty list as opposed to `null`.  These are the reasons I hadn't put it in as a validation check.  Would be curious if you feel it should still be checked (I'm open to either way)

jsotobroad:
@nh13 I think originally I actually had this as a separate tool but was pointed as this as "the filter tool" so I thought I would try to put it in here instead as arguments could be made either way.  Do you think theres any way (updating the help message, better doc strings for the new options, better names, etc etc) that these changes could live in this tool or is it too specific/simplified and should almost certainly be made its own tool.  I imagine it would look very similar to this tool is my only hesitation.

nh13:
@jsotobroad it's up to the Picard maintainers to consider the trade-off, but I would say that the added complexity cannot be fully accounted for in the documentation (who reads more than 2-3 lines anyhow?).  I'll refrain from commenting further, and let @yfarjoun take it from here.

src/main/java/picard/sam/FilterSamReads.java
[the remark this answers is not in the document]
yfarjoun: Yeah, I agree. though I'd emit a warning as it is almost certainly not what the user intended (similar to how DownsampleSam emits a warning when P=0 or P=1.)

src/test/java/picard/sam/FilterSamReadsTest.java
yfarjoun: You still do not want to see TAG_VALUE in anything but include/exclude TagValue...right?

yfarjoun:
My 2ct is that this is a somewhat niche use-case, and that I'd rather not "waste" space in the list of tools. similar to how we didn't split up IntervalListTools into many different CLPs....

src/main/java/picard/sam/FilterSamReads.java
```diff
@@ -276,6 +298,9 @@ protected int doWork() {
             final SamReader samReader = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(INPUT);
             final FilteringSamIterator filteringIterator;
 
+            // Used for exclude/include tag filter
+            List<Object> tagList = TAG_VALUE == null ? null: TAG_VALUE.stream().map(T -> (Object) T).collect(Collectors.toList());
```
jsotobroad: I had to add a null check here for the tests testing the other filters to pass, this will never be null when run as a clp.  I just feel slightly iffy about what I've done.

jsotobroad:
@yfarjoun could you give this a second look

src/main/java/picard/sam/FilterSamReads.java
```diff
@@ -289,6 +289,9 @@ protected int doWork() {
             final SamReader samReader = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(INPUT);
             final FilteringSamIterator filteringIterator;
 
+            // Used for exclude/include tag filter
+            List<Object> tagList = TAG_VALUE.stream().map(T -> (Object) T).collect(Collectors.toList());
```
yfarjoun: didn't you say you're looking for strings? what's this messing around with Objects about?

yfarjoun: ( I realize I asked for it to be refactored...but now, I don't understand why it's needed....)

jsotobroad: Because if I pass the TagFilter() `TAG_VALUE` which is a `List<String>` it will use this constructor https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/samtools/filter/TagFilter.java#L69 instead of the one it would choose if I pass it a `List<Object>` https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/samtools/filter/TagFilter.java#L80 which makes all the `filterOut` functionality not behave correctly.  There is probably a better/easier way to do this conversion

yfarjoun:
can't you do an unchecked cast to (List<Object>)?

jsotobroad:
cant directly use a `(List<Object>)` cast as far as I could tell but I did something similar.

src/main/java/picard/sam/FilterSamReads.java
```diff
@@ -276,6 +298,9 @@ protected int doWork() {
             final SamReader samReader = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(INPUT);
             final FilteringSamIterator filteringIterator;
 
+            // Used for exclude/include tag filter which expects a List<Object> input so casting here
+            List<Object> tagList = (List) TAG_VALUE;
```
pshapiro4broad: This could be better. The problem isn't here, it's in the definition of `TagFilter`. It was a mistake to override the `Object` parameter with `List<Object>`; what the author of that change should've done is used `List<?>`, which would work for any `List`. (Actually using `Collection<?>` would be even better since the code isn't specific to `List`)

jsotobroad: hmm that would be nice.  Should an issue be made in htsjdk?  I don't think I would want to hold up this PR for that change though.

yfarjoun:
If that's the plan, I'd open an issue (not PR) in htsjdk, and add a comment here pointing to that issue...that would enable follow up once the issue is fixed in htsjdk.

src/main/java/picard/sam/FilterSamReads.java
jsotobroad: ticket made in htsjdk for this change/fix - https://github.com/samtools/htsjdk/issues/1082

PR #1085, opened 2018-01-18
set debug argument WRITE_READS_FILES in FilterSamReads to false by default

jsotobroad:
A debugging option for FilterSamReads is set to true by default, this changes it to be false so not everyone gets the "debugging use only" *.reads file with read names.

This is a change that breaks how the command runs by default.

----

#### Content

#### Review

### DIFF
--- a/src/test/java/picard/sam/FilterSamReadsTest.java
+++ b/src/test/java/picard/sam/FilterSamReadsTest.java
@@ -24,6 +24,7 @@
 package picard.sam;
 
 import htsjdk.samtools.*;
+import htsjdk.samtools.util.IOUtil;
 import org.testng.Assert;
 import org.testng.annotations.BeforeTest;
 import org.testng.annotations.DataProvider;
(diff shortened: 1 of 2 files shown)


src/test/java/picard/sam/FilterSamReadsTest.java
```diff
@@ -237,10 +276,18 @@ private FilterSamReads setupProgram(final File inputFile, final File inputSam, f
             default:
                 throw new IllegalArgumentException("Not configured for filter=" + filter);
         }
+        if (writeDebugReads != null){
```
yfarjoun: space ) {

src/test/java/picard/sam/FilterSamReadsTest.java
```diff
@@ -237,10 +276,18 @@ private FilterSamReads setupProgram(final File inputFile, final File inputSam, f
...
+            program.WRITE_READS_FILES = writeDebugReads;
+        }
 
         return program;
     }
 
+    private FilterSamReads setupProgram(final File inputFile, final File inputSam, final FilterSamReads.Filter filter) throws Exception {
+        return setupProgram(inputFile, inputSam, filter, null);
+
```
yfarjoun: remove nl

src/test/java/picard/sam/FilterSamReadsTest.java
```diff
@@ -217,7 +218,45 @@ public void testPairedIntervalFilter(final String intervalFilename, final int ex
...
+                {true, true},
+                {false, false}
+        };
+    }
+
+    /**
+     * makes sure debug files are created properly
+     */
+    @Test(dataProvider = "dataTestDebugOption")
+    public void testDebugOption(Boolean writeDebugReads, Boolean isDebugFileExpected) throws Exception {
```
yfarjoun: can't you take in a boolean and make the code simpler?

src/test/java/picard/sam/FilterSamReadsTest.java
```diff
@@ -217,7 +218,45 @@ public void testPairedIntervalFilter(final String intervalFilename, final int ex
...
+        final File inputReadsFile = new File(filterTest.OUTPUT.getParentFile(), IOUtil.basename(filterTest.INPUT) + ".reads");
+        Assert.assertEquals(inputReadsFile.exists(), isDebugFileExpected.booleanValue());
+
+        final File outputReadsFile = new File(filterTest.OUTPUT.getParentFile(), IOUtil.basename(filterTest.OUTPUT) + ".reads");
+        Assert.assertEquals(outputReadsFile.exists(), isDebugFileExpected.booleanValue());
+
+        // We have to clean up the debug files after each test is run to make sure a clean state is preserved in between tests
+        // This mostly affects the input *.reads file because it will always be called "aligned.reads" and will cause future
+        // tests to fail if it sticks around and we dont expect it to be written
+        if (inputReadsFile.exists()) {
```
yfarjoun: use Files.deleteIfExists()

src/test/java/picard/sam/FilterSamReadsTest.java
jsotobroad: boolean cant be null and I want to test for the default case (the 2nd one could be boolean but it looked weird when I had it that way).  Happy to change the 2nd input though

src/test/java/picard/sam/FilterSamReadsTest.java
jsotobroad: This actually doesn't work in this case because it wont try to delete the file until all runs of the `testDebugOption` method complete but I need it to delete the files between runs.  At least that is what I think it is doing because I tried that and on my 3rd test always failed because the the `aligned.reads` file still existed from the second test where it is supposed to be written out.

jsotobroad: oh...I read your comment incorrectly.  I read `deleteOnExit()`....Sorry!  It took me forever to figure out why `deleteOnExit()` wasnt working

PR #1086, opened 2018-01-18
CrosscheckFingerprint speedup, extra functionality and NIO enabling

yfarjoun:
### Description

This rather large PR does several things that are needed for cross checking large (i.e. 20K) sample call-sets.

1. Multithreaded crosschecking
2. General speedups in the haplotypeBlock code
3. Enable NIO access to google cloud objects (though the JAR needed for this is not bundled into the picard jar....please read the updated ReadMe.md)
4. Enable "smarter" crosschecking to avoid the n^2 problem. In particular this enables one to compare a small set of samples "INPUT" against a vcf that has many samples "SECOND_INPUT" without calculating the result of cross-checking all the samples against themselves. Also, in this mode it will not collapse the data from "INPUT" with data from "SECOND_INPUT" allowing for a true Sample-level crosscheck even if the same sample-name is present in both INPUT and SECOND_INPUT.

I think it would be good to have several different "eyes" on this PR. 

----

#### Content

#### Review


src/main/java/picard/nio/GoogleStorageUtils.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.cloud.storage.contrib.nio.CloudStorageConfiguration;
+import com.google.cloud.storage.contrib.nio.CloudStorageFileSystemProvider;
+import shaded.cloud_nio.com.google.api.gax.retrying.RetrySettings;
+import shaded.cloud_nio.org.threeten.bp.Duration;
+
+
+/**
+ * Created by farjoun on 11/13/17.
+ */
+class GoogleStorageUtils {
```
magicDGS: Cannot GCS be configured in a different way? It looks wierd to me that there is support for the GCS in Picard, but not include the support by default.

It would be nice that both HTSJDK and Picard can use any `java.nio.Path` as a plugin, without extra classes supporting specific filesystems that are afterwards not included in the binary and requires special command line options...

yfarjoun: I agree that it would be good...but I'm not sure how it would be possible to accomplish that.

The current PR allows GCS access to (some) picard tools via GATK. The reason I didn't want to include the google-nio artifact is that it's huge...adding it more than doubles the size of the picard jar....but without the changes I have made here, it's effectively impossible to provide GCS access to picard tools....so I think that it's a good compromise, but I'm happy to hear other opinions.

src/main/java/picard/nio/PathHelper.java
```diff
@@ -0,0 +1,42 @@
...
+import picard.cmdline.ClassFinder;
+
+/**
+ * Created by farjoun on 11/13/17.
+ */
+public class PathHelper {
+
+    private PathHelper() {}
+
+    public static boolean isHasGoogle() {
```
nh13: isHas? That is all.

yfarjoun: I forsee a day in which we will have 

isHasGoogle(), isHasAws(), isHasAzure() etc....

yfarjoun: that said, I agree that the name is somewhat awkward...but I think it's good to keep google, or GCS in the name.

src/main/java/picard/nio/GoogleStorageUtils.java
nh13: It's a good compromise, require the dependency for compiling only, but don't package it by default (i.e. it's an optional dependency).

src/main/java/picard/nio/GoogleStorageUtils.java
```diff
@@ -0,0 +1,46 @@
...
+                .setRetrySettings(RetrySettings.newBuilder()
+                        .setMaxAttempts(15)
+                        .setMaxRetryDelay(Duration.ofMillis(256_000L))
+                        .setTotalTimeout(Duration.ofMillis(4000_000L))
+                        .setInitialRetryDelay(Duration.ofMillis(1000L))
+                        .setRetryDelayMultiplier(2.0)
+                        .setInitialRpcTimeout(Duration.ofMillis(180_000L))
+                        .setRpcTimeoutMultiplier(1.0)
+                        .setMaxRpcTimeout(Duration.ofMillis(180_000L))
+                        .build());
```
nh13: There are a A LOT of settings here, do you want to have them able to be set a la Defaults.java (or GoogleStorageDefaults.java)?

src/main/java/picard/nio/GoogleStorageUtils.java
```diff
@@ -0,0 +1,46 @@
+package picard.nio;
```
nh13: missing copyright

src/main/java/picard/nio/GoogleStorageUtils.java
```diff
@@ -0,0 +1,46 @@
...
+import shaded.cloud_nio.com.google.api.gax.retrying.RetrySettings;
+import shaded.cloud_nio.org.threeten.bp.Duration;
+
+
+/**
+ * Created by farjoun on 11/13/17.
+ */
+class GoogleStorageUtils {
+
+    public static void initialize() {
```
nh13: no javadoc

src/main/java/picard/nio/GoogleStorageUtils.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.cloud.http.HttpTransportOptions;
+import com.google.cloud.storage.StorageOptions;
+import com.google.cloud.storage.contrib.nio.CloudStorageConfiguration;
+import com.google.cloud.storage.contrib.nio.CloudStorageFileSystemProvider;
+import shaded.cloud_nio.com.google.api.gax.retrying.RetrySettings;
+import shaded.cloud_nio.org.threeten.bp.Duration;
+
+
+/**
+ * Created by farjoun on 11/13/17.
```
nh13: Needs a better description.  Include the decision to include this so that we can have an optional dependency on GCS.

src/main/java/picard/nio/PathHelper.java
```diff
@@ -0,0 +1,42 @@
+package picard.nio;
```
nh13: Copyright/license

src/main/java/picard/nio/PathHelper.java
```diff
@@ -0,0 +1,42 @@
...
+
+import picard.cmdline.ClassFinder;
+
+/**
+ * Created by farjoun on 11/13/17.
+ */
```
nh13: This whole class needs documentation.

src/main/java/picard/nio/PathHelper.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    private static boolean isGoogleOnPath() {
+        try {
+
+            new GoogleStorageUtils();
+            return true;
+        } catch (NoClassDefFoundError e) {
+            return false;
+        }
```
nh13: I am not a big fan of all these named methods for google, since this limits how we would add additional nio implementations in the future.  Could you instead have an enum for each supported nio plugin, and then change the methods to take in one or more enum values to either test for (`onPath(Enum provider...)`), initialize (`init(Enum provider..)`), or if supported (`supported(Enum provider..)`).

src/main/java/picard/nio/PathHelper.java
nh13: how about `supports()`?  (see my comments about not making the API google specific)

src/test/java/picard/fingerprint/CheckFingerprintTest.java
```diff
@@ -56,10 +56,10 @@ private void rmTemp() {
     @Test
     public void testBaseOutput() {
         String[] args = new String[]{
-                "I=" + TEST_INPUT_VCF1,
+                "I=" + TEST_INPUT_VCF1.getAbsolutePath(),
```
nh13: Why is this necessary? Why don't you move the `getAbsolutePath()` to where `TEST_INPUT_VCF1` is declared?

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -229,28 +229,34 @@ protected int doWork() {
             outputSummaryMetricsFile = new File(OUTPUT + FINGERPRINT_SUMMARY_FILE_SUFFIX);
         }
 
-        IOUtil.assertFileIsReadable(INPUT);
```
nh13: It would be nice to have this method only check if isn't a URL, or if it is, at least to try to open it!

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -229,28 +229,34 @@ protected int doWork() {
...
+        try {
+            inputPath = IOUtil.getPath(INPUT);
+            genotypesPath = IOUtil.getPath(GENOTYPES);
+        } catch (IOException e) {
+            throw new IllegalArgumentException(e);
+        }
```
nh13: seems like a hack so you don't have to change `IOUtil.assertFileIsReadable`.  Is this going to become the new normal for all of Picard if we want GCS support (it's a slippery slope).

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -187,7 +189,7 @@
 
     @Argument(shortName = "G", doc = "File of genotypes (VCF) to be used in comparison. May contain " +
             "any number of genotypes; CheckFingerprint will use only those that are usable for fingerprinting.")
-    public File GENOTYPES;
+    public String GENOTYPES;
```
nh13: Why not a URL?

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -380,7 +385,15 @@ protected int doWork() {
...
+        return isBamOrSam(new File(s));
+    }
+
+    static boolean isBamOrSam(final File f) {
+        return isBamOrSam(f.toPath());
+    }
+
+    static boolean isBamOrSam(final Path p) {
+        return (p.toUri().getRawPath().endsWith(BamFileIoUtils.BAM_FILE_EXTENSION) || p.toUri().getRawPath().endsWith(IOUtil.SAM_FILE_EXTENSION));
     }
```
nh13: This is really ugly since any other tool will need to do the same. Is there a better way?

yfarjoun: not many tools take in vcf or sam/bam to the same argument and then do different things with it....

src/main/java/picard/fingerprint/CheckFingerprint.java
yfarjoun: why yes? what's the advantage?

src/main/java/picard/fingerprint/CheckFingerprint.java
yfarjoun: See line 328 below.

src/main/java/picard/nio/GoogleStorageUtils.java
lbergelson: If the `Path` api provided a generic method for configuration of FileSystems properties, than we could use that instead.  It doesn't though, and this is the only way that we know to configure important settings in the google-nio library.  The problem is that the default settings that google chose make sense for general use, but not for jobs that last days and make millions of requests.  Since our use case is weird, we need to configure these weird settings.  I don't see any way to do that without including the library at compile time.  

It's likely that we'll want little performance tweaks to whatever nio libraries we work with.  They will work out of the box without them, like the google one would, but most will work better if we configure them specifically for our application.  

Maybe any confusion could be fixed by adding documentation here that would explain that this is for an optional dependency.

src/main/java/picard/fingerprint/CheckFingerprint.java
nh13: I'll respond with a question: what can be represented as a String but not a Path/URL?

src/main/java/picard/fingerprint/CheckFingerprint.java
nh13: got it, I'm dumb.

src/main/java/picard/fingerprint/CheckFingerprint.java
nh13: I don't see anything at 328.  I was talking about changing ` IOUtil.assertFileIsReadable` method, since presumably you'd need to do the same for any other tool in picard to support GCS

src/main/java/picard/nio/GoogleStorageUtils.java
yfarjoun: I think that eventually we will want to expose these...but as there's going to be a mechanism for exposing these parameters in GATK, where NIO is dealt with in a much better way, I'd like to consider options that make configuring picard and GATK easier...so I'd prefer to wait a little (especially given how few tools currently support NIO in picard). Is that reasonable?

src/main/java/picard/nio/PathHelper.java
yfarjoun: good idea. thanks.

src/test/java/picard/fingerprint/CheckFingerprintTest.java
yfarjoun: will do.

src/main/java/picard/fingerprint/CheckFingerprint.java
yfarjoun: well...that's another htsjdk change, which I hesitate to make unless there's an obvious need...so once we see several Picard tools making use of the Path interface, we can figure out how we'd like htsjdk to provide that....

src/main/java/picard/fingerprint/CheckFingerprint.java
yfarjoun: My hope is that we can come up with a wrapper of sorts that behaves like a Path but has a string constructor (which Path doesn't have) so it can be used as an `@Argument`. Until that time, we will need to use this "hack"...I think.

src/main/java/picard/fingerprint/CheckFingerprint.java
yfarjoun: The problem is that Path support for actual URLs (i.e. http://www.google.com is flaky at best...so calling the type a URL can cause confusion. I'm super-unhappy about using String and would love to come up with an alternative.

yfarjoun:
@nh13  can you re-review please?

@hensonc, i'm told this requires your approval, can you please review?

yfarjoun:
@nh13 bump! :-)

src/main/java/picard/fingerprint/CheckFingerprint.java
nh13: It should be pretty easy to add support for `Path` in Barclay similar to how it has support for `Enum`: https://github.com/broadinstitute/barclay/blob/master/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java#L1108

Even more interesting is that Barclay could handle `Optional` rather than relying on `null`, but that's for another day.

We do this for fgbio via our [commons](https://github.com/fulcrumgenomics/commons/blob/master/src/main/scala/com/fulcrumgenomics/commons/reflect/ReflectionUtil.scala#L370-L408) library.  Barclay isn't much different in how it goes about constructing arguments from strings.

README.md
```diff
@@ -83,10 +83,33 @@ During development in Picard, it is sometimes necessary to build locally against
...
 #### Releasing Picard
 
-Full instructions on how to create a new release of Picard are [here](https://github.com/broadinstitute/picard/wiki/How-to-release-Picard)
+Full instructions on how to create a new release of 
+Picard are [here](https://github.com/broadinstitute/picard/wiki/How-to-release-Picard)
 
-----
+#### Path providers
+
+Picard has limited support for reading from Path providers. Currently only google's api is supported, and only a few tools support this
```
nh13: missing period

README.md
```diff
@@ -83,10 +83,33 @@ During development in Picard, it is sometimes necessary to build locally against
...
+java -cp google-cloud-nio.jar -cp picard.jar picard.cmdline.PicardCommandLine CrosscheckFingerprints \
+I=gs://sample1.vcf \
+I=gs://sample2.vcf \
+CROSSCHECK_BY=FILE \
+H=Haplotype_db.txt \
+O=crosscheck.out
+```
+
+Alternatively, you can run the tool via the [GATK](https://software.broadinstitute.org/gatk/download/) which bundles the Google-Cloud
+jar, and should thus "Just Work".
```
nh13: Any way you can set some setting in gradle to bundle this in for you, but have it default to exclude this dependency (optional dependency)?

README.md
```diff
@@ -83,10 +83,33 @@ During development in Picard, it is sometimes necessary to build locally against
...
 
+#### GA4GH API
 It's also possible to build a version of Picard that supports reading from
```
nh13: Newline after 112.

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -243,21 +244,32 @@ public int instanceMain(final String[] argv) {
...
+                }
+
+
                 final boolean usingIntelDeflater = (BlockCompressedOutputStream.getDefaultDeflaterFactory() instanceof IntelDeflaterFactory &&
                         ((IntelDeflaterFactory)BlockCompressedOutputStream.getDefaultDeflaterFactory()).usingIntelDeflater());
                 final boolean usingIntelInflater = (BlockGunzipper.getDefaultInflaterFactory() instanceof IntelInflaterFactory &&
                         ((IntelInflaterFactory)BlockGunzipper.getDefaultInflaterFactory()).usingIntelInflater());
                 final String msg = String.format(
-                    "[%s] Executing as %s@%s on %s %s %s; %s %s; Deflater: %s; Inflater: %s; Picard version: %s",
+                    "[%s] Executing as %s@%s on %s %s %s; %s %s; Deflater: %s; Inflater: %s; %sPicard version: %s",
```
nh13: Can you add a space before `Picard` and modify how you build the path providers string not include the last one?  Optional.

src/main/java/picard/fingerprint/CheckFingerprint.java
nh13: I think this is such a slippery and steep slope I'd like others (who have more engineering background) to weigh in on this. ["With the first link, the chain is forged, ... <the first `File` to `String` conversion> ... chains us all, irrevocably"](http://memory-alpha.wikia.com/wiki/Aaron_Satie)

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -212,10 +214,8 @@
     public static final String FINGERPRINT_SUMMARY_FILE_SUFFIX = "fingerprinting_summary_metrics";
     public static final String FINGERPRINT_DETAIL_FILE_SUFFIX = "fingerprinting_detail_metrics";
 
-    // Stock main method
-    public static void main(final String[] args) {
-        new CheckFingerprint().instanceMainWithExit(args);
-    }
+    private Path inputPath;
+    private Path genotypesPath;
```
nh13: It's hard to see in all the sea of diffs, but could these two be defined in the `doWork` method?

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -229,28 +229,34 @@ protected int doWork() {
...
 
             // Verify that there's only one sample in the SAM/BAM.
-            final SamReader in = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(INPUT.toPath());
+            final SamReader in = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(inputPath);
             for (final SAMReadGroupRecord rec : in.getFileHeader().getReadGroups()) {
                 if (observedSampleAlias == null) {
                     observedSampleAlias = rec.getSample();
                 } else if (!observedSampleAlias.equals(rec.getSample())) {
-                    throw new PicardException("INPUT SAM/BAM file must not contain data from multiple samples.");
+                    throw new PicardException("inputPath SAM/BAM file must not contain data from multiple samples.");
```
nh13: Why was `INPUT` changed to `inputPath` in the exceptions here and below?

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
```diff
@@ -265,6 +291,10 @@
     @Argument(doc = "The number of threads to use to process files and generate fingerprints.")
     public int NUM_THREADS = 1;
 
+    @Argument(doc = "specifies whether the Tumor-aware result should be calculated. These are time consuming and can roughly double the " +
```
nh13: Capitalize `specifies`

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
```diff
@@ -265,6 +291,10 @@
...
+            "runtime of the tool. When crosschecking many groups not calculating the tumor-aware  results can result in a significant speedup.")
```
nh13: extra double space

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
```diff
@@ -291,71 +321,126 @@
...
         return super.customCommandLineValidation();
     }
 
+    enum CrosscheckMode implements CommandLineParser.ClpEnum {
+        CHECK_SAME_SAMPLE {
+            @Override
+            public String getHelpDoc() {
+                return "In this mode, each sample in INPUT will only be checked against a single corresponding sample in SECOND_INPUT. " +
+                        "If a corresponding sample cannot be found, the program will proceed, but report the missing samples" +
+                        " and return the value specified in EXIT_CODE_WHEN_MISMATCH. The corresponding samples are those that equal each other, after possible renaming " +
```
nh13: nit-pick, but the space is at the start of _this_ string, but not the others.

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
```diff
@@ -371,14 +456,82 @@ protected int doWork() {
...
+            log.warn(numUnexpected + " " + CROSSCHECK_BY + "s did not relate as expected.");
             return EXIT_CODE_WHEN_MISMATCH;
         } else {
-            log.info("All read groups related as expected.");
+            log.info("All " + CROSSCHECK_BY + "s are related as expected.");
             return 0;
         }
     }
 
+    /** inspects the contents of sampleMapFile building a map of Sample->Sample.
```
nh13: Capitalize `inspects`

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
```diff
@@ -371,14 +456,82 @@ protected int doWork() {
...
+     * Checks for sanity, and then replaces in the fpMap,
+     * @param fpMap
+     * @param sampleMapFile
+     */
+
```
nh13: found an extra newline!

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
```diff
@@ -371,14 +456,82 @@ protected int doWork() {
...
+    private void remapFingerprints(final Map<FingerprintIdDetails, Fingerprint> fpMap, final File sampleMapFile, final String inputFieldName) {
+        final Map<String, String> sampleMap = new HashMap<>();
+
+        TabbedInputParser parser = new TabbedInputParser(false, sampleMapFile);
```
nh13: can be `final` no?

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
```diff
@@ -371,14 +456,82 @@ protected int doWork() {
...
+                if (unique.add(s))
+                    duplicates.add(s);
+            });
+            throw new IllegalArgumentException("After applying the mapping found in the " + inputFieldName + " the resulting " +
+                    "sample names must be unique when taken together with the remaining unmapped samples. " +
+                    "Duplicates are: [" + String.join(",", duplicates) + "], " + inputFieldName);
+        }
+
+        // replace samples with their mapped values:
+        Set<FingerprintIdDetails> ids = fpMap.keySet();
```
nh13: `final`.

src/main/java/picard/fingerprint/CheckFingerprint.java
danxmoran: The ideal would be to parse directly to `Path`, but one downside we've hit in other projects is that `Paths.get` (etc) uses whatever default storage options are set at the time, so if you want to tweak anything in the options you have to do so before parsing. We've been using `URI` at the parsing level to work around it.

yfarjoun:
Thanks @nh13 . The File->String is unfortunate, I agree. using URL doesn't work as it isn't as general. (many "legal" characters for files are excluded for URL http://www.ietf.org/rfc/rfc2396.txt) 

I think that if really start sliding down the slope you are describing, there will be a strong need for a special type that can represent Files and gs://, s3://, etc. objects and be initialzable from a String. This would have to be implemented in htsjdk and then I'll happily switch to using it...in the meantime, I don't think that this project should wait.

README.md
yfarjoun: I don't know gradle well enough.. @cmnbroad can you advise here?

yfarjoun:
:-)

README.md
cmnbroad: I think the gradle way would be to create a separate ```cloud``` configuration and task for the nio-aware artifact:

```
configurations {
    cloudConfiguration {
        extendsFrom runtime
    }
}

task cloudJar(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar)  {
    configurations = [project.configurations.cloudConfiguration]
    from project.sourceSets.main.output
    archiveName 'picardcloud.jar'
}
```
Then add the conditional dependency to the list of dependencies (the configuration definition above must appear before this in the file):

```
    configurations.cloudConfiguration {
        compile ('com.google.cloud:google-cloud-nio:0.5.0')
    }
```

Then ```.gradlew cloudJar``` will create a picardcloud.jar that includes the, without affecting the other jars.

yfarjoun: thanks @cmnbroad. 
is there a reason you quoted `com.google.cloud:google-cloud-nio:0.5.0' as opposed to `com.google.cloud:google-cloud-nio:0.28.0-alpha:shaded` (which I think is the version we use)?

cmnbroad: No reason, I just took those coordinates off of maven as an example.

yfarjoun:
@tlangs want to review this?

src/main/java/picard/fingerprint/CheckFingerprint.java
tlangs: `INPUT` was changed to a String so that it would be parsed correctly by the Barclay argument parser. `gs://` paths aren't supported by the Barclay parser, since it only has support for `File`s.

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -243,21 +244,34 @@ public int instanceMain(final String[] argv) {
...
+        PathHelper.initilizeAll();
+
         if (!QUIET) {
             System.err.println("[" + new Date() + "] " + commandLine);
 
             // Output a one liner about who/where and what software/os we're running on
             try {
+                final StringBuilder pathProvidersBuilder = new StringBuilder();
+
+                for(PathHelper.PathProviders providers : PathHelper.PathProviders.values()) {
```
tlangs: minor formatting nitpick: `for` should have a space after it.

yfarjoun:
🎉

PR #1092, opened 2018-01-24
Add option to SamToFastq for tags

jsotobroad:
This is still a WIP and doesn't have any tests or great documentation yet but is functional.  Any feedback on what currently exists would be greatly appreciated.

This functionality gives the user the ability to create fastq files from record tags.  Some single cell sequencing (3', 5') start their pipelines with different numbers of fastqs.  One fastq being the sequenced read and then a couple of other fastqs usually for some combination of cell barcode, UMI barcode, sample index.  When converted to sam/bam, these barcode fastq files are converted to tags for each record.  These changes allow for someone to recreate the original set of fastq files from a sam/bam using tags for say a reanalysis or subsetting of data.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/sam/SamToFastq.java
+++ b/src/main/java/picard/sam/SamToFastq.java
@@ -51,7 +51,6 @@ import picard.cmdline.programgroups.ReadDataManipulationProgramGroup;
 
 import java.io.File;
 import java.util.*;
-import java.util.stream.Collector;
 import java.util.stream.Collectors;
 
 /**
@@ -191,6 +190,10 @@ public class SamToFastq extends CommandLineProgram {
     private final static String TAG_SPLIT_DEFAULT_SEP = "";
     private final static String TAG_SPLIT_DEFAULT_QUAL = "~";
 
+    private ArrayList<String[]> SPLIT_SEQUENCE_TAGS;
+    private ArrayList<String[]> SPLIT_QUALITY_TAGS;
+    private ArrayList<String> SPLIT_SEPARATOR_TAGS;
+
     public static void main(final String[] argv) {
         System.exit(new SamToFastq().instanceMain(argv));
     }
@@ -209,7 +212,7 @@ public class SamToFastq extends CommandLineProgram {
         }
 
         final ProgressLogger progress = new ProgressLogger(log);
-        setupTagSplitDefaults();
+        setupTagSplitValues();
         for (final SAMRecord currentRecord : reader) {
             if (currentRecord.isSecondaryOrSupplementary() && !INCLUDE_NON_PRIMARY_ALIGNMENTS)
                 continue;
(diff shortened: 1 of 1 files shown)


src/main/java/picard/sam/SamToFastq.java
```diff
@@ -168,15 +182,34 @@
...
             "is not comprehensive, so there may be exceptions if this is set to true and there are paired reads with non-primary alignments.")
     public boolean INCLUDE_NON_PRIMARY_ALIGNMENTS = false;
 
+    @Argument(shortName = "STG", doc = "List of comma separated tag values to extract from Input SAM/BAM to be used as read sequence", optional = true)
+    public List<String> SEQUENCE_TAG_GROUP;
+
+    @Argument(shortName = "QTG", doc = "List of comma separated tag values to extract from Input SAM/BAM to be used as read qualities", optional = true)
+    public List<String> QUALITY_TAG_GROUP;
+
+    @Argument(shortName = "SEP", doc = "List of sequences to put in between each comma separated list of sequence tags (QFT)", optional = true)
```
nh13: QFT?

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -93,6 +91,22 @@
...
+            "the \"CY\" and \"UY\" tags concatenated together." +
+            "<br />" +
+            "<pre>" +
+            "java -jar picard.jar SamToFastq <br />" +
+            "     I=input.bam<br />" +
+            "     FASTQ=output.fastq<br />" +
+            "     SEQUENCE_TAG_GROUP=CR<br />" +
+            "     QUALITY_TAG_GROUP=CY<br />" +
+            "     SEQUENCE_TAG_GROUP=\"CB,UR\"<br />" +
+            "     QUALITY_TAG_GROUP=\"CY,UY\"" +
```
nh13: I have no idea what the file names are for the FASTQs in this example without reading the code.

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -168,15 +182,34 @@
...
+    public Boolean COMPRESS_OUTPUTS_PER_TAG_GROUP = false;
+
     private final Log log = Log.getInstance(SamToFastq.class);
 
+    private final static String TAG_SPLIT_DEFAULT_SEP = "";
+    private final static String TAG_SPLIT_DEFAULT_QUAL = "~";
+
+    private ArrayList<String[]> SPLIT_SEQUENCE_TAGS;
+    private ArrayList<String[]> SPLIT_QUALITY_TAGS;
+    private ArrayList<String> SPLIT_SEPARATOR_TAGS;
```
nh13: This is just my opinion, but I feel like this tool is dying a death by a thousand cuts.  There are just so many options now, that I am have to read the code to know what it really does.  For example how does the newer option `OUTPUT_PER_RG` and this play together? 

How hard would it be to re-factor this tool so that you can implement another `SamToFastqWithTags` that sub-classes `SamToFastq`?  I worry we'll never detangle this code.

jsotobroad:
Thanks for your feedback @nh13 and I do agree this tool is due for some refactor / de-swiss knifing, it wasn't the easiest trying to figure out what the tool did originally.  I will be asking people around here for ideas on how to best refactor this as I cant cant think of a clear good solution myself.  There is some time component that I would like meet with this functionality but I at least want to explore what a refactor would look like first and reconvene.  If you have any ideas/thoughts that might help that would be useful.

src/main/java/picard/sam/SamToFastq.java
jsotobroad: agreed, I would hope when someone saw the output it would make sense but I totally understand.

src/main/java/picard/sam/SamToFastq.java
jsotobroad: copy paste from when I had written this as a separate tool ><

src/main/java/picard/sam/SamToFastq.java
jsotobroad: As far as `OUTPUT_PER_RG` it would actually just also split the tag fastqs by readgroup as well.  The one weird case is wanting paired fastqs with tag fastqs because then they all wouldnt have the same number of lines.

jacarey:
@jsotobroad @nh13 I'd be happy to help with the refactor of this. I agree that for a utility that seems simple (just a basic file conversion really) the list of options is daunting. I think the first step is to figure out which options are necessary for doing a basic conversion from sam to fastq. Once we have that list I think it will inform the best way to go about doing the refactor.

PR #1097, opened 2018-01-26
Fix tile checking to ensure we are checking against a tile based format.

jacarey:
Fix non deterministic check for tile counts per format.
----

#### Content

#### Review


jacarey:
Testing it is a little tricky but I think we could come up with something.  Ideally we should make it a treeset to remove the non determinism but I'm worried about downstream implications and would want to test more thoroughly

PR #1110, opened 2018-02-09
Add option to SamToFastq for tags

jacarey:
### Description

This functionality gives the user the ability to create fastq files from record tags. Some single cell sequencing (3', 5') start their pipelines with different numbers of fastqs. One fastq being the sequenced read and then a couple of other fastqs usually for some combination of cell barcode, UMI barcode, sample index. When converted to sam/bam, these barcode fastq files are converted to tags for each record. These changes allow for someone to recreate the original set of fastq files from a sam/bam using tags for say a reanalysis or subsetting of data.

#### Content

#### Review


kbergin:
@jacarey @jsotobroad Is this refactor ready for a review by @nh13 ?

jsotobroad:
We wanted someone on green team to review it in case we missed something, possibly @pshapiro4broad, but would happy to have @nh13 review it as well.

pshapiro4broad:
I can review but it would be nice to have a PR with a linked issue, description of what was changed in the PR, etc.

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -93,12 +78,28 @@
...
+            "     SEQUENCE_TAG_GROUP=\"CB,UR\"<br />" +
+            "     QUALITY_TAG_GROUP=\"CY,UY\"" +
+            "</pre>" +
             "<hr />";
     @Argument(doc = "Input SAM/BAM file to extract reads from", shortName = StandardOptionDefinitions.INPUT_SHORT_NAME)
     public File INPUT;
 
     @Argument(shortName = "F", doc = "Output FASTQ file (single-end fastq or, if paired, first end of the pair FASTQ).",
-            mutex = {"OUTPUT_PER_RG", "COMPRESS_OUTPUTS_PER_RG"})
+            mutex = {"OUTPUT_PER_RG", "COMPRESS_OUTPUTS_PER_RG", "OUTPUT_DIR"})
```
pshapiro4broad: Not sure exactly how `mutex` works but it seems like some of these are redundant, e.g. do you need a mutex on both sides of a relationship? (Is there a place where barclay docs are hosted?)

jsotobroad:
There was no issue for these changes but an explanation can be found in the original PR - https://github.com/broadinstitute/picard/pull/1092

src/main/java/picard/sam/SamToFastqWithTags.java
```diff
@@ -0,0 +1,213 @@
...
+
+    @Argument(shortName = "SEP", doc = "List of sequences to put in between each comma separated list of sequence tags in each SEQUENCE_TAG_GROUP (STG)", optional = true)
+    public List<String> TAG_GROUP_SEPERATOR;
+
+    @Argument(shortName = "GZOPTG", doc = "Compress output FASTQ files per Tag grouping using gzip and append a .gz extension to the file names.")
+    public Boolean COMPRESS_OUTPUTS_PER_TAG_GROUP = false;
+
+    private final Log log = Log.getInstance(SamToFastqWithTags.class);
+
+    private final static String TAG_SPLIT_DEFAULT_SEP = "";
```
pshapiro4broad: Correct modifier ordering is `private static final`; the IDE should highlight this and sort them for you.

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -286,21 +260,70 @@ protected int doWork() {
...
                 final Lazy<FastqWriter> lazySecondOfPairWriter = new Lazy<>(() -> INTERLEAVE ? firstOfPairWriter : factory.newWriter(makeReadGroupFile(rg, "_2")));
-                writerMap.put(rg, new FastqWriters(firstOfPairWriter, lazySecondOfPairWriter, firstOfPairWriter));
+
+                writerMap.put(rg, new FastqWriters(firstOfPairWriter, lazySecondOfPairWriter, firstOfPairWriter, additionalWriters));
             }
         }
         return writerMap;
     }
 
+    protected void initializeAdditionalWriters() {
```
pshapiro4broad: I'm not a fan of this kind of extension, as it's unclear what a subclass is supposed to do if they subclass `SamToFastqWithTags` and want to override this method again.

Given that this change creates a direct dependency between these two classes, I think you'd be better off using `super.` to override the base class functionality. I think you can do this by overriding `doWork()` or  `generateWriters()`. E.g. you could override `doWork()` in `SamToFastqWithTags` as:
```
    protected int doWork() {
        setupTagSplitValues();
        return super.doWork();
    }
```

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -286,21 +260,70 @@ protected int doWork() {
...
+                if (secondOfPairWriter == null) {
+                    throw new PicardException("Input contains paired reads but no SECOND_END_FASTQ specified.");
+                }
+                writeRecord(read2, 2, secondOfPairWriter, READ2_TRIM, READ2_MAX_BASES_TO_WRITE);
+            }
+        } else {
+            writeRecord(currentRecord, null, fq.getUnpaired(), READ1_TRIM, READ1_MAX_BASES_TO_WRITE);
+        }
+
+        handleAdditionalRecords(currentRecord, fq.additionalWriters, read1, read2);
```
pshapiro4broad: As with `initializeAdditionalWriters()` above, I think it would be clearer to have the subclass override `handleRecord()` and call `super.handleRecord()` to execute the code in the base class.

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -383,22 +404,22 @@ else if (CLIPPING_ACTION.equalsIgnoreCase("N")) {
...
+                    result.insert(0, replacement);
                 }
             }
         }
-        return result;
+        return result.toString();
     }
 
-    private void assertPairedMates(final SAMRecord record1, final SAMRecord record2) {
+    protected void assertPairedMates(final SAMRecord record1, final SAMRecord record2) {
```
pshapiro4broad: It looks like this can be static

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -383,22 +404,22 @@ else if (CLIPPING_ACTION.equalsIgnoreCase("N")) {
      */
     private String clip(final String src, final int point, final Character replacement, final boolean posStrand) {
```
pshapiro4broad: It looks like this can be static

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -414,50 +435,46 @@ private void assertPairedMates(final SAMRecord record1, final SAMRecord record2)
...
-                    "Cannot set INTERLEAVE to true and pass in a SECOND_END_FASTQ"
-            };
+            errors.add("Cannot set INTERLEAVE to true and pass in a SECOND_END_FASTQ");
         }
 
         if (UNPAIRED_FASTQ != null && SECOND_END_FASTQ == null) {
-            return new String[]{
-                    "UNPAIRED_FASTQ may only be set when also emitting read1 and read2 fastqs (so SECOND_END_FASTQ must also be set)."
-            };
+            errors.add("UNPAIRED_FASTQ may only be set when also emitting read1 and read2 fastqs (so SECOND_END_FASTQ must also be set).");
```
pshapiro4broad: Can this assertion be done using the `Argument` annotation?

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -414,50 +435,46 @@ private void assertPairedMates(final SAMRecord record1, final SAMRecord record2)
...
-            return new String[]{
-                    "Both or neither of CLIPPING_ATTRIBUTE and CLIPPING_ACTION should be set."};
+            errors.add("Both or neither of CLIPPING_ATTRIBUTE and CLIPPING_ACTION should be set.");
         }
 
         if (CLIPPING_ACTION != null) {
-            if (CLIPPING_ACTION.equals("N") || CLIPPING_ACTION.equals("X")) {
-                // Do nothing, this is fine
-            } else {
+            if (!CLIPPING_ACTION.equals("N") && !CLIPPING_ACTION.equals("X")) {
```
pshapiro4broad: Should have a named constant for these strings as they're repeated a few times here. E.g. `CLIP_TRIM`, `CLIP_TO_N`.

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -469,17 +486,25 @@ private void assertPairedMates(final SAMRecord record1, final SAMRecord record2)
...
-            this(firstOfPair, new Lazy<>(() -> secondOfPair), unpaired);
+        /**
+         * Simple constructor; all writers are pre-initialized..
+         */
+        private FastqWriters(final FastqWriter firstOfPair, final FastqWriter secondOfPair, final FastqWriter unpaired,
+                             Map<SAMReadGroupRecord, List<FastqWriter>> additionalWriters) {
+            this(firstOfPair, new Lazy<>(() -> secondOfPair), unpaired, additionalWriters);
         }
 
         public FastqWriter getFirstOfPair() {
```
pshapiro4broad: pretty sure all the methods in this nested class can be `private`

src/main/java/picard/sam/SamToFastqWithTags.java
```diff
@@ -0,0 +1,213 @@
...
+
+import java.io.File;
+import java.util.*;
+import java.util.stream.Collectors;
+
+@CommandLineProgramProperties(
+        summary = SamToFastqWithTags.USAGE_SUMMARY + SamToFastqWithTags.USAGE_DETAILS,
+        oneLineSummary = SamToFastqWithTags.USAGE_SUMMARY,
+        programGroup = ReadDataManipulationProgramGroup.class)
+public class SamToFastqWithTags extends SamToFastq {
```
pshapiro4broad: javadoc for class?

src/main/java/picard/sam/SamToFastqWithTags.java
```diff
@@ -0,0 +1,213 @@
...
+    static final String USAGE_SUMMARY = "Converts a SAM or BAM file to FASTQ.";
```
pshapiro4broad: Should its summary be the same as the one in `SamToFastq`?

src/main/java/picard/sam/SamToFastqWithTags.java
```diff
@@ -0,0 +1,213 @@
...
+            "<hr />";
+
+    @Argument(shortName = "STG", doc = "List of comma separated tag values to extract from Input SAM/BAM to be used as read sequence", minElements = 1)
+    public List<String> SEQUENCE_TAG_GROUP;
+
+    @Argument(shortName = "QTG", doc = "List of comma separated tag values to extract from Input SAM/BAM to be used as read qualities", optional = true)
+    public List<String> QUALITY_TAG_GROUP;
+
+    @Argument(shortName = "SEP", doc = "List of sequences to put in between each comma separated list of sequence tags in each SEQUENCE_TAG_GROUP (STG)", optional = true)
+    public List<String> TAG_GROUP_SEPERATOR;
```
pshapiro4broad: Would it be helpful to include an example that shows all the arguments? It's not clear to me what values would make sense here, although I don't know much about running this tool.

src/main/java/picard/sam/SamToFastqWithTags.java
```diff
@@ -0,0 +1,213 @@
...
+
+    @Argument(shortName = "GZOPTG", doc = "Compress output FASTQ files per Tag grouping using gzip and append a .gz extension to the file names.")
+    public Boolean COMPRESS_OUTPUTS_PER_TAG_GROUP = false;
+
+    private final Log log = Log.getInstance(SamToFastqWithTags.class);
+
+    private final static String TAG_SPLIT_DEFAULT_SEP = "";
+    private final static String TAG_SPLIT_DEFAULT_QUAL = "~";
```
pshapiro4broad: The name says `DEFAULT` but it doesn't appear to have a way to override it.

src/main/java/picard/sam/SamToFastqWithTags.java
```diff
@@ -0,0 +1,213 @@
...
+    private ArrayList<String> SPLIT_SEPARATOR_TAGS;
+
+    @Override
+    protected void initializeAdditionalWriters() {
+        setupTagSplitValues();
+    }
+
+    @Override
+    protected void handleAdditionalRecords(SAMRecord currentRecord, Map<SAMReadGroupRecord, List<FastqWriter>> tagWriters, SAMRecord read1, SAMRecord read2) {
+        if (currentRecord.isSecondaryOrSupplementary() && !INCLUDE_NON_PRIMARY_ALIGNMENTS)
```
pshapiro4broad: please use `{` `}` around the `if` body

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -64,13 +49,13 @@
...
  * the original read sequence as it was generated by the sequencer.</p>
  * <br />
- " <h4>Usage example:</h4>" +
- " <pre>" +
- " java -jar picard.jar SamToFastq \\<br />" +
- "     I=input.bam \\<br />" +
- "     FASTQ=output.fastq" +
- " </pre>" +
- " <hr />"
+ * " <h4>Usage example:</h4>" +
```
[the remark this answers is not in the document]
jsotobroad: Not sure why the double quotes are there, those were there before.

src/main/java/picard/sam/SamToFastq.java
jsotobroad: I actually like the redundancy because when you look at the descriptions of each argument you can see what each one is mutually exclusive to

src/main/java/picard/sam/SamToFastqWithTags.java
jsotobroad: my ide does not but changed

src/main/java/picard/sam/SamToFastq.java
jsotobroad: done

src/main/java/picard/sam/SamToFastq.java
jsotobroad: done

src/main/java/picard/sam/SamToFastq.java
jsotobroad: As far as i know you can only designate things as mutually exclusive but not mutually inclusive.

src/main/java/picard/sam/SamToFastq.java
jsotobroad: done

src/main/java/picard/sam/SamToFastq.java
jsotobroad: done

src/main/java/picard/sam/SamToFastqWithTags.java
jsotobroad: yup

src/main/java/picard/sam/SamToFastqWithTags.java
jsotobroad: whoops

src/main/java/picard/sam/SamToFastqWithTags.java
jsotobroad: I added some more info to doc for this input, hopefully that helps

src/main/java/picard/sam/SamToFastqWithTags.java
jsotobroad: SEP can be overwritten by providing a value for `TAG_GROUP_SEPERATOR`, qual cannot.  Change made

src/main/java/picard/sam/SamToFastqWithTags.java
jsotobroad: Is someone going to go through the codebase and fix these?  Its hard to tell what is and isnt proper style for this repo

src/main/java/picard/sam/SamToFastq.java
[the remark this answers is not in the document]
pshapiro4broad: Can you please remove them, from here and from the javadoc for `SamToFastqWithTags`? Even though it was written this way before, it is incorrect to have double quotes in this part of the javadoc. You also don't want the `+` symbols, the `\` should not be doubled, and the `<br/>` results in an extra newline (because it's used inside a `<pre>`, which already preserves newlines). Instead it should look like:
```
 * <h4>Usage example:</h4>
 *  <pre>
 *  java -jar picard.jar SamToFastq \
 *      I=input.bam \
 *      FASTQ=output.fastq
 *  </pre>
 * <hr />
```

It is easy in the IDE to render javadoc so you can see right away what the output will look like, and fix it there.

src/main/java/picard/sam/SamToFastqWithTags.java
pshapiro4broad: I think the desired target is the Sun/Oracle Java coding standard (Google also has a Java coding standard which also requires braces around if/else/etc bodies). Regarding fixing all the code, I don't know if it's worth the effort to reformat everything, personally I would reformat code as it's modified due to other work being done.

src/main/java/picard/sam/SamToFastq.java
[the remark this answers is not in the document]
jsotobroad: 👍  I didnt even try looking at it in the viewer

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -466,40 +474,51 @@ private void assertPairedMates(final SAMRecord record1, final SAMRecord record2)
...
             return unpaired;
         }
 
-        public void closeAll() {
-            final Set<FastqWriter> fastqWriters = new HashSet<FastqWriter>();
+        private void closeAll() {
+            final Set<FastqWriter> fastqWriters = new HashSet<>();
             fastqWriters.add(firstOfPair);
             fastqWriters.add(unpaired);
+            fastqWriters.addAll(additionalWriters.values().stream().flatMap(Collection::stream).collect(Collectors.toList()));
```
pshapiro4broad: I think `additionalWriters.values().forEach(fastqWriters::addAll);` is equivalent

src/main/java/picard/sam/SamToFastqWithTags.java
```diff
@@ -0,0 +1,237 @@
...
+    private ArrayList<String> SPLIT_SEPARATOR_TAGS;
+
+    @Override
+    protected void initializeAdditionalWriters() {
+        setupTagSplitValues();
+    }
+
+    @Override
+    protected void handleAdditionalRecords(SAMRecord currentRecord, Map<SAMReadGroupRecord, List<FastqWriter>> tagWriters, SAMRecord read1, SAMRecord read2) {
+        if (currentRecord.isSecondaryOrSupplementary() && !INCLUDE_NON_PRIMARY_ALIGNMENTS) {
```
pshapiro4broad: These appear to duplicate checks done in `SamToFastq`. Are they needed here too?

src/main/java/picard/sam/SamToFastqWithTags.java
```diff
@@ -0,0 +1,237 @@
...
+            }
+        } else {
+            baseFilename = "";
+        }
+
+        List<File> tagFiles = new ArrayList<>();
+        for (String tagSplit : SEQUENCE_TAG_GROUP) {
+            String fileName = baseFilename;
+
+            fileName += tagSplit.replace(",", "_");
```
pshapiro4broad: Why not write as:
```
String fileName = baseFilename + tagSplit.replace(",", "_");
```
That avoids using `+=` on `String`.

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -466,40 +474,51 @@ private void assertPairedMates(final SAMRecord record1, final SAMRecord record2)
      * Allows for lazy construction of the second-of-pair writer, since when we are in the "output per read group mode", we only wish to
      * generate a second-of-pair fastq if we encounter a second-of-pair read.
      */
-    static final class FastqWriters {
+    private static final class FastqWriters {
         private final FastqWriter firstOfPair, unpaired;
         private final Lazy<FastqWriter> secondOfPair;
+        private final Map<SAMReadGroupRecord, List<FastqWriter>> additionalWriters;
```
pshapiro4broad: This field is in an object that itself is in a `Map<SAMReadGroupRecord, FastqWriters>`. Do we need a map inside a map for this feature? It seems like what you really want is `writers` and `additionalWriters` as both being local variables in `doWork()`.

jsotobroad: nice catch, I split it out into its own variable in `doWork()` for now and we can change after we talk

jsotobroad:
@pshapiro4broad I believe the changes we talked about are in if you could take another look.  Thanks!

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -237,6 +200,13 @@ protected int doWork() {
             writerMapping.closeAll();
         }
 
+        // close all `additionalWriters` only once
+        final Set<FastqWriter> additionalWriterSet = new HashSet<>();
+        additionalWriters.values().forEach(additionalWriterSet::addAll);
+        for (final FastqWriter fastqWriter : additionalWriterSet) {
```
pshapiro4broad: It's done this way because you can have duplicate writers in the additional writers collection? Otherwise you could write
```
additionalWriters.values().stream().flatMap(Collection::stream).forEach(FastqWriter::close);
```

src/main/java/picard/sam/SamToFastq.java
```diff
@@ -245,11 +215,8 @@ protected int doWork() {
         return 0;
     }
 
-    /**
-     * Generates the writers for the given read groups or, if we are not emitting per-read-group, just returns the single set of writers.
-     */
-    private Map<SAMReadGroupRecord, FastqWriters> generateWriters(final List<SAMReadGroupRecord> samReadGroupRecords,
-                                                                  final FastqWriterFactory factory) {
+    private Map<SAMReadGroupRecord, FastqWriters> generateWriters(List<SAMReadGroupRecord> samReadGroupRecords,
+                                                                  FastqWriterFactory factory) {
```
pshapiro4broad: `factory` is no longer final? Also why remove javadoc?

src/main/java/picard/sam/SamToFastqWithTags.java
```diff
@@ -0,0 +1,225 @@
...
+    @Override
+    protected void initializeAdditionalWriters() {
+        setupTagSplitValues();
+    }
+
+    @Override
+    protected void handleAdditionalRecords(SAMRecord currentRecord, Map<SAMReadGroupRecord, List<FastqWriter>> tagWriters, SAMRecord read1, SAMRecord read2) {
+        final List<FastqWriter> rgTagWriters = tagWriters.get(currentRecord.getReadGroup());
+        if (currentRecord.getReadPairedFlag()) {
+            if (read1 != null && read2 !=null){
```
pshapiro4broad: need space before `{`

src/main/java/picard/sam/SamToFastqWithTags.java
```diff
@@ -0,0 +1,225 @@
...
+            for (final SAMReadGroupRecord rg : samReadGroupRecords) {
+                List<FastqWriter> tagWriters = makeTagWriters(rg).stream().map(factory::newWriter).collect(Collectors.toList());
+
+                writerMap.put(rg, tagWriters);
+            }
+        }
+        return writerMap;
+    }
+
+    // Creates fastq writers based on readgroup passed in and sequence tag groupings from command line
```
pshapiro4broad: may as well make this a javadoc comment

src/main/java/picard/sam/SamToFastqWithTags.java
```diff
@@ -0,0 +1,225 @@
...
+    private List<File> makeTagWriters(final SAMReadGroupRecord readGroup) {
```
pshapiro4broad: This is called twice and in both cases the result is passed to `.stream().map(factory::newWriter).collect(Collectors.toList())`. To avoid duplication that work could be done in this method, or another method that calls this method.

Or this method could be modified to call `factory.newWriter()` directly when before adding the new writer to the collection `tagFiles`.

src/main/java/picard/sam/SamToFastqWithTags.java
```diff
@@ -0,0 +1,225 @@
...
+                    : new File(FASTQ.getParent(), fileName);
+            IOUtil.assertFileIsWritable(result);
+            tagFiles.add(result);
+        }
+        return tagFiles;
+    }
+
+    // Setting up the Groupings of Sequence Tags, Quality Tags, and Separator Strings so we dont have to calculate them for every loop
+    private void setupTagSplitValues() {
+        if (SEQUENCE_TAG_GROUP.isEmpty()) return;
```
pshapiro4broad: please split across lines and add `{` `}`

src/main/java/picard/sam/SamToFastq.java
jsotobroad: exactly, if your bam has multiple readgroups but you have `OUTPUT_PER_RG` not set then you will have duplicate writers.

src/main/java/picard/sam/SamToFastq.java
jsotobroad: hmm I have no idea how i managed to do that.....

src/main/java/picard/sam/SamToFastqWithTags.java
jsotobroad: 👍

src/main/java/picard/sam/SamToFastqWithTags.java
jsotobroad: yea it would nice if the method returned what the comment/method name describe.....

src/main/java/picard/sam/SamToFastqWithTags.java
jsotobroad: I can actually get rid of this `if` entirely, thanks.

PR #1121, opened 2018-02-22
Get set of genes from OverlapDetector once rather than over and over, for speed.

alecw:
### Description

Change: Compute set of genes (return value of overlapDetector.getAll()) once in RnaSeqMetricsCollector ctor, rather than over and over again in pickTranscripts().

Motivation: For Drop-seq data, there can be tens of thousands of libraries rather than the handful of libraries in a convention Rna-seq experiment, so computing this over and over is very time-consuming.

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
+++ b/src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
@@ -38,6 +38,7 @@ public class RnaSeqMetricsCollector extends SAMRecordMultiLevelCollector<RnaSeqM
     final private Set<Integer> ignoredSequenceIndices;
 
     private final OverlapDetector<Gene> geneOverlapDetector;
+    private final Set<Gene> allGenes;
     private final OverlapDetector<Interval> ribosomalSequenceOverlapDetector;
     private final boolean collectCoverageStatistics;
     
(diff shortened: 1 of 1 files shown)


src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
```diff
@@ -38,6 +38,7 @@
     final private Set<Integer> ignoredSequenceIndices;
 
     private final OverlapDetector<Gene> geneOverlapDetector;
+    private final Set<Gene> allGenes;
```
pshapiro4broad: You could declare this as `Collection<>` to avoid an issue if the return type of `OverlapDetector.getAll()` ever changes, since your code doesn't care if it's a `Set` or not.

pshapiro4broad:
This is better, but it still results in `getAll()` being called twice as it's also called by `CollectRnaSeqMetrics.setup()`. It would be even better to cache the result in `OverlapDetector` itself so a caller doesn't have to worry about this implementation detail of the class. This would be slightly more complicated since you'd have to invalidate the cache in `OverlapDetector` when a `put` occurs. But that would also make it easier to know that the cached result is always correct; in this checkin it could become invalid.

alecw:
Thanks, @pshapiro4broad.  I'll submit a PR to htsjdk with the change you suggest, and if that is accepted I'll close this one.  OK?

alecw:
I'm going to close this and submit a different PR that will both enable caching and result in much better performance.

PR #1126, opened 2018-02-27
Change order of iteration in pickTranscripts for speed.

alecw:
### Description

Change order of iteration in pickTranscripts for speed, and create an overridable method so that OverlapDetector.getAll result can be cached.

### Motivation

For single-cell libraries, the number of transcripts in a library is much smaller than the global list of transcripts.  Therefore, when creating a list of genes in the library, it can be much faster to iterate over the transcripts in the library in order to arrive at a list of genes in the library, compared to iterating over the global list of transcripts and testing each transcript for its presence in the library.  

For conventional RNA-seq libraries, this change might not help as much, but it should not hurt performance because the number of transcripts in a library should never be larger than the global list of transcripts.

In addition, for single-cell experiments, there can be tens of thousands of libraries for which metrics are to be accumulated.  Calling geneOverlapDetector.getAll() for each library is very expensive, and generally is unnecessary given that the return value is the same each time.  Providing an overridable method for making this call allows extenders of this class to cache the list of genes and return the same list for each invocation.

For my test data set, these changes resulted in a **75x** performance improvement.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
+++ b/src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
@@ -430,12 +430,22 @@ public class RnaSeqMetricsCollector extends SAMRecordMultiLevelCollector<RnaSeqM
             return out;
         }
 
+        /**
+         * Derived class may override this method in order to return a different set of genes for
+         * picking transcripts for computing coverage, or to cache the value returned by geneOverlapDetector.getAll(),
+         * because it can be expensive to compute repeatedly.
+         */
+        protected Set<Gene> getGenesForPickTranscripts() {
+            return geneOverlapDetector.getAll();
+        }
+
         /** Picks the set of transcripts on which the coverage metrics are to be calculated. */
         public Map<Gene.Transcript, int[]> pickTranscripts(final Map<Gene.Transcript, int[]> transcriptCoverage) {
-            final Map<Gene.Transcript, Double> bestPerGene = new HashMap<Gene.Transcript, Double>();
+            final Set<Gene> genesWithTranscripts = getGenesWithTranscripts(transcriptCoverage.keySet());
 
             // Make a map of the best transcript per gene to it's mean coverage
-            for (final Gene gene : geneOverlapDetector.getAll()) {
+            final Map<Gene.Transcript, Double> bestPerGene = new HashMap<Gene.Transcript, Double>();
+            for (final Gene gene : genesWithTranscripts) {
                 Gene.Transcript best = null;
                 double bestMean = 0;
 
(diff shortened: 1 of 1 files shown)


src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
```diff
@@ -430,20 +430,34 @@ else if (this.metrics.SAMPLE != null) {
...
+        }
+
         /** Picks the set of transcripts on which the coverage metrics are to be calculated. */
         public Map<Gene.Transcript, int[]> pickTranscripts(final Map<Gene.Transcript, int[]> transcriptCoverage) {
-            final Map<Gene.Transcript, Double> bestPerGene = new HashMap<Gene.Transcript, Double>();
+            // Get the set of genes for which there are transcripts
+            // Note that the size of transcriptCoverage may be << than size of genesForPickTranscripts, so iterating
+            // over that set should be much faster.  If the size is not much smaller, it shouldn't be significantly
+            // larger.
+            final Set<Gene> genesWithTranscripts = new HashSet<>();
```
pshapiro4broad: It might be clearer to move this into its own method, since the only value that's used after this block is `genesWithTranscripts`. Then the comment can be changed to a javadoc for the method.

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
```diff
@@ -430,20 +430,34 @@ else if (this.metrics.SAMPLE != null) {
             return out;
         }
 
+        protected Set<Gene> getGenesForPickTranscripts() {
```
pshapiro4broad: Is the idea that a subclass would want to change this method, or would need access to it? In either case it would be helpful to add a javadoc about what it does (conceptually) in this class. If it can be made private, then it would be OK IMO to not bother adding a javadoc.

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
alecw: Good idea.

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
alecw: Yes, it is there for a subclass to override.  I added javadoc.

src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
```diff
@@ -430,20 +430,29 @@ else if (this.metrics.SAMPLE != null) {
...
+        }
+
         /** Picks the set of transcripts on which the coverage metrics are to be calculated. */
         public Map<Gene.Transcript, int[]> pickTranscripts(final Map<Gene.Transcript, int[]> transcriptCoverage) {
-            final Map<Gene.Transcript, Double> bestPerGene = new HashMap<Gene.Transcript, Double>();
+            final Set<Gene> genesWithTranscripts = getGenesWithTranscripts(transcriptCoverage.keySet());
 
             // Make a map of the best transcript per gene to it's mean coverage
-            for (final Gene gene : geneOverlapDetector.getAll()) {
+            final Map<Gene.Transcript, Double> bestPerGene = new HashMap<Gene.Transcript, Double>();
```
pshapiro4broad: You can use a diamond operator here, e.g. `new HashMap<>()`. the compiler infers the type parameters from the variable declaration.

PR #1127, opened 2018-03-01
CreateSequenceDictionary support for alternative names (@SQ-AN)

lindenb:
### Description

this PR add a new option **AN** in **CreateSequenceDictionary** to support the alternative name in a dictionary, as defined in the sam spec  https://github.com/samtools/hts-specs/blob/master/SAMv1.tex#L212

```
ALT_NAMES=File
AN=File                       Optional file containing the alternative names for the contigs. First column is the
                              original name, the second column is an alternative name. One contig may have more than one
                              alternative name.   Default value: null. 
```

In the code, I added a` Map<String,Set<String>>`  mapping the contigs to a set of aliases. This map is then used when a SamSequenceRecord is serialized.

a test was added in CreateSequenceDictionaryTest

Related PR  in htsjdk : https://github.com/samtools/htsjdk/pull/956/files

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/sam/CreateSequenceDictionary.java
+++ b/src/main/java/picard/sam/CreateSequenceDictionary.java
@@ -223,9 +223,9 @@ public class CreateSequenceDictionary extends CommandLineProgram {
             // read reference sequence one by one and write its metadata
             for (ReferenceSequence refSeq = refSeqFile.nextSequence(); refSeq != null; refSeq = refSeqFile.nextSequence()) {
                 final SAMSequenceRecord samSequenceRecord = makeSequenceRecord(refSeq);
-                //add aliases
-                if (aliasesByContig.containsKey(samSequenceRecord.getSequenceName())) {
-                    final Set<String> aliases = aliasesByContig.get(samSequenceRecord.getSequenceName());
+                // retrieve aliases, if any
+                final Set<String> aliases = aliasesByContig.get(samSequenceRecord.getSequenceName());
+                if (aliases != null) {
                     // "Alternative names is a comma separated list of alternative names"
                     samSequenceRecord.setAttribute(AN_ATTRIBUTE, String.join(",", aliases));
                 }
@@ -241,7 +241,7 @@ public class CreateSequenceDictionary extends CommandLineProgram {
         // check uniqueness of sequences names
         final CloseableIterator<String> iterator = sequenceNames.iterator();
 
-        if(!iterator.hasNext()) return 0;
+        if (!iterator.hasNext()) return 0;
 
         String current = iterator.next();
         while (iterator.hasNext()) {
@@ -324,51 +324,59 @@ public class CreateSequenceDictionary extends CommandLineProgram {
      * @throws PicardException if there is any error in the file ALT_NAMES
      */
     private Map<String, Set<String>> loadContigAliasesMap() throws PicardException {
-    	// return an empty map if no mapping file was provided
-    	if (this.ALT_NAMES == null) return Collections.emptyMap();
-    	// the map returned by the function
-    	final Map<String, Set<String>> aliasesByContig = new HashMap<>();
-        try {
-	        for(final String line :IOUtil.slurpLines(this.ALT_NAMES)) {
-	            if (StringUtil.isBlank(line)) continue;
-	            final int tab = line.indexOf('\t');
-	            if (tab == -1 ) throw new IOException("tabulation missing in " + line);
-	            final String contigName = line.substring(0,tab);
-	            final String altName = line.substring(tab + 1);
-	            //check for empty values
-	            if (StringUtil.isBlank(contigName)) {
-	                throw new IOException("empty contig in  " + line);
-	            }
-	            if (StringUtil.isBlank(altName)) {
-	                throw new IOException("empty alternative name in  " + line);
-	            }
-	            if (altName.equals(contigName)) continue;
-	            if (!ALTERNATIVE_CONTIG_NAME_PATTERN.matcher(altName).matches()) {
-	                throw new IOException("alternative name in  " + line +
-	                        " doesn't match the regular expression : " +
-	                        ALTERNATIVE_CONTIG_NAME_PATTERN.pattern());
-	            }
-	            //check alias not previously defined as contig
-	            if (aliasesByContig.containsKey(altName)) {
-	                throw new IOException("alternate name  " + altName +
-	                        "previously defined as a contig in " + line);
-	            }
-	            //check contig not previously defined as alias
-	            if (aliasesByContig.keySet().stream().
-	                    filter(K->!K.equals(contigName)). // not an error is defined twice for same contig
-	                    anyMatch(K->aliasesByContig.get(K).contains(contigName))) {
-	                throw new IOException("contig  " + contigName +
-	                        "previously defined as an alternate name in " + line);
-	            }
-	            //add alias
-	            if (!aliasesByContig.containsKey(contigName)) {
-	                aliasesByContig.put(contigName, new HashSet<>());
-	            }
-	            aliasesByContig.get(contigName).add(altName);
-	        }
-	        return aliasesByContig;  
+        // return an empty map if no mapping file was provided
+        if (this.ALT_NAMES == null) {
+            return Collections.emptyMap();
         }
-        catch (final IOException e) {
+        // the map returned by the function
+        final Map<String, Set<String>> aliasesByContig = new HashMap<>();
+        try {
+            for (final String line :IOUtil.slurpLines(this.ALT_NAMES)) {
+            if (StringUtil.isBlank(line)) {
+                    continue;
+                }
+                final int tab = line.indexOf('\t');
+                if (tab == -1 ) {
+                    throw new IOException("tabulation missing in " + line);
+                }
+                final String contigName = line.substring(0,tab);
+                final String altName = line.substring(tab + 1);
+                // check for empty values
+                if (StringUtil.isBlank(contigName)) {
+                    throw new IOException("empty contig in  " + line);
+                }
+                if (StringUtil.isBlank(altName)) {
+                    throw new IOException("empty alternative name in  " + line);
+                }
+                if (altName.equals(contigName)) {
+                    continue;
+                }
+                if (!ALTERNATIVE_CONTIG_NAME_PATTERN.matcher(altName).matches()) {
+                    throw new IOException("alternative name in  " + line +
+                            " doesn't match the regular expression : " +
+                            ALTERNATIVE_CONTIG_NAME_PATTERN.pattern());
+                }
+                // check alias not previously defined as contig
+                if (aliasesByContig.containsKey(altName)) {
+                    throw new IOException("alternate name  " + altName +
+                            "previously defined as a contig in " + line);
+                }
+                // check contig not previously defined as alias
+                if (aliasesByContig.keySet().stream().
+                        // not an error if defined twice for same contig
+                        filter(K->!K.equals(contigName)). 
+                        anyMatch(K->aliasesByContig.get(K).contains(contigName))) {
+                        throw new IOException("contig  " + contigName +
+                            "previously defined as an alternate name in " + line);
+                }
+                // add alias
+                if (!aliasesByContig.containsKey(contigName)) {
+                    aliasesByContig.put(contigName, new HashSet<>());
+                }
+                aliasesByContig.get(contigName).add(altName);
+            }
+            return aliasesByContig;  
+        } catch (final IOException e) {
             throw new PicardException("Can't read alias file " + ALT_NAMES, e);
         }
    
--- a/src/main/java/picard/sam/CreateSequenceDictionary.java
+++ b/src/main/java/picard/sam/CreateSequenceDictionary.java
@@ -52,6 +52,7 @@ import java.nio.file.Files;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -118,6 +119,18 @@ public class CreateSequenceDictionary extends CommandLineProgram {
     public File ALT_NAMES = null;
 
     private final MessageDigest md5;
+    
+    /** 
+     * Regular expression defined in the sam spec. Any alternative contig should match this regular expression 
+     * TODO: replace the pattern with a constant : see https://github.com/samtools/htsjdk/pull/956/files
+     */
+    private static final Pattern ALTERNATIVE_CONTIG_NAME_PATTERN = Pattern.compile("[0-9A-Za-z][0-9A-Za-z\\*\\+@\\|\\-]*");
+    
+    /**  
+     * 'AN' attribute in the dictionary
+     * TODO: replace "AN" with a constant : see https://github.com/samtools/htsjdk/pull/956/files
+     */
+    private static final String AN_ATTRIBUTE = "AN";
 
     public CreateSequenceDictionary() {
         try {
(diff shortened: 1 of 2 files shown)


magicDGS:
Maybe we can wait until https://github.com/samtools/htsjdk/pull/956 is in, but it's a good idea. I was looking forward that change for a while!

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -185,6 +196,58 @@ protected int doWork() {
                     " already exists.  Delete this file and try again, or specify a different output file.");
         }
 
+        // map for aliases mapping a contig to its' aliases
```
pshapiro4broad: By `its'` I think you really mean `its`

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -185,6 +196,58 @@ protected int doWork() {
...
+        final Map<String, Set<String>> contig2aliases = new HashMap<>();
```
pshapiro4broad: In general I wouldn't use numbers in variable names like this, in the place of the word "to". Looking at other Maps in picard, such maps are typically named by the value, not the key, such as `aliases`. In cases where both the key and value are mentioned, one convention is to use `by`, e.g. `aliasesByContig`.

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -185,6 +196,58 @@ protected int doWork() {
...
+        
+        // fill the alias map
+        if(this.ALT_NAMES != null) {
+            try {
```
pshapiro4broad: This is a lot of code to insert directly into a method. If the output of this block is the `Map<>` above, it shouldn't be hard to move it into a method that returns the `Map`.

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -185,6 +196,58 @@ protected int doWork() {
                     " already exists.  Delete this file and try again, or specify a different output file.");
         }
 
+        // map for aliases mapping a contig to its' aliases
+        final Map<String, Set<String>> contig2aliases = new HashMap<>();
+        
+        // fill the alias map
+        if(this.ALT_NAMES != null) {
```
pshapiro4broad: There should be a space between `if` and `(` here and below

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -185,6 +196,58 @@ protected int doWork() {
...
+            try {
+                // regex defined in the sam spec
+                final Pattern altNameRegex = Pattern.compile("[0-9A-Za-z][0-9A-Za-z\\*\\+@\\|\\-]*");
```
pshapiro4broad: Compiled patterns are thread safe so the usual convention is to store them in a `static final`. Also since this regex is somewhat confusing, either a better explanation would be helpful, or a comment that links to the place in the spec where this regex is specified.

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -185,6 +196,58 @@ protected int doWork() {
...
+                for(final String line :IOUtil.slurpLines(this.ALT_NAMES)) {
```
pshapiro4broad: need a space after the `:`

pshapiro4broad: Instead of `IOUtil.slurpLines()` I would use the NIO function `Files.readAllLines()` here.

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -185,6 +196,58 @@ protected int doWork() {
...
+                    if(StringUtil.isBlank(line)) continue;
```
pshapiro4broad: each `if` body needs to be split across lines and use `{` `}`

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -185,6 +196,58 @@ protected int doWork() {
...
+                    final int tab = line.indexOf('\t');
+                    if(tab == -1 ) throw new IOException("tabulation missing in " + line);
+                    final String contigName = line.substring(0,tab);
+                    final String altName = line.substring(tab + 1);
```
pshapiro4broad: I think `String.split()` would be a more natural way to express this.

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -185,6 +196,58 @@ protected int doWork() {
...
+                    }
+                    //check alias not previously defined as contig
+                    if(contig2aliases.containsKey(altName)) {
+                        throw new IOException("alternate name  " + altName +
+                                "previously defined as a contig in " + line);
+                    }
+                    //check contig not previously defined as alias
+                    if(contig2aliases.keySet().stream().
+                            filter(K->!K.equals(contigName)). // not an error is defined twice for same contig
+                            flatMap(K->contig2aliases.get(K).stream()).
```
pshapiro4broad: If you need to iterate over both keys and values, it's better to stream over `entrySet()` instead of looping over keys and doing a `get()` for each key.

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -185,6 +196,58 @@ protected int doWork() {
...
+                                altNameRegex.pattern());
+                    }
+                    //check alias not previously defined as contig
+                    if(contig2aliases.containsKey(altName)) {
+                        throw new IOException("alternate name  " + altName +
+                                "previously defined as a contig in " + line);
+                    }
+                    //check contig not previously defined as alias
+                    if(contig2aliases.keySet().stream().
+                            filter(K->!K.equals(contigName)). // not an error is defined twice for same contig
```
pshapiro4broad: Need spaces around `->` operator. Also `K`, `S` here are variable names (anonymous function parameters) so they should be named like parameters. Use `key` or `value` (or `entry` if you're streaming over `entrySet`

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -185,6 +196,58 @@ protected int doWork() {
...
+                                "previously defined as an alternate name in " + line);
+                    }
+                    //add alias
+                    if(!contig2aliases.containsKey(contigName)) {
+                        contig2aliases.put(contigName, new HashSet<>());
+                    }
+                    contig2aliases.get(contigName).add(altName);
+                }
+            }
+            catch (final IOException e) {
```
pshapiro4broad: `catch` should be on previous line, e.g. `} catch (final IOException e) {`

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -196,6 +259,12 @@ protected int doWork() {
             // read reference sequence one by one and write its metadata
             for (ReferenceSequence refSeq = refSeqFile.nextSequence(); refSeq != null; refSeq = refSeqFile.nextSequence()) {
                 final SAMSequenceRecord samSequenceRecord = makeSequenceRecord(refSeq);
+                //add aliases
+                if(contig2aliases.containsKey(samSequenceRecord.getSequenceName())) {
```
pshapiro4broad: To avoid hashing twice you could instead write
```
final Set<String> aliases = contig2aliases.get(samSequenceRecord.getSequenceName());
if (aliases != null) {
    ...
}
```

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -196,6 +259,12 @@ protected int doWork() {
             // read reference sequence one by one and write its metadata
             for (ReferenceSequence refSeq = refSeqFile.nextSequence(); refSeq != null; refSeq = refSeqFile.nextSequence()) {
                 final SAMSequenceRecord samSequenceRecord = makeSequenceRecord(refSeq);
+                //add aliases
```
pshapiro4broad: comments should have a space after the `//`

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -196,6 +259,12 @@ protected int doWork() {
...
+                if(contig2aliases.containsKey(samSequenceRecord.getSequenceName())) {
+                    final Set<String> aliases = contig2aliases.get(samSequenceRecord.getSequenceName());
+                    // "Alternative names is a comma separated list of alternative names"
```
pshapiro4broad: Not sure why this is in double quotes? Even if this is quoted from a spec, quotes are unnecessary in a comment like this.

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -196,6 +259,12 @@ protected int doWork() {
...
+                    samSequenceRecord.setAttribute("AN",String.join(",",aliases)); //TODO replace "AN" with constants/methods: https://github.com/samtools/htsjdk/pull/956/files
```
pshapiro4broad: In general, end-of-line comments are hard to read/maintain, it's better to put them on their own line. If you had a static final for the magic string `"AN"` then that would be a good place for a comment like this.

pshapiro4broad: Need to have spaces after `,`s here

src/test/java/picard/sam/CreateSequenceDictionaryTest.java
```diff
@@ -114,4 +122,53 @@ public void testNonUniqueSequenceName() throws Exception {
         Assert.assertEquals(runPicardCommandLine(argv), 0);
         Assert.fail("Exception should have been thrown.");
     }
+    
+    
+    @Test
+    public void testAltNames() throws Exception {
+        final File altFile = File.createTempFile("CreateSequenceDictionaryTest.", ".alt");
```
pshapiro4broad: first arg to `createTempFile()` shouldn't end in a `.`

src/test/java/picard/sam/CreateSequenceDictionaryTest.java
```diff
@@ -114,4 +122,53 @@ public void testNonUniqueSequenceName() throws Exception {
...
+        
+        // check chr2
+        ssr = dict.getSequence("chr2");
+        Assert.assertNotNull(ssr, "chr2 missing in dictionary");
+        an = ssr.getAttribute("AN");
+        Assert.assertNull(an, "AN Present");
+        
+        // check chrM
+        ssr = dict.getSequence("chrM");
+        Assert.assertNull(ssr, "chrM presnt in dictionary");
```
pshapiro4broad: typo `presnt`

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -185,6 +196,58 @@ protected int doWork() {
...
+                    //check alias not previously defined as contig
+                    if(contig2aliases.containsKey(altName)) {
+                        throw new IOException("alternate name  " + altName +
+                                "previously defined as a contig in " + line);
+                    }
+                    //check contig not previously defined as alias
+                    if(contig2aliases.keySet().stream().
+                            filter(K->!K.equals(contigName)). // not an error is defined twice for same contig
+                            flatMap(K->contig2aliases.get(K).stream()).
+                            anyMatch(S->S.contains(contigName))) {
```
pshapiro4broad: I think you could avoid the `flatMap` here by using `contains` on each set and then using `findFirst`. I can write something up if this doesn't make sense.

lindenb:
@pshapiro4broad thank you for the review ! :-)
back to you. 

https://github.com/broadinstitute/picard/pull/1127#discussion_r172935622 fixed typo
https://github.com/broadinstitute/picard/pull/1127#discussion_r172936629 replace contig2aliases by aliasesByContig
https://github.com/broadinstitute/picard/pull/1127#discussion_r172936958 moved code to its own function
https://github.com/broadinstitute/picard/pull/1127#discussion_r172937641 moved regex to static final
https://github.com/broadinstitute/picard/pull/1127#discussion_r172937138 moved the 'if(' to 'if ('
https://github.com/broadinstitute/picard/pull/1127#discussion_r172944679 moved "AN" to a static final String with comment about Todo
https://github.com/broadinstitute/picard/pull/1127#discussion_r172945325 add spaces
https://github.com/broadinstitute/picard/pull/1127#discussion_r172945618 remove dot in tmp name
https://github.com/broadinstitute/picard/pull/1127#discussion_r172946563 fixed typo
https://github.com/broadinstitute/picard/pull/1127#discussion_r172947848 remove flatMap

gbggrant:
@yfarjoun look ok?

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -284,6 +317,63 @@ private String md5Hash(final byte[] bytes) {
...
 
+    /**
+     * Load the file ALT_NAMES containing the alternative contig names
+     * @author Pierre Lindenbaum
+     * @return a <code>Map&lt;src_contig,Set&lt;new_names&gt;&gt;</code>. Never null. May be empty if ALT_NAMES is null.
+     * @throws PicardException if there is any error in the file ALT_NAMES
+     */
+    private Map<String, Set<String>> loadContigAliasesMap() throws PicardException {
+    	// return an empty map if no mapping file was provided
+    	if (this.ALT_NAMES == null) return Collections.emptyMap();
```
pshapiro4broad: Please make the formatting changes from my first set of comments:
- `if` statements should be multi-line and have `{` `}` around the body
- `for(` should be written as `for (`
- `catch` should be wrapped up to the line before it, as `} catch (...) {`
- `//` comments should have a space between the `//` and the first letter of the comment

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -284,6 +317,63 @@ private String md5Hash(final byte[] bytes) {
...
+    	// the map returned by the function
+    	final Map<String, Set<String>> aliasesByContig = new HashMap<>();
+        try {
```
pshapiro4broad: Spacing looks odd here; `try` should match up with the line before it, and the body of the `try` appears to have the wrong indent level.

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -196,6 +223,12 @@ protected int doWork() {
             // read reference sequence one by one and write its metadata
             for (ReferenceSequence refSeq = refSeqFile.nextSequence(); refSeq != null; refSeq = refSeqFile.nextSequence()) {
                 final SAMSequenceRecord samSequenceRecord = makeSequenceRecord(refSeq);
+                //add aliases
+                if (aliasesByContig.containsKey(samSequenceRecord.getSequenceName())) {
```
pshapiro4broad: As in my previous comment, you can avoid hashing twice by calling `.get()` and then checking for `null` instead of using `containsKey()`. You only need to use `containsKey()` if your map might contain `null` as a value, and you wish to detect that case.

lindenb:
@pshapiro4broad thanks again.
back to you.
I hope I've fixed all your points.

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -106,7 +112,25 @@
...
+            + "First column is the original name, the second column is an alternative name. "
+            + "One contig may have more than one alternative name. " ,
+            optional=true)
+    public File ALT_NAMES = null;
+
     private final MessageDigest md5;
+    
+    /** 
+     * Regular expression defined in the sam spec. Any alternative contig should match this regular expression 
+     * TODO: replace the pattern with a constant : see https://github.com/samtools/htsjdk/pull/956/files
```
pshapiro4broad: Is this supposed to be the same as `SAMSequenceRecord.ALTERNATIVE_SEQUENCE_NAME_REGEXP`? It's not the same, it has a `|` character in it.

If it is supposed to be the same, you should make the one in `SamSequenceRecord` public and use it instead of defining another copy here.

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -106,7 +112,25 @@
...
+     */
+    private static final Pattern ALTERNATIVE_CONTIG_NAME_PATTERN = Pattern.compile("[0-9A-Za-z][0-9A-Za-z\\*\\+@\\|\\-]*");
```
pshapiro4broad: The `\\` characters in this string aren't necessary since the characters `*` `+` and `|` do not require quoting. `-` would require quoting but doesn't since it occurs last in the character class.

src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -106,7 +112,25 @@
     @Argument(doc = "Stop after writing this many sequences.  For testing.")
     public int NUM_SEQUENCES = Integer.MAX_VALUE;
 
+    @Argument(shortName = "AN", doc = "Optional file containing the alternative names for the contigs. "
```
yfarjoun: Explain how this input will be used.

yfarjoun:
@lindenb are you going to respond to the comments? I know it may seem like an arduous process, but we would like to include your contribution!

lindenb:
@yfarjoun sorry, I forgot about that one, I'll try to answer tomorrow.

lindenb:
@yfarjoun  @pshapiro4broad 

Thank you again for your reviews and sorry for my late answer

 I fixed the following points, Back to you.

https://github.com/broadinstitute/picard/pull/1127#discussion_r174872620  indented four spaces from the for
https://github.com/broadinstitute/picard/pull/1127#discussion_r174872721 missing space ':'
https://github.com/broadinstitute/picard/pull/1127#discussion_r174873066 missing space
https://github.com/broadinstitute/picard/pull/1127#discussion_r174873174 missing space
https://github.com/broadinstitute/picard/pull/1127#discussion_r174875051 remove extra space
https://github.com/broadinstitute/picard/pull/1127#discussion_r174875098 need space after the ,
https://github.com/broadinstitute/picard/pull/1127#discussion_r174875429 fix indent
https://github.com/broadinstitute/picard/pull/1127#discussion_r174875756 need space

https://github.com/broadinstitute/picard/pull/1127#discussion_r174877910

> Is this supposed to be the same as SAMSequenceRecord.ALTERNATIVE_SEQUENCE_NAME_REGEXP? It's not the same, it has a | character in it.

I just checked: there is a '|' in the SAM spec.  https://samtools.github.io/hts-specs/SAMv1.pdf

https://github.com/broadinstitute/picard/pull/1127#discussion_r174878872

> The \\ characters in this string aren't necessary since the characters * +

they' required by the java compiler because \+ is treated as `invalid escape sequence`

https://github.com/broadinstitute/picard/pull/1127#discussion_r181113518  Explain how this input will be used.

jmarshall:
Regarding ALTERNATIVE_SEQUENCE_NAME_REGEXP: indeed, `|` is listed in the spec, very intentionally so that NCBI-style pipe-containing contig names can be relegated to being aliases.

Seems like the regexp is missing `.` and `_` compared to the one in the SAM spec.

lindenb:
test failed for a travis-related problem IMHO

magicDGS:
I updated https://github.com/samtools/htsjdk/pull/956 to reflect the regexp changes too. You can use that one, @lindenb

lindenb:
@magicDGS :+1:  thanks ! I suppose I've to wait for your PR to be merged

yfarjoun:
I agree that the failures in travis are not due to the PR. The master branch is currently failing...there's a PR to fix that ( #1178 )

src/main/java/picard/sam/CreateSequenceDictionary.java
yfarjoun: To be clear, I mean in the documentation...something like: "The alternative names will be put into the appropriate `@AN` annotation for each contig". Also, please state how the "columns" need to be separated ("tab", I presume) and whether the columns should have a "header" (I presume that they do not)

lindenb: @yfarjoun I added the sentences.

yfarjoun:
👍 thanks @lindenb

PR #1130, opened 2018-03-06
Tl dsdegp 1641 check illumina directory return code

tlangs:
### Description
`CheckIlluminaDirectory` currently returns the number of errors encountered as its return code. The problem with this is that the standard is that only 8 bits (255) are used for return codes, and some return codes have special meanings. For example, there was recently a situation where the return code, modulus 255, meant that Sun GridEngine retried the job thousands of times because a failing return code was interpreted as something else. 

This PR changes `CheckIlluminaDirectory` to return `1` on error and `0` on success. If there are errors, the number of errors is written to `errors.count` in the current working directory. 

While I was in there, I replaced some unsafe operations with try with resources to make things safer.
----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/illumina/CheckIlluminaDirectory.java
+++ b/src/main/java/picard/illumina/CheckIlluminaDirectory.java
@@ -23,6 +23,7 @@ import picard.illumina.parser.readers.CbclReader;
 import picard.illumina.parser.readers.LocsFileReader;
 
 import java.io.File;
+import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.ArrayList;
@@ -56,7 +57,8 @@ public class CheckIlluminaDirectory extends CommandLineProgram {
             "and are reasonably sized for every tile and cycle.  Reasonably sized means non-zero sized for files that exist per tile and " +
             "equal size for binary files that exist per cycle or per tile. If DATA_TYPES {Position, BaseCalls, QualityScores, PF," +
             " or Barcodes} are not specified, then the default data types used by IlluminaBasecallsToSam are used.  " +
-            "CheckIlluminaDirectory DOES NOT check that the individual records in a file are well-formed.</p>" +
+            "CheckIlluminaDirectory DOES NOT check that the individual records in a file are well-formed. If there are errors, " +
+            "the number of errors is written in a file called 'errors.count' in the working directory</p>" +
             "" +
             "<h4>Usage example:</h4> " +
             "<pre>" +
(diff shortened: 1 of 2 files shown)


src/main/java/picard/illumina/CheckIlluminaDirectory.java
```diff
@@ -230,7 +236,12 @@ protected int doWork() {
         if (totalFailures == 0) {
             log.info("SUCCEEDED!  All required files are present and non-empty.");
         } else {
-            status = totalFailures;
+            status = 1;
+            try {
+                Files.write(Paths.get(".", "errors.count"), Integer.toString(totalFailures).getBytes());
+            } catch (IOException e) {
+                log.error("Unable to write number off errors to file");
```
jacarey: of not off

src/main/java/picard/illumina/CheckIlluminaDirectory.java
```diff
@@ -161,42 +165,44 @@ protected int doWork() {
...
                 //check s.locs
                 final File locsFile = new File(BASECALLS_DIR.getParentFile(), AbstractIlluminaPositionFileReader.S_LOCS_FILE);
-                final LocsFileReader locsFileReader = new LocsFileReader(locsFile);
-                final List<AbstractIlluminaPositionFileReader.PositionInfo> locs = new ArrayList<>();
-                while (locsFileReader.hasNext()) {
-                    locs.add(locsFileReader.next());
+                final List<AbstractIlluminaPositionFileReader.PositionInfo> locs;
+                final Map<Integer, File> filterFileMap;
+                try (LocsFileReader locsFileReader = new LocsFileReader(locsFile)) {
+                    locs = new ArrayList<>();
```
danxmoran: Woah... this works even when it's `final`?

Why not just assign the `new ArrayList` at the point where you declare the var?

src/main/java/picard/illumina/CheckIlluminaDirectory.java
```diff
@@ -161,42 +165,44 @@ protected int doWork() {
...
+                    while (locsFileReader.hasNext()) {
+                        locs.add(locsFileReader.next());
+                    }
                 }
 
-                final Map<Integer, File> filterFileMap = new HashMap<>();
+                filterFileMap = new HashMap<>();
```
danxmoran: Same question, why not assign at the point where you declare it?

src/main/java/picard/illumina/CheckIlluminaDirectory.java
tlangs: its declared final and then assigned. It just can't be re-assigned.

src/main/java/picard/illumina/CheckIlluminaDirectory.java
tlangs: I'm not sure why these got changed. I didn't change them... I least I don't remember changing them.

PR #1084, opened 2018-01-17
RevertSam does not accept /dev/stdin - fixed.

grozhnev:
1 - setDestructiveIteration property of SortingCollection of SAMRecords in RevertSamSorter constructor  to 'false' value:
			singleSorter.setDestructiveIteration(false);

	2 - createReadGroupFormatMap method refactoring; // processing previously filled SortingCollections by RevertSamSorter iterators instead of reading files from /dev/stdin
		2.2 - deletion of unnecessary properties;
		2.3 - use iteration inside each of the ReadGroup;
		2.4 - readGroup NPE fix;

### Description
This pull request contains fix to the issue at https://github.com/broadinstitute/picard/issues/1008.

The problem was that **RevertSam tool didn't work with /dev/stdin in a proper way**, it couldn't define file qualities to restore the previous state of BAM-file on input.

I've **changed the structure of readGroupToFormat()** mapping in RevertSam class. Previously this map was filling by the second iteration of /dev/stdin. I **made filling this map by SortingCollection** that was build by the first iteration of /dev/stdin.
Actually the second iteration of /dev/stdin led to NullPointerException as buffer gets empty after filling sorting collection. 

**In readGroupToFormat()** method i've **changed input parameters**, replaced the unnecessary SAMReadGroupRecord iterator with iterator of Sorting Collection.

I also **changed setDestructiveIteration() parameter** of SortingCollection in RevertSamSorter constructor for the reason that iteration on SortingCOllection began from the element with 0 index that was null, instead of the element with index 1.

----

#### Content
- [ ] Added or modified tests to cover changes and any new functionality
- [ ] Edited the README / documentation (if applicable)
- [ ] All tests passing on Travis

#### Review
- [ ] Final thumbs-up from reviewer
- [ ] Rebase, squash and reword as applicable

For more detailed guidelines, see https://github.com/broadinstitute/picard/wiki/Guidelines-for-pull-requests

PR #1131, opened 2018-03-07
Test for piping htsjdk data

tlangs:
### Description
A recent change in htsjdk broke the ability of pipe BAM output from stdin into a Picard command line tool. This PR adds a test to make sure that piping into picard is tested.
----

#### Content

#### Review


src/test/java/picard/sam/PipedDataTest.java
```diff
@@ -0,0 +1,67 @@
...
+import picard.cmdline.CommandLineProgramTest;
+
+public class PipedDataTest extends CommandLineProgramTest {
+
+    private String classPath = "\"" + System.getProperty("java.class.path") + "\" ";
+
+
+    @Override
+    public String getCommandLineProgramName() {
+        return SortSam.class.getSimpleName();
```
magicDGS: I guess that this is wrong...maybe an exception saying that "this shouldn't be called" is better than assigning a name for a random tool...

danxmoran: Do we need to extend `CommandLineProgramTest` at all? It doesn't seem a good fit if it forces picking one tool name (but I'm not sure of all the magic it provides).

magicDGS: I think that in this case the "magic" (making command line with arguments and the tool name, or running the tool using `Main`) is not used anywhere, so probably is better to don't extend...

tlangs: This is a remnant of an older revision that I forgot to delete.

danxmoran:
@tlangs I took another look and you were right that #1084 seems like a separate, more complicated error than just piping a bam. Can you remove the `RevertSam` test so we can merge this? HTSJDK [fixed the issue](https://github.com/samtools/htsjdk/commit/2b8919b50479c1046cae80be6dd766b3db7424eb) in 2.14.3 (which we're using now) so I'd expect the simple `SortSam` test to pass.

kbergin:

PR #1133, opened 2018-03-08
Made unpaired reads not fall through to paired read checks

tlangs:
### Description

RevertSam had a bug in the sanitation step where unpaired records would fall through to code blocks that required paired records, which caused sanitation to fail. This PR fixes that.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/sam/RevertSam.java
+++ b/src/main/java/picard/sam/RevertSam.java
@@ -429,13 +429,15 @@ public class RevertSam extends CommandLineProgram {
                         if (unpairedRecord == null) unpairedRecord = rec;
                         ++unpaired;
                     }
-                    if (rec.getFirstOfPairFlag()) {
-                        if (firstRecord == null) firstRecord = rec;
-                        ++firsts;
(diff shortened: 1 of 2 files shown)


src/main/java/picard/sam/RevertSam.java
```diff
@@ -429,13 +429,15 @@ public void revertSamRecord(final SAMRecord rec) {
...
                         ++unpaired;
                     }
-                    if (rec.getFirstOfPairFlag()) {
-                        if (firstRecord == null) firstRecord = rec;
-                        ++firsts;
-                    }
-                    if (rec.getSecondOfPairFlag()) {
-                        if (secondRecord == null) secondRecord = rec;
-                        ++seconds;
+                    else {
```
pshapiro4broad: This code is not very good java... For the code you're updating, let's at least format the `else` correctly and put it on the previous line, e.g. `} else {`.

The `if`s should be formatted as:
```
if (firstRecord == null) {
    firstRecord = rec;
}
```

PR #1140, opened 2018-03-22
Include google-cloud-nio only in the cloudJar, not in the shadowJar.

cmnbroad:
google-cloud-nio is currently being included in the shadowJar (picard.jar) but should only be included in the cloudJar configuration. Before this change the two jars (picard.jar and picardcloud.jar) were the about the same size. With this:

-rw-r--r--  1 cnorman  CHARLES\Domain Users  13257211 Mar 22 11:34 picard.jar
-rw-r--r--  1 cnorman  CHARLES\Domain Users  30166495 Mar 22 11:34 picardcloud.jar

#### Content

#### Review

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,6 +58,9 @@ final javadocJDKFiles = files(((URLClassLoader) ToolProvider.getSystemToolClassL
 configurations {
     cloudConfiguration {
         extendsFrom runtime
+        dependencies {
+            cloudConfiguration('com.google.cloud:google-cloud-nio:' + googleNioVersion)
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


cmnbroad:
@lbergelson Can you also take a look if you get a chance ? thx.

yfarjoun:
I don't understand the changes in build.xml but I can attest that the resulting jars do and do not contain google's cloud libary as appropriate. 👍

droazen:
@cmnbroad Have you confirmed that this dependency in Picard doesn't collide with the GATK dependency on the NIO library (since we use a custom-patched version, and we take in Picard as a maven dependency).

cmnbroad:
@droazen I haven't, but this isn't adding the dependency - that already exists and was introduced in picard 2.18.0 - its removing it from everything except the cloudJar, which isn't a maven artifact. It may be that there is currently a collision, but gatk has never taken picard 2.18.0, we're still on 2.17.11.

droazen:
@cmnbroad Can you file an issue in the GATK repo? We are going to have to blacklist the Picard NIO dependency in our `build.gradle`.

cmnbroad:
@droazen Yeah, we can force-resolve it to the one we want.

droazen:
We might also need to consider having both projects depend on a build of htsjdk that comes with the NIO plugin, as this blacklisting approach is untenable over the longer term.

Issue #1149, opened 2018-04-01
Can LiftoverVcf TAGS_TO_REVERSE be made more automatic?


bbimber:
### Affected tool(s)

LiftoverVcf

### Affected version(s)
- [ x] Latest public release version [version?]
- [ ] Latest development/master branch as of [date of test?]

### Description 

You recently added code to automatically try to reverse the REF/ALT alleles if the lifted variant is reverse complemented.  As you recognized, this would sometimes invalidate INFO field tags.  You special-cased AF and MAX_AF and added TAGS_TO_REVERSE; however, this is still pretty manual and easy for less sophisticated users to get wrong.

Can you actually inspect the header and use tags of type VCFHeaderLineType.A and VCFHeaderLineType.R to more automatically determine what needs to get reversed?  This will not help with the MAX_AF situation, but it would help considerably over the current code.

#### Actual behavior

----

### Tool(s) involved
_Tool name(s), special parameters?_

### Description
_Specify whether you want a modification of an existing behavior or addition of a new capability._
_Provide **examples**, **screenshots**, where appropriate._

----

### Tool(s) involved
_Tool name(s), parameters?_

### Description 
_Describe what needs to be added or modified._

----

bbimber:
For example, AF is special-cased, but AC is also more or less a built-in field that is also allele level, but it does not get this special treatment.

yfarjoun:
AC is a reasonable tag to add to the default, but I'm not sure how to automate and at the same time allow for over-ride. I'm happy to discuss possibilities or review a PR if you have a proposed implementation.

bbimber:
The tool cant reasonably be expected to get everything right automatically.  However, if a field is either A (per allele) or R (per allele plus ref), there is near certainty that swapping REF/ALT is going to leave them invalid.  I understand the goal of lifting the maximum number of variants, but VCF file format is complex, and the current default behavior of this tool is quite likely to leave the user with invalid data, in a way that is pretty cryptic and hard to realize.  

It's useful to special-case AF (which is an admittedly important, but also annotation use-case), but this seem like an instance where this tool special-cases whatever narrow scenario drove the addition of the feature, but isnt doing an adequate job of thinking about the broader space.

For solutions:

- As you saw, I put in a separate PR to add an argument to disable this behavior, since at least in some of our cases the problems this introduced for annotations is greater than the additional variants.

[...]

PR #1151, opened 2018-04-02
Add arguments and reporting around recovering REF/ALT swapped alleles

bbimber:
### Description

Relatively recently LiftoverVcf has a feature added to automatically recover lifted variants with swapped REF/ALT alleles.  See #1149 for some discussion around this.  This PR:

1) Adds an argument to LiftoverVcf to allow users to disable that feature, due to the problems it can cause with allele-based annotations.

2) Adds a new opt-in feature to record the ORIGINAL_ALLELES, in addition to ORIGINAL_START and ORIGINAL_CONTIG.  This is something not directly related to the swapping feature, but as the allele-level modifications get more complicated, simply stashing the original alleles can be useful.  For example, reverse complemented indels might get left-aligned, and reversing that whole process isnt simple.  

3) This is perhaps the one controversial thing I propose: I'd like to see the allele swapping feature default to false, for the reasons noted in #1149, since I think that can cause subtle but real problems in VCFs and there isnt a good solution in LiftoverVcf to handle that.  Generally speaking, it seems like public tools such as Picard should err on the more conservative side, and therefore I believe that's where the default should be unless the feature becomes more robust.

4) I added/augmented tests

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/util/LiftoverUtils.java
+++ b/src/main/java/picard/util/LiftoverUtils.java
@@ -63,7 +63,7 @@ public class LiftoverUtils {
      * @param target The target interval
      * @param refSeq The reference sequence, which should match the target interval
      * @param writeOriginalPosition If true, INFO field annotations will be added to store the original position and contig
-     * @param writeOriginalAlleles If true, INFO an field annotation will be added to store the original alleles in order.  This can be useful in the case of more complex liftovers, like reverse complements, left aligned indels or swapped REF/ALT
+     * @param writeOriginalAlleles If true, an INFO field annotation will be added to store the original alleles in order.  This can be useful in the case of more complex liftovers, like reverse complements, left aligned indels or swapped REF/ALT
      * @return The lifted VariantContext.  This will be null if the input VariantContext could not be lifted.
      */
     public static VariantContext liftVariant(final VariantContext source, final Interval target, final ReferenceSequence refSeq, final boolean writeOriginalPosition, final boolean writeOriginalAlleles){
@@ -102,7 +102,7 @@ public class LiftoverUtils {
         }
 
         if (writeOriginalAlleles && !source.getAlleles().equals(builder.getAlleles())) {
-            List<String> alleles = new ArrayList<>();
+            final List<String> alleles = new ArrayList<>();
             source.getAlleles().forEach(a -> alleles.add(a.getDisplayString()));
             builder.attribute(LiftoverVcf.ORIGINAL_ALLELES, alleles);
         }
--- a/src/main/java/picard/util/LiftoverUtils.java
+++ b/src/main/java/picard/util/LiftoverUtils.java
@@ -108,9 +108,16 @@ public class LiftoverUtils {
         return builder.make();
     }
 
+    /**
+     * This is a utility method that will convert a list of alleles into a list of base strings.  Reference status
+     * is ignored when creating these strings (i.e. 'A', not 'A*').  These strings should be sufficient
+     * to recreate an Allele using Allele.create()
+     * @param alleles The list of alleles
+     * @return A list of strings representing the bases of the input alleles.
+     */
     protected static List<String> allelesToStringList(final List<Allele> alleles) {
         final List<String> ret = new ArrayList<>();
-        alleles.forEach(a -> ret.add(a.getDisplayString()));
+        alleles.forEach(a -> ret.add(a.isNoCall() ? Allele.NO_CALL_STRING : a.getDisplayString()));
         return ret;
     }
 
--- a/src/main/java/picard/util/LiftoverUtils.java
+++ b/src/main/java/picard/util/LiftoverUtils.java
@@ -63,9 +63,10 @@ public class LiftoverUtils {
      * @param target The target interval
      * @param refSeq The reference sequence, which should match the target interval
      * @param writeOriginalPosition If true, INFO field annotations will be added to store the original position and contig
+     * @param writeOriginalAlleles If true, INFO an field annotation will be added to store the original alleles in order.  This can be useful in the case of more complex liftovers, like reverse complements, left aligned indels or swapped REF/ALT
      * @return The lifted VariantContext.  This will be null if the input VariantContext could not be lifted.
      */
-    public static VariantContext liftVariant(final VariantContext source, final Interval target, final ReferenceSequence refSeq, final boolean writeOriginalPosition){
+    public static VariantContext liftVariant(final VariantContext source, final Interval target, final ReferenceSequence refSeq, final boolean writeOriginalPosition, final boolean writeOriginalAlleles){
         if (target == null) {
             return null;
         }
@@ -98,6 +99,12 @@ public class LiftoverUtils {
         if (writeOriginalPosition) {
             builder.attribute(LiftoverVcf.ORIGINAL_CONTIG, source.getContig());
             builder.attribute(LiftoverVcf.ORIGINAL_START, source.getStart());
+        }
+
+        if (writeOriginalAlleles && !source.getAlleles().equals(builder.getAlleles())) {
+            List<String> alleles = new ArrayList<>();
+            source.getAlleles().forEach(a -> alleles.add(a.getDisplayString()));
+            builder.attribute(LiftoverVcf.ORIGINAL_ALLELES, alleles);
         }
 
         return builder.make();
--- a/src/main/java/picard/util/LiftoverUtils.java
+++ b/src/main/java/picard/util/LiftoverUtils.java
@@ -102,12 +102,16 @@ public class LiftoverUtils {
         }
 
         if (writeOriginalAlleles && !source.getAlleles().equals(builder.getAlleles())) {
-            final List<String> alleles = new ArrayList<>();
-            source.getAlleles().forEach(a -> alleles.add(a.getDisplayString()));
-            builder.attribute(LiftoverVcf.ORIGINAL_ALLELES, alleles);
+            builder.attribute(LiftoverVcf.ORIGINAL_ALLELES, allelesToStringList(source.getAlleles()));
         }
 
         return builder.make();
+    }
+
+    protected static List<String> allelesToStringList(final List<Allele> alleles) {
+        final List<String> ret = new ArrayList<>();
+        alleles.forEach(a -> ret.add(a.getDisplayString()));
+        return ret;
     }
 
     protected static VariantContextBuilder liftSimpleVariantContext(VariantContext source, Interval target){
--- a/src/main/java/picard/vcf/LiftoverVcf.java
+++ b/src/main/java/picard/vcf/LiftoverVcf.java
@@ -238,14 +238,14 @@ public class LiftoverVcf extends CommandLineProgram {
     private static final List<VCFInfoHeaderLine> ATTRS = CollectionUtil.makeList(
             new VCFInfoHeaderLine(ORIGINAL_CONTIG, 1, VCFHeaderLineType.String, "The name of the source contig/chromosome prior to liftover."),
             new VCFInfoHeaderLine(ORIGINAL_START, 1, VCFHeaderLineType.String, "The position of the variant on the source contig prior to liftover."),
-            new VCFInfoHeaderLine(ORIGINAL_ALLELES, VCFHeaderLineCount.R, VCFHeaderLineType.String, "A list of the original alleles (including REF) of the variant prior to liftover.")
+            new VCFInfoHeaderLine(ORIGINAL_ALLELES, VCFHeaderLineCount.R, VCFHeaderLineType.String, "A list of the original alleles (including REF) of the variant prior to liftover.  If the alleles are identical, this attribute will be omitted.")
             );
 
     private VariantContextWriter rejects;
     private final Log log = Log.getInstance(LiftoverVcf.class);
     private SortingCollection<VariantContext> sorter;
 
-    private long failedLiftover = 0, failedAlleleCheck = 0, totalRecoveredBySwapRefAlt = 0, totalNotRecoveredSwapRefAlt = 0;
+    private long failedLiftover = 0, failedAlleleCheck = 0, totalTrackedAsSwapRefAlt = 0;
     private Map<String, Long> rejectsByContig = new TreeMap<>();
     private Map<String, Long> liftedByDestContig = new TreeMap<>();
     private Map<String, Long> liftedBySourceContig = new TreeMap<>();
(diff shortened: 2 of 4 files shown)


src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -63,9 +63,10 @@
      * @param target The target interval
      * @param refSeq The reference sequence, which should match the target interval
      * @param writeOriginalPosition If true, INFO field annotations will be added to store the original position and contig
+     * @param writeOriginalAlleles If true, INFO an field annotation will be added to store the original alleles in order.  This can be useful in the case of more complex liftovers, like reverse complements, left aligned indels or swapped REF/ALT
```
eitanbanks: fix the English here

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -100,6 +101,12 @@ public static VariantContext liftVariant(final VariantContext source, final Inte
             builder.attribute(LiftoverVcf.ORIGINAL_START, source.getStart());
         }
 
+        if (writeOriginalAlleles && !source.getAlleles().equals(builder.getAlleles())) {
```
eitanbanks: So you're calling equals on 2 ArrayLists, which means that for them to be equal they must be the exact same Object.  I suppose that's what is happening in the code, but you may want to use a more explicit matching routine (or at the very least add a comment to this effect).

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -100,6 +101,12 @@ public static VariantContext liftVariant(final VariantContext source, final Inte
...
+            List<String> alleles = new ArrayList<>();
```
eitanbanks: Should be a 'final' List

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -222,14 +237,15 @@
      */
     private static final List<VCFInfoHeaderLine> ATTRS = CollectionUtil.makeList(
             new VCFInfoHeaderLine(ORIGINAL_CONTIG, 1, VCFHeaderLineType.String, "The name of the source contig/chromosome prior to liftover."),
-            new VCFInfoHeaderLine(ORIGINAL_START, 1, VCFHeaderLineType.String, "The position of the variant on the source contig prior to liftover.")
+            new VCFInfoHeaderLine(ORIGINAL_START, 1, VCFHeaderLineType.String, "The position of the variant on the source contig prior to liftover."),
+            new VCFInfoHeaderLine(ORIGINAL_ALLELES, VCFHeaderLineCount.R, VCFHeaderLineType.String, "The position of the variant on the source contig prior to liftover.")
```
eitanbanks: The description is wrong here

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -440,6 +463,12 @@ private void trackLiftedVariantContig(Map<String, Long> map, String contig) {
         map.put(contig, ++val);
     }
 
+    private void addAndTrack(VariantContext toAdd, VariantContext source){
```
eitanbanks: 'final' x 2

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -155,12 +155,22 @@
...
     public double LIFTOVER_MIN_MATCH = 1.0;
 
     @Argument(doc = "Allow INFO and FORMAT in the records that are not found in the header", optional = true)
     public boolean ALLOW_MISSING_FIELDS_IN_HEADER = false;
 
+
+    @Argument(doc = "If the REF allele of the lifted site does not match the target genome, that variant is normally rejected. " +
+            "For bi-allelic SNPs, if this is set to true and the ALT allele equals the new REF allele, the REF and ALT alleles will be swapped.  This can rescue " +
+            "some variants; however, do this carefully as some annotations may become invalid, such as any that are alelle-specifc.  See also TAGS_TO_REVERSE and TAGS_TO_DROP.", optional = true)
+    public boolean RECOVER_SWAPPED_REF_ALT = false;
```
eitanbanks: Out of curiosity, why are we changing the default behavior of this tool?  Why not keep it the way it currently is and add a DO_NOT_RECOVER_SWAPPED_REF_ALT option instead?

src/main/java/picard/util/LiftoverUtils.java
bbimber: ok

src/main/java/picard/util/LiftoverUtils.java
bbimber: ok

src/main/java/picard/vcf/LiftoverVcf.java
bbimber: ok

src/main/java/picard/vcf/LiftoverVcf.java
bbimber: ok

src/main/java/picard/vcf/LiftoverVcf.java
bbimber: Like I noted in the description of the PR, I figured this might need some discussion.  My reasoning is in #1149.  The main reason I believe the current default is a bad idea is b/c of the high likelihood of leaving the VCF in a bad/inaccurate state with respect to annotations.  As a general rule, I think tools like picard should take the more conservative default position.  This seems like a reasonable feature a user might want in some cases, hence the additional logging I added in the PR around features that could be rescued if this was turned on.  However, that user needs to think about the annotations specific to their VCF and the consequences of this swap before blindly doing it.

src/main/java/picard/util/LiftoverUtils.java
bbimber: I dont think this is correct.  ArrayList (from AbstractList) will return true for equals() if the elements are identical, not only if they are the exact same Object:

https://docs.oracle.com/javase/7/docs/api/java/util/List.html#equals(java.lang.Object)

src/main/java/picard/vcf/LiftoverVcf.java
eitanbanks: Okay, fair enough.  I would just want to hear from @yfarjoun to make sure he does not object (since he made the original changes).

src/main/java/picard/util/LiftoverUtils.java
eitanbanks: Okay, thanks for the correction.

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -404,6 +420,13 @@ protected int doWork() {
             log.info("no successfully lifted variants");
         }
 
+        if (RECOVER_SWAPPED_REF_ALT){
```
yfarjoun: space between `)` and `{`

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -404,6 +420,13 @@ protected int doWork() {
...
+            log.info(totalRecoveredBySwapRefAlt, " variants were lifted by swapping REF/ALT alleles.");
+        }
+        else {
```
yfarjoun: `} else {`

bbimber: done

src/main/java/picard/vcf/LiftoverVcf.java
bbimber: done

bbimber:
the travis build seems to have timed out - i've never seen that happen before.  do you have the ability to re-trigger that?

bbimber:
I believe that addresses all the requested changes and travis has now passed - are there any other changes or questions on this?  Thanks.

eitanbanks:
@yfarjoun needs to approve.

bbimber:
Hello - just wanted to check back on this.  It is just pending a final approval.  If there are additional changes you want please let me know.

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -404,6 +420,12 @@ protected int doWork() {
             log.info("no successfully lifted variants");
         }
 
+        if (RECOVER_SWAPPED_REF_ALT) {
+            log.info(totalRecoveredBySwapRefAlt, " variants were lifted by swapping REF/ALT alleles.");
+        } else {
+            log.info(totalNotRecoveredSwapRefAlt, " variants with a swapped REF/ALT were identified, but were not recovered.  See RECOVER_SWAPPED_REF_ALT.");
```
yfarjoun: This should be a warning, I think.

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -222,14 +237,15 @@
...
+            new VCFInfoHeaderLine(ORIGINAL_START, 1, VCFHeaderLineType.String, "The position of the variant on the source contig prior to liftover."),
+            new VCFInfoHeaderLine(ORIGINAL_ALLELES, VCFHeaderLineCount.R, VCFHeaderLineType.String, "A list of the original alleles (including REF) of the variant prior to liftover.")
             );
 
     private VariantContextWriter rejects;
     private final Log log = Log.getInstance(LiftoverVcf.class);
     private SortingCollection<VariantContext> sorter;
 
-    private long failedLiftover = 0, failedAlleleCheck = 0;
+    private long failedLiftover = 0, failedAlleleCheck = 0, totalRecoveredBySwapRefAlt = 0, totalNotRecoveredSwapRefAlt = 0;
```
yfarjoun: seems like these variables can be refactored into one variable, `nRefAltSwapVariants`, which can then be used in two different ways in the code depending on `RECOVER_SWAPPED_REF_ALT`.

src/test/java/picard/util/LiftoverVcfTest.java
```diff
@@ -545,7 +549,40 @@ public void testFlipIndel(final VariantContext source, final ReferenceSequence r
...
+        final VariantContext flipped = LiftoverUtils.liftVariant(source, target, reference, false, false);
+
+        VcfTestUtils.assertEquals(flipped, result);
+    }
+
+    @DataProvider(name = "indelFlipDataWithOriginalAllele")
+    public Iterator<Object[]> indelFlipDataWithOriginalAllele() {
+        final List<Object[]> tests = new ArrayList<>();
+        indelFlipData().forEachRemaining(x -> {
+            if (x[2] == null){
```
yfarjoun: `) {`

bbimber: ok

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -222,14 +237,15 @@
      */
     private static final List<VCFInfoHeaderLine> ATTRS = CollectionUtil.makeList(
             new VCFInfoHeaderLine(ORIGINAL_CONTIG, 1, VCFHeaderLineType.String, "The name of the source contig/chromosome prior to liftover."),
-            new VCFInfoHeaderLine(ORIGINAL_START, 1, VCFHeaderLineType.String, "The position of the variant on the source contig prior to liftover.")
+            new VCFInfoHeaderLine(ORIGINAL_START, 1, VCFHeaderLineType.String, "The position of the variant on the source contig prior to liftover."),
+            new VCFInfoHeaderLine(ORIGINAL_ALLELES, VCFHeaderLineCount.R, VCFHeaderLineType.String, "A list of the original alleles (including REF) of the variant prior to liftover.")
```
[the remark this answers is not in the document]
bbimber: That is the behavior (is was documented better for the argument WRITE_ORIGINAL_ALLELES).  I updated that description

src/test/java/picard/util/LiftoverVcfTest.java
```diff
@@ -545,7 +549,40 @@ public void testFlipIndel(final VariantContext source, final ReferenceSequence r
...
+                tests.add(x);
+            }
+            else {
```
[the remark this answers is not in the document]
bbimber: ok

src/main/java/picard/vcf/LiftoverVcf.java
bbimber: I dont feel all that strongly about this, but I am not sure I agree.  I assume you only really think the warning matters for line 426?  I flipped that to warn() and slightly updated the verbiage.

src/main/java/picard/vcf/LiftoverVcf.java
bbimber: True, though it seemed clearer in the code to separate them as two variables, so the intent was clearer.  I renamed it to totalTrackedAsSwapRefAlt and merged them into one variable.

src/main/java/picard/vcf/LiftoverVcf.java
```diff
@@ -238,14 +238,14 @@
     private static final List<VCFInfoHeaderLine> ATTRS = CollectionUtil.makeList(
             new VCFInfoHeaderLine(ORIGINAL_CONTIG, 1, VCFHeaderLineType.String, "The name of the source contig/chromosome prior to liftover."),
             new VCFInfoHeaderLine(ORIGINAL_START, 1, VCFHeaderLineType.String, "The position of the variant on the source contig prior to liftover."),
-            new VCFInfoHeaderLine(ORIGINAL_ALLELES, VCFHeaderLineCount.R, VCFHeaderLineType.String, "A list of the original alleles (including REF) of the variant prior to liftover.")
+            new VCFInfoHeaderLine(ORIGINAL_ALLELES, VCFHeaderLineCount.R, VCFHeaderLineType.String, "A list of the original alleles (including REF) of the variant prior to liftover.  If the alleles are identical, this attribute will be omitted.")
```
yfarjoun: I think that `If the alleles haven't changed...` would be more clear.

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -102,14 +102,18 @@ public static VariantContext liftVariant(final VariantContext source, final Inte
...
-            final List<String> alleles = new ArrayList<>();
-            source.getAlleles().forEach(a -> alleles.add(a.getDisplayString()));
-            builder.attribute(LiftoverVcf.ORIGINAL_ALLELES, alleles);
+            builder.attribute(LiftoverVcf.ORIGINAL_ALLELES, allelesToStringList(source.getAlleles()));
         }
 
         return builder.make();
     }
 
+    protected static List<String> allelesToStringList(final List<Allele> alleles) {
```
yfarjoun: could you add a few small independent tests to this new static function? Also, despite it being non-public, could you add a javadoc so it's clear what this function intends to do?

src/main/java/picard/vcf/LiftoverVcf.java
bbimber: added "If the alleles were not changed during liftover"

src/main/java/picard/util/LiftoverUtils.java
bbimber: done

bbimber:
@yfarjoun does that cover everything?

PR #1153, opened 2018-04-09
Make picardcloud accessible in docker image

tlangs:
### Description

The cloud version of Picard should be accessible from the docker image. By using the `-c` option when invoking the docker image, the picardcloud jar is now run instead of the normal picard jar

----

#### Content

#### Review

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,7 @@ FROM broadinstitute/java-baseimage
 MAINTAINER Broad Institute DSDE <dsde-engineering@broadinstitute.org>
 
 ARG build_command=shadowJar
+ARG jar_name=picard.jar
 
 # Install ant, git for building
 RUN apt-get update && \
@@ -17,7 +18,7 @@ WORKDIR /usr/picard
 
 # Build the distribution jar, clean up everything else
 RUN ./gradlew ${build_command} && \
-    mv build/libs/picard*.jar picard.jar && \
+    mv build/libs/${jar_name} ${jar_name} && \
     mv src/main/resources/picard/docker_helper.sh docker_helper.sh && \
     ./gradlew clean && \
     rm -rf src && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,9 @@ WORKDIR /usr/picard
 
 # Build the distribution jar, clean up everything else
 RUN ./gradlew shadowJar && \
+    ./gradlew cloudJar && \
     mv build/libs/picard.jar picard.jar && \
+    mv build/libs/picardcloud.jar picardcloud.jar && \
     mv src/main/resources/picard/docker_helper.sh docker_helper.sh && \
     ./gradlew clean && \
     rm -rf src && \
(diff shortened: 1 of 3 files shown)


lbergelson:
I'm curious, why are there two picard jars in the docker?  One packages the google cloud library and the other doesn't?  If so, why not just use the cloud one only?

danxmoran:
> If so, why not just use the cloud one only?

We've seen the cloud one log huge stack traces at startup if it's not configured correctly, and it's nice to be able to avoid the noise if you know you're not going to be using NIO.

Dockerfile
```diff
@@ -15,7 +15,9 @@ WORKDIR /usr/picard
 
 # Build the distribution jar, clean up everything else
 RUN ./gradlew shadowJar && \
+    ./gradlew cloudJar && \
     mv build/libs/picard.jar picard.jar && \
+    mv build/libs/picardcloud.jar picardcloud.jar && \
```
danxmoran: I do agree though that it's weird to package both jars into one Docker, when you'd only ever want one or the other. I'd think it'd be better to have two images: one for the shadowJar, and one for the cloudJar.

tlangs:
I've made it so that the Dockerfile can build both `picard` and `picardcloud` docker images. The Dockerfile is paramaterized, but it defaults to building the non-cloud JAR. There's also a new script which builds both docker images and pushes them to DockerHub. This will be run when we release new versions of picard so that both DockerHub repositories are updated.

build_push_docker.sh
```diff
@@ -0,0 +1,27 @@
...
+echo "Will build and push the following docker images:"
+echo "broadinstitute/picard:${TAG}"
+echo "broadinstitute/picard-cloud:${TAG}"
+
+read -p "Is this really what you want to do? " -n 1 -r
+echo    # (optional) move to a new line
+if [[ $REPLY =~ ^[Yy]$ ]]
+then
+    docker build -t broadinstitute/picard:${TAG} --build-arg build_command=shadowJar --build-arg jar_name=picard.jar .
+    docker build -t broadinstitute/picard-cloud:${TAG} --build-arg build_command=cloudJar --build-arg jar_name=picardcloud.jar .
```
danxmoran: Bike-shedding: I wonder if it should be `/picard:${TAG}-cloud` instead of `/picard-cloud:${TAG}`

danxmoran:
Ah, can you also be sure to add the step to run the script to the release instructions?

build_push_docker.sh
tlangs: I think we should be keeping the two versions in different repos since they do different things. The automated `latest` build would be ambiguous as to which tool it refers to.

tlangs:
The release information has been updated.

Dockerfile
```diff
@@ -14,8 +17,8 @@ COPY / /usr/picard/
 WORKDIR /usr/picard
 
 # Build the distribution jar, clean up everything else
-RUN ./gradlew shadowJar && \
-    mv build/libs/picard.jar picard.jar && \
+RUN ./gradlew ${build_command} && \
+    mv build/libs/${jar_name} ${jar_name} && \
```
danxmoran: Oh! Last thing, could we always name the target `picard.jar`? That way you can switch the docker in and out without needing to edit your commands.

kbergin:

Issue #605, opened 2016-07-20
Make picard build properly even no .git directory is available.


lbergelson:
Currently the build.gradle uses some plugins which both require a .git directory.  

The first case is gradleGit, I've opened a ticket ( https://github.com/ajoberstar/gradle-git/issues/205 ) to track that and the developer responded positively.  

The second is with the version plugin we're using.  We can either request a change there or simply catch the exception it's throwing when we call getVersion.

droazen:
Why do we need to be able to build without a git checkout?

lbergelson:
We've had 3 public tickets about it already across picard and htsjdk.  It's not a huge deal but if we can make it work I think we should.

droazen:
I don't really see the value in it -- it's simpler if we just tell people to clone and build within their clone.

lbergelson:
We do tell people that, but people are still confused.  Since it didn't used to be a requirement and it isn't really necessarily one I think it's worth just fixing it if/when gradleGit gets patched to do its initialization lazily.

vdauwera:
Hey @lbergelson, is this still on the radar/table at all? I agree it would be better to enable git-less builds, if possible and not developmentally onerous.

e4p:
Adding a +1 to this issue since I encountered it today;
A common use case for building without git is building the released source tarballs packaged by github. This specific use is especially important for users trying to include picard in a dockerfile since requiring the git repo increases download time, storage, and adds git as a dependency for your image.

Issue #604, opened 2016-07-19
Gradle build fails


ycl6:
Build failed using the source from the 2.5.0-gradle tar.gz file. I googled for similar issue, it seems I have to git clone the source from github to avoid this error. 

```
$ gradle build --stacktrace

FAILURE: Build failed with an exception.

* Where:
Build file 'picard-2.5.0-gradle/build.gradle' line: 19

* What went wrong:
An exception occurred applying plugin request [id: 'org.ajoberstar.grgit', version: '1.4.2']
> Failed to apply plugin [id 'org.ajoberstar.grgit']
   > One of setGitDir or setWorkTree must be called.

* Try:
Run with --info or --debug option to get more log output.

* Exception is:
org.gradle.api.plugins.InvalidPluginException: An exception occurred applying plugin request [id: 'org.ajoberstar.grgit', version: '1.4.2']
        at org.gradle.plugin.use.internal.DefaultPluginRequestApplicator.applyPlugin(DefaultPluginRequestApplicator.java:216)
        at org.gradle.plugin.use.internal.DefaultPluginRequestApplicator.applyPlugins(DefaultPluginRequestApplicator.java:168)
        at org.gradle.configuration.DefaultScriptPluginFactory$ScriptPluginImpl.apply(DefaultScriptPluginFactory.java:157)
        at org.gradle.configuration.project.BuildScriptProcessor.execute(BuildScriptProcessor.java:38)
        at org.gradle.configuration.project.BuildScriptProcessor.execute(BuildScriptProcessor.java:25)
        at org.gradle.configuration.project.ConfigureActionsProjectEvaluator.evaluate(ConfigureActionsProjectEvaluator.java:34)
        at org.gradle.configuration.project.LifecycleProjectEvaluator.evaluate(LifecycleProjectEvaluator.java:55)
        at org.gradle.api.internal.project.AbstractProject.evaluate(AbstractProject.java:540)
        at org.gradle.api.internal.project.AbstractProject.evaluate(AbstractProject.java:93)
        at org.gradle.execution.TaskPathProjectEvaluator.configureHierarchy(TaskPathProjectEvaluator.java:42)
        at org.gradle.configuration.DefaultBuildConfigurer.configure(DefaultBuildConfigurer.java:35)
        at org.gradle.initialization.DefaultGradleLauncher$2.run(DefaultGradleLauncher.java:124)
        at org.gradle.internal.Factories$1.create(Factories.java:22)
[...]

PR #1156, opened 2018-04-12
Add to gradle build pre-requisites check

magicDGS:
### Description

Due to the usage of the git repository to get the version, Picard fails to build if the zipped source is downloaded. Based on the same strategy as GATK, this commit adds a check for the build pre-requisites.

This provides a better error message for cases like https://github.com/broadinstitute/picard/issues/604#issuecomment-380765514, being a temporary solution until https://github.com/broadinstitute/picard/issues/605 is implemented (adding better logging for the user).

----

#### Content

#### Review

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,7 @@ jacoco {
 }
 
 final requiredJavaVersion = "8"
-final buildPrerequisitesMessage = "See https://github.com/broadinstitute/picard/blob/master/README.md for information on how to build picard"
+final buildPrerequisitesMessage = "See https://github.com/broadinstitute/picard/blob/master/README.md#building-picard for information on how to build picard"
 // Ensure that we have the right JDK version, a clone of the git repository
 def ensureBuildPrerequisites(requiredJavaVersion, largeResourcesFolder, buildPrerequisitesMessage) {
     // Make sure we can get a ToolProvider class loader. If not we may have just a JRE, or a JDK from the future.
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -47,6 +47,23 @@ jacoco {
     toolVersion = "0.7.5.201505241946"
 }
 
+final requiredJavaVersion = "8"
+final buildPrerequisitesMessage = "See https://github.com/broadinstitute/picard/blob/master/README.md#building-picard for information on how to build picard"
+// Ensure that we have the right JDK version, a clone of the git repository
+def ensureBuildPrerequisites(requiredJavaVersion, buildPrerequisitesMessage) {
+    // Make sure we can get a ToolProvider class loader. If not we may have just a JRE, or a JDK from the future.
+    if (ToolProvider.getSystemToolClassLoader() == null) {
```
danxmoran: I'm not super clear on the differences between JDK versions... Is `ToolProvider` something new in 8, or could this check pass for 7? If it'll pass, is there a way we can detect 8?

magicDGS: @cmnbroad - you did this change in GATK. Do you know something about it?

cmnbroad: This was put in to protect against NPEs in the case where someone tries to build using a JRE (it has happened) or tries to use a JDK from the future (I haven't tried it but I think it will fail on JDK9). The build file has a `targetCompatibility` declaration that should prevent building with older JDKs.

PR #1167, opened 2018-05-02
Avoid overflow when computing barcode metrics.

alecw:
### Description

For instruments like NovaSeq, total number of reads in a lane can be greater than will fit into a 32-bit signed int, causing count of reads to be negative.  This causes various PCT metrics in BarcodeMetric to be 0.  Changing these counters from int to long solves the problem.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/illumina/ExtractIlluminaBarcodes.java
+++ b/src/main/java/picard/illumina/ExtractIlluminaBarcodes.java
@@ -371,9 +371,9 @@ public class ExtractIlluminaBarcodes extends CommandLineProgram {
     public static void finalizeMetrics(final Map<String, BarcodeMetric> barcodeToMetrics,
                                        final BarcodeMetric noMatchMetric) {
         // Finish metrics tallying.
-        int totalReads = noMatchMetric.READS;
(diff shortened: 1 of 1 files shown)


src/main/java/picard/illumina/ExtractIlluminaBarcodes.java
```diff
@@ -371,9 +371,9 @@ protected int doWork() {
     public static void finalizeMetrics(final Map<String, BarcodeMetric> barcodeToMetrics,
                                        final BarcodeMetric noMatchMetric) {
         // Finish metrics tallying.
-        int totalReads = noMatchMetric.READS;
-        int totalPfReads = noMatchMetric.PF_READS;
-        int totalPfReadsAssigned = 0;
```
nh13: I assume you'll want to change `BarcodeMetric` to store `long` instead of `int` as well?  One or two samples on a NovaSeq high-throughput flowcell would causethis.

alecw: Hi @nh13 , 
Good point.  Will do.

-Alec

PR #1173, opened 2018-05-20
Allowing static fields in MergableMetrics to be not annotated 

### LINKED ISSUE #1172: MergeableMetricsBase shouldn't require or allow static fields to be annotated
### Affected tool(s)
MergeableMetricsBase

### Affected version(s)
- [x] Latest public release version 2.18.4
- [x] Latest development/master branch as of 5/18/18

### Description 
`MetricsFiles` that implement `MergeableMetricsBase` require all fields to be annotated with a merge strategy.  This should not include `static` fields.   It's easy to work around by specifying `@NoMergingKeepsValue` but it would be better to not have to.  It seems like it should instead require that they NOT be annotated, because combining static fields is nonsensical.

#### Steps to reproduce
1. add a static final constant to a subclass of MergeableMetrics
2. run a tool that tries to merge those metrics

#### Expected behavior
static fields are ignored during merging

#### Actual behavior
crash with `IllegalStateException` on line 151 of (MergeableMetricsBase)[https://github.com/broadinstitute/picard/blob/master/src/main/java/picard/analysis/MergeableMetricBase.java]

### Other even more minor annoyance:
  The error message that's reported lists several options for annotating but not the one you actually want.  "All fields of this class must be annotated with @MergeByAdding, @NoMergingIsDerived, or @MergeByAssertEquals"

yfarjoun:
or they can be annotated but not with @MergeByAdding...

fixes #1172

### DIFF
--- a/src/main/java/picard/analysis/MergeableMetricBase.java
+++ b/src/main/java/picard/analysis/MergeableMetricBase.java
@@ -39,30 +39,31 @@ import java.util.Collection;
 import java.util.List;
 
 /**
- * An extension of MetricBase that knows how to merge-by-adding fields that are appropriately annotated ({@link MergeByAdding}). It also provides an interface
- * for calculating derived fields {@link #calculateDerivedFields()}  (and an annotation that informs that said fields are derived {@link NoMergingIsDerived}). Finally, it also allows for an annotation
- * that suggests that a field will be used as an ID and thus merging will simply ensure that these fields are equal {@link MergeByAssertEquals}. Other annotations are available,
- * though they all currently imply that no <it>implicit</it> action will be taken {@link MergingIsManual} and {@link NoMergingKeepsValue}.
+ * An extension of MetricBase that knows how to merge-by-adding fields that are appropriately annotated ({@link MergeByAdding MergeByAdding}). It also provides an interface
+ * for calculating derived fields {@link #calculateDerivedFields}  (and an annotation that informs that said fields are derived {@link NoMergingIsDerived NoMergingIsDerived}). Finally, it also allows for an annotation
+ * that suggests that a field will be used as an ID and thus merging will simply ensure that these fields are equal {@link MergeByAssertEquals MergeByAssertEquals}. Other annotations are available,
+ * though they all currently imply that no <it>implicit</it> action will be taken {@link MergingIsManual MergingIsManual} and {@link NoMergingKeepsValue NoMergingKeepsValue}.
  *<p>
- * {@link MergeByAdding} is only enabled for the following types: int, Integer, float, Float, double, Double, short, Short, long, Long, byte, Byte.
+ * {@link MergeByAdding MergeByAdding} is only enabled for the following types: int, Integer, float, Float, double, Double, short, Short, long, Long, byte, Byte.
  * Overflow will be detected (for the short, and byte types) and an exception thrown.
- *<p><
- * Every (non-static) field in this class <it>must</it> be @Annotated by one of: {@link MergeByAdding}, {@link MergeByAssertEquals}, {@link NoMergingIsDerived}, {@link MergingIsManual}, {@link NoMergingKeepsValue}.
- * Static fields may be annotated, but not with {@link MergeByAdding}, there will be no automatic modification of static fields
+ *<p>
+ * Every (non-static) field in this class <it>must</it> be @Annotated by one of: {@link MergeByAdding MergeByAdding},
+ * {@link MergeByAssertEquals MergeByAssertEquals}, {@link NoMergingIsDerived NoMergingIsDerived}, {@link MergingIsManual MergingIsManual}, {@link NoMergingKeepsValue NoMergingKeepsValue}.
+ * Static fields may be annotated, but not with {@link MergeByAdding MergeByAdding}, there will be no automatic modification of static fields
  *
  * <dl>
- * <dt>{@link MergeByAdding}</dt>
+ * <dt>{@link MergeByAdding MergeByAdding}</dt>
  * <dd>When merging another metric into this one, the value of the field in the other class will be added to the value in this.
  * This will happen automatically!</dd>
- * <dt>{@link MergeByAssertEquals}</dt>
+ * <dt>{@link MergeByAssertEquals MergeByAssertEquals}</dt>
  * <dd>When merging another metric into this one, the code will assert that value of the field in the other class is equal to the value in this.</dd>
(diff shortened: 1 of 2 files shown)


src/main/java/picard/analysis/MergeableMetricBase.java
```diff
@@ -143,15 +145,26 @@ public MergeableMetricBase merge(final MergeableMetricBase other) {
...
+                    field.getAnnotationsByType(MergeByAdding.class).length +
                     field.getAnnotationsByType(MergeByAssertEquals.class).length +
                     field.getAnnotationsByType(NoMergingIsDerived.class).length +
                     field.getAnnotationsByType(MergingIsManual.class).length +
-                    field.getAnnotationsByType(NoMergingKeepsValue.class).length == 0) {
-                throw new IllegalStateException("All fields of this class must be annotated with @MergeByAdding, @NoMergingIsDerived, or @MergeByAssertEquals. " +
+                    field.getAnnotationsByType(NoMergingKeepsValue.class).length != 0;
+
+            final boolean isStatic = Modifier.isStatic(field.getModifiers());
+            final boolean isAnnotatedActive = field.getAnnotationsByType(MergeByAdding.class).length != 0;
```
pshapiro4broad: This code computes both `isAnnotated` and `isAnnotatedActive` but these are only used on a single branch of the `if`. How about:
```
if (Modifier.isStatic(field.getModifiers()) {
    final boolean isAnnotatedMergeByAdding = field.getAnnotationsByType(MergeByAdding.class).length != 0;
    if (isAnnotatedMergeByAdding) {
        throw new IllegalStateException(...);
    }
} else {
    final boolean isAnnotated = ...;
    if (!isAnnotated) {
        throw new IllegalStateException(...);
    }
}
```
This also flips the order of the ifs and renames `isAnnotatedActive` since that name doesn't seem to reflect its value.

yfarjoun:
by "document" what do you mean? in the commit message?

pshapiro4broad:
> by "document" what do you mean? in the commit message?

I was thinking javadoc or wherever we document how stuff works, so someone using this API knows what to expect without having to read the code. At least that's the goal, am not sure how well things are documented in the first place in this code.

src/main/java/picard/analysis/MergeableMetricBase.java
```diff
@@ -45,6 +46,18 @@
  * merge-by-adding is only enabled for the following types: int, Integer, float, Float, double, Double, short, Short, long, Long, byte, Byte.
  * Overflow will be detected (for the short, and byte types) and an exception thrown.
  *
+ * Every (public, non-static) field in this class _must_ be @Annotated by one of: MergeByAdding, MergeByAssertEquals, NoMergingIsDerived, MergingIsManual, NoMergingKeepsValue.
+ * Static fields may be annotated, but not with @MergeByAdding.
+ *
+ * MergeByAdding: When merging another metric into this one, the value of the field in the other class will be added to the value in this.
```
pshapiro4broad: This is better although javadoc will ignore all your formatting here and present all the text as a single paragraph. If you want to fix this, I would use a description list. It uses three tags, `<dl>`, `<dt>` and `<dd>`. E.g.
```
<dl>
<dt>MergeByAdding
<dd>When merging another metric into this one, the value of the field in the other class will be added to the value in this. This will happen automatically!
<dt>MergeByAssertEquals
<dd> When merging another metric into this one, the code will assert that value of the field in the other class is equal to the value in this.
...
</dl>
```

pshapiro4broad: If you want to get even fancier you can use `@link` here to link from your dt's to the object, e.g.
```
<dt>{@link MergeByAdding MergeByAdding}
```

src/main/java/picard/analysis/MergeableMetricBase.java
```diff
@@ -39,24 +39,32 @@
...
- *
- * merge-by-adding is only enabled for the following types: int, Integer, float, Float, double, Double, short, Short, long, Long, byte, Byte.
+ * An extension of MetricBase that knows how to merge-by-adding fields that are appropriately annotated ({@link MergeByAdding}). It also provides an interface
+ * for calculating derived fields {@link #calculateDerivedFields()}  (and an annotation that informs that said fields are derived {@link NoMergingIsDerived}). Finally, it also allows for an annotation
+ * that suggests that a field will be used as an ID and thus merging will simply ensure that these fields are equal {@link MergeByAssertEquals}. Other annotations are available,
+ * though they all currently imply that no <it>implicit</it> action will be taken {@link MergingIsManual} and {@link NoMergingKeepsValue}.
+ *<p>
+ * {@link MergeByAdding} is only enabled for the following types: int, Integer, float, Float, double, Double, short, Short, long, Long, byte, Byte.
  * Overflow will be detected (for the short, and byte types) and an exception thrown.
+ *<p><
```
pshapiro4broad: html typo

src/main/java/picard/analysis/MergeableMetricBase.java
```diff
@@ -39,24 +39,32 @@
...
- * that suggests that a field will be used as an ID and thus merging will simply require that these fields are equal.
- *
- * merge-by-adding is only enabled for the following types: int, Integer, float, Float, double, Double, short, Short, long, Long, byte, Byte.
+ * An extension of MetricBase that knows how to merge-by-adding fields that are appropriately annotated ({@link MergeByAdding}). It also provides an interface
+ * for calculating derived fields {@link #calculateDerivedFields()}  (and an annotation that informs that said fields are derived {@link NoMergingIsDerived}). Finally, it also allows for an annotation
+ * that suggests that a field will be used as an ID and thus merging will simply ensure that these fields are equal {@link MergeByAssertEquals}. Other annotations are available,
+ * though they all currently imply that no <it>implicit</it> action will be taken {@link MergingIsManual} and {@link NoMergingKeepsValue}.
+ *<p>
+ * {@link MergeByAdding} is only enabled for the following types: int, Integer, float, Float, double, Double, short, Short, long, Long, byte, Byte.
```
pshapiro4broad: If you use this form instead of `{@link MergeByAdding MergeByAdding}` then it will show up as `MergableMetricBase.MergeByAdding` in the javadoc. I think the shorter form is easier to read.

To see what your javadoc will look like, put the cursor on the class name and hit control-J (or use View/Quick Documentation).

yfarjoun:
Thanks @pshapiro4broad !! TILx2 (ctrl-J and the extra text after {@link class})

PR #1176, opened 2018-05-23
DSDEGP-2354 id correct tile metrics file in parser for novaseq

ApChagi:
### Description

TileMetricsUtil was relying on a hierarchy in the Illumina directory structure to determine which TileMetricsOut.bin to use when collecting lane metrics. This hierarchy has turned out to be unreliable. The correct TileMetricsOut.bin to use should be the one with cluster records. This change has the collector parse all of the TileMetricsOut.bin files in hierarchical order until it finds one with cluster records.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
+++ b/src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
@@ -159,8 +159,8 @@ public class CollectIlluminaLaneMetrics extends CommandLineProgram {
                     tiles = TileMetricsUtil.parseClusterRecordsFromTileMetricsV3(
                             tileMetricsOutFiles,
                             TileMetricsUtil.renderPhasingMetricsFilesFromBasecallingDirectory(illuminaRunDirectory),
-                            readStructure,
-                            validationStringency);
+                            readStructure
+                    );
                 } else {
                     tiles = TileMetricsUtil.parseTileMetrics(tileMetricsOutFiles.get(0),
                             readStructure,
--- a/src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
+++ b/src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
@@ -154,15 +154,15 @@ public class CollectIlluminaLaneMetrics extends CommandLineProgram {
                                                                              final boolean isNovaSeq) {
             final Collection<Tile> tiles;
             try {
-                File tileMetricsOutFile = TileMetricsUtil.renderTileMetricsFileFromBasecallingDirectory(illuminaRunDirectory, readStructure.totalCycles, isNovaSeq);
+                List<File> tileMetricsOutFiles = TileMetricsUtil.renderTileMetricsFileFromBasecallingDirectory(illuminaRunDirectory, readStructure.totalCycles, isNovaSeq);
                 if (isNovaSeq) {
-                    tiles = TileMetricsUtil.parseTileMetrics(
-                            tileMetricsOutFile,
+                    tiles = TileMetricsUtil.parseClusterRecordsFromTileMetricsV3(
+                            tileMetricsOutFiles,
                             TileMetricsUtil.renderPhasingMetricsFilesFromBasecallingDirectory(illuminaRunDirectory),
                             readStructure,
                             validationStringency);
                 } else {
-                    tiles = TileMetricsUtil.parseTileMetrics(tileMetricsOutFile,
+                    tiles = TileMetricsUtil.parseTileMetrics(tileMetricsOutFiles.get(0),
                             readStructure,
                             validationStringency
                     );
--- a/src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
+++ b/src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
@@ -154,7 +154,7 @@ public class CollectIlluminaLaneMetrics extends CommandLineProgram {
                                                                              final boolean isNovaSeq) {
             final Collection<Tile> tiles;
             try {
-                List<File> tileMetricsOutFiles = TileMetricsUtil.renderTileMetricsFileFromBasecallingDirectory(illuminaRunDirectory, readStructure.totalCycles, isNovaSeq);
+                List<File> tileMetricsOutFiles = TileMetricsUtil.renderTileMetricsFilesFromBasecallingDirectory(illuminaRunDirectory, readStructure.totalCycles, isNovaSeq);
                 if (isNovaSeq) {
                     tiles = TileMetricsUtil.parseClusterRecordsFromTileMetricsV3(
                             tileMetricsOutFiles,
--- a/src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
+++ b/src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
@@ -154,7 +154,7 @@ public class CollectIlluminaLaneMetrics extends CommandLineProgram {
                                                                              final boolean isNovaSeq) {
             final Collection<Tile> tiles;
             try {
-                List<File> tileMetricsOutFiles = TileMetricsUtil.findTileMetricsFiles(illuminaRunDirectory, readStructure.totalCycles, isNovaSeq);
+                final List<File> tileMetricsOutFiles = TileMetricsUtil.findTileMetricsFiles(illuminaRunDirectory, readStructure.totalCycles, isNovaSeq);
                 if (isNovaSeq) {
                     tiles = TileMetricsUtil.parseClusterRecordsFromTileMetricsV3(
                             tileMetricsOutFiles,
@@ -162,7 +162,8 @@ public class CollectIlluminaLaneMetrics extends CommandLineProgram {
                             readStructure
                     );
                 } else {
-                    tiles = TileMetricsUtil.parseTileMetrics(tileMetricsOutFiles.get(0),
+                    tiles = TileMetricsUtil.parseTileMetrics(
+                            tileMetricsOutFiles.get(0),
                             readStructure,
                             validationStringency
                     );
--- a/src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
+++ b/src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
@@ -154,7 +154,7 @@ public class CollectIlluminaLaneMetrics extends CommandLineProgram {
                                                                              final boolean isNovaSeq) {
             final Collection<Tile> tiles;
             try {
-                List<File> tileMetricsOutFiles = TileMetricsUtil.renderTileMetricsFilesFromBasecallingDirectory(illuminaRunDirectory, readStructure.totalCycles, isNovaSeq);
+                List<File> tileMetricsOutFiles = TileMetricsUtil.findTileMetricsFiles(illuminaRunDirectory, readStructure.totalCycles, isNovaSeq);
                 if (isNovaSeq) {
                     tiles = TileMetricsUtil.parseClusterRecordsFromTileMetricsV3(
                             tileMetricsOutFiles,
--- a/src/main/java/picard/illumina/parser/TileMetricsUtil.java
+++ b/src/main/java/picard/illumina/parser/TileMetricsUtil.java
@@ -111,7 +111,7 @@ public class TileMetricsUtil {
 
     private static Collection<Tile> getTileClusterRecordsV3(
             Map<String, ? extends Collection<IlluminaTileMetrics>> locationToMetricsMap,
-            final Map<Integer, Collection<TilePhasingValue>> phasingValues,
+            final Map<Integer, Map<Integer, Collection<TilePhasingValue>>> phasingValues,
             final float density) {
 
         final Collection<Tile> tiles = new LinkedList<>();
@@ -122,7 +122,7 @@ public class TileMetricsUtil {
 
             //only create for cluster records
             if (record.isClusterRecord()) {
-                final Collection<TilePhasingValue> tilePhasingValues = phasingValues.get(record.getTileNumber());
+                final Collection<TilePhasingValue> tilePhasingValues = phasingValues.get(record.getLaneNumber()).get(record.getTileNumber());
                 tiles.add(new Tile(record.getLaneNumber(), record.getTileNumber(), density, record.getMetricValue(),
                         tilePhasingValues.toArray(new TilePhasingValue[tilePhasingValues.size()])));
             }
@@ -143,7 +143,7 @@ public class TileMetricsUtil {
     ) throws FileNotFoundException {
         // Snag and cache the phasing data for each read in the read structure for each tile.
         // For both types of phasing values, this is the median of all of the individual values seen
-        Map<Integer, Collection<TilePhasingValue>> phasingValues = getTilePhasingValues(phasingMetricsFiles, readStructure);
+        Map<Integer, Map<Integer, Collection<TilePhasingValue>>> phasingValues = getTilePhasingValues(phasingMetricsFiles, readStructure);
         for (File tileMetricsOutFile : tileMetricsOutFiles) {
             TileMetricsOutReader tileMetricsIterator = new TileMetricsOutReader(tileMetricsOutFile, TileMetricsOutReader.TileMetricsVersion.THREE);
             final Collection<IlluminaTileMetrics> tileMetrics = determineLastValueForLaneTileMetricsCode(tileMetricsIterator);
@@ -179,7 +179,7 @@ public class TileMetricsUtil {
                                                     final ReadStructure readStructure,
                                                     final ValidationStringency validationStringency)
             throws FileNotFoundException {
-        Map<Integer, Collection<TilePhasingValue>> phasingValues = getTilePhasingValues(phasingMetricsFiles, readStructure);
+        Map<Integer, Map<Integer, Collection<TilePhasingValue>>> phasingValues = getTilePhasingValues(phasingMetricsFiles, readStructure);
         TileMetricsOutReader tileMetricsIterator = new TileMetricsOutReader(tileMetricsOutFile, TileMetricsOutReader.TileMetricsVersion.THREE);
         final Collection<IlluminaTileMetrics> tileMetrics = determineLastValueForLaneTileMetricsCode(tileMetricsIterator);
         final Map<String, ? extends Collection<IlluminaTileMetrics>> locationToMetricsMap = partitionTileMetricsByLocation(tileMetrics);
@@ -228,10 +228,10 @@ public class TileMetricsUtil {
         return Collections.unmodifiableCollection(tiles);
     }
 
-    private static Map<Integer, Collection<TilePhasingValue>> getTilePhasingValues(
+    private static Map<Integer, Map<Integer, Collection<TilePhasingValue>>> getTilePhasingValues(
                                                                      Map<Integer, File> phasingMetricFiles,
                                                                      final ReadStructure readStructure) {
-        final Map<Integer, Collection<TilePhasingValue>> phasingValues = new HashMap<>();
+        final Map<Integer, Map<Integer, Collection<TilePhasingValue>>> phasingValues = new HashMap<>();
         int totalCycleCount = 0;
 
         boolean isFirstRead = true;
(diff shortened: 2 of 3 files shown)


src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -121,14 +126,47 @@ private static File fileTileMetricsFile(File illuminaRunDirectory, Integer numCy
...
+            }
+        }
+        final String pathsString = tileMetricsOutFiles
+                .stream()
+                .map(File::getAbsolutePath)
+                .collect(Collectors.joining(", "));
+        throw new RuntimeException("None of the following input files contained cluster records:" + pathsString);
+    }
+
+    @Deprecated // Use parseClusterRecordsFromTileMetricsV3 instead
```
pshapiro4broad: Instead of a comment, add a javadoc. https://stackoverflow.com/a/9031854

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -91,24 +92,28 @@ private static File fileTileMetricsFile(File illuminaRunDirectory, Integer numCy
...
-            throws FileNotFoundException {
-        TileMetricsOutReader tileMetricsIterator = new TileMetricsOutReader(tileMetricsOutFile, TileMetricsOutReader.TileMetricsVersion.THREE);
-        final Collection<IlluminaTileMetrics> tileMetrics = determineLastValueForLaneTileMetricsCode(tileMetricsIterator);
-        final Map<String, ? extends Collection<IlluminaTileMetrics>> locationToMetricsMap = partitionTileMetricsByLocation(tileMetrics);
+    private static Collection<Tile> getTileClusterRecordsV3(
+            Map<String, ? extends Collection<IlluminaTileMetrics>> locationToMetricsMap,
+            final Map<Integer, File> phasingMetricsFiles,
+            final ReadStructure readStructure,
+            final ValidationStringency validationStringency,
+            final Float density) {
```
pshapiro4broad: Since `density` can't be null this should be `float` not `Float`.

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -74,11 +74,12 @@
...
      * Returns the path to the TileMetrics file given the basecalling directory.
      */
+    @Deprecated // Use findTileMetricsFiles instead; the illumina directory hierarchy for novaSeq is unreliable.
     public static File renderTileMetricsFileFromBasecallingDirectory(final File illuminaRunDirectory, Integer numCycles, boolean isNovaSeq) {
-        return fileTileMetricsFile(illuminaRunDirectory, numCycles, isNovaSeq);
+        return findTileMetricsFiles(illuminaRunDirectory, numCycles, isNovaSeq).get(0);
     }
 
-    private static File fileTileMetricsFile(File illuminaRunDirectory, Integer numCycles, boolean isNovaSeq) {
+    public static List<File> findTileMetricsFiles(File illuminaRunDirectory, Integer numCycles, boolean isNovaSeq) {
```
pshapiro4broad: Since `numCycles` can't be null this should be `int` not `Integer`.

src/test/java/picard/illumina/parser/TileMetricsUtilTest.java
```diff
@@ -8,67 +8,90 @@
...
+import java.util.List;
+import java.util.Random;
 import java.util.stream.IntStream;
 
 public class TileMetricsUtilTest {
     @Test
     public void testFindTileMetrics() throws IOException {
-        Path baseDir = createTestDirs(0, true);
-        File tileMetricsFile = TileMetricsUtil.renderTileMetricsFileFromBasecallingDirectory(
+        List<Integer> populatedCycleDirs = new ArrayList<>();
```
pshapiro4broad: It looks like this collection is never modified and is always empty. If so, it would be more efficient to use `Collections.emptyList()` instead of allocating a new list.

src/test/java/picard/illumina/parser/TileMetricsUtilTest.java
```diff
@@ -8,67 +8,90 @@
...
-        Assert.assertEquals(tileMetricsFile.getAbsolutePath(),
-                baseDir.resolve(TileMetricsUtil.INTEROP_SUBDIRECTORY_NAME)
-                        .resolve(TileMetricsUtil.TILE_METRICS_OUT_FILE_NAME).toString());
-
+        List<File> expected = generateFindTileMetricsExpected(baseDir, populatedCycleDirs, true);
+        Assert.assertEquals(tileMetricsFiles, expected);
         IOUtil.deleteDirectoryTree(baseDir.toFile());
     }
 
     @Test(expectedExceptions = IllegalStateException.class)
```
pshapiro4broad: If this test throws an exception, the `deleteDirectoryTree()` operation at the end will never execute.

Not sure what the best solution here is, using a shared test directory that the class creates and then deletes, using an `@AfterClass` method, might be best.

src/test/java/picard/illumina/parser/TileMetricsUtilTest.java
```diff
@@ -8,67 +8,90 @@
...
+    // put tile metrics file in several permutations of random cycle dirs
+    // and make sure we get all metricsFiles each time
     @Test
-    public void testFindTileMetricsNovaSeqInBaseInterOp() throws IOException {
-        Path baseDir = createTestDirs(1, true);
-        File tileMetricsFile = TileMetricsUtil.renderTileMetricsFileFromBasecallingDirectory(
-                baseDir.toFile(), 1, true);
+    public void testFindTileMetricsNovaSeq() throws IOException {
+        int numCycles = 15;
+        int numRandomDirectoryPermutations = 5;
```
pshapiro4broad: Why is randomness necessary here? Or is this just instead of creating names by hand? In general randomness in a unit test is not great, since a test can pass and then fail based on a random number.

src/test/java/picard/illumina/parser/TileMetricsUtilTest.java
```diff
@@ -8,67 +8,90 @@
...
+        for (int i=0; i<numRandomDirectoryPermutations; i++) {
```
pshapiro4broad: reformat so there are spaces around `=` `<` operators

src/test/java/picard/illumina/parser/TileMetricsUtilTest.java
```diff
@@ -80,4 +103,27 @@ private Path createTestDirs(int numCycleMetricsFiles, boolean baseMetricsFile) t
...
         return baseDir;
     }
+
+    private List<File> generateFindTileMetricsExpected(Path baseDir, List<Integer> populatedCycleDirs, boolean baseMetricsFile) {
+        List<File> expected = new ArrayList<>();
+        if (baseMetricsFile) {
+            expected.add(new File(
+                baseDir
+                    .resolve(TileMetricsUtil.INTEROP_SUBDIRECTORY_NAME)
+                    .resolve(TileMetricsUtil.TILE_METRICS_OUT_FILE_NAME).toUri())
```
pshapiro4broad: why `new File(<path expression>.toUri()` instead of `<path expression>.toFile()`?

src/test/java/picard/illumina/parser/TileMetricsUtilTest.java
```diff
@@ -2,90 +2,87 @@
...
+            {"novaSeqBaseAndSomeCycleFiles", 5, someCycleFiles, true, true}
+        };
     }
 
     @Test(expectedExceptions = IllegalStateException.class)
     public void testMissingTileMetrics() throws IOException {
-        List<Integer> populatedCycleDirs = new ArrayList<>();
-        Path baseDir = createTestDirs(0, populatedCycleDirs, false);
+        List<Integer> populatedCycleDirs = Collections.emptyList();
+        Path baseDir = createTestDirs("missingTileMetrics",0, populatedCycleDirs, false);
```
pshapiro4broad: missing space after `,`

src/test/java/picard/illumina/parser/TileMetricsUtilTest.java
```diff
@@ -2,73 +2,93 @@
...
+        return new Object[][]{
+            //specificTestDir, numCycleMetricsFiles, populatedCycleDirs, baseMetricsFile, isNovaSeq
+            {"notNovaSeq", 0, Collections.emptyList(), true, false},
+            {"novaSeqBaseFileOnly", 0, Collections.emptyList(), true, true},
+            {"novaSeqSingleCycleFile", 1, oneCycleFile, false, true},
+            {"novaSeqBaseAndSingleCycleFile", 1, oneCycleFile, true, true},
+            {"novaSeqAllCycleFiles", 5, allCycleFiles, false, true},
+            {"novaSeqBaseAndAllCycleFiles", 5, allCycleFiles, true, true},
+            {"novaSeqSomeCycleFiles", 5, someCycleFiles, false, true},
+            {"novaSeqBaseAndSomeCycleFiles", 5, someCycleFiles, true, true}
```
tlangs: What is going on here?

ApChagi: These are test inputs that are run through testFilePresencePermutation. Whether the base directory is populated or not, and whether all, none, one or some non-sequential cycle directories contain TileMetricsOut.bin files, findTileMetricsFiles should identify and return a list of all of them

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -117,18 +122,70 @@ private static File fileTileMetricsFile(File illuminaRunDirectory, Integer numCy
...
+        return record.isClusterRecord();
+    }
+
+    public static Collection<Tile> parseClusterRecordsFromTileMetricsV3(
+            final Collection<File> tileMetricsOutFiles,
+            final Map<Integer, File> phasingMetricsFiles,
+            final ReadStructure readStructure
+    ) throws FileNotFoundException {
+        // Snag and cache the phasing data for each read in the read structure for each tile.
+        // For both types of phasing values, this is the median of all of the individual values seen
```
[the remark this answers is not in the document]
ApChagi: this is happening on line 281 and 282. This comment was before the call to getTilePhasingValues before, would it be clearer if it were javadoc for the method (getTilePhasingValues)?

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -117,18 +122,70 @@ private static File fileTileMetricsFile(File illuminaRunDirectory, Integer numCy
...
+        final IlluminaTileMetrics record = CollectionUtil.getSoleElement(entry.getValue());
+        return record.isClusterRecord();
+    }
+
+    public static Collection<Tile> parseClusterRecordsFromTileMetricsV3(
+            final Collection<File> tileMetricsOutFiles,
+            final Map<Integer, File> phasingMetricsFiles,
+            final ReadStructure readStructure
+    ) throws FileNotFoundException {
+        // Snag and cache the phasing data for each read in the read structure for each tile.
```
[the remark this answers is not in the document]
ApChagi: keep phasingValues in memory so we don't reparse for each tile in the tileMetricsOut.bin we end up parsing. Not sure what "phasing data" is exactly, but we collect it per read per tile

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -117,18 +122,70 @@ private static File fileTileMetricsFile(File illuminaRunDirectory, Integer numCy
...
+        // For both types of phasing values, this is the median of all of the individual values seen
+        Map<Integer, Map<Integer, Collection<TilePhasingValue>>> phasingValues = getTilePhasingValues(phasingMetricsFiles, readStructure);
+        for (File tileMetricsOutFile : tileMetricsOutFiles) {
```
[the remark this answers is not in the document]
ApChagi: For NovaSeq runs, multiple TileMetricsOut.bin files are generated. Theoretically we should be able to rely on the directory structure to decided which one to use, but this turned out to be unreliable. Ultimately, we want the file that contains cluster records ([file spec](http://illumina.github.io/interop/tile_v3.html)). In the worst case, making sure a file does not contain cluster records means parsing the file and iterating through each record it contains. This loop takes a quick peek into the file to try and find a cluster record (I believe a file may contain either just 'r' type records, just 't' type cluster records, or an equal number of both.). If it finds a cluster record, it collects and returns all cluster records in the current file. The next loop goes through every tile record in every tileMetricsFile until it has collected a non-zero number of records from a file, at which point it returns those tiles.

Since file parsing has now been significantly sped up, it may be the case that we aren't gaining much from this first pass. What do you think, @jacarey ?

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -229,7 +281,21 @@ private static File fileTileMetricsFile(File illuminaRunDirectory, Integer numCy
                         float[] phasingSlopeAndOffset = computeLinearFit(cycleNumWithData.toArray(new Float[0]), phasingMetrics.toArray(new Float[0]), phasingMetrics.size());
                         float[] prePhasingSlopeAndOffset = computeLinearFit(cycleNumWithData.toArray(new Float[0]), prephasingMetrics.toArray(new Float[0]), phasingMetrics.size());
                         TilePhasingValue value = new TilePhasingValue(tileTemplateRead, phasingSlopeAndOffset[0], prePhasingSlopeAndOffset[0]);
-                        tilePhasingValues.add(value);
+                        if (phasingValues.containsKey(laneNum)) {
```
pshapiro4broad: I believe all this logic can be replaced with two calls to `computeIfAbsent`, e.g.
```
phasingValues.computeIfAbsent(laneNum, k -> new HashMap<>()).computeIfAbsent(tileNum, k -> new ArrayList<>()).add(value);
```

ApChagi: Ah! I was hoping there might be something like this

src/main/java/picard/illumina/parser/TileMetricsUtil.java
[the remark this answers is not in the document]
als364: "Phasing data" is more what I meant, unfortunately

src/main/java/picard/illumina/parser/TileMetricsUtil.java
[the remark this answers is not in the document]
als364: Probably, yes.

src/main/java/picard/illumina/parser/TileMetricsUtil.java
pshapiro4broad: It looks like the code above could also benefit from this API
```
int laneNumber = laneTileCode.getLaneNumber();
phasing.computeIfAbsent(tileNumber, k -> new HashMap<>()).computeIfAbsent(laneNumber, k -> new ArrayList<>()).add(phasingMetrics.phasingWeight);
prePhasing.computeIfAbsent(tileNumber, k -> new HashMap<>()).computeIfAbsent(laneNumber, k -> new ArrayList<>()).add(phasingMetrics.prephasingWeight);
```

src/main/java/picard/illumina/parser/TileMetricsUtil.java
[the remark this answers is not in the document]
ApChagi: Google is my friend! Apparently [phasing values](http://lab.loman.net/high-throughput%20sequencing/2013/11/21/diagnosing-problems-with-phasing-and-pre-phasing-on-illumina-platforms/) are some measure of lost signal due to some molecules losing sync with one another. Apparently, if it's measured, it can be corrected for. Still not 100% clear on what it means for molecules to lose sync...

src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
```diff
@@ -154,15 +154,15 @@ public static void main(final String[] args) {
...
+                    tiles = TileMetricsUtil.parseClusterRecordsFromTileMetricsV3(
+                            tileMetricsOutFiles,
                             TileMetricsUtil.renderPhasingMetricsFilesFromBasecallingDirectory(illuminaRunDirectory),
-                            readStructure,
-                            validationStringency);
+                            readStructure
+                    );
                 } else {
-                    tiles = TileMetricsUtil.parseTileMetrics(tileMetricsOutFile,
+                    tiles = TileMetricsUtil.parseTileMetrics(tileMetricsOutFiles.get(0),
```
danxmoran: Nitpick: push `tileMetricsOutFiles.get(0)` to the next line, to match the formatting of the call in the `if` branch?

src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
```diff
@@ -154,15 +154,15 @@ public static void main(final String[] args) {
                                                                              final boolean isNovaSeq) {
             final Collection<Tile> tiles;
             try {
-                File tileMetricsOutFile = TileMetricsUtil.renderTileMetricsFileFromBasecallingDirectory(illuminaRunDirectory, readStructure.totalCycles, isNovaSeq);
+                List<File> tileMetricsOutFiles = TileMetricsUtil.findTileMetricsFiles(illuminaRunDirectory, readStructure.totalCycles, isNovaSeq);
```
danxmoran: Could be `final`? (Not sure if it matters)

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -91,24 +94,26 @@ private static File fileTileMetricsFile(File illuminaRunDirectory, Integer numCy
...
-    public static Collection<Tile> parseTileMetrics(final File tileMetricsOutFile,
-                                                    final Map<Integer, File> phasingMetricsFiles,
-                                                    final ReadStructure readStructure,
-                                                    final ValidationStringency validationStringency)
-            throws FileNotFoundException {
-        TileMetricsOutReader tileMetricsIterator = new TileMetricsOutReader(tileMetricsOutFile, TileMetricsOutReader.TileMetricsVersion.THREE);
-        final Collection<IlluminaTileMetrics> tileMetrics = determineLastValueForLaneTileMetricsCode(tileMetricsIterator);
-        final Map<String, ? extends Collection<IlluminaTileMetrics>> locationToMetricsMap = partitionTileMetricsByLocation(tileMetrics);
+    private static Collection<Tile> getTileClusterRecordsV3(
+            Map<String, ? extends Collection<IlluminaTileMetrics>> locationToMetricsMap,
```
danxmoran: Could be `final`?

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -117,18 +122,68 @@ private static File fileTileMetricsFile(File illuminaRunDirectory, Integer numCy
...
+        final IlluminaTileMetrics record = CollectionUtil.getSoleElement(entry.getValue());
+        return record.isClusterRecord();
+    }
+
+    public static Collection<Tile> parseClusterRecordsFromTileMetricsV3(
+            final Collection<File> tileMetricsOutFiles,
+            final Map<Integer, File> phasingMetricsFiles,
+            final ReadStructure readStructure
+    ) throws FileNotFoundException {
+        Map<Integer, Map<Integer, Collection<TilePhasingValue>>> phasingValues = getTilePhasingValues(phasingMetricsFiles, readStructure);
```
danxmoran: Could be `final`?

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -117,18 +122,68 @@ private static File fileTileMetricsFile(File illuminaRunDirectory, Integer numCy
...
+        for (File tileMetricsOutFile : tileMetricsOutFiles) {
+            TileMetricsOutReader tileMetricsIterator = new TileMetricsOutReader(tileMetricsOutFile, TileMetricsOutReader.TileMetricsVersion.THREE);
```
danxmoran: Could be `final`?

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -117,18 +122,68 @@ private static File fileTileMetricsFile(File illuminaRunDirectory, Integer numCy
...
+            final Collection<IlluminaTileMetrics> tileMetrics = determineLastValueForLaneTileMetricsCode(tileMetricsIterator);
+            final Map<String, ? extends Collection<IlluminaTileMetrics>> locationToMetricsMap = partitionTileMetricsByLocation(tileMetrics);
+            if (v3TileMetricsHasClusterRecordsFirstPass(locationToMetricsMap)) {
+                final float density = tileMetricsIterator.getDensity();
+                return getTileClusterRecordsV3(locationToMetricsMap, phasingValues, density);
+            }
+        }
+        for (File tileMetricsOutFile : tileMetricsOutFiles) {
```
danxmoran: Could the logic in this loop be combined with the loop above? There's some duplication.

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -117,18 +122,68 @@ private static File fileTileMetricsFile(File illuminaRunDirectory, Integer numCy
...
+    /**
+     * @deprecated use {@link #parseClusterRecordsFromTileMetricsV3(Collection, Map, ReadStructure)} instead
+     */
+    @Deprecated
+    public static Collection<Tile> parseTileMetrics(final File tileMetricsOutFile,
+                                                    final Map<Integer, File> phasingMetricsFiles,
+                                                    final ReadStructure readStructure,
+                                                    final ValidationStringency validationStringency)
+            throws FileNotFoundException {
+        Map<Integer, Map<Integer, Collection<TilePhasingValue>>> phasingValues = getTilePhasingValues(phasingMetricsFiles, readStructure);
```
danxmoran: Could be `final`?

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -117,18 +122,68 @@ private static File fileTileMetricsFile(File illuminaRunDirectory, Integer numCy
...
+        TileMetricsOutReader tileMetricsIterator = new TileMetricsOutReader(tileMetricsOutFile, TileMetricsOutReader.TileMetricsVersion.THREE);
```
danxmoran: Could be `final`?

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -117,18 +122,53 @@ private static File fileTileMetricsFile(File illuminaRunDirectory, Integer numCy
...
+            final Collection<Tile> tiles = getTileClusterRecordsV3(locationToMetricsMap, phasingValues, density);
+            if (!tiles.isEmpty()) {
+                return tiles;
+            }
+        }
+        final String pathsString = tileMetricsOutFiles
+                .stream()
+                .map(File::getAbsolutePath)
+                .collect(Collectors.joining(", "));
+        throw new RuntimeException("None of the following input files contained cluster records:" + pathsString);
```
pshapiro4broad: I think a space after the `:` would look better in the output

PR #1178, opened 2018-05-25
Fixes intermittent failures in TheoreticalHetSensitivityTests.

yfarjoun:
The problem was a hidden random seed that was initialized from the system time and memory location.

### DIFF
--- a/src/main/java/picard/analysis/CollectWgsMetrics.java
+++ b/src/main/java/picard/analysis/CollectWgsMetrics.java
@@ -159,7 +159,7 @@ static final String USAGE_DETAILS = "<p>This tool collects metrics about the fra
                 optional = true)
         public File INTERVALS;
 
-        public File getIntervalFile() { return INTERVALS; };
+        public File getIntervalFile() { return INTERVALS; }
     };
 
     /** Metrics for evaluating the performance of whole genome sequencing experiments. */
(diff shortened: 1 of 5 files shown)


yfarjoun:
increased the sampling counts as it was too small for the tests to converge.

src/main/java/picard/analysis/directed/CollectTargetedMetrics.java
```diff
@@ -165,6 +165,7 @@ protected int doWork() {
         if (THEORETICAL_SENSITIVITY_OUTPUT != null) {
             // Write out theoretical sensitivity results.
             final MetricsFile<TheoreticalSensitivityMetrics, ?> theoreticalSensitivityMetrics = getMetricsFile();
+            log.info("Calculating theoretical sentitivity at " + ALLELE_FRACTION.size() + " allele fractions. ");
```
fleharty: Do you want the space after "allele fractions."?
Also, if you are going to put this log here, can you put another one in CollectWgsMetrics?

src/main/java/picard/analysis/CollectWgsMetrics.java
```diff
@@ -501,6 +501,7 @@ protected int doWork() {
         if (THEORETICAL_SENSITIVITY_OUTPUT != null) {
             // Write out theoretical sensitivity results.
             final MetricsFile<TheoreticalSensitivityMetrics, ?> theoreticalSensitivityMetrics = getMetricsFile();
+            log.info("Calculating theoretical sentitivity at " + ALLELE_FRACTION.size() + " allele fractions. ");
```
fleharty: Serious nit-pick, but is there a reason for the space after "fractions."?

PR #1195, opened 2018-07-09
slight changes to MarkDuplciates

yfarjoun:
### Description
- Making MarkDuplicates work on query-grouped (not only sorted) input. This can speed up pipelines by not requiring to sort the output of the sequencer prior to Marking duplicates. 
- tweak the tie-brake function (for determining the representative read) so that file-order is trumped by a hash of the tile-and-position. This should make parallel version of MarkDuplicates easier to write in a way that is identical with this version (the file-position tie break should never actually occur)
- added tests

#### Content

#### Review


src/test/java/picard/sam/markduplicates/UmiAwareMarkDuplicatesWithMateCigarTest.java
```diff
@@ -54,75 +54,75 @@ protected UmiAwareMarkDuplicatesWithMateCigarTester getTester(final boolean allo
     private Object[][] testUmiSetsDataProvider() {
         return new Object[][] {{
                 // Test basic error correction using edit distance of 1
-                Arrays.asList(new String[] {"AAAA", "AAAA", "ATTA", "AAAA", "AAAT"}), // Observed UMI
-                Arrays.asList(new String[] {"AAAA", "AAAA", "ATTA", "AAAA", "AAAA"}), // Expected inferred UMI
-                Arrays.asList(new Boolean[] {false, true, false, true, true}), // Should it be marked as duplicate?
+                Arrays.asList("AAAA", "AAAA", "ATTA", "AAAA", "AAAT"), // Observed UMI
```
pshapiro4broad: I would revert these changes as they will collide with the same changes from @fleharty 's branch, and it doesn't look like the tests were changed here, unlike Mark's changes.

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -925,6 +933,14 @@ public int compare(final ReadEndsForMarkDuplicates lhs, final ReadEndsForMarkDup
             if (compareDifference == 0) {
                 compareDifference = lhs.read2Coordinate - rhs.read2Coordinate;
             }
+
+            if (compareDifference == 0)
```
pshapiro4broad: add `{` `}` around `if` body please

lbergelson:
@jamesemery Take a look at this.

src/main/java/picard/sam/markduplicates/util/PhysicalLocationForMateCigar.java
```diff
@@ -75,11 +75,9 @@ public boolean equals(Object other) {
 
     @Override
     public int hashCode() {
-        int result = getLibraryId();
+        int result = super.hashCode();
```
pshapiro4broad: I think we should start the practice of using `Objects.hash()` instead of computing hashes by hand, e.g. replace this code with
```
return Objects.hash(super.hashCode(), getLibraryId(), getReadGroup());
```

src/main/java/picard/sam/util/PhysicalLocationInt.java
```diff
@@ -34,4 +34,11 @@
 
     public void setLibraryId(final short libraryId) { throw new PicardException("Not Implemented"); }
 
+    @Override
+    public int hashCode() {
```
pshapiro4broad: If you do this, you also need to override `equals()` so it also works this way.

src/test/java/picard/sam/markduplicates/MarkDuplicatesWithMateCigarTest.java
```diff
@@ -37,6 +37,16 @@ protected AbstractMarkDuplicatesCommandLineProgramTester getTester() {
         return new MarkDuplicatesWithMateCigarTester();
     }
 
+    /**
+     * Currently this class fails this test due to
+     * a small difference in the selection of representative read. To fix this
+     * a change is required in htsjdk.samtools.DuplicateScoringStrategy.compare
```
pshapiro4broad: Create a github issue for this? Or is it not worth fixing?

src/test/java/picard/sam/markduplicates/AsIsMarkDuplicatesTester.java
```diff
@@ -63,6 +63,31 @@ public void testSameUnclipped5PrimeOppositeStrand(final File input) {
...
+    public void testQueryGroupedInput(final File input) {
+
+        final AbstractMarkDuplicatesCommandLineProgramTester tester = new BySumOfBaseQAndInOriginalOrderMDTester();
+
+        final SamReader reader = SamReaderFactory.makeDefault().open(input);
+
+        tester.setHeader(reader.getFileHeader());
+        reader.iterator().stream().forEach(tester::addRecord);
+
+        CloserUtil.close(reader);
```
pshapiro4broad: use try-with-resources and remove explicit `close()`

src/main/java/picard/sam/util/PhysicalLocationInt.java
```diff
@@ -34,4 +34,11 @@
...
+        int result = (int) tile;
+        result = 31 * result + x;
+        result = 31 * result + y;
+        return result;
```
pshapiro4broad: can also replace with `Objects.hash()`

src/test/java/picard/sam/markduplicates/MarkDuplicatesWithMateCigarTest.java
yfarjoun: done.

src/test/java/picard/sam/markduplicates/AsIsMarkDuplicatesTester.java
yfarjoun: done.

yfarjoun:
@jamesemery my original thought is that using the hash might be faster...but thinking about it again, I agree with you that it it's not better.

nh13:
@yfarjoun more explanation is needed how this achieves what you want in bullet 2 (changing the representative).  I thought that `ReadEndsMDComparator` was just used to resort the reads so that "duplicates" were grouped together (see `pairSort` and `fragSort`).  Then the `DUPLICATE_SCORING_STRATEGY` was used to choose among the duplicates.

yfarjoun:
you are correct @nh13. But if theres'a tie in the comparator, a representative still needs to be chosen... one of the last tiebreakers is the original order of the reads in the file , but since for spark this is not good (it will require remembering another int or long for every read, just for this obscure purpose) I suggested a different tie-break mechanism, based on the tile and xy coordinates.

nh13:
@yfarjoun I don’t see how this changes the representative as the comparator you changed isn’t used to pick the representative. Sorry if I am missing something obvious. Can you point me to the code in MarkDuplicates that chooses the representative and then where it uses the comparator you changed. I can’t see it.

yfarjoun:
So, MarkDupilcates is indeed complicated:

1. Upon the first traversal, the read(end)s are sorted by ReadEndsMDComparator order and tiebreak order (I've modified slightly the tiebreak order)
2. after the readends are sorted, they are iterated over (in that order) and collected into duplicate sets (keeping the original order)
3. the reads are compared according to their "score" and the **first** read that has the highest score is chosen as the representative.

This should explain how the original order in the ReadEnds Comparator affects the chosen representative during tie-breaks of the scoring function.

PR #1201, opened 2018-07-19
Fix an issue where the google NIO code would try to load even when the NIO library wasn't present.

pshapiro4broad:
Due to the way `GCS.isAvailable()` worked, a false positive could be generated when the google storage APIs were present, but the contrib NIO APIs were not. It turned out that the presence of the class `com.google.cloud.TransportOptions` would cause `isAvailable()` to return true.

This change requires that `initialize()` is successful, which is a better test of the presence of the NIO code.

It is slightly different than before in that it now initializes GCS immediately on startup, instead of waiting until `PathHelper.initilizeAll()` is called.

#### Content

#### Review


src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -244,21 +245,15 @@ public int instanceMain(final String[] argv) {
...
             try {
-                final StringBuilder pathProvidersBuilder = new StringBuilder();
-
-                for (PathHelper.PathProviders providers : PathHelper.PathProviders.values()) {
-                    pathProvidersBuilder.append(String.format("Provider %s is%s available; ",
-                            providers.name(), providers.isAvailable() ? "" : " not"));
-                }
-                final int lastSpacePos = pathProvidersBuilder.lastIndexOf(" ");
-                pathProvidersBuilder.delete(lastSpacePos, lastSpacePos + 1);
+                String pathProvidersMessage =
```
danxmoran: Could be `final` (seems like the convention for locals in this method).

pshapiro4broad:
I verified this fix locally by writing a test program that has the gradle dependency:
```
        compile 'com.google.cloud:google-cloud-storage:1.35.0'
```
Without this fix I get an error invoking picard `CommandLineProgram`:
```
Exception in thread "main" java.lang.NoClassDefFoundError: com/google/cloud/storage/contrib/nio/CloudStorageConfiguration
	at picard.nio.GoogleStorageUtils.getCloudStorageConfiguration(GoogleStorageUtils.java:58)
	at picard.nio.GoogleStorageUtils.initialize(GoogleStorageUtils.java:52)
	at picard.nio.PathHelper$PathProviders$1.initialize(PathHelper.java:40)
	at picard.nio.PathHelper.initilizeAll(PathHelper.java:71)
	at picard.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:247)
	at picard.cmdline.PicardCommandLine.instanceMain(PicardCommandLine.java:103)
	at edu.mit.broad.picard.cmdline.PicardPrivateCommandLine.main(PicardPrivateCommandLine.java:24)
Caused by: java.lang.ClassNotFoundException: com.google.cloud.storage.contrib.nio.CloudStorageConfiguration
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 7 more
```
With the fix, no error is generated -- the code correctly doesn't attempt to initialize the GCS storage object.

PR #1202, opened 2018-07-20
Tool to add `OA` sam tag to reads

jsotobroad:
### Description

We are realigning autosomal reads to the mitochondria and we wanted to store the original alignment information in a tag to be used by filtering later on.  There is a PR in hts-specs for a new tag `OA`, https://github.com/samtools/hts-specs/pull/193, that stores the information that we want to store.  This tools allows someone to set the `OA` tag for all aligned reads or within an interval if desired.

#### Content

#### Review


src/main/java/picard/sam/AddOATag.java
```diff
@@ -0,0 +1,103 @@
...
+
+        OverlapDetector overlapDetector = getOverlapDetectorFromIntervalListFile(INTERVAL_LIST, 0, 0);
+        for (SAMRecord rec : reader) {
+            if (!rec.getReadUnmappedFlag() && (overlapDetector == null || overlapDetector.overlapsAny(rec))) {
+                setOATag(rec, OVERWRITE_TAG);
+            }
+            writer.addAlignment(rec);
+        }
+
+        CloserUtil.close(reader);
```
magicDGS: Instead of closing like this, I would suggest to have a try-with-resources:

```java
try(final SamReader reader = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(INPUT)) {
    ...
}
```

You can also include the writer there...

src/test/java/picard/sam/AddOATagTest.java
```diff
@@ -0,0 +1,112 @@
...
+            args.add("INTERVAL_LIST=" + intervalList);
+        }
+        if (overwriteTag != null) {
+            args.add("OVERWRITE_TAG=" + overwriteTag);
+        }
+        AddOATag addOATag = new AddOATag();
+        Assert.assertEquals(addOATag.instanceMain(args.toArray(new String[args.size()])), 0, "Running addOATag did not succeed");
+    }
+
+    private void validateBamForOA(final File input, final File intervalList, Boolean overwriteTag) {
```
magicDGS: I would prefer to have tests files to compare with; here the code is similar to the tool one; for example, if an htsjdk upgraded version has a bug which changes the results of this tool, there is not going to be detected by this test...

src/test/java/picard/sam/AddOATagTest.java
```diff
@@ -0,0 +1,112 @@
...
+        final SamReader reader = SamReaderFactory.makeDefault().open(input);
+        final SAMRecordIterator iterator = reader.iterator();
+
+        OverlapDetector<Interval> detector = AddOATag.getOverlapDetectorFromIntervalListFile(intervalList, 0, 0);
+        while (iterator.hasNext()){
+            SAMRecord rec = iterator.next();
+            if (!rec.getReadUnmappedFlag() && (detector == null || detector.overlapsAny(rec))){
+                Assert.assertNotNull(rec.getAttribute("OA"));
```
magicDGS: Non-null does not mean that is correct, just that it was set; I prefer the file-based test due to that.

src/test/java/picard/sam/AddOATagTest.java
jsotobroad: Feel the same way, will change it up 👍

jsotobroad:
@magicDGS if you could take a second look when you get a chance, thanks!

src/main/java/picard/sam/AddOATag.java
```diff
@@ -0,0 +1,105 @@
...
+            OverlapDetector<Interval> detector = new OverlapDetector<>(lhsBuffer, rhsBuffer);
+            detector.addAll(intervals, intervals);
+            return detector;
+        }
+    }
+
+    // format OA tag string according to the spec
+    private void setOATag(SAMRecord rec, Boolean overrideTag) {
+        String OAValue = String.format("%s,%s,%s,%s,%s,%s;",
+                (rec.getReferenceName().replace(",", "_")),
```
magicDGS: This replacement of `,` for `_` should be documented somewhere. Maybe @yfarjoun could add a recommendation in the specs (I've already added a comment there).

src/main/java/picard/sam/AddOATag.java
```diff
@@ -0,0 +1,105 @@
...
+        } else {
+            List<Interval> intervals = IntervalList.fromFile(intervalList).uniqued().getIntervals();
+            OverlapDetector<Interval> detector = new OverlapDetector<>(lhsBuffer, rhsBuffer);
+            detector.addAll(intervals, intervals);
+            return detector;
+        }
+    }
+
+    // format OA tag string according to the spec
+    private void setOATag(SAMRecord rec, Boolean overrideTag) {
```
magicDGS: Would be nice to add a TODO to backport to htsjdk in an utility class for `SAMRecord`

src/main/java/picard/sam/AddOATag.java
```diff
@@ -0,0 +1,105 @@
...
+    private static final Log log = Log.getInstance(AddOATag.class);
+
+    public static void main(final String[] argv) {
+        System.exit(new AddOATag().instanceMain(argv));
+    }
+
+    @Override
+    protected int doWork() {
+        try (final SamReader reader = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(INPUT)) {
+            try (final SAMFileWriter writer = new SAMFileWriterFactory().makeSAMOrBAMWriter(reader.getFileHeader(), true, OUTPUT)) {
```
magicDGS: You can put the constructors in the same `try` clause:

```java
       try (final SamReader reader = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(INPUT);
             final SAMFileWriter writer = new SAMFileWriterFactory().makeSAMOrBAMWriter(reader.getFileHeader(), true, OUTPUT)) {
...
```

magicDGS: Also, if you use `makeWriter` instead of `makeSAMOrBAMWriter` you can support CRAM output, and you just need to add the `REFERENCE_SEQUENCE` arg at the end:

```java
new SAMFileWriterFactory().makeWriter(reader.getFileHeader(), true, OUTPUT, REFERENCE_SEQUENCE)
```

src/main/java/picard/sam/AddOATag.java
```diff
@@ -0,0 +1,105 @@
...
+                writer.setProgressLogger(
+                        new ProgressLogger(log, (int) 1e7, "Wrote", "records"));
```
magicDGS: Isn't it better to have a progress logger for the processing instead for the writing? It doesn't really matter, but depends where you wanna put the focus on the logger.

src/main/java/picard/sam/AddOATag.java
```diff
@@ -0,0 +1,105 @@
...
+
+                OverlapDetector overlapDetector = getOverlapDetectorFromIntervalListFile(INTERVAL_LIST, 0, 0);
+                for (SAMRecord rec : reader) {
+                    if (!rec.getReadUnmappedFlag() && (overlapDetector == null || overlapDetector.overlapsAny(rec))) {
```
magicDGS: It will be nice to have a `SamRecordFilter` here instead here to simpify code: instead of `getOverlapDetectorFromIntervalListFile`, a `makeFilter` will return a `SAMRecordFilter` consisting only on `AlignedFilter` if no intervals were specified or a `AggregateFilter` with a list including `AlignedFilter` and a "OverlapDetectorFilter" (not present in htsjdk or picard although is easy to implement). Maybe this is out-of-scope for this PR, but it will make the code simpler and readable.

src/main/java/picard/sam/AddOATag.java
```diff
@@ -0,0 +1,105 @@
...
+                        setOATag(rec, OVERWRITE_TAG);
```
magicDGS: Actually, I realized now that using overwrite and intervals might have an interesting effect on unmapped pairs where the mate is mapped: if a `OA` tag already exists for the unmapped read and for the mate, it will be overwrite only for the mate and the unmapped one will conserve the original tag. For evaluation of a downstream realigner this can be problematic, as they can assume that the unmapped read was originally mapped...

src/test/java/picard/sam/AddOATagTest.java
```diff
@@ -0,0 +1,142 @@
...
+                if (overwriteTag != null){
+                    validateOverwriteTag(rec, overwriteTag);
+                }
+            } else {
+                Assert.assertNull(rec.getAttribute("OA"));
+            }
+        }
+    }
+
+    private void validateOverwriteTag(SAMRecord rec, Boolean overwriteTag) {
```
magicDGS: I would suggest to have instead of a boolean an integer indicating how many OA tags are expected - this is more flexible for adding new tests.

src/test/java/picard/sam/AddOATagTest.java
```diff
@@ -0,0 +1,142 @@
...
+        Assert.assertTrue(truthOAValues.equals(testOAValues));
+    }
+
+    private void validateBamHasOA(final File input, final File intervalList, Boolean overwriteTag) {
+        final SAMRecordIterator iterator = SamReaderFactory.makeDefault().open(input).iterator();
+
+        OverlapDetector<Interval> detector = AddOATag.getOverlapDetectorFromIntervalListFile(intervalList, 0, 0);
+        while (iterator.hasNext()){
+            SAMRecord rec = iterator.next();
+            if (!rec.getReadUnmappedFlag() && (detector == null || detector.overlapsAny(rec))){
```
magicDGS: I still prefer that all the validation is done with real output files, as this for example does not catch the problem that I pinpoint about unmapped pairs with paired mate (and thus, it is unclear if the behaviour is expected).

src/test/java/picard/sam/AddOATagTest.java
```diff
@@ -0,0 +1,142 @@
...
+                new Object[]{new File("testdata/picard/sam/aligned.sam"), new File("testdata/picard/analysis/directed/CollectHsMetrics/chrM.interval_list")},
+                new Object[]{new File("testdata/picard/sam/aligned.sam"), new File("testdata/picard/sam/contiguous.interval_list")}
+        };
+    }
+
+    @DataProvider(name="testOverwriteData")
+    Object[][] overwriteTestData() {
+        return new Object[][]{
+                new Object[]{new File("testdata/picard/sam/aligned.sam"), null, true},
+                new Object[]{new File("testdata/picard/sam/aligned.sam"), null, false}
```
magicDGS: Because the default is overwrite, I guess that the test cases here should be for not overwriting, with and without interval list and with/without original OA tag.

src/test/java/picard/sam/AddOATagTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+
+public class AddOATagTest {
+    @DataProvider(name="testOAData")
+    Object[][] testOAData() {
+        return new Object[][]{
+                new Object[]{new File("testdata/picard/sam/aligned_withoutOA.sam"), new File("testdata/picard/sam/aligned_withOA.sam")},
```
magicDGS: Also, a test with an aligned with original OA to test the default overwrite will be nice here. Also with unmapped reads with OA tag because the current code won't overwrite them (and I guess it should).

src/test/java/picard/sam/AddOATagTest.java
```diff
@@ -0,0 +1,142 @@
...
+        };
+    }
+
+    @DataProvider(name="testIntervalListData")
+    Object[][] testIntervalListData() {
+        return new Object[][]{
+                new Object[]{new File("testdata/picard/sam/aligned.sam"), null},
```
magicDGS: I guess that this test cases should also be with intervals only.

src/test/java/picard/sam/AddOATagTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+
+public class AddOATagTest {
+    @DataProvider(name="testOAData")
```
magicDGS: Maybe instead of the three separated test cases, a single one containing all of them is enought unless there is a specific reason for separate interval/overwrite tests. In that case, a single test method will has an input file, overwrite boolean, interval list (or null if none) and expected output. This will allow to share expected files and input files (e.g., the `aligned_withoutOA.sam` and `aligned_withOtherOA.sam` as inputs should produce both the same `expected_overwriteOA.sam` if they contain the same records), and be a good integration test to check if they are really the same.

src/main/java/picard/sam/AddOATag.java
yfarjoun: commas are going to be outlawed shortly....I'd be happy with a exception here.

src/main/java/picard/sam/AddOATag.java
meganshand: I added an OA tag for unmapped reads which should help resolve this issue.

src/test/java/picard/sam/AddOATagTest.java
meganshand: I left these as separate tests, but now they are all file based integration tests. Each case has an expected sam file.

src/main/java/picard/sam/AddOATag.java
```diff
@@ -0,0 +1,105 @@
...
+            List<Interval> intervals = IntervalList.fromFile(intervalList).uniqued().getIntervals();
+            OverlapDetector<Interval> detector = new OverlapDetector<>(lhsBuffer, rhsBuffer);
+            detector.addAll(intervals, intervals);
+            return detector;
+        }
+    }
+
+    // format OA tag string according to the spec
+    private void setOATag(SAMRecord rec, Boolean overrideTag) {
+        String OAValue = String.format("%s,%s,%s,%s,%s,%s;",
```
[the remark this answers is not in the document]
meganshand: I don't understand this comment, are there supposed to be parenthesis in this tag? It's hard to read the spec when it's still in a PR so maybe I'm confused about what this tag should look like.

meganshand:
Sorry for the delay. I'm going to take over this PR from @jsotobroad 

@yfarjoun @magicDGS Back to you.

src/main/java/picard/sam/AddOATag.java
meganshand: I'm not sure I understand the use case for either of these options (an OVERWRITE_TAG that only overwrites what's within the intervals or REMOVE_OLD_TAG). I don't love that REMOVE_OLD_TAG would change records outside of a provided interval and I'm not entirely sure what problem it's trying to solve. I might remove the option entirely until there's a need for it.

meganshand:
Thanks @yfarjoun, back to you!

src/main/java/picard/sam/AddOATag.java
```diff
@@ -0,0 +1,129 @@
...
+            "<h4>Usage example:</h4>" +
+            "<pre>" +
+            "java -jar picard.jar AddOATag \\<br />" +
+            "      L=some_picard.interval_list \\<br />" +
+            "      I=sorted.bam \\<br />" +
+            "      O=fixed.bam <br />"+
+            "</pre>";
+
+    @Argument(shortName = StandardOptionDefinitions.INPUT_SHORT_NAME, doc = "SAM or BAM input file")
+    public File INPUT;
```
lbergelson: why not use path?

src/main/java/picard/sam/AddOATag.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Argument(shortName = StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "SAM or BAM file to write merged result to")
+    public File OUTPUT;
```
lbergelson: path

src/main/java/picard/sam/AddOATag.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Argument(shortName = "OV", doc = "Whether or not to override any existing OA tag values, otherwise the current alignment will be appended to the tag.", optional = true)
+    public Boolean OVERWRITE_TAG = true;
+
+    @Argument(shortName = "L", doc = "An interval list file for which records that overlap will have the OA tag added", optional = true)
```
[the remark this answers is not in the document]
lbergelson: I'm curious what this is for  When do you want to add it to a specific region of the bam?

src/main/java/picard/sam/AddOATag.java
lbergelson: You can't use path directly really, you have to go through string -> path but you get the idea...

src/test/java/picard/sam/AddOATagTest.java
```diff
@@ -0,0 +1,110 @@
...
+        clpOutput.deleteOnExit();
+
+        runAddOATag(inputSam, clpOutput, intervalList);
+
+        validateOATag(clpOutput, truthSam);
+    }
+
+    private void runAddOATag(final File inputSam, final File output, final File intervalList) throws IOException {
+        final ArrayList<String> args = new ArrayList<String>(){
+            {
```
pshapiro4broad: Using `{{ .. add(...) ... }}` is a pretty obscure way to initialize an array. I don't think it's currently done in the picard code base. Instead I would write:
```
final List<String> args = new ArrayList<>(Arrays.asList("INPUT=" + inputSam, "OUTPUT=" + output));
```

src/test/java/picard/sam/AddOATagTest.java
```diff
@@ -0,0 +1,110 @@
...
+                add("INPUT=" + inputSam);
+                add("OUTPUT=" + output);
+            }
+        };
+        if (intervalList != null) {
+            args.add("INTERVAL_LIST=" + intervalList);
+        }
+        AddOATag addOATag = new AddOATag();
+        Assert.assertEquals(addOATag.instanceMain(args.toArray(new String[args.size()])), 0, "Running addOATag did not succeed");
```
pshapiro4broad: In general it's better to pass an empty array in instead of a preallocated array
```suggestion
        Assert.assertEquals(addOATag.instanceMain(args.toArray(new String[0])), 0, "Running addOATag did not succeed");
```

src/test/java/picard/sam/AddOATagTest.java
```diff
@@ -0,0 +1,110 @@
...
+    public void testIntervalList(final File inputSam, final File intervalList, final File truthSam) throws IOException {
+        final File clpOutput = File.createTempFile("AddOATag", ".bam");
+        clpOutput.deleteOnExit();
+
+        runAddOATag(inputSam, clpOutput, intervalList);
+
+        validateOATag(clpOutput, truthSam);
+    }
+
+    private void runAddOATag(final File inputSam, final File output, final File intervalList) throws IOException {
```
pshapiro4broad: can be static
```suggestion
    private static void runAddOATag(final File inputSam, final File output, final File intervalList) throws IOException {
```

src/test/java/picard/sam/AddOATagTest.java
```diff
@@ -0,0 +1,110 @@
...
+            }
+        };
+        if (intervalList != null) {
+            args.add("INTERVAL_LIST=" + intervalList);
+        }
+        AddOATag addOATag = new AddOATag();
+        Assert.assertEquals(addOATag.instanceMain(args.toArray(new String[args.size()])), 0, "Running addOATag did not succeed");
+    }
+
+    private void validateOATag(final File testSam, final File truthSam) {
```
pshapiro4broad: can be static
```suggestion
    private static void validateOATag(final File testSam, final File truthSam) {
```

src/test/java/picard/sam/AddOATagTest.java
```diff
@@ -0,0 +1,110 @@
...
+        final ArrayList<String> truthOAValues = new ArrayList<>();
+        final ArrayList<String> testOAValues = new ArrayList<>();
+
+        SAMRecordIterator iterator = SamReaderFactory.makeDefault().open(truthSam).iterator();
+        while (iterator.hasNext()){
```
pshapiro4broad: missing space
```suggestion
        while (iterator.hasNext()) {
```

src/test/java/picard/sam/AddOATagTest.java
```diff
@@ -0,0 +1,110 @@
...
+            SAMRecord rec = iterator.next();
+            truthOAValues.add(rec.getStringAttribute(AddOATag.OA));
+        }
+
+        iterator = SamReaderFactory.makeDefault().open(testSam).iterator();
+        while (iterator.hasNext()){
```
pshapiro4broad:
```suggestion
        while (iterator.hasNext()) {
```

meganshand:
@pshapiro4broad and @lbergelson back to you!

src/main/java/picard/sam/AddOATag.java
```diff
@@ -0,0 +1,131 @@
...
+        return detector;
+    }
+
+    // format OA tag string according to the spec
+    //TODO: Move this to htsjdk once https://github.com/samtools/hts-specs/pull/193 is merged
+    private void setOATag(SAMRecord rec) {
+        if (rec.getReferenceName().contains(",")) {
+            throw new PicardException(String.format("Reference name for record %s contains a comma character.", rec.getReadName()));
+        }
+        final String oAValue;
```
yfarjoun: This is unreadable (to me)...if we are sticking to lower-case first letter, this shoudl be `oaValue`...sorry.

PR #1204, opened 2018-07-26
Warning message about upcoming command line syntax transition.

cmnbroad:
Once we determine a target transition date, we'll need to decide at what point we want to include this. Proposed warning message about upcoming command line syntax transition. Displays an informational message with a link to a wiki page (under construction):

image

Edit: A couple of people have suggested that we do something like this warning, but it looks like it breaks PipedDataTest.testSortSam.

### DIFF
--- a/src/main/java/picard/cmdline/CommandLineProgram.java
+++ b/src/main/java/picard/cmdline/CommandLineProgram.java
@@ -203,8 +203,9 @@ public abstract class CommandLineProgram {
             Log.getInstance(this.getClass()).info(
                     String.format("\n\n********** NOTE: In a future release, the command line syntax used by Picard will change, and the existing syntax \n" +
                             "********** will no longer be accepted. In the future release, the syntax for the command line arguments would be specified as:\n" +
-                            "**********\n**********\t %s\n**********\n" +
+                            "**********\n**********\t%s %s\n**********\n" +
                             "********** See https://github.com/broadinstitute/picard/wiki/Command-Line-Syntax-Transition-For-Users-(Pre-Transition) for more information.\n\n",
+                           this.getClass().getSimpleName(),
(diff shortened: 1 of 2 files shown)


src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -195,6 +195,18 @@ public int instanceMain(final String[] argv) {
...
+            // a link to a Picard wiki page with more detail. For now this is only an informational message letting
+            // users know a change is coming. In a future release of Picard, when the default parser is Barclay, we'll
+            // update this message to link to a different page describing the options for that release.
+            Log.getInstance(this.getClass()).info(
+                    String.format("\n\n********** NOTE: In a future release, the command line syntax used by Picard will change, and the existing syntax \n" +
+                            "********** will no longer be accepted. In the future release, the syntax for the command line arguments would be specified as:\n" +
+                            "**********\n**********\t %s\n**********\n" +
+                            "********** See https://github.com/broadinstitute/picard/wiki/Command-Line-Syntax-Transition-For-Users-(Pre-Transition) for more information.\n\n",
+                           Arrays.stream(CommandLineSyntaxTranslater.translatePicardStyleToPosixStyle(argv)).collect(Collectors.joining(" ")))
```
yfarjoun: Is there a way to add the CLP name to this message?

yfarjoun: Do you want this to pause for (say) 5 seconds to let the message sink in?

tfenne: If you're worried about people ignoring it, you could print it at the start and end of execution.  But I'm with @nh13 that a forced pause would be super-annoying.

kbergin:

PR #530, opened 2016-05-02
Adding a tool FixVcfHeader.

nh13:
This tool replaces or fixes up the VCF header. In the latter case, it
adds dummy header lines for INFO or FORMAT fields found in the records
but not in the input VCF.

PR #1205, opened 2018-08-01
FixVcfHeader can now add missing FILTER header lines.

nh13:
@yfarjoun care to take a peek since you looked at the original? https://github.com/broadinstitute/picard/pull/530


src/main/java/picard/vcf/FixVcfHeader.java
```diff
@@ -121,41 +114,54 @@ public static void main(final String[] args) {
...
             }
             else {
-                outHeader = new VCFHeader(headerReader.getFileHeader().getMetaDataInInputOrder(), reader.getFileHeader().getSampleNamesInOrder());
+                outHeader = new VCFHeader(inputHeader.getMetaDataInInputOrder(), existingHeader.getSampleNamesInOrder());
             }
             CloserUtil.close(headerReader);
+            outHeader.getFilterLines()
+                    .stream()
+                    .filter(filterHeaderLine -> !existingHeader.hasFilterLine(filterHeaderLine.getID()))
+                    .forEach(filterHeaderLine -> log.info("FILTER line found in HEADER will be added to OUTPUT: " + filterHeaderLine));
```
pshapiro4broad: The other cases log `headerLine.getID()` but this case logs `headerLine`. Is that difference intentional?

src/main/java/picard/vcf/FixVcfHeader.java
```diff
@@ -121,41 +114,54 @@ public static void main(final String[] args) {
...
             outHeader.getInfoHeaderLines()
                     .stream()
-                    .filter(infoHeaderLine -> !reader.getFileHeader().hasInfoLine(infoHeaderLine.getID()))
+                    .filter(infoHeaderLine -> !existingHeader.hasInfoLine(infoHeaderLine.getID()))
                     .forEach(infoHeaderLine -> log.info("INFO line found in HEADER will be added to OUTPUT: " + infoHeaderLine.getID()));
```
pshapiro4broad: It would be slightly more streamlined to convert the stream to ID, e.g.
```
outHeader.getInfoHeaderLines()
        .stream()
        .map(VCFCompoundHeaderLine::getID)
        .filter(id -> !reader.getFileHeader().hasInfoLine(id))
        .forEach(id -> log.info("INFO line found in HEADER will be added to OUTPUT: " + id));
```

src/main/java/picard/vcf/FixVcfHeader.java
```diff
@@ -121,41 +114,54 @@ public static void main(final String[] args) {
...
-                outHeader = headerReader.getFileHeader();
-                enforceSameSamples(reader.getFileHeader(), outHeader);
+                outHeader = inputHeader;
+                enforceSameSamples(existingHeader, outHeader);
             }
             else {
-                outHeader = new VCFHeader(headerReader.getFileHeader().getMetaDataInInputOrder(), reader.getFileHeader().getSampleNamesInOrder());
+                outHeader = new VCFHeader(inputHeader.getMetaDataInInputOrder(), existingHeader.getSampleNamesInOrder());
             }
             CloserUtil.close(headerReader);
```
pshapiro4broad: this code should use try-with-resources

src/main/java/picard/vcf/FixVcfHeader.java
```diff
@@ -121,41 +114,54 @@ public static void main(final String[] args) {
...
                     .stream()
-                    .filter(formatHeaderLine -> !reader.getFileHeader().hasInfoLine(formatHeaderLine.getID()))
+                    .filter(formatHeaderLine -> !existingHeader.hasInfoLine(formatHeaderLine.getID()))
                     .forEach(formatHeaderLine -> log.info("FORMAT line found in HEADER will be added to OUTPUT: " + formatHeaderLine.getID()));
         }
         else { // read the input
             final ProgressLogger progress = new ProgressLogger(log, 1000000, "read");
-            final VCFFileReader in = new VCFFileReader(INPUT, false);
-            final VCFHeader inHeader = in.getFileHeader();
+            final VCFFileReader in        = new VCFFileReader(INPUT, false);
```
pshapiro4broad: `in` is never closed

src/main/java/picard/vcf/FixVcfHeader.java
nh13: Done

src/main/java/picard/vcf/FixVcfHeader.java
nh13: I blame IntelliJ for not finding this.  Thanks!

src/main/java/picard/vcf/FixVcfHeader.java
nh13: Done, but I decided to use `existingHeader` instead of `reader.getFileHeader`, since I have it on hand.

src/main/java/picard/vcf/FixVcfHeader.java
nh13: Closed on line 176/192 below but I have changed this block to be a try-with-resources

nh13:
@yfarjoun fixing tests, though tests pass on my machine, good times.

nh13:
@pshapiro4broad @yfarjoun I could use some help.  I have run the tests locally and they pass.  I built the picard JAR and ran the command line invocation myself, and it does what I expect, which is `testMissingFilterNotInFirstNRecords` should throw an exception because writing a record that has a filter field that isn't in the header is invalid.  Any ideas?

nh13:
@cmnbroad thank-you kindly for looking into it, and that's surely the issue.  Once tests pass, I'll merge.  Thanks again!

PR #1206, opened 2018-08-04
Yf fix scatter with overflow mode

yfarjoun:
### Description

While refactoring the scattering code in a different branch I noticed this bug. it seems that sometime OVERFLOW mode "forgets" to overflow.

fixed and  added some tests. 

Not sure if there's a point in merging this, but it can serve as an explanation for my other PR (coming up)

### DIFF
--- a/src/test/java/picard/util/IntervalListScattererTest.java
+++ b/src/test/java/picard/util/IntervalListScattererTest.java
@@ -158,6 +158,19 @@ public class IntervalListScattererTest {
         ));
 
         testcases.add(new Testcase(
+                LIST_TO_SCATTER, 2, IntervalListScatterer.Mode.BALANCING_WITHOUT_INTERVAL_SUBDIVISION_WITH_OVERFLOW,
+                Arrays.asList(
+                        composeIntervalList(LIST_TO_SCATTER, "1",
+                                30000, 30098,
+                                30100, 30150
+                        ),
+                        composeIntervalList(LIST_TO_SCATTER, "1",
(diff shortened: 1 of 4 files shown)


src/test/java/picard/util/IntervalListScattererTest.java
```diff
@@ -40,9 +40,11 @@
  */
 public class IntervalListScattererTest {
     private static final IntervalList LIST_TO_SCATTER;
+    private static final IntervalList LIST_TO_SCATTER_WITH_OVERFLOW;
 
     static {
         LIST_TO_SCATTER = IntervalList.fromFile(new File("testdata/picard/util/scatterable.interval_list"));
+        LIST_TO_SCATTER_WITH_OVERFLOW = IntervalList.fromFile(new File("testdata/picard/util/scatterable_with_overflow.interval_list"));
```
pshapiro4broad: Does it not work to initialize this field at its declaration, e.g. 
```
private static final IntervalList LIST_TO_SCATTER_WITH_OVERFLOW =
    IntervalList.fromFile(new File("testdata/picard/util/scatterable_with_overflow.interval_list"));
```

pshapiro4broad:
While looking at this code I noticed, on line 57:
```
```
which can be written instead as 
```
final int widestIntervalLength = Collections.max(uniquedList.getIntervals(), Comparator.comparingInt(Interval::length)).length();
```

yfarjoun:
I only added test-cases, so I'm taking this as a thumbs up.

PR #1207, opened 2018-08-06
Fix output file naming in SplitSamByNumberOfReads

cboursnell:
### Description
Moved the OUT_PREFIX String out of the DecimalFormat and into the File name creation. This means if the user adds any of the DecimalFormat Special Pattern Characters in the prefix it won't mess up the pattern. 

For example
If OUT_PREFIX contains any zeroes then the file name gets messed up

if OUT_PREFIX is "test01"
then output filename is "test00001.1_bam"
with this fix filename is "test01_0001.bam"

The default of "shard" as the prefix means that this behaviour wasn't covered by tests.

#### Content
- [no] Added or modified tests to cover changes and any new functionality
sorry I don't know how to write tests and run them. 
- [n/a] Edited the README / documentation (if applicable)
- [probably] All tests passing on Travis

#### Review

### DIFF
--- a/src/main/java/picard/sam/SplitSamByNumberOfReads.java
+++ b/src/main/java/picard/sam/SplitSamByNumberOfReads.java
@@ -124,14 +124,14 @@ public class SplitSamByNumberOfReads extends CommandLineProgram {
         final long totalReads = TOTAL_READS_IN_INPUT == 0 ? firstPassProgress.getCount() : TOTAL_READS_IN_INPUT;
 
         final SAMFileWriterFactory writerFactory = new SAMFileWriterFactory();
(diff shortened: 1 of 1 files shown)


meganshand:
@cboursnell Thanks for catching this and submitting a fix! It looks like the current tests are not passing because the same change you applied to line 134 also needs to go on line 140. 

If you'd like I can take over this issue and add a new test/fix the current tests. 

Or if you'd like to see this PR through you could add a new test to the file https://github.com/broadinstitute/picard/blob/master/src/test/java/picard/sam/SplitSamByNumberOfReadsTest.java.

Thanks again for the PR, let me know what you'd like to do.

yfarjoun:
@meganshand it looks like there's no response from the OP...could you please take over this one?

meganshand:
I opened a new PR with these changes and a test (#1222). Thanks again @cboursnell.

PR #1208, opened 2018-08-06
Add Scatter by IntervalCount (IntervalListTools)

yfarjoun:
### Description

In order to scatter interval lists for tools that operate on intervals (rather than bases) we need a new scatter mode in IntervalListTools. For example, we would like to use this for scattering CNV calling.

The IntervalScatterer class had gotten to convoluted, that I refactored it and then added another implementation that scatters by interval count.


yfarjoun:
@samuelklee can you take a look? This is for you.

samuelklee:
@yfarjoun Thanks for putting this together.  I might be missing something, but is it possible to run the tool by specifying the number of intervals per (complete) scatter block, rather than the number of scatter blocks?

src/main/java/picard/util/IntervalList/IntervalListScatterer.java
```diff
@@ -0,0 +1,132 @@
...
+
+        // if last list isn't empty, add it to return value.
+        if (!runningIntervalList.getIntervals().isEmpty()) {
+            accumulatedIntervalLists.add(runningIntervalList);
+        }
+
+        return accumulatedIntervalLists;
+    }
+
+    default IntervalList preprocessIntervalList(final IntervalList inputList) {
```
samuelklee: Document these methods.

src/main/java/picard/util/IntervalList/IntervalListScattererByInterval.java
```diff
@@ -0,0 +1,63 @@
...
+import htsjdk.samtools.util.IntervalList;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Scatters intervalLists by Interval so that resulting interval lists have same number of intervals in them.
+ * Final interval can have up to the scatter number extra intervals.
+ */
+public class IntervalListScattererByInterval implements IntervalListScatterer {
```
samuelklee: IntervalListScattererByInterval -> IntervalListScattererByIntervalCount seems more consistent

src/main/java/picard/util/IntervalList/IntervalListScattererByInterval.java
```diff
@@ -0,0 +1,63 @@
...
+
+import htsjdk.samtools.util.CollectionUtil;
+import htsjdk.samtools.util.Interval;
+import htsjdk.samtools.util.IntervalList;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Scatters intervalLists by Interval so that resulting interval lists have same number of intervals in them.
```
samuelklee: Interval -> interval count

samuelklee: have same number -> have the same number

src/main/java/picard/util/IntervalList/IntervalListScattererByInterval.java
```diff
@@ -0,0 +1,63 @@
...
+ * Final interval can have up to the scatter number extra intervals.
```
samuelklee: Perhaps something along the lines of: Final interval can have up to the scatter number extra intervals. -> An exception is made for the final interval list, which will always contain at least this number of intervals, but may also contain the remainder.

src/main/java/picard/util/IntervalList/IntervalListScatterMode.java
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * A simple scatter approach in which all output intervals have size equal to the total base count of the source list divide by the
+     * scatter count (except for possible variance in the final interval list).
+     */
+    INTERVAL_SUBDIVISION(IntervalListScattererWithSubdivision::new, "Scatter the interval list into similarly sized interval lists " +
+            "(by base count), breaking up intervals as needed.") {
+
+    },
+    /**
+     * A scatter approach that differs from {@link IntervalListScatterMode#INTERVAL_SUBDIVISION} in a few ways.
```
pshapiro4broad: You don't need to qualify the member in the javadoc link because it's in scope. e.g. you can write `{@link #INTERVAL_SUBDIVISION}`.

Also the java coding guidelines for javadoc recommend not to link every reference in a javadoc. See `Use in-line links economically` in https://www.oracle.com/technetwork/java/javase/tech/index-137868.html

src/main/java/picard/util/IntervalList/IntervalListScattererByBaseCount.java
```diff
@@ -0,0 +1,51 @@
...
+
+package picard.util.IntervalList;
+
+import htsjdk.samtools.util.Interval;
+import htsjdk.samtools.util.IntervalList;
+
+/**
+ * a Baseclass for scatterers that scatter by uniqued base count.
+ */
+abstract public class IntervalListScattererByBaseCount implements IntervalListScatterer {
```
pshapiro4broad: Instead of `abstract public`, the conventional order is `public abstract`. You can make the IDE highlight and fix this for you by enabling Editor/Inspections Java/Code style issues/Missorted modifiers

src/main/java/picard/util/IntervalList/IntervalListScattererByIntervalCount.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    public long listWeight(final IntervalList intervalList) {
+        return intervalList.size();
+    }
+
+    @Override
+    public List<Interval> takeSome(final Interval interval, final long idealSplitWeight, final long currentSize, final double projectSizeOfRemaining) {
+        final long amount = idealSplitWeight - currentSize;
+        if (amount > 0) {
+            return CollectionUtil.makeList(interval, null);
```
pshapiro4broad: Why not use `Arrays.asList()` here? It looks like the result of this method is never modified.

src/main/java/picard/util/IntervalList/IntervalListScattererWithoutSubdivision.java
```diff
@@ -0,0 +1,67 @@
...
+
+    protected boolean shouldIncludeInterval(long idealSplitWeight, double projectedSizeOfRemaining, long projectedSize) {
+        return projectedSize <= idealSplitWeight;
+    }
+
+    @Override
+    public int deduceIdealSplitWeight(final IntervalList intervalList, final int nCount) {
+        final int splitWidth = super.deduceIdealSplitWeight(intervalList, nCount);
+        final int widestIntervalLength = intervalList.getIntervals().stream()
+                .map(Interval::length)
```
pshapiro4broad: Slight improvement, use `mapToInt()`, e.g.
```
        final int widestIntervalLength = intervalList.getIntervals().stream()
                .mapToInt(Interval::length)
                .max()
                .orElse(1);
```

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -13,14 +37,16 @@
...
+import picard.util.IntervalList.IntervalListScatterer;
 
 import java.io.File;
 import java.text.DecimalFormat;
 import java.util.*;
 
 /**
  * Performs various {@link IntervalList} manipulations.
- *
+ * <p>
```
pshapiro4broad: You don't need `<p>` before any `<hX>` tag, unless you want an extra blank line. There should already be enough vertical spacing. You can use control-J in the IDE to see what it will look like. Unless you're looking at the html output and it needs this to look OK?

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -70,7 +96,7 @@
  *       I=input_2.interval_list \\
```
pshapiro4broad: This code could avoid quoting like this by wrapping the code in `{@code}`. e.g.
```
 * <pre>
 * {@code
 * java -jar picard.jar IntervalListTools \
 *       ACTION=CONCAT \
 *       I=input.interval_list \
 *       I=input_2.interval_list \
 *       O=new.interval_list
 * }
 * </pre>
```
This is just a good trick to remember in general for code blocks, not really required here though.

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -236,13 +261,22 @@
             " broken up at integer multiples of this value. Set to 0 to NOT break up intervals.", optional = true)
     public int BREAK_BANDS_AT_MULTIPLES_OF = 0;
 
-    @Argument(shortName = "M", doc = "Selects between various ways in which scattering of the interval-list can happen.")
-    public IntervalListScatterer.Mode SUBDIVISION_MODE = IntervalListScatterer.Mode.INTERVAL_SUBDIVISION;
+    @Argument(shortName = "M", doc = "Selects the way in which scattering of the interval-list will happen.")
```
pshapiro4broad: Not sure what the style guide is for picard docs is. Maybe `The mode used to scatter the interval list.` would be less awkward.

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -396,28 +424,38 @@ protected int doWork() {
...
-            totalUniqueBaseCount += finalInterval.getUniqueBaseCount();
-            intervalCount += finalInterval.size();    
+            totalBaseCount += finalInterval.getBaseCount();
+            intervalCount += finalInterval.size();
         }
 
-        LOG.info("Produced " + intervalCount + " intervals totalling " + totalUniqueBaseCount + " unique bases.");
-        
+        LOG.info("Produced " + intervalCount + " intervals totalling " + totalBaseCount + " bases.");
+        switch (OUTPUT_VALUE) {
```
pshapiro4broad: You can avoid the switch and `throw` by putting this in the enum itself. It's a little more code but it's nice to avoid a switch, IMO.
```
    enum Output {
        NONE() {
            void output(long totalBaseCount, long intervalCount) {
            }
        },
        BASES {
            void output(long totalBaseCount, long intervalCount) {
                System.out.println(totalBaseCount);
            }
        },
        INTERVALS {
            void output(long totalBaseCount, long intervalCount) {
                System.out.println(intervalCount);
            }
        };
        abstract void output(long totalBaseCount, long intervalCount);
    }
...
        OUTPUT_VALUE.output(totalBaseCount, intervalCount);
```

src/main/java/picard/util/IntervalList/IntervalListScattererByIntervalCount.java
yfarjoun: I'm not sure I understand the difference between using makeList versus asList...

src/main/java/picard/util/IntervalListTools.java
yfarjoun: the ide added it automagically for me when I asked it to rearrange the lines...

src/main/java/picard/util/IntervalList/IntervalListScattererByIntervalCount.java
pshapiro4broad: The basic difference is that `CollectionUtil.makeList()` returns a modifiable list and `Arrays.asList()` returns an unmodifiable one.

Also, `Arrays` is part of the Java API, and `CollectionUtil` is part of htsjdk. In general I think it's better to use a Java API when possible.

src/main/java/picard/util/IntervalListTools.java
pshapiro4broad: I wonder what settings you're using that's causing that behavior.. when I try reformatting the copy in /master, the only changes I get are adding `<p>` on lines 43 and 54, which are the only places where it's actually needed. Maybe this is a bug in the IDE that's fixed in a newer version?

src/main/java/picard/util/IntervalListTools.java
yfarjoun: nice. we should show this to the comms team...

yfarjoun:
@samuelklee I added SCATTER_CONTENT argument as you asked.

yfarjoun:
🎉

samuelklee:
Thanks again!  Forgot to explicitly approve changes, hope I wasn't holding anything up!

src/main/java/picard/util/IntervalListTools.java
sooheelee: So use `{@code ... }` and remove the extra`\` per line?

pshapiro4broad: Using `@code` has other benefits as it also allows `<` `>` and `&` in javadoc without quoting.

Issue #258, opened 2015-08-12
ValidateSamFile should warn that NM validation cannot be done without REFERENCE_SEQUENCE


jmthibault79:
I was unable to observe an error in NM tag calculation when running ValidateSamFile without setting the REFERENCE_SEQUENCE.  Adding that parameter solved my problem.

Was NM tag validation done differently, or not performed at all in the first case?  Either way, a warning or other notification is necessary to inform the user of this situation.

yfarjoun:
NM validation cannot be performed without the reference. I guess an easy fix would be to add a warning if reference is null.

yfarjoun:
fixed by https://github.com/broadinstitute/picard/pull/1220

PR #1220, opened 2018-08-28
Fix to ValidateSamFile to warn when NM validation occurs without reference (#258)

kvinter1:
This addresses #258 https://github.com/broadinstitute/picard/issues/258.

Gives warning if REFERENCE_SEQUENCE is null.

### DIFF
--- a/src/main/java/picard/sam/ValidateSamFile.java
+++ b/src/main/java/picard/sam/ValidateSamFile.java
@@ -206,7 +206,7 @@ public class ValidateSamFile extends CommandLineProgram {
                 reference = ReferenceSequenceFileFactory.getReferenceSequenceFile(REFERENCE_SEQUENCE);
 
             } else {
-                log.warn("NM validation cannot be performed without the reference");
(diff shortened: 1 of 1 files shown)


src/main/java/picard/sam/ValidateSamFile.java
```diff
@@ -205,6 +205,8 @@ protected int doWork() {
                 IOUtil.assertFileIsReadable(REFERENCE_SEQUENCE);
                 reference = ReferenceSequenceFileFactory.getReferenceSequenceFile(REFERENCE_SEQUENCE);
 
+            } else {
+                log.warn("NM validation cannot be performed without the reference");
```
yfarjoun: is that the only validation that doesn't occur? 

Also, please add a period in the end, and make it clear that all other validations will still happen.

kvinter1: I believe that is the only validation that needs a warning if a reference file isn't provided.

I've fixed the warning message based on your comments.

PR #1245, opened 2018-10-30
Fixed order dependance bug in graphUtils.cluster() algorithm. 

jamesemery:
The documentation claims that all nodes in the same cluster will have the same representative read as output. As it turns out this would not have been the case for certain graphs with many branching edges and pathological input ordering. I have added a test that demonstrates this problem on the old branch and added an extra step to resolve transitive node parenthood so that we are consistent with the stated behavior.

### DIFF
--- a/src/main/java/picard/util/GraphUtils.java
+++ b/src/main/java/picard/util/GraphUtils.java
@@ -37,6 +37,11 @@ public class GraphUtils {
         final private List<Node> nodes;
         final private List<List<Integer>> neighbors;
 
+        // A public getter for the nodes in the list for iteration, note that this is no
+        public List<Node> getNodes() {
+            return Collections.unmodifiableList(nodes);
+        }
+
         /**
          * returns the cluster map of connected components
          *
@@ -50,7 +55,7 @@ public class GraphUtils {
                     neighbors.get(i).stream().forEach(j -> joinNodes(cluster, j, i)));
 
             // Must call findRepNode() here in case nodes are orphaned and become only transitively equal to roots in their cluster.
-            return nodes.stream().collect(Collectors.toMap(n -> n, n -> findRepNode(cluster, cluster[nodes.indexOf(n)])));
+            return nodes.stream().collect(Collectors.toMap(n -> n, n -> findRepNode(cluster, nodes.indexOf(n))));
 
         }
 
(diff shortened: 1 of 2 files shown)


jamesemery:
@yfarjoun Can I get a review on this? The results look promising thus far, I think this will finally fix the correctness for MarkDuplicatesSpark optical duplicate marking.

src/main/java/picard/util/GraphUtils.java
```diff
@@ -49,7 +49,8 @@
             IntStream.range(0, neighbors.size()).forEach(i ->
                     neighbors.get(i).stream().forEach(j -> joinNodes(cluster, j, i)));
 
-            return nodes.stream().collect(Collectors.toMap(n -> n, n -> cluster[nodes.indexOf(n)]));
+            // Must call findRepNode() here in case nodes are orphaned and become only transitively equal to roots in their cluster.
+            return nodes.stream().collect(Collectors.toMap(n -> n, n -> findRepNode(cluster, cluster[nodes.indexOf(n)])));
```
yfarjoun: given that you're calling findRepNode, can you drop the cluster[.] call? 
```suggestion
            return nodes.stream().collect(Collectors.toMap(n -> n, n -> findRepNode(cluster, nodes.indexOf(n))));
```

src/test/java/picard/util/GraphUtilsTest.java
```diff
@@ -73,4 +73,39 @@ public void secondTest() {
...
+        graph.addEdge(e, f);
+        graph.addEdge(g, a);
+
+        final Map<Integer, Integer> clusters = graph.cluster();
+
+        // 9 nodes
+        assertEquals(clusters.size(), 8);
+
+        // Asserting that this connected graph has every node pointing to the same place
+        final Integer repCluster = clusters.get(a);
```
yfarjoun: hmm. perhaps better to add `getNodes` and `getEdges` methods to Graph and then use `getNodes().stream()` instead of the clunky `range(0,7)`?

src/main/java/picard/util/GraphUtils.java
```diff
@@ -37,6 +37,11 @@
         final private List<Node> nodes;
         final private List<List<Integer>> neighbors;
 
+        // A public getter for the nodes in the list for iteration, note that this is no
```
yfarjoun: complete sentence in the comment...

kbergin:

PR #1248, opened 2018-11-02
Fixes off by one error in trimDistribution that causes CollectHsMetrics to crash on edge case

fleharty:
### Description

trimDistribution() in TheoreticalSensitivity an off by one bug that can cause tools that use TheoreticalSensitivity to crash.  This bug-fix resolve this.
----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/analysis/TheoreticalSensitivity.java
+++ b/src/main/java/picard/analysis/TheoreticalSensitivity.java
@@ -244,6 +244,11 @@ public class TheoreticalSensitivity {
      * @return Theoretical sensitivity for the given arguments at a constant depth.
      */
     public static double sensitivityAtConstantDepth(final int depth, final Histogram<Integer> qualityHistogram, final double logOddsThreshold, final int sampleSize, final double alleleFraction, final long randomSeed) {
+        // If the depth is 0 at a particular locus, the sensitivity is trivially 0.0.
+        if (depth == 0) {
+            return 0.0;
+        }
+
         final RouletteWheel qualityRW = new RouletteWheel(trimDistribution(normalizeHistogram(qualityHistogram)));
         final Random randomNumberGenerator = new Random(randomSeed);
         final RandomGenerator rg = new Well19937c(randomSeed);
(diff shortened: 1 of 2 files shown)


src/main/java/picard/analysis/TheoreticalSensitivity.java
```diff
@@ -354,9 +359,13 @@ public static double theoreticalSensitivity(final Histogram<Integer> depthHistog
      * @return Distribution of base qualities removing any trailing zeros
      */
      static double[] trimDistribution(final double[] distribution) {
-         int endOfDistribution = distribution.length - 1;
-         while(distribution[endOfDistribution] == 0) {
+         int endOfDistribution = distribution.length;
+         if (distribution.length == 0) {
+             return distribution;
+         }
+         while(distribution[endOfDistribution - 1] == 0) {
```
yfarjoun: what if the distribution is empty (as in all the entries are zero) should you protect against that as well?

src/test/java/picard/analysis/TheoreticalSensitivityTest.java
```diff
@@ -268,7 +268,7 @@ public void testSensitivityAtConstantDepth(final double expected, final File met
         // keeping the compute time of the tests short.
         return new Object[][]{
                 {0.90, wgsMetricsFile, 0.5, 400},
-                {0.77, wgsMetricsFile, 0.3, 400},
+                {0.78, wgsMetricsFile, 0.3, 400},
```
yfarjoun: why did this change?

fleharty: The reason is because it was on the edge between .77 and .78.  The bug-fix causes the value to go up because it no longer trims the highest base quality score from the distribution.

It's an integration test so I didn't calculate this value independently.  In a way, that makes me wonder about just how valuable this particular test is, as well as the test that uses theoreticalSensitivityDataProvider in CollectTargetedMetricsTest.java.

src/main/java/picard/analysis/TheoreticalSensitivity.java
fleharty: I believe it already handled the case you describe, but I added some test cases to cover it and modified the comment to hopefully better describe what is happening.

PR #1259, opened 2018-12-27
Trim descriptions to shorten --list display width.

cmnbroad:
Picard part of https://github.com/broadinstitute/gatk/issues/5421.

Before:
screen shot 2018-12-27 at 9 04 18 am

After:
screen shot 2018-12-27 at 9 07 45 am

### DIFF
--- a/src/test/java/picard/cmdline/PicardCommandLineTest.java
+++ b/src/test/java/picard/cmdline/PicardCommandLineTest.java
@@ -4,10 +4,25 @@ import org.broadinstitute.barclay.argparser.CommandLineArgumentParser;
 import org.broadinstitute.barclay.argparser.CommandLineException;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
 import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 import java.util.*;
 
 public class PicardCommandLineTest {
+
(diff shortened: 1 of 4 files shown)


cmnbroad:
After some discussion on the GTAK part of this, we decided to add a test to ensure we don't exceed the length, rather than just silently truncating.

src/test/java/picard/cmdline/PicardCommandLineTest.java
```diff
@@ -63,4 +63,19 @@ public void testLaunchAllCommandLineProgramsWithBarclayParser() {
     public void testPrintUsage() {
         Assert.assertEquals(new PicardCommandLine().instanceMain(new String[]{"-h"}), 1);
     }
+
+    @Test
+    public void testCommandlineProgramPropertiesOneLineSummaryLength() {
+        // Find and test each command line tool to ensure that one line
+        // summaries don't exceed the maximum allowable length
+        PicardCommandLine.processAllCommandLinePrograms(
+                Collections.singletonList("picard"),
```
pshapiro4broad: It's not too bad, but it does seem wasteful that the tests generates the list of all picard CLPs three times here. If you generated it once in a `BeforeClass` you could use it in all three tests.

cmnbroad: Done. Consolidated these into a single `BeforeClass` discovery method.

yfarjoun:
why isn't this merged?

PR #1265, opened 2019-01-11
Determine arg parser to use.

jacarey:
### Description

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/cmdline/CommandLineProgram.java
+++ b/src/main/java/picard/cmdline/CommandLineProgram.java
@@ -52,7 +52,6 @@ import picard.cmdline.argumentcollections.OptionalReferenceArgumentCollection;
 import picard.cmdline.argumentcollections.ReferenceArgumentCollection;
 import picard.cmdline.argumentcollections.RequiredReferenceArgumentCollection;
 import picard.nio.PathProvider;
-import picard.util.PropertyUtils;
 
 import java.io.File;
 import java.net.InetAddress;
@@ -63,7 +62,6 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Properties;
 import java.util.stream.Collectors;
 
 /**
@@ -83,8 +81,6 @@ import java.util.stream.Collectors;
  *
  */
 public abstract class CommandLineProgram {
-    // Picard CmdLine properties file resource path, placed at this path by the gradle build script
-    private static String PICARD_CMDLINE_PROPERTIES_FILE = "picard/picardCmdLine.properties";
     private static String PROPERTY_USE_LEGACY_PARSER = "picard.useLegacyParser";
     private static String PROPERTY_CONVERT_LEGACY_COMMAND_LINE = "picard.convertCommandLine";
     private static Boolean useLegacyParser;
@@ -135,7 +131,7 @@ public abstract class CommandLineProgram {
 
     @ArgumentCollection(doc="Special Arguments that have meaning to the argument parsing system.  " +
                 "It is unlikely these will ever need to be accessed by the command line program")
-    public Object specialArgumentsCollection = useLegacyParser(getClass()) ?
+    public Object specialArgumentsCollection = useLegacyParser() ?
             new Object() : // legacy parser does not require these
             new SpecialArgumentsCollection();
 
@@ -342,7 +338,7 @@ public abstract class CommandLineProgram {
     */
     protected boolean parseArgs(final String[] argv) {
 
-        commandLineParser = determineArgParser(argv);
+        commandLineParser = getCommandLineParserForArgs(argv);
 
         boolean ret;
         try {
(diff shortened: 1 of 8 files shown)


src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -358,6 +358,18 @@ protected boolean parseArgs(final String[] argv) {
...
 
+    private CommandLineParser determineArgParser(String[] argv) {
+        // If there are args that don't start with dashes and contain equals then we use the old argument parser
+        boolean hasEqualsNoDash = Arrays.stream(argv).anyMatch(arg -> arg.contains("=") && !arg.startsWith("-"));
+
+        commandLineParser = (hasEqualsNoDash) ?
+                new LegacyCommandLineArgumentParser(this) :
+                new CommandLineArgumentParser(this,
+                        Collections.emptyList(),
+                        new HashSet<>(Collections.singleton(CommandLineParserOptions.APPEND_TO_COLLECTIONS)));
```
pshapiro4broad: `Collections.singletion()` already is a set (and `CommandLineArgumentParser` doesn't modify its arguments)  so the `new HashSet<>()` is unnecessary
```suggestion
                       Collections.singleton(CommandLineParserOptions.APPEND_TO_COLLECTIONS));
```

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -329,7 +329,7 @@ public int instanceMain(final String[] argv) {
     */
     protected boolean parseArgs(final String[] argv) {
 
-        commandLineParser = getCommandLineParser();
+        commandLineParser = determineArgParser(argv);
```
pshapiro4broad: The code here is a bit sloppy with regard to when `commandLineParser` is initialized. It looks like there are a few different call chains where `getCommandLineParser()` is called, so it's not immediately obvious to me that this line will occur before any other call to `getCommandLineParser()` occurs. It would be much better if we had control over the first place where this is initialized and used the field directly in all other places, to make it clear that this initialization must occur first.

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -358,6 +358,18 @@ protected boolean parseArgs(final String[] argv) {
         return true;
     }
 
+    private CommandLineParser determineArgParser(String[] argv) {
+        // If there are args that don't start with dashes and contain equals then we use the old argument parser
+        boolean hasEqualsNoDash = Arrays.stream(argv).anyMatch(arg -> arg.contains("=") && !arg.startsWith("-"));
+
+        commandLineParser = (hasEqualsNoDash) ?
```
pshapiro4broad: If the user specifies a different command line flavor than the one inferred, do you want to raise an error? Alternatively, should we ignore the flavor specified and only infer?

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -380,7 +392,7 @@ public CommandLineParser getCommandLineParser() {
             commandLineParser = useLegacyParser(getClass()) ?
                         new LegacyCommandLineArgumentParser(this) :
                         new CommandLineArgumentParser(this,
-                            Collections.EMPTY_LIST,
+                            Collections.emptyList(),
```
pshapiro4broad: If we end up keeping both this code and the new code above, it should be refactored (e.g. as a method that accepts `boolean useLegacyParser`) so the code is only written once.

yfarjoun: the plan is NOT to keep both parsers for long....

src/main/java/picard/cmdline/CommandLineProgram.java
yfarjoun: I think that if the user specifies then no inference should occur.

src/main/java/picard/cmdline/CommandLineProgram.java
cmnbroad: Per @pshapiro4broad comment above, this change should be integrated into `getCommandLineParser`.

yfarjoun:
is there a reason this isn't being merged?

yfarjoun:
@jacarey is this on your radar?

yfarjoun:
@cmnbroad can you commandeer this PR?

yfarjoun:
@gbggrant can someone _please_ move this so that we can get to a unified parser ?

cmnbroad:
I'm updating, rebasing, resolving conflicts and testing locally. Once tests pass it will need to be tested with GATK as well.

cmnbroad:
Changes in this PR:
- use the legacy parser when the args look legacy-ish
- otherwise, use the Barclay parser, including when no args have been provided (without this change, the command line help that gets displayed when no args are specified, or when `-h` is specified, would be legacy syntax rather than Barclay syntax)
- remove the barclayShadowJar artifact (we never published it anyway) and corresponding properties templates
- fix the picardDoc task to include the compileRuntimeClassPath so the doclet process can resolve googleNio classes

Finally, note that the Picard source still contains embedded usage examples that show legacy syntax. These are displayed in the command line help, and in the online doc - see for example the VcfFormatConverter "Usage example" in the [GATK doc](https://software.broadinstitute.org/gatk/documentation/tooldocs/current/picard_vcf_VcfFormatConverter.php). If the picardDoc output is published, these really should be updated.

kbergin:
Hi there! Can I ask how things are going on this PR? Is it blocked by something?

cmnbroad:
I think @gbggrant and @jessicaway wanted to do some testing with it (Zamboni ?), which is probably a good idea. Otherwise its ready from my perspective.

kbergin:
Got it, thank you Chris! I'll make sure we get to that as soon as we can.

droazen:
@kbergin @gbggrant @jessicaway Any news on when the additional testing on this branch will occur? It's been a couple of months, so I wanted to check in. It looks like this branch is otherwise approved for merge by @cmnbroad.

nikellepetrillo:
@cmnbroad i think this needs your approval so we can merge. Let me know if there is anything else you need! Thanks

yfarjoun:
🎉

PR #1266, opened 2019-01-12
Liftover all indels in flipped strand

yfarjoun:
As it turns out, the present liftover code is capable of lifting over multi-allelic indels and mixed variants, even in regions of the genome that are reverse-complemented during liftover. The was code that restricted the liftover to snps or biallelic indels (for RevComp'ed regions) but it was unneeded. 

This PR removes these restrictions and as a result more variants will be lifted over.


yfarjoun:
I'll take it. :-)

PR #1281, opened 2019-02-13
demultiplex only when we should

ApChagi:
### Description

Currently, IlluminaBasecallsToSam always demultiplexes. We should only demultiplex when the read structure indicates we should.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/illumina/IlluminaBasecallsToSam.java
+++ b/src/main/java/picard/illumina/IlluminaBasecallsToSam.java
@@ -305,17 +305,18 @@ public class IlluminaBasecallsToSam extends CommandLineProgram {
             adapters.add(new CustomAdapterPair(FIVE_PRIME_ADAPTER, THREE_PRIME_ADAPTER));
         }
 
+        final boolean demultiplex = !readStructure.sampleBarcodes.isEmpty();
         if (IlluminaFileUtil.hasCbcls(BASECALLS_DIR, LANE)) {
             if (BARCODES_DIR == null) BARCODES_DIR = BASECALLS_DIR;
             basecallsConverter = new NewIlluminaBasecallsConverter<>(BASECALLS_DIR, BARCODES_DIR, LANE, readStructure,
-                    barcodeSamWriterMap, true, Math.max(1, MAX_READS_IN_RAM_PER_TILE / numOutputRecords),
+                    barcodeSamWriterMap, demultiplex, Math.max(1, MAX_READS_IN_RAM_PER_TILE / numOutputRecords),
                     TMP_DIR, NUM_PROCESSORS,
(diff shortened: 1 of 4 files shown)


src/main/java/picard/illumina/IlluminaBasecallsToSam.java
```diff
@@ -305,17 +305,18 @@ private void initialize() {
             adapters.add(new CustomAdapterPair(FIVE_PRIME_ADAPTER, THREE_PRIME_ADAPTER));
         }
 
+        final boolean demultiplex = !readStructure.sampleBarcodes.isEmpty();
```
pshapiro4broad: It looks like this expression is used elsewhere, maybe it should be a method on `ReadStructure`, e.g. `hasSampleBarcode()`?

src/main/java/picard/illumina/IlluminaBasecallsToSam.java
```diff
@@ -305,17 +305,18 @@ private void initialize() {
...
-                    barcodeSamWriterMap, true, Math.max(1, MAX_READS_IN_RAM_PER_TILE / numOutputRecords),
+                    barcodeSamWriterMap, demultiplex, Math.max(1, MAX_READS_IN_RAM_PER_TILE / numOutputRecords),
                     TMP_DIR, NUM_PROCESSORS,
                     FIRST_TILE, TILE_LIMIT, new QueryNameComparator(),
                     new Codec(numOutputRecords),
                     SAMRecordsForCluster.class, bclQualityEvaluationStrategy, IGNORE_UNEXPECTED_BARCODES);
         } else {
             basecallsConverter = new IlluminaBasecallsConverter<>(BASECALLS_DIR, BARCODES_DIR, LANE, readStructure,
-                    barcodeSamWriterMap, true, MAX_READS_IN_RAM_PER_TILE / numOutputRecords, TMP_DIR, NUM_PROCESSORS, FORCE_GC,
+                    barcodeSamWriterMap, demultiplex, MAX_READS_IN_RAM_PER_TILE / numOutputRecords, TMP_DIR, NUM_PROCESSORS, FORCE_GC,
```
pshapiro4broad: Anytime I see two parameters being passed to a method, where one can be derived from the other, I wonder if both are necessary. E.g. you're passing both `readStructure` and `!readStructure.sampleBarcodes.isEmpty()`. In the other case where this is called, will the value of `demultplex` follow the value that's computed here?

ApChagi: I think the idea here is that there may be a use case for the IlluminaBasecallsConverter where there are barcodes in the readstructure, but the caller does not want to demultiplex?

tbl3rd:
Thank you, Charlie and Phil.

PR #1284, opened 2019-02-20
Updated the documenation for SortSam where it was somewhat misleading

### LINKED ISSUE #1283: SortSam documentation is misleading
I've noticed that in the tool documentation for SortSam the following line is present:
```
For queryname-sorted alignments, all alignments are grouped using the queryname field but the alignments are not necessarily sorted within these groups. Reads having the same queryname are derived from the same template.
```
This seems to be wrong however, as it looks to me that SortSam passes the SortOrder to the writer which ultimately ends up instantiating a `SAMRecordQueryNameComparator` object and invoking the compare() function. While the first criteria for comparison is indeed `compareReadNames()`, there are many other criteria it uses after that (important ones being strandedness followed by primary/secondary). This line should be amended.

jamesemery:
Fixes #1283

### DIFF
--- a/src/main/java/picard/sam/SortSam.java
+++ b/src/main/java/picard/sam/SortSam.java
@@ -55,7 +55,8 @@ import java.io.File;
  * Alignments within these subgroups are secondarily sorted using the left-most mapping position of the read (POS).
  * Subsequent to this sorting scheme, alignments are listed arbitrarily.</p> For queryname-sorted alignments, the tool
  * orders records deterministically by queryname field followed by record strand orientation flag, primary record flag, and secondary
- * alignment flag. (See {@link htsjdk.samtools.SAMRecordQueryNameComparator#compare(SAMRecord, SAMRecord)}} for details)
+ * alignment flag. (See {@link htsjdk.samtools.SAMRecordQueryNameComparator#compare(SAMRecord, SAMRecord)}} for details).
+ * This ordering may change in the future.
  *
  * <h3> Inputs</h3>
  * <ul>
@@ -88,7 +89,7 @@ public class SortSam extends CommandLineProgram {
             "reference sequence dictionary (@SQ tag).  Alignments within these subgroups are secondarily sorted using the left-most mapping " +
             "position of the read (POS).  Subsequent to this sorting scheme, alignments are listed arbitrarily.</p>" +
             "<p> For queryname-sorted alignments, the tool orders records deterministically by queryname field followed by " +
-            "record strand orientation flag, primary record flag, and secondary alignment flag.</p>" +
+            "record strand orientation flag, primary record flag, and secondary alignment flag. This ordering may change in future versions. </p>" +
             "<hr /><hr />" +
             "<h4>Usage example:</h4>" +
             "<pre>" +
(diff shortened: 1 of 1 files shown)


src/main/java/picard/sam/SortSam.java
```diff
@@ -54,8 +54,8 @@
  * field using the reference sequence dictionary tag labeled SQ.
  * Alignments within these subgroups are secondarily sorted using the left-most mapping position of the read (POS).
  * Subsequent to this sorting scheme, alignments are listed arbitrarily.</p> For queryname-sorted alignments, all
- * alignments are grouped using the queryname field but the alignments are not necessarily sorted within these groups.
- * Reads having the same queryname are derived from the same template.
+ * alignments are grouped using the queryname field followed by record strand orientation flag, primary record flag, and secondary
```
sooheelee: Since you say it is deterministic @jamesemery, I suggest: 
'all alignments are grouped using the' --> 'the tool orders records deterministically by'

src/main/java/picard/sam/SortSam.java
```diff
@@ -53,9 +53,9 @@
  * <p> For a coordinate sorted SAM/BAM file, read alignments are sorted first by the reference sequence name (RNAME)
  * field using the reference sequence dictionary tag labeled SQ.
  * Alignments within these subgroups are secondarily sorted using the left-most mapping position of the read (POS).
- * Subsequent to this sorting scheme, alignments are listed arbitrarily.</p> For queryname-sorted alignments, all
- * alignments are grouped using the queryname field but the alignments are not necessarily sorted within these groups.
- * Reads having the same queryname are derived from the same template.
+ * Subsequent to this sorting scheme, alignments are listed arbitrarily.</p> For queryname-sorted alignments,'the tool
```
sooheelee: Minor: remove `'`

src/main/java/picard/sam/SortSam.java
```diff
@@ -53,9 +53,9 @@
  * <p> For a coordinate sorted SAM/BAM file, read alignments are sorted first by the reference sequence name (RNAME)
  * field using the reference sequence dictionary tag labeled SQ.
  * Alignments within these subgroups are secondarily sorted using the left-most mapping position of the read (POS).
- * Subsequent to this sorting scheme, alignments are listed arbitrarily.</p> For queryname-sorted alignments, all
- * alignments are grouped using the queryname field but the alignments are not necessarily sorted within these groups.
- * Reads having the same queryname are derived from the same template.
+ * Subsequent to this sorting scheme, alignments are listed arbitrarily.</p> For queryname-sorted alignments, the tool
+ * orders records deterministically by queryname field followed by record strand orientation flag, primary record flag, and secondary
+ * alignment flag. (See {@link htsjdk.samtools.SAMRecordQueryNameComparator#compare(SAMRecord, SAMRecord)}} for details)
```
sooheelee: To address @yfarjoun's request, add:
"This ordering may change in the future. " to the end of this line.

sooheelee: @jamesemery ^^

sooheelee:
For this PR to pass checks, it needs approval from @yfarjoun, @jamesemery.

PR #1286, opened 2019-02-27
Update CreateSequenceDictionary.java

nh13:
In particular, alternate contig names like `NC_000001.11` where not allowed (period should be allowed).

### DIFF
--- a/src/main/java/picard/sam/CreateSequenceDictionary.java
+++ b/src/main/java/picard/sam/CreateSequenceDictionary.java
@@ -23,6 +23,7 @@
  */
 package picard.sam;
 
+import htsjdk.samtools.SAMException;
 import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.SAMSequenceDictionaryCodec;
 import htsjdk.samtools.SAMSequenceRecord;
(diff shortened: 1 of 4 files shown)


src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -125,7 +125,7 @@
      * Regular expression defined in the sam spec. Any alternative contig should match this regular expression 
      * TODO: replace the pattern with a constant : see https://github.com/samtools/htsjdk/pull/956/files
      */
-    private static final Pattern ALTERNATIVE_CONTIG_NAME_PATTERN = Pattern.compile("[0-9A-Za-z][0-9A-Za-z\\*\\+@\\|\\-]*");
+    private static final Pattern ALTERNATIVE_CONTIG_NAME_PATTERN = Pattern.compile("[0-9A-Za-z][0-9A-Za-z*+.@_|-]*");
```
yfarjoun: I think that there is no special alternative_contig_name pattern anymore, so you can use SAMSequenceRecord.LEGAL_RNAME_PATTERN

nh13:
thanks @yfarjoun, back to you!

PR #1292, opened 2019-03-07
Simplify BedToIntervalList by not reimplementing coordinate conversion

lbergelson:
### Description

* BedToIntervalList spends a lot of comments complaining about coordinate conversions after explicitly setting the codec to not perform the conversion from [0,..) to [1,..]
* Simplified by just using the baked in conversion instead of disabling it.

Maybe I'm missing something clever that it's doing here...

----

#### Content

#### Review


src/main/java/picard/util/BedToIntervalList.java
```diff
@@ -124,25 +124,14 @@ protected int doWork() {
             header.setSortOrder(SAMFileHeader.SortOrder.coordinate);
             final IntervalList intervalList = new IntervalList(header);
 
-            /**
-             * NB: BED is zero-based, but a BEDCodec by default (since it is returns tribble Features) has an offset of one,
-             * so it returns 1-based starts.  Ugh.  Set to zero.
-             */
-            final FeatureReader<BEDFeature> bedReader = AbstractFeatureReader.getFeatureReader(INPUT.getAbsolutePath(), new BEDCodec(BEDCodec.StartOffset.ZERO), false);
+            final FeatureReader<BEDFeature> bedReader = AbstractFeatureReader.getFeatureReader(INPUT.getAbsolutePath(), new BEDCodec(), false);
```
yfarjoun: will this be able to read a bed that has position 0?

yfarjoun: (the tests are not very good...they just test that the dictionary is right and that the program didn't crash...)

lbergelson: Good question.. The tests do actually test the output interval see `testBedToIntervalListDataProvider`, that's how we noticed that there was a bug with merging 0-length intervals.  

With some experimentation: 0 intervals that start at 0 work, 0-length intervals that start as 0 have never worked...  I added a fix and a new test.

To be complete we might want to allow the weird edge case of the zero-length interval AFTER the end of the contig... My guess is that doesn't work either... GATK will almost certainly explode if it sees one of those as well.

PR #1298, opened 2019-03-19
Add IntervalListTools COUNT action

kachulis:
Adds a COUNT action to IntervalListTools.  This action behaves identically to CONCAT, but including it as a duplicate provides an intuitively named way to count the bases or intervals in an interval list.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/util/IntervalListTools.java
+++ b/src/main/java/picard/util/IntervalListTools.java
@@ -275,24 +275,29 @@ public class IntervalListTools extends CommandLineProgram {
     public boolean INVERT = false;
 
     @Argument(doc = "What value to output to COUNT_OUTPUT file (for scripting)")
-    public Output OUTPUT_VALUE = Output.BASES;
+    public Output OUTPUT_VALUE = Output.NONE;
 
-    @Argument(doc = "File to which to print count of bases or intervals in final output interval list.", optional = true)
+    @Argument(doc = "File to which to print count of bases or intervals in final output interval list.  When this parameter is set OUTPUT_VALUE must be set to a value other than NONE.", optional = true)
     public File COUNT_OUTPUT;
 
     enum Output {
+        NONE {
+            void output(final long totalBaseCount, final long intervalCount, final PrintWriter writer){
+
+            }
+        },
         BASES {
-            void output(long totalBaseCount, long intervalCount, PrintWriter writer) {
+            void output(final long totalBaseCount, final long intervalCount, final PrintWriter writer) {
                 writer.println(totalBaseCount);
             }
         },
         INTERVALS {
-            void output(long totalBaseCount, long intervalCount, PrintWriter writer) {
+            void output(final long totalBaseCount, final long intervalCount, final PrintWriter writer) {
                 writer.println(intervalCount);
             }
         };
 
-        abstract void output(long totalBaseCount, long intervalCount, PrintWriter writer);
+        abstract void output(final long totalBaseCount, final long intervalCount, final PrintWriter writer);
     }
 
     private static final Log LOG = Log.getInstance(IntervalListTools.class);
@@ -461,6 +466,7 @@ public class IntervalListTools extends CommandLineProgram {
 
         LOG.info("Produced " + intervalCount + " intervals totalling " + totalBaseCount + " bases.");
         if (COUNT_OUTPUT != null) {
+            IOUtil.assertFileIsWritable(COUNT_OUTPUT);
             try {
                 final PrintWriter countWriter = new PrintWriter(new FileWriter(COUNT_OUTPUT));
                 OUTPUT_VALUE.output(totalBaseCount, intervalCount,countWriter);
@@ -500,6 +506,12 @@ public class IntervalListTools extends CommandLineProgram {
         }
         if ((SECOND_INPUT != null && !SECOND_INPUT.isEmpty()) && !ACTION.takesSecondInput) {
             errorMsgs.add("SECOND_INPUT was provided but action " + ACTION + " doesn't take a second input.");
+        }
+        if(COUNT_OUTPUT != null && OUTPUT_VALUE == Output.NONE) {
+            errorMsgs.add("COUNT_OUTPUT was provided but OUTPUT_VALUE is set to NONE.");
+        }
+        if(COUNT_OUTPUT == null && OUTPUT_VALUE != Output.NONE) {
+            errorMsgs.add("OUTPUT_VALUE is set to "+OUTPUT_VALUE+" but COUNT_OUTPUT was not provided.");
         }
 
         return errorMsgs.isEmpty() ? null : errorMsgs.toArray(new String[errorMsgs.size()]);
--- a/src/main/java/picard/util/IntervalListTools.java
+++ b/src/main/java/picard/util/IntervalListTools.java
@@ -271,30 +271,33 @@ public class IntervalListTools extends CommandLineProgram {
             "by any of the input intervals. Will merge abutting intervals first. Output will be sorted.", optional = true)
     public boolean INVERT = false;
 
-    @Argument(doc = "What value to output to COUNT_OUTPUT file (for scripting)")
+    @Argument(doc = "What value to output to COUNT_OUTPUT file or stdout (for scripting).  If COUNT_OUTPUT is provided, this parameter must not be NONE.")
     public Output OUTPUT_VALUE = Output.NONE;
 
-    @Argument(doc = "File to which to print count of bases or intervals in final output interval list.  When this parameter is set OUTPUT_VALUE must be set to a value other than NONE.", optional = true)
+    @Argument(doc = "File to which to print count of bases or intervals in final output interval list.  When not set, value indicated by OUTPUT_VALUE will be printed to stdout.  " +
+            "If this parameter is set, OUTPUT_VALUE must not be NONE.", optional = true)
     public File COUNT_OUTPUT;
 
     enum Output {
         NONE {
-            void output(final long totalBaseCount, final long intervalCount, final PrintStream writer){
-
+            void output(final long totalBaseCount, final long intervalCount, final PrintStream stream){
+                if (stream != System.out) {
+                    throw new PicardException("Asked to output NONE with COUNT_OUTPUT set.");
+                }
             }
         },
         BASES {
-            void output(final long totalBaseCount, final long intervalCount, final PrintStream writer) {
-                writer.println(totalBaseCount);
+            void output(final long totalBaseCount, final long intervalCount, final PrintStream stream) {
+                stream.println(totalBaseCount);
             }
         },
         INTERVALS {
-            void output(final long totalBaseCount, final long intervalCount, final PrintStream writer) {
-                writer.println(intervalCount);
+            void output(final long totalBaseCount, final long intervalCount, final PrintStream stream) {
+                stream.println(intervalCount);
             }
         };
 
-        abstract void output(final long totalBaseCount, final long intervalCount, final PrintStream writer);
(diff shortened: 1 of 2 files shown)


src/main/java/picard/util/IntervalListTools.java
```diff
@@ -335,6 +343,12 @@ IntervalList act(final List<IntervalList> list1, final List<IntervalList> list2)
             IntervalList act(final List<IntervalList> list1, final List<IntervalList> list2) {
                 return IntervalList.overlaps(list1, list2);
             }
+        },
+        COUNT("Count number of unique bases or intervals in INPUT.  This action behaves identically to CONCAT and is provided as an intuitively named convenience.",false) {
```
yfarjoun: this seem to be identical to the CONCAT action.

yfarjoun:
I wonder if this could be implemented as an argument, rather than an ACTION. The argument would be BASE_COUNT_OUTPUT and if non-null the program would emit the total bases in the outputfile to that file.

kachulis:
That's reasonable.  I think in that case it will make the most sense to combine it with the functionality of the `OUTPUT_VALUE` argument, so that the base (or interval) count is output to the given file instead of to stdout.

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -271,26 +274,25 @@
...
-    @Argument(doc = "What value (if anything) to output to stdout (for scripting)")
-    public Output OUTPUT_VALUE = Output.NONE;
+    @Argument(doc = "What value to output to COUNT_OUTPUT file (for scripting)")
+    public Output OUTPUT_VALUE = Output.BASES;
+
+    @Argument(doc = "File to which to print count of bases or intervals in final output interval list.", optional = true)
+    public File COUNT_OUTPUT;
 
     enum Output {
-        NONE() {
```
yfarjoun: I'd leave the none option around...

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -271,26 +274,25 @@
             "by any of the input intervals. Will merge abutting intervals first. Output will be sorted.", optional = true)
     public boolean INVERT = false;
 
-    @Argument(doc = "What value (if anything) to output to stdout (for scripting)")
-    public Output OUTPUT_VALUE = Output.NONE;
+    @Argument(doc = "What value to output to COUNT_OUTPUT file (for scripting)")
```
yfarjoun: add a commandlinevalidation that checks that COUNT_OUTPUT != null iff OUTPUT_VALUE != None

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -271,26 +274,25 @@
...
+    public Output OUTPUT_VALUE = Output.BASES;
```
yfarjoun: default to None

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -271,26 +274,25 @@
...
 
     enum Output {
-        NONE() {
-            void output(long totalBaseCount, long intervalCount) {
-            }
-        },
         BASES {
-            void output(long totalBaseCount, long intervalCount) {
-                System.out.println(totalBaseCount);
+            void output(long totalBaseCount, long intervalCount, PrintWriter writer) {
```
yfarjoun: finals in the parameters...(not your fault)

src/main/java/picard/util/IntervalListTools.java
yfarjoun: and also validate that COUNT_OUTPUT is writable

kachulis:
@yfarjoun back to you!

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -275,24 +275,29 @@
...
+    public Output OUTPUT_VALUE = Output.NONE;
 
-    @Argument(doc = "File to which to print count of bases or intervals in final output interval list.", optional = true)
+    @Argument(doc = "File to which to print count of bases or intervals in final output interval list.  When this parameter is set OUTPUT_VALUE must be set to a value other than NONE.", optional = true)
     public File COUNT_OUTPUT;
 
     enum Output {
+        NONE {
+            void output(final long totalBaseCount, final long intervalCount, final PrintWriter writer){
+
```
yfarjoun: throw an error here?

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -490,6 +507,12 @@ protected int doWork() {
         if ((SECOND_INPUT != null && !SECOND_INPUT.isEmpty()) && !ACTION.takesSecondInput) {
             errorMsgs.add("SECOND_INPUT was provided but action " + ACTION + " doesn't take a second input.");
         }
+        if(COUNT_OUTPUT != null && OUTPUT_VALUE == Output.NONE) {
```
yfarjoun:
```suggestion
        if (COUNT_OUTPUT != null && OUTPUT_VALUE == Output.NONE) {
```

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -490,6 +507,12 @@ protected int doWork() {
...
+            errorMsgs.add("COUNT_OUTPUT was provided but OUTPUT_VALUE is set to NONE.");
+        }
+        if(COUNT_OUTPUT == null && OUTPUT_VALUE != Output.NONE) {
```
yfarjoun:
```suggestion
        if (COUNT_OUTPUT == null && OUTPUT_VALUE != Output.NONE) {
```

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -490,6 +507,12 @@ protected int doWork() {
...
+            errorMsgs.add("OUTPUT_VALUE is set to "+OUTPUT_VALUE+" but COUNT_OUTPUT was not provided.");
```
yfarjoun:
```suggestion
            errorMsgs.add("OUTPUT_VALUE is set to " + OUTPUT_VALUE + " but COUNT_OUTPUT was not provided.");
```

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -271,26 +274,30 @@
             "by any of the input intervals. Will merge abutting intervals first. Output will be sorted.", optional = true)
     public boolean INVERT = false;
 
-    @Argument(doc = "What value (if anything) to output to stdout (for scripting)")
+    @Argument(doc = "What value to output to COUNT_OUTPUT file (for scripting)")
     public Output OUTPUT_VALUE = Output.NONE;
 
+    @Argument(doc = "File to which to print count of bases or intervals in final output interval list.  When this parameter is set OUTPUT_VALUE must be set to a value other than NONE.", optional = true)
+    public File COUNT_OUTPUT;
```
yfarjoun: if the default previously was that output goes to /dev/stdout, then perhaps we should leave it that way...

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -458,7 +465,17 @@ protected int doWork() {
...
-        OUTPUT_VALUE.output(totalBaseCount, intervalCount);
+        if (COUNT_OUTPUT != null) {
+            IOUtil.assertFileIsWritable(COUNT_OUTPUT);
+            try {
+                final PrintWriter countWriter = new PrintWriter(new FileWriter(COUNT_OUTPUT));
+                OUTPUT_VALUE.output(totalBaseCount, intervalCount,countWriter);
+                countWriter.close();
+            }
+            catch (final IOException e) {
+                throw new PicardException("There was a problem writing count to "+COUNT_OUTPUT.getAbsolutePath());
```
yfarjoun:
```suggestion
                throw new PicardException("There was a problem writing count to " + COUNT_OUTPUT.getAbsolutePath());
```

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -458,7 +465,17 @@ protected int doWork() {
         }
 
         LOG.info("Produced " + intervalCount + " intervals totalling " + totalBaseCount + " bases.");
-        OUTPUT_VALUE.output(totalBaseCount, intervalCount);
+        if (COUNT_OUTPUT != null) {
+            IOUtil.assertFileIsWritable(COUNT_OUTPUT);
+            try {
+                final PrintWriter countWriter = new PrintWriter(new FileWriter(COUNT_OUTPUT));
```
yfarjoun: ur try with resources

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -458,7 +465,17 @@ protected int doWork() {
         }
 
         LOG.info("Produced " + intervalCount + " intervals totalling " + totalBaseCount + " bases.");
-        OUTPUT_VALUE.output(totalBaseCount, intervalCount);
+        if (COUNT_OUTPUT != null) {
+            IOUtil.assertFileIsWritable(COUNT_OUTPUT);
```
yfarjoun: these asserts should be added to customCommandLineValidation

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -490,6 +507,12 @@ protected int doWork() {
         if ((SECOND_INPUT != null && !SECOND_INPUT.isEmpty()) && !ACTION.takesSecondInput) {
             errorMsgs.add("SECOND_INPUT was provided but action " + ACTION + " doesn't take a second input.");
         }
+        if (COUNT_OUTPUT != null && OUTPUT_VALUE == Output.NONE) {
```
yfarjoun: should be in customCommandLineValidation

src/main/java/picard/util/IntervalListTools.java
kachulis: I suppose the `assertFileIsWritable` could be in `customCommandLineValidation`, though this is inconsistent with the treatment of other inputs/outputs.  All the assertions on readability/writability of `INPUT`, `SECOND_INPUT`, and `OUTPUT` are done in the `doWork()` method.  I am happy to leave them where they are, or move them all to `customCommandLineValidation`, just think it should be consistent.

src/main/java/picard/util/IntervalListTools.java
kachulis: I believe this is in `customCommanLineValidation`...

kachulis: not sure why github claims this is in `doWork` when viewing changes... if you view the full file it is in `customCommandLineValidation`.

yfarjoun: odd.

src/main/java/picard/util/IntervalListTools.java
kachulis: That's fair, I have adjusted this so that by default the values are printed to standard out (and adjusted the command line validation to account for this as well).

src/main/java/picard/util/IntervalListTools.java
kachulis: With the default changed to print to stdout, it becomes only reasonable to throw an exception here if COUNT_OUTPUT is set.  As written, `customCommandLineValidation` should guarantee that state never reaches here, but I suppose checking and throwing an exception here provides additional safety.

kachulis:
@yfarjoun back to you

src/main/java/picard/util/IntervalListTools.java
yfarjoun: as long as it happens in the beginning of the invocation, rather than just before writing out, I'm fine.

kachulis: got it, moved to beginning of `doWork()`

yfarjoun:
please delete branch if no longer needed

PR #1312, opened 2019-04-13
Add fingerprint Metric (not FingerprintingMetric)

yfarjoun:
This adds a CLP that collects metrics on a given fingerprint (not a comparison between two fingerprints) The idea is to be able to detect when the fingerprint itself is "funny", i.e. doesn't seem to follow the assumptions that we have on fingerprints, namely that they are derived from a diploid mendelian draw from a distribution of genotypes with a prescribed frequency. Several different metrics and statistics are collected, since different "failure modes" could result in different assumptions being broken.

### DIFF
--- a/src/main/java/picard/fingerprint/CalculateFingerprintMetrics.java
+++ b/src/main/java/picard/fingerprint/CalculateFingerprintMetrics.java
@@ -161,7 +161,7 @@ public class CalculateFingerprintMetrics extends CommandLineProgram {
         fingerprintMetrics.HAPLOTYPES = fingerprint.values().size();
         fingerprintMetrics.HAPLOTYPES_WITH_EVIDENCE = fingerprint.values().stream().filter(HaplotypeProbabilities::hasEvidence).count();
         fingerprintMetrics.DEFINITE_GENOTYPES = fingerprint.values().stream().filter(h -> h.getLodMostProbableGenotype() > GENOTYPE_LOD_THRESHOLD).count();
-        fingerprintMetrics.NUM_HOM_ALLELE1 = roundedGenotypeCounts[0];
+        fingerprintMetrics.NUM_HOM_ALLELE1 = roundedGenotypeCounts[HOM_ALLELE1.v];
         fingerprintMetrics.NUM_HET = roundedGenotypeCounts[1];
         fingerprintMetrics.NUM_HOM_ALLELE2 = roundedGenotypeCounts[2];
 
--- a/src/main/java/picard/fingerprint/CalculateFingerprintMetrics.java
+++ b/src/main/java/picard/fingerprint/CalculateFingerprintMetrics.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2016 The Broad Institute
+ * Copyright (c) 2018 The Broad Institute
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -24,43 +24,32 @@
 
 package picard.fingerprint;
 
-import htsjdk.samtools.BamFileIoUtils;
-import htsjdk.samtools.SAMReadGroupRecord;
-import htsjdk.samtools.SamReader;
-import htsjdk.samtools.SamReaderFactory;
 import htsjdk.samtools.metrics.MetricsFile;
-import htsjdk.samtools.util.*;
-import htsjdk.variant.utils.SAMSequenceDictionaryExtractor;
-import htsjdk.variant.vcf.VCFFileReader;
-import htsjdk.variant.vcf.VCFHeader;
+import htsjdk.samtools.util.IOUtil;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
 import org.broadinstitute.barclay.help.DocumentedFeature;
-import picard.PicardException;
-import picard.analysis.FingerprintingDetailMetrics;
-import picard.analysis.FingerprintingSummaryMetrics;
 import picard.cmdline.CommandLineProgram;
 import picard.cmdline.StandardOptionDefinitions;
 import picard.cmdline.programgroups.DiagnosticsAndQCProgramGroup;
 
 import java.io.File;
-import java.io.IOException;
 import java.nio.file.Path;
-import java.text.Normalizer;
-import java.util.Collections;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 /**
- * Checks the sample identity
+ * Calculates vrious metrics on a sample fingerprint, indicating whether the fingerprint satisfies the assumptions we have.
+ * For example, if too many sites are hetrozygous, that would get flagged.
  *
  * @author Yossi Farjoun
  */
 
 @CommandLineProgramProperties(
-        summary = CalculateFingerprintMetrics.USAGE_DETAILS,
-        oneLineSummary = "blah",
+        summary = CalculateFingerprintMetrics.USAGE_SUMMARY + CalculateFingerprintMetrics.USAGE_DETAILS,
+        oneLineSummary = CalculateFingerprintMetrics.USAGE_SUMMARY,
         programGroup = DiagnosticsAndQCProgramGroup.class
 )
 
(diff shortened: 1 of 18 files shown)


src/main/java/picard/fingerprint/FingerprintChecker.java
```diff
@@ -96,6 +96,14 @@ public FingerprintChecker(final HaplotypeMap haplotypes) {
         this.haplotypes = haplotypes;
     }
 
+    /**
+     * Creates a fingerprint checker that will work with the set of haplotyped provided.
+     */
+    public FingerprintChecker(final Fingerprint fingerprint) {
```
meganshand: It doesn't look like you're using this.

src/main/java/picard/fingerprint/CalculateFingerprintMetrics.java
```diff
@@ -0,0 +1,223 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2018 The Broad Institute
```
meganshand: 2019

src/main/java/picard/fingerprint/FingerprintMetrics.java
```diff
@@ -0,0 +1,99 @@
...
+    /* Number of haplotypes that had evidence in the source file */
+    public long HAPLOTYPES_WITH_EVIDENCE;
+
+    /* Number of haplotypes that had enough evidence to make a definite genotype call */
+    public long DEFINITE_GENOTYPES;
+
+    /* Number of major allele homozygous calls*/
+    public long NUM_HOM_ALLELE1;
+
+    /* Number of definite heterozygous calls*/
```
meganshand: What does definite mean here? These are the number of het calls (that haven't been filtered by `GENOTYPE_LOD_THRESHOLD`) so this is different from `DEFINITE_GENOTYPES`, is that right? (And same for `NUM_HOM_ALLELE2 below`)

src/main/java/picard/fingerprint/FingerprintMetrics.java
```diff
@@ -0,0 +1,99 @@
...
+    public long NUM_HET;
+
+    /* Number of definite minor allele homozygous calls*/
+    public long NUM_HOM_ALLELE2;
+
+    /* The Chi-squared pvalue of the counts vector relative to the expected counts, (3x2 table) */
```
meganshand:
```suggestion
    /* The Chi-squared pvalue of the observed counts vector relative to the expected counts, (3x2 table) */
```

src/main/java/picard/fingerprint/FingerprintMetrics.java
```diff
@@ -0,0 +1,99 @@
...
+import htsjdk.samtools.metrics.MetricBase;
+import picard.analysis.FingerprintingDetailMetrics;
+import picard.analysis.FingerprintingSummaryMetrics;
+
+/**
+ * Class for holding metrics on a single fingerprint.
+ * Note: this is distinct from {@link FingerprintingDetailMetrics} and
+ * {@link FingerprintingSummaryMetrics} in that it is calculated on a single fingerprint,
+ * and attempts to describe how likely that fingerprint is to have arisen from an actual sample
+ * as opposed to having artifacts such as contamination, or strong bias towards homozygous genotypes.
```
meganshand: Can you note here that "expected" is based on the allele frequency from the `HAPLOTYPE_MAP`? I suppose this is noted in the doc for `CalculateFingerprintMetrics`, but it would be nice to also note it here since the docs for the metrics file aren't next to the docs for the tool.

src/main/java/picard/fingerprint/CalculateFingerprintMetrics.java
```diff
@@ -0,0 +1,223 @@
...
+                    "a non-contaminated, diploid human, whose genotypes are modelled by the probabilities given in the " +
+                    "HAPLOTYPE_MAP file." +
+                    "<p>" +
+                    "<h3>Example</h3>\n" +
+                    "<pre>\" +\n" +
+                    "java -jar picard.jar CalculateFingerprintMetrics \\\n" +
+                    "      INPUT=sample.bam \\\n" +
+                    "      HAPLOTYPE_DATABASE=fingerprinting_haplotype_database.txt \\\n" +
+                    "      OUTPUT=sample.fingerprint_metrics\n" +
+                    " </pre>\n";
```
meganshand: Can you also add something like:

```
"<p>Please see the FingerprintMetrics " +
"<a href='http://broadinstitute.github.io/picard/picard-metric-definitions.html#FingerprintMetrics'>definitions</a> " +
"for a complete description of the metrics produced by this tool.</p>" +
"<hr />";
```

src/main/java/picard/fingerprint/HaplotypeProbabilities.java
```diff
@@ -24,6 +24,9 @@
 
 package picard.fingerprint;
 
+import org.apache.commons.math3.distribution.ChiSquaredDistribution;
```
meganshand: unused imports

src/main/java/picard/util/MathUtil.java
```diff
@@ -62,7 +64,11 @@ public static double mean(final double[] in, final int start, final int stop) {
         return total / (stop - start);
     }
 
-    /** Calculated the standard deviation of an array of doubles. */
+    public static double mean(final double[] in) {
+        return mean(in,0, in.length);
+    }
+
+        /** Calculated the standard deviation of an array of doubles. */
```
meganshand: weird indent

src/main/java/picard/util/MathUtil.java
```diff
@@ -77,7 +83,15 @@ public static double stddev(final double[] in, final int start, final int stop,
...
-    public static int compare(final int v1, final int v2) {
+
+    /** Calculated the standard deviation of an array of doubles. */
+    public static double stddev(final double[] in, final double mean) {
+        return stddev(in, 0, in.length, mean);
+    }
+
+
+
+        public static int compare(final int v1, final int v2) {
```
meganshand: extra indent here too

src/main/java/picard/fingerprint/CalculateFingerprintMetrics.java
```diff
@@ -0,0 +1,223 @@
...
+        // calculate LOD (cross-entropy)
+        fingerprintMetrics.HOM_CROSS_ENTROPY_LOD = MathUtil.klDivergance(homAllele1VsAllele2Counts, homAllele1VsAllele2Expect);;
+
+        final MathUtils.RunningStat randomTrials = new MathUtils.RunningStat();
+
+        // get a bunch of random permutations of the fingerprint and compare to self
+        IntStream.range(0, NUMBER_OF_SAMPLING).forEach(i ->
+                randomTrials.push(FingerprintChecker.calculateMatchResults(fingerprint, randomizeFingerprint(fingerprint, rg)).getLOD()));
+
+        fingerprintMetrics.LOD_SELF_CHECK = FingerprintChecker.calculateMatchResults(fingerprint, fingerprint).getLOD();
```
meganshand: The point of this is to make sure the evidence of the genotypes are strong enough for this fingerprint, is that right? Since a self check should always be a higher LOD than any permutation of the PLs? If that's the case, why have `DISCRIMINATORY_POWER` be the `LOD_SELF_CHECK - mean`, rather than say `LOD_SELF_CHECK - max`?

src/main/java/picard/fingerprint/FingerprintMetrics.java
```diff
@@ -0,0 +1,99 @@
...
+    /* The Chi-squared pvalue for the number of HOM-Allele1s and HOM_Allele2s relative to the expected counts (2x2 table) */
+    public double HOM_CHI_SQUARED_PVALUE;
+
+    /* The log10 of the Chi-squared pvalue for the number of HOM-Allele1s and HOM_Allele2s */
+    public double HOM_LOG10_CHI_SQUARED_PVALUE;
+
+    /* The categorical cross entropy of the HOM-Allele1s and HOM_Allele2s relative to the expected counts (big is bad)*/
+    public double HOM_CROSS_ENTROPY_LOD;
+
+    /* The fingerprinting LOD score this sample gets when compared to itself*/
```
meganshand: add `(big is good)`

src/main/java/picard/fingerprint/CalculateFingerprintMetrics.java
```diff
@@ -0,0 +1,223 @@
...
+        fingerprintMetrics.HAPLOTYPES_WITH_EVIDENCE = fingerprint.values().stream().filter(HaplotypeProbabilities::hasEvidence).count();
+        fingerprintMetrics.DEFINITE_GENOTYPES = fingerprint.values().stream().filter(h -> h.getLodMostProbableGenotype() > GENOTYPE_LOD_THRESHOLD).count();
+        fingerprintMetrics.NUM_HOM_ALLELE1 = roundedGenotypeCounts[0];
+        fingerprintMetrics.NUM_HET = roundedGenotypeCounts[1];
+        fingerprintMetrics.NUM_HOM_ALLELE2 = roundedGenotypeCounts[2];
+
+        // calculate p-value
+        final double chiSquaredTest = chiSquareTest.chiSquareTest(expectedRatios, roundedGenotypeCounts);
+        fingerprintMetrics.CHI_SQUARED_PVALUE = chiSquaredTest;
+        fingerprintMetrics.LOG10_CHI_SQUARED_PVALUE = Math.log10(chiSquaredTest);
```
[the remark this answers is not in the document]
yfarjoun: My sense is that people are used to thinking about log10(p-values) and phed-scale(error-rates)...

src/main/java/picard/fingerprint/CalculateFingerprintMetrics.java
yfarjoun:
```suggestion
 * Copyright (c) 2019 The Broad Institute
```

src/main/java/picard/fingerprint/CalculateFingerprintMetrics.java
yfarjoun: my rational is that using the max is less stable, while using the mean I expect will converge after ~100 iterations...

src/main/java/picard/fingerprint/FingerprintMetrics.java
yfarjoun:
```suggestion
    /* The fingerprinting LOD score this sample gets when compared to itself (big is good)*/
```

src/main/java/picard/fingerprint/CalculateFingerprintMetrics.java
```diff
@@ -0,0 +1,223 @@
...
+
+    static final String USAGE_SUMMARY="Calculate statistics on fingerprints, checking their viability";
+
+    @Argument(shortName = StandardOptionDefinitions.INPUT_SHORT_NAME, doc = "One or more input files (SAM/BAM/CRAM or VCF).")
+    public List<String> INPUT;
+
+    @Argument(shortName = StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "The output files to write.")
+    public File OUTPUT;
+
+    @Argument(shortName = "H", doc = "The file lists a set of SNPs, optionally arranged in high-LD blocks, to be used for fingerprinting. See " +
```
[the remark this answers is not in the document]
yfarjoun: it's H in crosscheck and checkfingerprints...

src/main/java/picard/fingerprint/CalculateFingerprintMetrics.java
```diff
@@ -0,0 +1,223 @@
...
+            "https://software.broadinstitute.org/gatk/documentation/article?id=9526 for details.")
+    public File HAPLOTYPE_MAP;
+
+    @Argument(doc = "Specificies which data-type should be used as the basic unit. Fingerprints from readgroups can " +
+            "be \"rolled-up\" to the LIBRARY, SAMPLE, or FILE level before being used." +
+            " Fingerprints from VCF can be be examined by SAMPLE or FILE.")
+    public CrosscheckMetric.DataType CALCULATE_BY = CrosscheckMetric.DataType.READGROUP;
+
+    @Argument(doc="LOD score threshold for considering a genotype to be definitive.")
+    public final double GENOTYPE_LOD_THRESHOLD = 3;
```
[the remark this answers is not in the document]
yfarjoun: yeah...here it's more important that I get counts. in the other tools the LOD isn't affected by the threshold, IIANM

src/main/java/picard/fingerprint/FingerprintChecker.java
yfarjoun: yeah...not sure why I added it.

src/main/java/picard/fingerprint/FingerprintMetrics.java
yfarjoun: same.... DEFINITE_GENOTYPES= NUM_HOM_ALLELE2 + NUM_HOM_ALLELE1 + NUM_HET

src/main/java/picard/fingerprint/CalculateFingerprintMetrics.java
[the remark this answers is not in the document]
meganshand: I know some VCF annotations in GATK have phred-scaled p-values, but I don't have a strong opinion either way, so you might as well leave it as is.

src/main/java/picard/fingerprint/FingerprintMetrics.java
yfarjoun: actually, that's not true...I'll make a better note of it.

yfarjoun:
I did test it on some samples. it mostly flagged fingerprints that were suspect for other reasons and RNA bams, probably due to differential expression (the HET counts were lower than expected)

yfarjoun:
@meganshand ,please re-review

src/main/java/picard/fingerprint/FingerprintMetrics.java
meganshand: I still don't understand this distinction. It looks like `NUM_HET`, `NUM_HOM_ALLELE1`, and `NUM_HOM_ALLELE2` are not "definite" because they aren't filtered by `GENOTYPE_LOD_THRESHOLD`. I would either remove the word definite here, or make it clearer [here](https://github.com/broadinstitute/picard/blob/yf_add_pvalue_to_fingerprinting/src/main/java/picard/fingerprint/CalculateFingerprintMetrics.java#L169) why these metrics are "definite" calls rather than just calls.

PR #1331, opened 2019-05-11
Add levenshtein distance to eibc

yfarjoun:
### Description
Adds the ability to extract Illumina Barcodes by using "FREE" (like levenshtein but doesn't count indels in the end) distance which counts the number of SNV and *unforced* indel changes between two sequences of the same length. 

Default behavior should remain the same, but when using "FREE" distance code, it has been observed to pick up about 1% more reads.


src/main/java/picard/util/StringDistanceUtils.java
```diff
@@ -0,0 +1,252 @@
...
+     * allocating an entire n by m matrix. This requires a few minor
+     * changes, such as immediately returning when it's detected that the
+     * stripe has run off the matrix and initially filling the arrays with
+     * large values so that entries we don't compute are ignored.
+     *
+     * See Algorithms on Strings, Trees and Sequences by Dan Gusfield for
+     * some discussion.
+     */
+
+    public static int levenshteinDistance(final byte[] barcode, final byte[] read, final int threshold) {
```
lbergelson: Is there a reason you can't use the htsjdk's `StringUtil.levenshteinDistance`

yfarjoun: yes...because it isn't exactly levenshtein...I should change the name.

src/main/java/picard/util/StringDistanceUtils.java
```diff
@@ -0,0 +1,253 @@
...
+package picard.util;
+
+import htsjdk.samtools.util.SequenceUtil;
+
+import java.util.Arrays;
+
+/**
+ * Algorithms for calculating various edit distances.
+ */
+public class StringDistanceUtils {
```
pshapiro4broad: Is there a better term than `String` for this concept, e.g. `EditDistanceUtils`? It's hard for me to not read `String` as the java type `String`.

src/main/java/picard/util/StringDistanceUtils.java
```diff
@@ -0,0 +1,253 @@
...
+
+    /**
+     * Compare barcode sequence to bases from read
+     *
+     * @return how many bases did not match
+     */
+    public static int countMismatches(final byte[][] barcodeBytes, final byte[][] readSubsequence, final byte[][] qualities, final int minimumBaseQuality) {
```
pshapiro4broad: It looks like many methods here accept the same argument list, `final byte[][] barcodeBytes, final byte[][] readSubsequence, final byte[][] qualities, final int minimumBaseQuality`.  If there was an object that had these fields, then these would become methods on that object that no longer accepted any arguments, which would be easier to read.

src/main/java/picard/util/StringDistanceUtils.java
```diff
@@ -0,0 +1,253 @@
...
+            if (readQualities != null && readQualities[i] < minimumBaseQuality) {
+                ++numMismatches;
+            }
+        }
+        return numMismatches;
+    }
+
+    static boolean anySmaller(final byte[] values, final int minValue) {
+        for (int i = 0; i < values.length; i++) {
+            if (values[i] < minValue) return true;
```
pshapiro4broad: please format across lines, e.g.
```
if (values[i] < minValue) {
    return true;
}
````

File: src/main/java/picard/util/StringDistanceUtils.java [hunk 5]
```diff
@@ -0,0 +1,253 @@
...
+     * length 3. The matrix would look like so:
+     * <p>
+     * <pre>
+     *    1 2 3 4 5
+     * 1 |#|#| | | |
+     * 2 |#|#|#| | |
+     * 3 | |#|#|#| |
+     * 4 | | |#|#|#|
+     * 5 | | | |#|#|
+     *
```
pshapiro4broad: missing end tag for `<pre>`

src/main/java/picard/util/StringDistanceUtils.java
```diff
@@ -0,0 +1,253 @@
...
+     * k allows us to reduce it to O(km) time by only computing a diagonal
+     * stripe of width 2k + 1 of the cost table. It is also possible to use
+     * this to compute the unbounded Levenshtein distance by starting the
+     * threshold at 1 and doubling each time until the distance is found;
+     * this is O(dm), where d is the distance.
+     * <p>
+     * One subtlety comes from needing to ignore entries on the border of
+     * our stripe eg. p[] = |#|#|#|* d[] = *|#|#|#| We must ignore the entry
+     * to the leftRev of the leftmost member We must ignore the entry above the
+     * rightmost member
```
pshapiro4broad: typo? missing `.`

src/main/java/picard/util/StringDistanceUtils.java
```diff
@@ -0,0 +1,253 @@
...
+    public static int freeDistanceForBarcodes(final byte[] barcodeBases, final byte[] readBases, final byte[] readQualities, final int minimumBaseQuality) {
+        final byte[] maskedReadBases = getBytesMasked(readBases, readQualities, minimumBaseQuality);
+
+        return freeDistance(barcodeBases, maskedReadBases, 10);
+    }
+
+    /**
+     * +++lifted from Commons Lang Text +++
+     * <p>
+     * modified to specific to comparing barcodes to reads. This means
```
pshapiro4broad: typo?

src/main/java/picard/util/StringDistanceUtils.java
```diff
@@ -0,0 +1,253 @@
...
+            }
+
+            // copy current distance counts to 'previous row' distance counts
+            tempD = previousCost;
+            previousCost = cost;
+            cost = tempD;
+        }
+
+        // if previousCost[n] is greater than the threshold, there's no guarantee on it
+        // being the correct
```
pshapiro4broad: odd formatting of comment

src/test/java/picard/util/TestNGUtil.java
```diff
@@ -96,6 +100,15 @@ public static void assertEqualDoubleArrays(final double[] lhs, final double[] rh
         }
     }
 
+    public static void assertEquals(final byte[] actual, final byte[] expected) {
```
pshapiro4broad: TestNG already has `Assert.assertEquals(byte[] actual, byte[] expected)`, is there a reason why that wouldn't work here?

src/main/java/picard/util/StringDistanceUtils.java
```diff
@@ -0,0 +1,253 @@
...
+        // distance
+        if (previousCost[n] <= threshold) {
```
pshapiro4broad: Is there a way to convert this into `return Math.min(previousCost[n], threshold) + 1;`? Would it work if you removed the `1 + ` from line 235 above?

src/main/java/picard/util/StringDistanceUtils.java
```diff
@@ -0,0 +1,253 @@
...
+     * @param readBases The bases that the read has
+     * @param readQualities the Read qualities
+     * @param minimumBaseQuality The minimum base quality in the read that will be considered informative.
+     * @return the FREE distance (allowing for non-informative bases) between the read and the barcode.
+     * This uses a threshold of 10 meaning that the distances obtained will never exceed 10.
+     */
+    public static int freeDistanceForBarcodes(final byte[] barcodeBases, final byte[] readBases, final byte[] readQualities, final int minimumBaseQuality) {
+        final byte[] maskedReadBases = getBytesMasked(readBases, readQualities, minimumBaseQuality);
+
+        return freeDistance(barcodeBases, maskedReadBases, 10);
```
kachulis: Why the hardcoded threshold of 10?  Would be nice to have as a parameter of the method.

kachulis:
@yfarjoun I have pushed the eibc caching code you requested to this branch.

src/main/java/picard/util/StringDistanceUtils.java
yfarjoun: no..it continues in the next line.

src/test/java/picard/util/TestNGUtil.java
yfarjoun: yes, IINM TestNG doesn't show you the offending difference....which is less helpful for debugging.

src/main/java/picard/util/StringDistanceUtils.java
yfarjoun: not really, the `+1` above is inside a loop...

src/main/java/picard/util/StringDistanceUtils.java
pshapiro4broad: The first sentence is ```modified to specific to comparing barcodes to reads.``` I don't understand what this is trying to say.

src/test/java/picard/util/TestNGUtil.java
pshapiro4broad: Pretty sure that's not true; looking at the code (`org.testng.Assert`) I see ```  public static final String ARRAY_MISMATCH_TEMPLATE = "arrays differ firstly at element [%d]; "
      + "expected value is <%s> but was <%s>. %s";
```

Comment (yfarjoun) [reply src/test/java/picard/util/TestNGUtil.java]: hmm. I keep getting pointed at `Assert.assertEquals(Object, Object)`

Comment (yfarjoun) [reply src/test/java/picard/util/TestNGUtil.java]: are you seeing that signature in intellij or in a website?

Comment (pshapiro4broad) [reply src/test/java/picard/util/TestNGUtil.java]: Sorry, I was looking at the copy that htsjdk uses. If you update testng to the same version that htsjdk uses, then you should have the byte compare assert that prints out the difference.
`build.gradle`:
```
    testCompile 'org.testng:testng:6.14.3'
```

File: src/main/java/picard/illumina/ExtractIlluminaBarcodes.java [hunk 12]
```diff
@@ -824,116 +848,122 @@ synchronized public void run() {
...
-                totalBarcodeReadBases += bc.length;
-                for (final byte b : bc) if (SequenceUtil.isNoCall(b)) ++numNoCalls;
+            boolean canUseLookupTable = true;
+            if (qualityScores != null) {
+                barcodes:
+                for (final byte[] qs : qualityScores) {
+                    for (final byte q : qs) {
+                        if (q < minimumBaseQuality) {
+                            canUseLookupTable = false;
+                            break barcodes;
```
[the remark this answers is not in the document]
yfarjoun: com'on! I worked so hard to find a use for this obscure construct!!!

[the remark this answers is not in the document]
yfarjoun: (jk I'll do it..)

src/main/java/picard/illumina/ExtractIlluminaBarcodes.java
```diff
@@ -206,6 +209,9 @@
             "the number available on the machine less NUM_PROCESSORS.")
     public int NUM_PROCESSORS = 1;
 
+    @Argument()
+    public DistanceMetric DISTANCE_MODE = DistanceMetric.HAMMING;
```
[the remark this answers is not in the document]
yfarjoun: oops

src/main/java/picard/util/SingleBarcodeDistanceMetric.java
```diff
@@ -0,0 +1,270 @@
...
+                maskedBases[i] = '.';
+            }
+        }
+        return maskedBases;
+
+    }
+
+    private byte[] getMaskedReadBases() {
+        final byte[] maskedReadBases;
+        if (readQualities == null || !anySmaller(readQualities, minimumBaseQuality)) {
```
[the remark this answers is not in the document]
yfarjoun: the expectation is that in the vast majority of cases there will not be a smaller quality...and so this will prevent non-needed data copying. I'll smoosh the two functions together though as you suggest.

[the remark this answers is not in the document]
yfarjoun: I thin kI found a bug in the process :-/

yfarjoun:
thanks for the comments @pshapiro4broad. Back to you.

src/main/java/picard/util/BarcodeEditDistanceQuery.java
```diff
@@ -28,19 +28,19 @@
  */
 public class BarcodeEditDistanceQuery {
 
-    // list of barcodes (e.g. for dual-indexed barcodes)
+    /* list of barcodes (e.g. for dual-indexed barcodes) */
```
pshapiro4broad: ugh, sorry about this but you need two **s for javadoc comments
```
    /** list of barcodes (e.g. for dual-indexed barcodes) */
```

src/main/java/picard/util/SingleBarcodeDistanceMetric.java
```diff
@@ -85,7 +87,12 @@ private static boolean anySmaller(final byte[] values, final int minValue) {
         return false;
     }
 
-    private byte[] maskSmaller() {
+    static private byte[] maskIfAnySmaller(final byte [] readBases, final byte[] readQualities, final int minimumBaseQuality) {
+
+        if (!anySmaller(readQualities,minimumBaseQuality)) {
```
pshapiro4broad: If `values` is long and you're concerned about performance you could make `anySmaller` return the index where the first smaller element was found, and start the `for` loop below at that index. (And use `-1` to indicate no smaller item found.)

yfarjoun: I don't think that this is a huge source of compute time.

src/main/java/picard/util/BarcodeEditDistanceQuery.java
yfarjoun: right.

PR #1340, opened 2019-06-03
Update CollectSamErrorMetrics.java

nh13:
@yfarjoun care to take a look at this?  Happy for there to be a test for it, but it ain't my bug.


nh13:
A few more things I've noticed:
- assumes that the `.dict` file is present (it may not be)
- assumes base qualities are present (since the default error+stratifiers include them)

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -137,7 +137,7 @@
     public ReadBaseStratification.Stratifier STRATIFIER_VALUE;
 
     @Argument(shortName = "V", doc = "VCF of known variation for sample. program will skip over polymorphic sites in this VCF and " +
-            "avoid collecting data on these loci.")
+            "avoid collecting data on these loci.", optional = true)
```
yfarjoun: hmmm.  Collecting error metrics over variants doesn't make much sense...you could give it dbSNP or something like that if you dont' have truth for that sample....

nh13: What if my sample has no variants relative the reference? That’s my real use case! No VCF is required. Think sequencing bacteria or synthetic sequences.

yfarjoun:
OK. could you at least log a warning so that the (less-advanced) user who might be using this against human will be informed that they should be supplying dbSNP or another appropriate source of known variation for that sample?

nh13:
@yfarjoun I have been trying to think of a way to say that you must intentionally say not to use a VCF.  Perhaps we could instead make the option required, but if the value is `"<EMPTY>"` then it means there are no variants to consider?  I'd prefer that to an empty VCF.

yfarjoun:
Alternatively you could have a hidden boolean variable
ALLOW_MISSING_VCF_DESPITE_THIS_NOT_BEING_RECOMMENDED (or something)

kockan:
Superseded by #1476

PR #1341, opened 2019-06-11
Recover from side effects introduced in CollectIndependentReplicatesMetric tests.

cmnbroad:
This fix is needed as a prerequisite to upgrade Picard to the next version of htsjdk.

`CollectIndependentReplicatesMetricTest` has an `@BeforeTest` method that runs `MergeSamFiles` as a command line program, with CREATE_INDEX=true. `CommandLineProgram` in turn has a side effect that leaves htsjdk's global SAMFileWriterFactory state mutated with non-default values. The next version of htsjdk has code that rejects attempts to create an index for a CRAM file that isn't coordinate sorted. `RevertSamTest.testOutputByReadGroupWithOutputMap` creates CRAM files that are queryname sorted, and usually succeeds because it doesn't attempt to write an index, but fails intermittently with the new htsjdk code if it happens to be run after the `CollectIndependentReplicatesMetricTest` `@BeforeTest` method has executed, due to the leftover htsjdk state.

Longer term we should fix `CommandLineProgram` to not set the global defaults.

### DIFF
--- a/src/test/java/picard/analysis/replicates/CollectIndependentReplicatesMetricTest.java
+++ b/src/test/java/picard/analysis/replicates/CollectIndependentReplicatesMetricTest.java
@@ -1,6 +1,7 @@
 package picard.analysis.replicates;
 
 import com.google.common.collect.ImmutableMap;
+import htsjdk.samtools.SAMFileWriterFactory;
 import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.util.IOUtil;
 import htsjdk.samtools.util.TestUtil;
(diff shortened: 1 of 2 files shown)


src/test/java/picard/analysis/replicates/CollectIndependentReplicatesMetricTest.java
```diff
@@ -208,15 +209,19 @@ public void simpleTest(final String vcf, final String bam, final Map<String, Obj
...
+            final int returnCode = msf.instanceMain(
+                    new String[]{
+                            "INPUT=" + (new File(testdir, sam).getAbsolutePath()),
+                            "CREATE_INDEX=true",
+                            "OUTPUT=" + bam.getAbsolutePath()});
+            Assert.assertEquals(returnCode, 0);
+            return bam;
+        } finally {
+            // reset the global defaults to eliminate side effects in downstream code
+            SAMFileWriterFactory.setDefaultCreateIndexWhileWriting(false);
```
yfarjoun: hmmm. this is quite insidious....how can we make sure that it doesn't happen again?

cmnbroad: Not sure, short of fixing https://github.com/broadinstitute/picard/issues/1342. One other option I considered was to change `CommandLineProgram` so that instead of [this](https://github.com/broadinstitute/picard/blob/master/src/main/java/picard/cmdline/CommandLineProgram.java#L243), it just does this:

`SAMFileWriterFactory.setDefaultCreateIndexWhileWriting(CREATE_INDEX);`
 
But that seems just as sketchy, and overwrites the value even in the case where it was set intentionally via a property or whatever. And there are several such options (IIRC VariantContextWriterBuilder has some of these too).

cmnbroad: There already exist other non-test code paths that can leave these side effects, such as just running SortVcfs or MergeVcfs with default args, since they default to CREATE_INDEX=true.

cmnbroad:
I reverted the previous commit on this branch, and updated it to handle reverting side effects right in `CommandLineProgram`, since there were otherwise too many code paths to repair.

PR #1348, opened 2019-06-20
Adds a buffer to the index stream

### LINKED ISSUE #1175: CrosscheckFingerprints NIO streaming of BAMs hangs tool indefinitely
## Bug Report
DavidR says to bring this to your attention @yfarjoun.

### Affected tool(s)
CrosscheckFingerprints called from GATK4.0.4.0

### Affected version(s)
- [x] Latest public release version [version?]
- [ ] Latest development/master branch as of [date of test?]

### Description 
- Original forum report from the researcher describing the tool behavior: https://gatkforums.broadinstitute.org/gatk/discussion/11999/problem-with-crosscheckfingerprints-when-using-nio/p1
    The researcher has no problems running the tool on local BAM files. This finishes in a couple of hours, including copying time for files in GCS bucket.
    The tool hangs ~20hrs without error nor producing any results.
- dsde-docs discussion: https://github.com/broadinstitute/dsde-docs/issues/3096#issuecomment-390526302

#### Steps to reproduce
Leaving this to Picard folks to try to reproduce. 

#### Expected behavior
Looks like either (i) the tool isn't compatible with NIO and should produce an error or (ii) there is a bug in the NIO process for this tool that should be fixed.

yfarjoun:
This makes reading the index over a network possible, which in turn, fixes #1175.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/fingerprint/FingerprintChecker.java
+++ b/src/main/java/picard/fingerprint/FingerprintChecker.java
@@ -24,11 +24,21 @@
 
 package picard.fingerprint;
 
-import htsjdk.samtools.*;
-import htsjdk.samtools.filter.NotPrimaryAlignmentFilter;
+import com.google.cloud.storage.contrib.nio.SeekableByteChannelPrefetcher;
+import htsjdk.samtools.SAMFileHeader;
+import htsjdk.samtools.SAMReadGroupRecord;
+import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.SamReader;
+import htsjdk.samtools.SamReaderFactory;
+import htsjdk.samtools.ValidationStringency;
 import htsjdk.samtools.filter.SamRecordFilter;
 import htsjdk.samtools.filter.SecondaryAlignmentFilter;
-import htsjdk.samtools.util.*;
+import htsjdk.samtools.util.Interval;
+import htsjdk.samtools.util.IntervalList;
+import htsjdk.samtools.util.Log;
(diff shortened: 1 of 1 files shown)


yfarjoun:
@lbergelson can you take a look?

src/main/java/picard/fingerprint/FingerprintChecker.java
```diff
@@ -448,6 +484,7 @@ public IntervalList getLociToGenotype(final Collection<Fingerprint> fingerprints
         // Now go through the data at each locus and figure stuff out!
         for (final SamLocusIterator.LocusInfo info : iterator) {
 
+            log.debug("At locus " + info.toString());
```
lbergelson: This will do a pointless string combine at each position which could be kind of expensive if you don't want this output.  If there's a way check if we're in debug mode first than I would test that before doing the debug command.

src/main/java/picard/fingerprint/FingerprintChecker.java
```diff
@@ -402,7 +438,7 @@ public IntervalList getLociToGenotype(final Collection<Fingerprint> fingerprints
         final SamReader in = SamReaderFactory.makeDefault()
                 .enable(SamReaderFactory.Option.CACHE_FILE_BASED_INDEXES)
                 .referenceSequence(referenceFasta)
-                .open(samFile);
+                .open(samFile, null, seekableChannelFunction);
```
lbergelson: I would add a comment here describing why this is needed and pointing at the htsjdk issue number ( is there one... there should be.)

PR #1351, opened 2019-06-26
Bugfix: Update default platform for IlluminaBasecallsToSam

nh13:
The SAM specification requires the platform to be upper-case.

@yfarjoun assigning it to you, but feel free to change the reviewer


pshapiro4broad:
Does this change affect other tools which may depend on the value? Taking a quick look at the picard sources I see:

https://github.com/broadinstitute/picard/blob/master/src/main/java/picard/sam/AddOrReplaceReadGroups.java#L49
```
            "      RGPL=illumina \\\n" +
```

https://github.com/broadinstitute/picard/blob/master/src/main/java/picard/sam/AddOrReplaceReadGroups.java#L111
```
    @Argument(shortName = "PL", doc = "Read-Group platform (e.g. illumina, solid)")
```

https://github.com/broadinstitute/picard/blob/master/src/main/java/picard/sam/FastqToSam.java#L181
```
    @Argument(shortName="PL", doc="The platform type (e.g. illumina, solid) to insert into the read group header", optional=true)
```

It looks like the test are failing due to the many `.sam` files checked in with `PL:illumina`

nh13:
Bad tests should be fixed; I’ll do that. As for other tools, I’ll update their descriptions but folks could still screw things up manually: garbage in garbage out. They could specify the platform as ‘iLlUmInA’!

yfarjoun:
IIANM Validation is done with uppercased values.
Tests are probably failing since comparisons to the resulting files are
case-sensitive and the test files were invalid.

nh13:
@yfarjoun looks like htsjdk is comparing the platform values after [upper-casing the target value](https://github.com/samtools/htsjdk/blob/15ec7da7c67e7369ed2980a308bc3c3208d06d95/src/main/java/htsjdk/samtools/SamFileValidator.java#L618).  I'd bet this breaks a lot of BAMs in the wild that have "PL:illumina",  so while strictly we should change that _in htsjdk_ to make it spec-compliant, I don't think practically it is a good idea.  I'll leave that to others to pick up the discussion.

@pshapiro4broad I originally encountered this issue (lower-case `illumina`) in a tool that examined read groups, and was comparing the platform _case sensitive_.  I looked in Picard for usages of `SAMReadGroupRecord.getPlatform` and I don't see the same issue (in fact it is used only in two places!). 

Once tests I am going to merge; Running them locally is a bit of a PITA tbh, so I am relying on travis-ci.

nh13:
@pshapiro4broad thanks! My earlier comment did link to that method in htsjdk!

pshapiro4broad:
> @pshapiro4broad thanks! My earlier comment did link to that method in htsjdk!

I missed that, I think I skimmed the part of your comment that wasn't addressed to me. Sorry!

PR #1368, opened 2019-07-31
Added a MIN_HISTOGRAM_WIDTH to CollectInsertSizeMetrics.

tfenne:
### Description

Adds a new option to CollectInsertSizeMetrics to specify a _minimum_ width for the histogram.  This differs from `HISTOGRAM_WIDTH` which _always_ sets the width to the given value.

The motivation for this is that I have a process and pipeline that yields libraries with a very predictable insert size distribution most of the time, and for those libraries I've been using `HISTOGRAM_WIDTH` so that all the plots come out with the same X-axis scaling, making visual comparisons much easier and less error-prone.  However, occasionally odd-ball libraries come through which have significantly larger insert sizes, and the plots are then un-usable.

My intention with this option is to allow usage that uses the `DEVIATIONS` argument in combination with `MIN_HISTOGRAM_WIDTH` in such a way that _most_ libraries get plotted using `MIN_HISTOGRAM_WIDTH` but that libraries that have sufficiently higher insert sizes will end up using the calculated value and expanding the plot width.

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/analysis/CollectInsertSizeMetrics.java
+++ b/src/main/java/picard/analysis/CollectInsertSizeMetrics.java
@@ -58,7 +58,7 @@ import java.util.Set;
         programGroup = DiagnosticsAndQCProgramGroup.class
 )
 @DocumentedFeature
-public class    CollectInsertSizeMetrics extends SinglePassSamProgram {
+public class CollectInsertSizeMetrics extends SinglePassSamProgram {
     static final String USAGE_BRIEF = "Collect metrics about the insert size distribution of a paired-end library.";
     static final String USAGE_SUMMARY = "<p>This tool provides useful metrics for validating library construction including " +
             "the insert size distribution and read orientation of paired-end libraries.</p>" +
--- a/src/main/java/picard/analysis/CollectInsertSizeMetrics.java
+++ b/src/main/java/picard/analysis/CollectInsertSizeMetrics.java
@@ -29,6 +29,7 @@ import htsjdk.samtools.SAMRecord;
 import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.reference.ReferenceSequence;
 import htsjdk.samtools.util.CollectionUtil;
+import htsjdk.samtools.util.Histogram;
 import htsjdk.samtools.util.IOUtil;
 import htsjdk.samtools.util.Log;
 import org.broadinstitute.barclay.argparser.Argument;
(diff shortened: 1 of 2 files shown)


src/main/java/picard/analysis/CollectInsertSizeMetrics.java
```diff
@@ -54,7 +58,7 @@
         programGroup = DiagnosticsAndQCProgramGroup.class
 )
 @DocumentedFeature
-public class CollectInsertSizeMetrics extends SinglePassSamProgram {
+public class    CollectInsertSizeMetrics extends SinglePassSamProgram {
```
yfarjoun: spurious spaces

src/main/java/picard/analysis/CollectInsertSizeMetrics.java
```diff
@@ -158,23 +166,18 @@
...
-                    String.valueOf(HISTOGRAM_WIDTH) ); //Histogram_WIDTH is passed because R automatically sets Histogram width to the last
-                                                         //bin that has data, which may be less than Histogram_WIDTH and confuse the user.
+            final List<String> plotArgs = new ArrayList<>();
+            Collections.addAll(plotArgs, OUTPUT.getAbsolutePath(), Histogram_FILE.getAbsolutePath(), INPUT.getName());
+
+            if (HISTOGRAM_WIDTH != null) {
+                plotArgs.add(String.valueOf(HISTOGRAM_WIDTH));
+            }
+            else if (MIN_HISTOGRAM_WIDTH != null) {
+                final int max = (int) file.getAllHistograms().stream().mapToDouble(Histogram::getMax).max().getAsDouble();
```
yfarjoun: In the absence of MIN_HISTOGRAM_WIDTH, this would make the largest insert-size that appears in the histogram the de-facto width, right?
 If so that is a departure from the current behavior that looks at the median and add a few deviations...

tfenne: A couple of things.  Firstly, this line is only triggered if a) `HISTOGRAM_WIDTH` is null and b) `MIN_HISTOGRAM_WIDTH` is not null.  This is akin to the logic in `if (HISTOGRAM_WIDTH == null)` previously, where no 4th argument is passed to limit the range of the plot.  So it's actually identical to the previous behaviour.

The way the limiting works (I had to look to remind myself when writing this PR) is that the data in the `Histogram` is actually trimmed using median + deviations.  So while the plot does fall back to using the full range of the data written to disk, the range of that data has already been truncated.

yfarjoun: OK. thanks for the detail.

yfarjoun:
@tfenne you can merge this yourself, right?

tfenne:
@yfarjoun I forgot I could do that!

PR #1376, opened 2019-08-09
Improvements to the locus iteration

jonn-smith:
* Adds in a progress meter to `CollectSamErrorMetrics`.
* Now will jump to the next read or variant rather than iterating through each position even if it does not exist in either.  This saves time for the use case that doesn't cover a whole genome / exome.

### DIFF
--- a/src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
+++ b/src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
@@ -180,7 +180,7 @@ public class CollectSamErrorMetrics extends CommandLineProgram {
 
     @Argument(
             fullName = "INTERVAL_ITERATOR",
-            doc = "Iterate through the file assuming they are a pre-created subset interval of the full genome.  " +
+            doc = "Iterate through the file assuming it consists of a pre-created subset interval of the full genome.  " +
                     "This enables fast processing of files with reads at disperate parts of the genome.  " +
                     "Requires that the provided VCF file is indexed. ",
             optional = true
@@ -279,7 +279,7 @@ public class CollectSamErrorMetrics extends CommandLineProgram {
             vcfFileReader = new VCFFileReader(VCF, true);
             // Make sure we can query our file for interval mode:
             if (!vcfFileReader.isQueryable()) {
-                throw new PicardException("Cannot query VCF File!  VCF Files must be queryable!");
+                throw new PicardException("Cannot query VCF File!  VCF Files must be queryable!  Please index input VCF and re-run.");
             }
         }
         else {
@@ -544,10 +544,12 @@ public class CollectSamErrorMetrics extends CommandLineProgram {
         }
 
         //same SequenceIndex, can compare by genomic position
-        if (locus.getPosition() < variantContext.getStart())
+        if (locus.getPosition() < variantContext.getStart()) {
             return variantContext.getStart() - locus.getPosition();
-        if (locus.getPosition() > variantContext.getEnd())
+        }
+        if (locus.getPosition() > variantContext.getEnd()) {
             return variantContext.getEnd() - locus.getPosition();
+        }
         return 0;
     }
 
@@ -567,8 +569,7 @@ public class CollectSamErrorMetrics extends CommandLineProgram {
             try (final CloseableIterator<VariantContext> vcfIterator = vcfFileReader.query(locusInfo)) {
 
                 while (vcfIterator.hasNext()) {
-                    final VariantContext variantContext = vcfIterator.next();
-                    if (variantContext.isFiltered()) {
+                    if (vcfIterator.next().isFiltered()) {
                         continue;
                     }
                     overlaps = true;
--- a/src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
+++ b/src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
@@ -91,6 +91,8 @@ import java.util.stream.Collectors;
 @DocumentedFeature
 public class CollectSamErrorMetrics extends CommandLineProgram {
 
+    // =====================================================================
+
     private static final int MAX_DIRECTIVES = ReadBaseStratification.Stratifier.values().length + 1;
     private static final Log log = Log.getInstance(CollectSamErrorMetrics.class);
 
@@ -169,11 +171,43 @@ public class CollectSamErrorMetrics extends CommandLineProgram {
     public double PROBABILITY = 1;
 
     @Argument(
-            fullName = "progressStepInterval",
+            fullName = "PROGRESS_STEP_INTERVAL",
             doc = "The interval between which progress will be displayed.",
             optional = true
     )
-    public int progressStepInterval = 100000;
+    public int PROGRESS_STEP_INTERVAL = 100000;
+
+    @Argument(
+            fullName = "INTERVAL_ITERATOR",
+            doc = "Iterate through the files assuming they are a pre-created subset interval of the full genome.  " +
+                    "This enables fast processing of files with reads at disperate parts of the genome.  " +
+                    "Requires that the provided VCF file is indexed. ",
+            optional = true
+    )
+    public boolean INTERVAL_ITERATOR = false;
+
+    // =====================================================================
+
+    /** Random object from which to pull pseudo-random numbers.  Initialized in {@link #initializeAggregationState()}.*/
+    private Random random;
+
+    /** Aggregator list to store errors.  Initialized in {@link #initializeAggregationState()}.*/
+    private Collection<BaseErrorAggregation> aggregatorList;
+
+    /** Logger with which to keep the user apprised of our progress.  Initialized in {@link #initializeAggregationState()}.*/
+    private ProgressLogger progressLogger;
+
+    /** Count of the total number of loci visited.  Initialized in {@link #initializeAggregationState()}.*/
+    private long nTotalLoci;
+
+    /** Count of the number of skipped loci.  Initialized in {@link #initializeAggregationState()}.*/
+    private long nSkippedLoci;
+
+    /** Count of the number of processed loci.  Initialized in {@link #initializeAggregationState()}.*/
+    private long nProcessedLoci;
+
+    /** The {@link SAMSequenceDictionary} for the reference, variants, and reads being processed. */
+    private SAMSequenceDictionary sequenceDictionary;
 
     // =====================================================================
 
@@ -228,68 +262,30 @@ public class CollectSamErrorMetrics extends CommandLineProgram {
         }
     }
 
-    @Override
-    protected int doWork() {
+    /**
+     * Process our input data by iterating through input variants and reads.
+     * Useful for going through whole genomes / exomes, but slow for sets of reads not beginning at the start of chr1.
+     *
+     * @return {@code 0} if all inputs were processed successfully.  {@code 1} if there was any kind of error.
+     */
+    private int processDataNormalIteration() {
 
-        final Random random = new Random(42);
-
-        final ProgressLogger progressLogger = new ProgressLogger(log, progressStepInterval);
-        long nTotalLoci = 0;
-        long nSkippedLoci = 0;
-        long nProcessedLoci = 0;
-
-        IOUtil.assertFileIsReadable(INPUT);
-        IOUtil.assertFilesAreReadable(INTERVALS);
-
-        final Collection<BaseErrorAggregation> aggregatorList = getAggregatorList();
         // Open up the input resources:
         try (
-                final SamReader sam = SamReaderFactory.makeDefault()
-                        .referenceSequence(REFERENCE_SEQUENCE)
-                        .setOption(SamReaderFactory.Option.INCLUDE_SOURCE_IN_RECORDS, true)
-                        .open(INPUT);
-                final ReferenceSequenceFileWalker referenceSequenceFileWalker = new ReferenceSequenceFileWalker(REFERENCE_SEQUENCE);
-                final VCFFileReader vcfFileReader = new VCFFileReader(VCF, true)
+            final SamReader sam = SamReaderFactory.makeDefault()
+                    .referenceSequence(REFERENCE_SEQUENCE)
+                    .setOption(SamReaderFactory.Option.INCLUDE_SOURCE_IN_RECORDS, true)
+                    .open(INPUT);
+
+            final ReferenceSequenceFileWalker referenceSequenceFileWalker = new ReferenceSequenceFileWalker(REFERENCE_SEQUENCE);
+
+            final PeekableIterator<VariantContext> vcfIterator = new PeekableIterator<>(
+                    VCF == null ? Collections.emptyIterator() : new VCFFileReader(VCF, true).iterator())
         ) {
+            // Go through our reads and variants now:
+            final SamLocusAndReferenceIterator iterator = createSamLocusAndReferenceIterator(sam, referenceSequenceFileWalker);
 
-            // Make sure we can query our file:
-            if (!vcfFileReader.isQueryable()) {
-                throw new PicardException("Cannot query VCF File!  VCF Files must be queryable!");
-            }
-
-            final SAMSequenceDictionary sequenceDictionary = referenceSequenceFileWalker.getSequenceDictionary();
-            if (sam.getFileHeader().getSortOrder() != SAMFileHeader.SortOrder.coordinate) {
-                throw new PicardException("Input BAM must be sorted by coordinate");
-            }
-
-            sequenceDictionary.assertSameDictionary(sam.getFileHeader().getSequenceDictionary());
-
-            final IntervalList regionOfInterest = getIntervals(sequenceDictionary);
-
-            log.info("Getting SamLocusIterator");
-
-            final SamLocusIterator samLocusIterator = new SamLocusIterator(sam, regionOfInterest);
-
-            // Now go through and compare information at each of these sites
-            samLocusIterator.setEmitUncoveredLoci(false);
-            samLocusIterator.setMappingQualityScoreCutoff(MIN_MAPPING_Q);
-            samLocusIterator.setQualityScoreCutoff(MIN_BASE_Q);
-
-            log.info("Using " + aggregatorList.size() + " aggregators.");
-
-            aggregatorList.forEach(la ->
-                    IOUtil.assertFileIsWritable(new File(OUTPUT + la.getSuffix())));
-
-            // iterate over loci
-            log.info("Starting iteration over loci");
-
-            final SamLocusAndReferenceIterator iterator = new SamLocusAndReferenceIterator(referenceSequenceFileWalker, samLocusIterator);
-
-            // This hasNext() call has side-effects. It loads up the index and makes sure that the iterator is really ready for
-            // action. Calling this allows for the logging to be more accurate.
-            iterator.hasNext();
             log.info("Really starting iteration now.");
-
             for (final SAMLocusAndReference info : iterator) {
                 if (random.nextDouble() > PROBABILITY) {
                     continue;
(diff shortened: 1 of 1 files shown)


src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -162,6 +168,15 @@
     @Argument(shortName = "P", doc = "The probability of selecting a locus for analysis (for downsampling).", optional = true)
     public double PROBABILITY = 1;
 
+    @Argument(
+            fullName = "progressStepInterval",
+            doc = "The interval between which progress will be displayed.",
+            optional = true
+    )
+    public int progressStepInterval = 100000;
```
yfarjoun: the standard for Picard arguments is UPPER_CASED.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -232,12 +244,19 @@ protected int doWork() {
...
         try (
-                final SamReader sam = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(INPUT);
+                final SamReader sam = SamReaderFactory.makeDefault()
+                        .referenceSequence(REFERENCE_SEQUENCE)
+                        .setOption(SamReaderFactory.Option.INCLUDE_SOURCE_IN_RECORDS, true)
+                        .open(INPUT);
                 final ReferenceSequenceFileWalker referenceSequenceFileWalker = new ReferenceSequenceFileWalker(REFERENCE_SEQUENCE);
-                final PeekableIterator<VariantContext> vcfIterator = new PeekableIterator<>(
-                        VCF == null ? Collections.emptyIterator() : new VCFFileReader(VCF, true).iterator())
+                final VCFFileReader vcfFileReader = new VCFFileReader(VCF, true)
```
yfarjoun: why did you remove the option to not provide a VCF input? That would be a change in API...

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -232,12 +244,19 @@ protected int doWork() {
...
         ) {
 
+            // Make sure we can query our file:
+            if (!vcfFileReader.isQueryable()) {
```
yfarjoun: This is an API change. we may think that it's OK, but in that case, the PR message needs to state so clearly, and the error message needs to inform the user how to move forward (e.g. Please index the file using ...)

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -328,21 +347,41 @@ private void writeMetricsFileForAggregator(final BaseErrorAggregation locusAggre
...
+     * @param locusInfo          a {@link SamLocusIterator.LocusInfo} at which to examine the variants
      * @return true if there's a variant over the locus, false otherwise.
      */
-    private static boolean advanceIteratorAndCheckLocus(final PeekableIterator<VariantContext> vcfIterator, final Locus locus, final SAMSequenceDictionary sequenceDictionary) {
-        while (vcfIterator.hasNext() && (vcfIterator.peek().isFiltered() ||
-                CompareVariantContextToLocus(sequenceDictionary, vcfIterator.peek(), locus) < 0)) {
-            vcfIterator.next();
+    private static boolean checkLocus(final VCFFileReader vcfFileReader, final SamLocusIterator.LocusInfo locusInfo) {
+
+        boolean overlaps = false;
```
yfarjoun: why so many empty lines?

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -328,21 +347,41 @@ private void writeMetricsFileForAggregator(final BaseErrorAggregation locusAggre
...
+
+        if (locusInfo != null) {
```
yfarjoun: querying every locus is faster than iterating over the entire VCF? I'm very confused about this.

jonn-smith: In giving dbSNP to filter out known variants, by default this tool will iterate over the VCF and ignore where the reads are.

This may be fine for when you go over a whole exome, but if I have a read on chr22 I don't want to have to wait for it to get there.  Since our VCFs are indexed by default it is trivial to just jump to the next read or variant, whichever is closer.  This saves a lot of time.

I don't think it's reasonable to have to subset dbSNP every time I want to test a specific region of the genome.  If I had to do this, then virtually every time I call this tool I would be subsetting dbSNP again by the extents of the reads.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: Fixed!

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: Oversight.  Fixed.

jonn-smith:
@yfarjoun I'll make this an option so it doesn't always happen, but I've found it universally useful.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: I'll make this feature optional so that it only requires a queryable VCF if it is enabled.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: I've found that leaving a little extra whitespace around pays dividends in readability.  

I'll remove a few.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: @yfarjoun Actually as of 7/4/2018 VCF is not an optional argument.  (See line 147).

So this is not an API change.

jonn-smith:
@yfarjoun OK, first round complete - back to you for round 2!

I updated it to have the new iteration as an optional argument.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -213,73 +262,99 @@ protected boolean requiresReference() {
...
-            log.info("Getting SamLocusIterator");
+                if (MAX_LOCI != 0 && nProcessedLoci >= MAX_LOCI) {
+                    log.warn("Early stopping due to having processed MAX_LOCI loci.");
+                    break;
+                }
+            }
 
-            final SamLocusIterator samLocusIterator = new SamLocusIterator(sam, regionOfInterest);
+        } catch (IOException e) {
+            log.error(e, "A problem occurred:");
```
yfarjoun: I know you just copied the code..could you fix this by including the error message from the exception? here and below as well.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -213,73 +262,99 @@ protected boolean requiresReference() {
...
-        final Collection<BaseErrorAggregation> aggregatorList = getAggregatorList();
         // Open up the input resources:
         try (
-                final SamReader sam = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(INPUT);
-                final ReferenceSequenceFileWalker referenceSequenceFileWalker = new ReferenceSequenceFileWalker(REFERENCE_SEQUENCE);
-                final PeekableIterator<VariantContext> vcfIterator = new PeekableIterator<>(
-                        VCF == null ? Collections.emptyIterator() : new VCFFileReader(VCF, true).iterator())
+            final SamReader sam = SamReaderFactory.makeDefault()
+                    .referenceSequence(REFERENCE_SEQUENCE)
+                    .setOption(SamReaderFactory.Option.INCLUDE_SOURCE_IN_RECORDS, true)
```
yfarjoun: why are you putting the file source into the reads? you don't seems to be using this anywhere...

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -299,14 +374,113 @@ protected int doWork() {
...
+
         log.info("Iteration complete, generating metric files");
 
         aggregatorList.forEach(this::writeMetricsFileForAggregator);
 
         log.info(String.format("Examined %d loci, Processed %d loci, Skipped %d loci.\n" +
                 "Computation took %d seconds.", nTotalLoci, nProcessedLoci, nSkippedLoci, progressLogger.getElapsedSeconds()));
 
-        return 0;
+        return returnValue;
```
yfarjoun: at this point `returnValue==0`, so why change?

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -299,14 +374,113 @@ protected int doWork() {
...
+     * @return An open {@link SamLocusAndReferenceIterator} ready to iterate.
+     */
+    private SamLocusAndReferenceIterator createSamLocusAndReferenceIterator(final SamReader sam, final ReferenceSequenceFileWalker referenceSequenceFileWalker) {
+
+        sequenceDictionary = referenceSequenceFileWalker.getSequenceDictionary();
+        if (sam.getFileHeader().getSortOrder() != SAMFileHeader.SortOrder.coordinate) {
+            throw new PicardException("Input BAM must be sorted by coordinate");
+        }
+
+        // Make sure our reference and reads have teh same sequence dictionary:
```
yfarjoun:
```suggestion
        // Make sure our reference and reads have the same sequence dictionary:
```

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -299,14 +374,113 @@ protected int doWork() {
...
+
+    @Override
+    protected int doWork() {
+
+        // Initialize our iteration:
+        initializeAggregationState();
+
+        // Make sure we can read our files:
+        IOUtil.assertFileIsReadable(INPUT);
+        IOUtil.assertFilesAreReadable(INTERVALS);
```
yfarjoun: might as well assert that VCF is readable.

yfarjoun: and reference too

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: Fixed.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: Fixed.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: In general it's clearer to avoid hard-coded values, even simple ones.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: Fixed.

jonn-smith:
@yfarjoun OK, pushed latest changes and ready for round 3.  I consolidated the reused code and broke out some methods for handling the default / iterative cases differently.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -296,17 +374,115 @@ protected int doWork() {
             }
 
         } catch (IOException e) {
-            log.error(e, "A problem occurred:");
+            log.error(e, "A problem occurred:", e.getMessage());
```
yfarjoun: yikes. thanks.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
yfarjoun: in that case, can you swap the if and only `return returnValue once in the end?

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -345,6 +521,69 @@ private static boolean advanceIteratorAndCheckLocus(final PeekableIterator<Varia
...
+                overlaps = true;
+
+                boolean allFiltered = true;
+
+                while (vcfIterator.hasNext()) {
+                    final VariantContext vcf = vcfIterator.next();
+                    if (vcf.isFiltered()) {
+                        continue;
+                    }
+                    allFiltered = false;
```
yfarjoun: you can break here right?

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -345,6 +521,69 @@ private static boolean advanceIteratorAndCheckLocus(final PeekableIterator<Varia
...
+     */
+    private static boolean checkLocus(final VCFFileReader vcfFileReader, final SamLocusIterator.LocusInfo locusInfo) {
+        boolean overlaps = false;
+
+        if (locusInfo != null) {
+            try (final CloseableIterator<VariantContext> vcfIterator = vcfFileReader.query(locusInfo)) {
+
+                overlaps = true;
+
+                boolean allFiltered = true;
```
yfarjoun: I think this would be more clear if you reversed the logic and looked for `anyPassing` rather than `allFiltered`....

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -162,6 +171,53 @@
...
+    @Argument(
+            fullName = "PROGRESS_STEP_INTERVAL",
+            doc = "The interval between which progress will be displayed.",
+            optional = true
+    )
+    public int PROGRESS_STEP_INTERVAL = 100000;
+
+    @Argument(
+            fullName = "INTERVAL_ITERATOR",
+            doc = "Iterate through the files assuming they are a pre-created subset interval of the full genome.  " +
```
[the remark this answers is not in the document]
jonn-smith: Fixed.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -213,73 +269,95 @@ protected boolean requiresReference() {
...
+    /**
+     * Check if the given locus overlaps a variant from our variant source.
+     *
+     * If running in {@link #INTERVAL_ITERATOR} mode:
+     * This will query the input variant file for overlapping variants, rather than just iterating through them.
+     * It should be only a little slower than the normal iteration, and much faster if a subset of a genome is used.
+     *
+     * If running in default mode:
+     * Process our input data by iterating through input variants and reads.
+     * Useful for going through whole genomes / exomes, but slow for sets of reads not beginning at the start of chr1.
```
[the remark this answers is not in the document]
jonn-smith: Essentially yes, though for user documentation it should be explained more.  I've now added more description.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: No - It's not guaranteed that there will be only 1 variant that overlaps the locus in the given variant file.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: sure.

jonn-smith: In fact, I simplified the method a bit.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -180,7 +180,7 @@
 
     @Argument(
             fullName = "INTERVAL_ITERATOR",
-            doc = "Iterate through the files assuming they are a pre-created subset interval of the full genome.  " +
+            doc = "Iterate through the file assuming they are a pre-created subset interval of the full genome.  " +
```
yfarjoun: ...." assuming it consists of a pre-created subset ...."

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -564,20 +566,13 @@ private static boolean checkLocus(final VCFFileReader vcfFileReader, final SamLo
...
 
-                overlaps = true;
-
-                boolean allFiltered = true;
-
                 while (vcfIterator.hasNext()) {
-                    final VariantContext vcf = vcfIterator.next();
-                    if (vcf.isFiltered()) {
+                    final VariantContext variantContext = vcfIterator.next();
+                    if (variantContext.isFiltered()) {
```
yfarjoun: you could inline and skip the local variable, right?

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -213,73 +269,100 @@ protected boolean requiresReference() {
...
+     * Initialize the source for {@link VariantContext} information.
+     * If running in {@link #INTERVAL_ITERATOR} mode, will initialize the {@link #vcfFileReader}.
+     * Otherwise, will initialize the {@link #vcfIterator}.
+     */
+    private void initializeVcfDataSource() {
+        if ( INTERVAL_ITERATOR ) {
+            vcfFileReader = new VCFFileReader(VCF, true);
+            // Make sure we can query our file for interval mode:
+            if (!vcfFileReader.isQueryable()) {
+                throw new PicardException("Cannot query VCF File!  VCF Files must be queryable!");
```
yfarjoun: "...Please index input VCF and re-run."

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
```diff
@@ -345,6 +523,62 @@ private static boolean advanceIteratorAndCheckLocus(final PeekableIterator<Varia
...
+     */
+    public static int CompareVariantContextToLocus(final SAMSequenceDictionary dictionary, final VariantContext variantContext, final Locus locus) {
+
+        final int indexDiff = dictionary.getSequenceIndex(variantContext.getContig()) - locus.getSequenceIndex();
+        if (indexDiff != 0) {
+            return indexDiff < 0 ? Integer.MIN_VALUE : Integer.MAX_VALUE;
+        }
+
+        //same SequenceIndex, can compare by genomic position
+        if (locus.getPosition() < variantContext.getStart())
```
yfarjoun: add {} to avoid future bugs.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: Fixe.d

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: Yes.  Fixed.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: Fixed.

src/main/java/picard/sam/SamErrorMetric/CollectSamErrorMetrics.java
jonn-smith: Fixed.

PR #1394, opened 2019-09-05
Throw exception in MergeBamAlignments if UNMAPPED_BAM has mapped reads

kachulis:
### Description
Currently, MergeBamAlignments will run through without complaint and produce a merged bam if a bam with mapped reads is used in the UNMAPPED_BAM argument.  However, any read which was aligned to the reverse strand in both the UNMAPPED_BAM and the ALIGNED_BAM will have its bases reverse complemented, resulting in a record with alignment on the reverse stand but with bases recorded for the forward strand.  See `MergeBamAlignmentTest.testReverseStrandMappedReadInUnmappedBam()` on branch [ck_Test_MBA_ReverseStrandMappedReadInUnmappedBam](https://github.com/broadinstitute/picard/tree/ck_Test_MBA_ReverseStrandMappedReadInUnmappedBam) for example.
This PR adds a check that all records in UNMAPPED_BAM are unmapped, and throws an exception if it finds any that are mapped. 

#### Content

#### Review

### DIFF
--- a/src/test/java/picard/sam/MergeBamAlignmentTest.java
+++ b/src/test/java/picard/sam/MergeBamAlignmentTest.java
@@ -1927,7 +1927,7 @@ public class MergeBamAlignmentTest extends CommandLineProgramTest {
     }
 
     @Test(dataProvider = "mappedReadInUnmappedBamDataProvider", expectedExceptions = PicardException.class)
-    public void testSingleEndMappedReadInUnmappedBam(final boolean paired, final boolean firstUnMapped, final boolean secondUnMapped) throws IOException {
+    public void testMappedReadInUnmappedBam(final boolean paired, final boolean firstUnMapped, final boolean secondUnMapped) throws IOException {
         final SAMRecordSetBuilder samRecordSetBuilderUnmappedBam = new SAMRecordSetBuilder(true, SAMFileHeader.SortOrder.queryname);
         samRecordSetBuilderUnmappedBam.setRandomSeed(12345);
         final SAMFileHeader header = samRecordSetBuilderUnmappedBam.getHeader();
--- a/src/test/java/picard/sam/MergeBamAlignmentTest.java
+++ b/src/test/java/picard/sam/MergeBamAlignmentTest.java
@@ -35,6 +35,7 @@ import htsjdk.samtools.SAMProgramRecord;
 import htsjdk.samtools.SAMReadGroupRecord;
 import htsjdk.samtools.SAMRecord;
 import htsjdk.samtools.SAMRecordIterator;
+import htsjdk.samtools.SAMRecordSetBuilder;
 import htsjdk.samtools.SAMTag;
 import htsjdk.samtools.SamPairUtil;
 import htsjdk.samtools.SamReader;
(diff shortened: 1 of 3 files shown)


src/test/java/picard/sam/MergeBamAlignmentTest.java
```diff
@@ -23,22 +23,7 @@
...
-import htsjdk.samtools.SAMFileWriterFactory;
-import htsjdk.samtools.SAMProgramRecord;
-import htsjdk.samtools.SAMReadGroupRecord;
-import htsjdk.samtools.SAMRecord;
-import htsjdk.samtools.SAMRecordIterator;
-import htsjdk.samtools.SAMTag;
-import htsjdk.samtools.SamPairUtil;
-import htsjdk.samtools.SamReader;
-import htsjdk.samtools.SamReaderFactory;
+import htsjdk.samtools.*;
```
yfarjoun: There seems to be a preference for explicit imports, please revert this. (you can also setup intelliJ so that it doesn't do this for you)

src/test/java/picard/sam/MergeBamAlignmentTest.java
```diff
@@ -1930,5 +1915,69 @@ public void testHeaderFromMappedBreaks(final String filename) throws IOException
...
+        output.deleteOnExit();
+        final List<String> args = new ArrayList<>(Arrays.asList(
+                "UNMAPPED_BAM=" + unmappedSam.getAbsolutePath(),
+                "ALIGNED_BAM=" + alignedSam.getAbsolutePath(),
+                "OUTPUT=" + output.getAbsolutePath(),
+                "REFERENCE_SEQUENCE=" + fasta.getAbsolutePath())
+        );
+
+        //should throw PicardException when run because mapped read found in UNMAPPED_BAM input
+        runPicardCommandLine(args);
```
yfarjoun: Could you add a test-case that doesn't fail to verify that the test isn't bogus (maybe it fails no matter what you do!)

src/test/java/picard/sam/MergeBamAlignmentTest.java
```diff
@@ -1930,5 +1931,73 @@ public void testHeaderFromMappedBreaks(final String filename) throws IOException
...
+        final List<String> args = new ArrayList<>(Arrays.asList(
+                "UNMAPPED_BAM=" + unmappedSam.getAbsolutePath(),
+                "ALIGNED_BAM=" + alignedSam.getAbsolutePath(),
+                "OUTPUT=" + output.getAbsolutePath(),
+                "REFERENCE_SEQUENCE=" + fasta.getAbsolutePath())
+        );
+
+        //should throw PicardException when run because mapped read found in UNMAPPED_BAM input
+        if (shouldThrowException) {
+            Assert.assertThrows(PicardException.class, () -> runPicardCommandLine(args));
```
yfarjoun: this didn't help...

refactor the insides of the test into a private static method, then have two tests call it, one which expects an exception and one that doesn't.

kachulis: I may be confused about your comment, but I think the additional test case, where all reads are unmapped and `shouldThrowException` is set to false, does distinguish between properly failing and always failing.  The test no longer expects an exception through the annotation, so it is testing that an exception is thrown when it should be, and not when it's shouldn't.  
But I will refactor as requested, seems like a more elegant solution anyway.

yfarjoun: upon second read, I agree that your solution works... :-) sorry for the confusion.

PR #1420, opened 2019-11-03
Yf optimize query for independent replicate metrics

yfarjoun:
CollectIndependentReplicateMetrics was using a non-optimized query which led to be slow runtimes...this significantly improves runtime, enables running on single-ended reads and also makes the code little nicer.
----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/analysis/replicates/CollectIndependentReplicateMetrics.java
+++ b/src/main/java/picard/analysis/replicates/CollectIndependentReplicateMetrics.java
@@ -198,7 +198,7 @@ public class CollectIndependentReplicateMetrics extends CommandLineProgram {
 
         log.info("Querying BAM for sites.");
 
-        final SAMRecordIterator samRecordIterator = in.query(intervalAlleleMap.keySet().toArray(new QueryInterval[intervalAlleleMap.size()]), false);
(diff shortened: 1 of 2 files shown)


src/main/java/picard/analysis/replicates/CollectIndependentReplicateMetrics.java
```diff
@@ -159,6 +160,13 @@ protected int doWork() {
         IOUtil.assertFileIsReadable(VCF);
         IOUtil.assertFileIsReadable(INPUT);
 
+        // get an iterator to reads that overlap the heterozygous sites
+        final SamReader in = SamReaderFactory.makeDefault().open(INPUT);
+
+        if (!in.hasIndex()) {
```
gbggrant: Shouldn't this (and the IOUtil.assertFile...) be in a CustomCommandLineValidation method?

yfarjoun: the problem is that figuring out if a file has an index involved opening a reader against it...which can be time consuming (think about doing it over NIO)...so I figured I prefer only to open the file once when creating the reader...

PR #1430, opened 2019-11-19
Don't not add a period to output basenames

tlangs:
### Description

A change was introduced last year to `CheckFingerprint` which made it only add a period to the end of the `OUTPUT` argument if it did not already end with a period. This causes errors in our pipeline when sample aliases end in a period, because we derive our cromwell outputs based on the sample alias concatenated with "fingerprinting_summary_metrics" and "fingerprinting_detail_metrics". If sample_alias is `MyCoolSampleG.T.`, the output files should be `MyCoolSampleG.T..fingerprinting_summary_metrics` and `MyCoolSampleG.T..fingerprinting_detail_metrics`

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/fingerprint/CheckFingerprint.java
+++ b/src/main/java/picard/fingerprint/CheckFingerprint.java
@@ -223,9 +223,7 @@ public class CheckFingerprint extends CommandLineProgram {
             outputDetailMetricsFile = DETAIL_OUTPUT;
             outputSummaryMetricsFile = SUMMARY_OUTPUT;
         } else {
-            if (!OUTPUT.endsWith(".")) {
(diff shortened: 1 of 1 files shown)


lbergelson:
@tlangs It's pretty confusing to have a sample name that ends in a period.  Can we prevent that?

tlangs:
The tool should respect what it's given, no matter if it looks ugly. Changing the output name is not documented and results in a lot of confusion when the tool spits out files with different names from what was specified and expected. I don't think its up to us dictate how people name their files.

lbergelson:
I agree we should fix the tool, but it would also be nice if we did dictate what people put in their sample names so they didn't have dumb things like a period as the last character.

Issue #1441, opened 2019-12-11
MarkDuplicates Missing Optical Duplicates


whitwham:
### Affected tool(s)
MarkDuplicates OPTICAL_DUPLICATE_PIXEL_DISTANCE=2500 TAGGING_POLICY=All

### Affected version(s)
Latest test on 2.21.4-2-ga3021a7-SNAPSHOT.  Older versions display the same problem.

### Description 
For HiSeq X runs the coordinate part of the query name can be greater than 32K.  The coordinates are stored using PhysicalLocationShort which converts the read in integer to a short.  As you would expect, integers over 32767 become negative.  This leads duplicates that should taken as being physically close together instead being calculated as far apart.

From my example file the numbers should look like this in the distance calculation:
abs (33023 - 32760) = 263
instead become:
abs (-32513 - 32760) = 65273

Which is a bit bigger than the optical duplicate distance of 2500.
(_This is from the calculations in closeEnough in OpticalDuplicateFinder_)

Any duplicates with X/Y coordinates around the short size boundary stand a chance of not being identified as optical duplicates.  In the production file I was testing on that amounted to about 22000 missed optical duplicates.  The included example file is based off the real data.  

#### Steps to reproduce
MarkDuplicates I=short_test.sam O=short_test_out.sam OPTICAL_DUPLICATE_PIXEL_DISTANCE=2500 M=short_test_stat.txt TAGGING_POLICY=All TAG_DUPLICATE_SET_MEMBERS=true

Input file (added .txt suffix to allow upload to GitHub):
[short_test.sam.txt](https://github.com/broadinstitute/picard/files/3950144/short_test.sam.txt)
Output file:
[short_test_out.sam.txt](https://github.com/broadinstitute/picard/files/3950145/short_test_out.sam.txt)

#### Expected behavior
The duplicates should be tagged DT:Z:SQ.

#### Actual behavior
The duplicates are tagged DT:Z:LB.

yfarjoun:
Thanks for making this an official issue. 

If I understand correctly, the effect is localized to a few bands of coordinates where, as your example shows, one read is under the unsigned-short maximum value and the other is over it.  

This has been discussed offline for a while, but it was deemed not big enough of an issue to warrant the increase in memory and thus disk space/file handles needed to mark duplicates. 

[...]

PR #1442, opened 2019-12-13
Partial fix of coordinate casting to short in MarkDuplicates

yfarjoun:
- Fixed short overflow issue in MarkDuplicates by internally dividing the values so that they are not so large, and a smarter distance metric that is aware of the possibility of short overflow

- Included example reads from #1441 as test

### Description

----

#### Content

#### Review


yfarjoun:
@jamesemery this will require a parallel change to GATK's MarkDuplicatesSpark to keep them in sync...right?

jamesemery:
@yfarjoun Yes we will. This shouldn't be too hard though as the optical duplicate finder is already called into by Spark so it would simply be a matter of making sure the scale factor is applied and porting the test. I'll make a ticket to track it and when we bump our picard version we can push this change.

yfarjoun:
@jamesemery could you review this PR?

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -483,14 +483,15 @@ private void buildSortedReadEndLists(final boolean useBarcodes) {
         log.info("Will retain up to " + maxInMemory + " data points before spilling to disk.");
 
         final ReadEndsForMarkDuplicatesCodec fragCodec, pairCodec, diskCodec;
+        final double scale = OPTICAL_DUPLICATE_PIXEL_DISTANCE/(double) OpticalDuplicateFinder.DEFAULT_OPTICAL_DUPLICATE_DISTANCE;
```
jamesemery: In the typical use case this might actually make a difference since we are compressing the pixel distance by 2500/100 or 25 and then int scaling, which will end up introducing approximately a 1% truncation in every case. That is probably fine given how unlikely it is to substantially change the optical duplicate finding. I do not however think this should not be defined by `OpticalDuplicateFinder.DEFAULT_OPTICAL_DUPLICATE_DISTANCE` and we should fix some other variable instead since the error range shouldn't really have anything to do with the default pixel distance. 

It would also seem prudent to ensure that we don't ever set a value below 1 here (i.e. we don't ever want to "inflate" the pixel coordinates if the user selects a pixel distance < 100).

yfarjoun: I was trying to think of a scaling factor that keeps the default behaviour the same....I guess I could add an additional argument, but I thought this admitted hack is good enough....if there are other ideas, I'm happy to consider.

jamesemery: That seems reasonable. I do think it seems bizzare that if we were to change the default to say 2500 (which I think is the standard for most of our pipelines) that we would be compressing by a factor of 25 less than we were before which could cause confusion of unintended consequences. I think we should make a separate variable that pegs the to 100 to make it a little clearer why we are doing it this way. I still think we should avoid ever scaling by fractional value however. 

 `final double scale = Math.max(OPTICAL_DUPLICATE_PIXEL_DISTANCE/(double) OpticalDuplicateFinder.OPTICAL_DUPLICATE_PIXEL_SCALING_FACTOR, 1.0);`

yfarjoun:
@jamesemery another look?

src/main/java/picard/sam/markduplicates/util/OpticalDuplicateFinder.java
```diff
@@ -327,16 +330,19 @@ private PhysicalLocation keeperOrNull(final List<? extends PhysicalLocation> lis
...
+                lhs.getReadGroup() == rhs.getReadGroup() &&      // must be in the same RG to be optical duplicates
+                lhs.getTile() == rhs.getTile() &&      // and the same tile
+                closeEnoughShort(lhs, rhs, distance);
     }
 
     private boolean closeEnoughShort(final PhysicalLocation lhs, final PhysicalLocation rhs, final int distance) {
         return lhs != rhs &&
-                Math.abs(lhs.getX() - rhs.getX()) <= distance &&
-                Math.abs(lhs.getY() - rhs.getY()) <= distance;
+                // Since the X and Y coordinates are constrained to short values, they can also become negative when
```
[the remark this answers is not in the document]
jamesemery: I had to do some arithmetic to convince myself that this works. Assuming your goal is to make it so that both 32768 and 32767 adjacent without disturbing the  65536 and 65535 adjacency I think this succeeded. However I do not think it is necessary to perform the operation forwards and backwards, the absolute value operation (even accounting for overflowing from the shorts) will not be changed by flipping the order by virtue of the fact that this calculation is being performed in integer space whereas the underlying data overflowed in short space. Unless your scaling factor is in the region of `Short.MAX_VALUE/2` you should never have an overflow here thus you should be able to get away without flipping the comparison.

PR #1458, opened 2020-01-28
Fixed a bug where the merging of fingerprints changes

yfarjoun:
The original list of fingerprints leading to incorrect results in the self-LOD case when using Crosscheck without SECOND_INPUT.

HT: David Wine and David Loginov for reporting.

### Description

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/fingerprint/HaplotypeProbabilities.java
+++ b/src/main/java/picard/fingerprint/HaplotypeProbabilities.java
@@ -24,9 +24,8 @@
 
 package picard.fingerprint;
 
-import static java.lang.Math.log10;
-import static picard.util.MathUtil.multiply;
-import static picard.util.MathUtil.pNormalizeVector;
+import htsjdk.utils.ValidationUtils;
+import picard.util.MathUtil;
 
 /**
  * Abstract class for storing and calculating various likelihoods and probabilities
@@ -63,7 +62,7 @@ public abstract class HaplotypeProbabilities {
      * Returns the posterior probabilities using the population frequency as a prior.
      */
     public double[] getPosteriorProbabilities() {
-        return pNormalizeVector(multiply(getLikelihoods(), getPriorProbablities()));
+        return MathUtil.pNormalizeVector(MathUtil.multiply(getLikelihoods(), getPriorProbablities()));
     }
 
     /**
@@ -75,9 +74,9 @@ public abstract class HaplotypeProbabilities {
 
     public double[] getLogLikelihoods() {
         final double[] likelihoods = getLikelihoods();
-        final double[] lLikelihoods = new double[likelihoods.length];
-        for (int i = 0; i < likelihoods.length; ++i) {
-            lLikelihoods[i] = Math.log10(likelihoods[i]);
+        final double[] lLikelihoods = new double[NUM_GENOTYPES];
+        for (final Genotype g : Genotype.values()){
+            lLikelihoods[g.v] = Math.log10(likelihoods[g.v]);
         }
         return lLikelihoods;
     }
@@ -131,13 +130,13 @@ public abstract class HaplotypeProbabilities {
     /**
      * Merges in the likelihood information from the supplied haplotype probabilities object.
      */
-    public abstract void merge(final HaplotypeProbabilities other);
+    public abstract HaplotypeProbabilities merge(final HaplotypeProbabilities other);
 
     /**
      * Returns the index of the highest probability which can then be used to look up
      * DiploidHaplotypes or DiploidGenotypes as appropriate.
      */
-    int getMostLikelyIndex() {
+    private int getMostLikelyIndex() {
         final double[] probs = getPosteriorProbabilities();
 
         if (probs[0] > probs[1] && probs[0] > probs[2]) return 0;
@@ -190,17 +189,18 @@ public abstract class HaplotypeProbabilities {
 
     public double scaledEvidenceProbabilityUsingGenotypeFrequencies(final double[] genotypeFrequencies) {
         final double[] likelihoods = getLikelihoods();
-        assert (genotypeFrequencies.length == likelihoods.length);
+        ValidationUtils.validateArg(genotypeFrequencies.length == NUM_GENOTYPES, "provided genotype frequencies must be length 3");
+        ValidationUtils.validateArg(likelihoods.length == NUM_GENOTYPES, "internal liklihoods must be length 3");
 
         double result = 0;
-        for (int i = 0; i < likelihoods.length; ++i) {
-            result += likelihoods[i] * genotypeFrequencies[i];
+        for (Genotype g : Genotype.values()) {
+            result += likelihoods[g.v] * genotypeFrequencies[g.v];
         }
         return result;
     }
 
     public double shiftedLogEvidenceProbabilityUsingGenotypeFrequencies(final double[] genotypeFrequencies) {
-        return log10(scaledEvidenceProbabilityUsingGenotypeFrequencies(genotypeFrequencies));
+        return Math.log10(scaledEvidenceProbabilityUsingGenotypeFrequencies(genotypeFrequencies));
     }
 
     /**
@@ -252,8 +252,10 @@ public abstract class HaplotypeProbabilities {
                 secondBiggest = prob;
             }
         }
-        return log10(biggest) - log10(secondBiggest);
+        return Math.log10(biggest) - Math.log10(secondBiggest);
     }
+
+    abstract public HaplotypeProbabilities deepCopy();
 
     /**
      * Log10(P(evidence| haplotype)) for the 3 different possible haplotypes
@@ -272,4 +274,6 @@ public abstract class HaplotypeProbabilities {
             this.v = v;
         }
     }
+    // A bit of shorthand since writing out "genotype.values().length" can get old.
+    static final int NUM_GENOTYPES = Genotype.values().length;
 }
--- a/src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromContaminatorSequence.java
+++ b/src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromContaminatorSequence.java
@@ -25,9 +25,8 @@
 package picard.fingerprint;
 
 import htsjdk.samtools.util.QualityUtil;
+import htsjdk.utils.ValidationUtils;
 import picard.util.MathUtil;
-
-import static java.lang.Math.log10;
 
 /**
  * Represents the probability of the underlying haplotype of the contaminating sample given the data.
@@ -42,21 +41,31 @@ public class HaplotypeProbabilitiesFromContaminatorSequence extends HaplotypePro
 
     // for each model (contGenotype, mainGenotype) there's a likelihood of the data. These need to be collected separately
     // and only collated once all the data is in.
-    double[][] likelihoodMap = {{1, 1, 1}, {1, 1, 1}, {1, 1, 1}};
+    private final double[][] likelihoodMap = {{1, 1, 1}, {1, 1, 1}, {1, 1, 1}};
 
     public HaplotypeProbabilitiesFromContaminatorSequence(final HaplotypeBlock haplotypeBlock, final double contamination) {
         super(haplotypeBlock);
 
-        assert (contamination <= 1.0);
-        assert (contamination >= 0.0);
+        ValidationUtils.validateArg(contamination <= 1.0, () -> "contamination must be <=1, found " + contamination);
+        ValidationUtils.validateArg(contamination >= 0.0, () -> "contamination must be >=0, found " + contamination);
 
         this.contamination = contamination;
     }
 
-    /**
-     * Adds a base observation with the observed quality to the evidence for this haplotype
-     * based on the fact that the SNP is part of the haplotype.
-     */
+    public HaplotypeProbabilitiesFromContaminatorSequence(HaplotypeProbabilitiesFromContaminatorSequence other) {
+        super(other);
+
+        contamination = other.contamination;
+        for (final Genotype g : Genotype.values()) {
+            System.arraycopy(other.likelihoodMap[g.v], 0, likelihoodMap[g.v], 0, NUM_GENOTYPES);
+        }
+    }
+
+
+        /**
+         * Adds a base observation with the observed quality to the evidence for this haplotype
+         * based on the fact that the SNP is part of the haplotype.
+         */
     public void addToProbs(final Snp snp, final byte base, final byte qual) {
         assertSnpPartOfHaplotype(snp);
 
@@ -90,16 +99,21 @@ public class HaplotypeProbabilitiesFromContaminatorSequence extends HaplotypePro
 
     //a function needed to update the logLikelihoods from the likelihoodMap.
     private void updateLikelihoods() {
-        final double[] ll = new double[Genotype.values().length];
+        final double[] ll = new double[NUM_GENOTYPES];
         for (final Genotype contGeno : Genotype.values()) {
             // p(a | g_c) = \sum_g_m { P(g_m) \prod_i P(a_i| g_m, g_c)}
-            ll[contGeno.v] = log10(MathUtil.sum(MathUtil.multiply(this.getPriorProbablities(), likelihoodMap[contGeno.v])));
+            ll[contGeno.v] = Math.log10(MathUtil.sum(MathUtil.multiply(this.getPriorProbablities(), likelihoodMap[contGeno.v])));
         }
         setLogLikelihoods(ll);
     }
 
     @Override
-    public void merge(final HaplotypeProbabilities other) {
+    public HaplotypeProbabilitiesFromContaminatorSequence deepCopy() {
+        return new HaplotypeProbabilitiesFromContaminatorSequence(this);
+    }
+
+    @Override
+    public HaplotypeProbabilitiesFromContaminatorSequence merge(final HaplotypeProbabilities other) {
         super.merge(other);
 
         if (!this.getHaplotype().equals(other.getHaplotype())) {
@@ -119,6 +133,7 @@ public class HaplotypeProbabilitiesFromContaminatorSequence extends HaplotypePro
         for (final Genotype contGeno : Genotype.values()) {
             this.likelihoodMap[contGeno.v] = MathUtil.multiply(this.likelihoodMap[contGeno.v], o.likelihoodMap[contGeno.v]);
         }
+        return this;
     }
 
     @Override
--- a/src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromGenotype.java
+++ b/src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromGenotype.java
@@ -32,6 +32,16 @@ public class HaplotypeProbabilitiesFromGenotype extends HaplotypeProbabilities {
     private final Snp snp;
     private final double[] likelihoods;
 
+    @Override
+    public HaplotypeProbabilitiesFromGenotype deepCopy()  {
+        return new HaplotypeProbabilitiesFromGenotype(this);
+    }
+
+    @SuppressWarnings("CopyConstructorMissesField")
+    public HaplotypeProbabilitiesFromGenotype(final HaplotypeProbabilitiesFromGenotype other) {
+        this(other.snp, other.getHaplotype(), other.likelihoods[0], other.likelihoods[1], other.likelihoods[2]);
+    }
+
     public HaplotypeProbabilitiesFromGenotype(final Snp snp, final HaplotypeBlock haplotypeBlock,
                                               final double AA, final double Aa, final double aa) {
         super(haplotypeBlock);
@@ -42,18 +52,13 @@ public class HaplotypeProbabilitiesFromGenotype extends HaplotypeProbabilities {
     /** Returns the SNP who's genotype was used to construct the likelihoods. */
     @Override public Snp getRepresentativeSnp() { return snp; }
 
-
-    // TODO: this can't be right in general. At least one needs to divide by the prior to set things straight.
-    // TODO: The only saving grace is that this is normally used for cases where the priors are large and similar to each other.
-
-
     // simply returns the _likelihoods_ that were passed into the constructor.
     public double[] getLikelihoods() {
         return likelihoods;
     }
 
     @Override
-    public void merge(final HaplotypeProbabilities other) {
+    public HaplotypeProbabilitiesFromGenotype merge(final HaplotypeProbabilities other) {
         if (!this.getHaplotype().equals(other.getHaplotype())) {
             throw new IllegalArgumentException("Mismatched haplotypes in call to HaplotypeProbabilities.merge(): " +
                     getHaplotype() + ", " + other.getHaplotype());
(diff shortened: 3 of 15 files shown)


pshapiro4broad:
Is there a reason why you're using the `clone()` API to create a copy of something? I think it would be better to add a new method to `HaplotypeProbabilities` that has a default implementation and instead use that method to create a copy of the object. I don't see any benefit to using Java `clone()`, only downsides (such as `Cloneable` and the related clone exception). Am I missing something?

src/main/java/picard/fingerprint/HaplotypeProbabilitiesUsingLogLikelihoods.java
```diff
@@ -26,33 +26,46 @@
...
+    private double[] shiftedLogPosteriors = new double[Genotype.values().length];
 
     public HaplotypeProbabilitiesUsingLogLikelihoods(final HaplotypeBlock haplotypeBlock) {
         super(haplotypeBlock);
     }
 
+    @Override
+    public Object clone() throws CloneNotSupportedException {
+        final HaplotypeProbabilitiesUsingLogLikelihoods c = (HaplotypeProbabilitiesUsingLogLikelihoods)super.clone();
+        c.likelihoods = this.likelihoods.clone();
```
pshapiro4broad: Is there a reason why you don't use `System.arrayCopy()` here too? Or call `setLogLikelihoods()`? it would be nice to keep the field `final` if possible.

src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromGenotype.java
```diff
@@ -28,10 +28,17 @@
...
  */
-public class HaplotypeProbabilitiesFromGenotype extends HaplotypeProbabilities {
+public class HaplotypeProbabilitiesFromGenotype extends HaplotypeProbabilities implements Cloneable{
     private final Snp snp;
     private final double[] likelihoods;
 
+    @Override
+    public Object clone() throws CloneNotSupportedException {
+        HaplotypeProbabilitiesFromGenotype c = (HaplotypeProbabilitiesFromGenotype) super.clone();
+        System.arraycopy(likelihoods,0,c.likelihoods,0,likelihoods.length);
```
pshapiro4broad: more whitespace, please
```suggestion
        System.arraycopy(likelihoods, 0, c.likelihoods, 0, likelihoods.length);
```

dinvlad:
Awesome, this seems to have had the desired effect, in all aspects.

dinvlad:
@pshapiro4broad, are you suggesting that a copy constructor should be used here instead? My take is as long as our behavior doesn't change, either solution is fine by us.

src/main/java/picard/fingerprint/HaplotypeProbabilitiesUsingLogLikelihoods.java
yfarjoun: super.clone does a shallow copy...in order to have these arrays be actually different, I need to clone them. I could make `new` arrays and arrayCopy, but I'm not sure why that is better. I don't think it's possible to keep the field final...not the clone, anyway...but I'm happy to learn something new.

yfarjoun: It's actually, the other way around...I need to get rid of the other arrayCopy and convert that to a clone..but I'll write a test the proves it first.

yfarjoun:
regarding using the "clone" function. I don't actually care. it's convenient, but perhaps I can write a better "deepCopy" method...not sure how to do that and keep finals though, I guess I just need to invoke a copy constructor that does a deep copy...right?

pshapiro4broad:
> regarding using the "clone" function. I don't actually care. it's convenient, but perhaps I can write a better "deepCopy" method...not sure how to do that and keep finals though, I guess I just need to invoke a copy constructor that does a deep copy...right?

Java doesn't have a notion of copy constructor like C++ does, but you can define a constructor to take an object as an argument to make a new copy from it.

But I don't think a copy constructor is the right approach since you have an object you want to copy without worrying about which constructor you need to call. So it makes sense to have an overrideable clone-like method for a class, but I don't see a reason to use the Java `clone()`/`Cloneable` interface to do it. If you define your own method then you can use covariant return types so no return casting is necessary, and you don't need a try/catch, since every subclass of your interface must have your copy method defined.

yfarjoun:
pushed a new commit

src/main/java/picard/fingerprint/HaplotypeProbabilities.java
```diff
@@ -190,17 +188,19 @@ void assertSnpPartOfHaplotype(final Snp snp) {
 
     public double scaledEvidenceProbabilityUsingGenotypeFrequencies(final double[] genotypeFrequencies) {
         final double[] likelihoods = getLikelihoods();
-        assert (genotypeFrequencies.length == likelihoods.length);
+        assert (genotypeFrequencies.length == nGeno);
```
pshapiro4broad: I don't think `assert` is what you want here, since they have no effect unless the code is run with the `-ea` flag. E.g. if you had a unit test that tried to trigger this assert, it would fail.

src/main/java/picard/fingerprint/HaplotypeProbabilities.java
```diff
@@ -190,17 +188,19 @@ void assertSnpPartOfHaplotype(final Snp snp) {
...
+        assert (likelihoods.length == nGeno);
+
 
         double result = 0;
-        for (int i = 0; i < likelihoods.length; ++i) {
-            result += likelihoods[i] * genotypeFrequencies[i];
+        for (Genotype g: Genotype.values()){
```
pshapiro4broad: whitespace
```suggestion
        for (Genotype g : Genotype.values()) {
```

src/main/java/picard/fingerprint/HaplotypeProbabilities.java
```diff
@@ -272,4 +274,5 @@ public double getLodMostProbableGenotype() {
             this.v = v;
         }
     }
+    static final int nGeno = Genotype.values().length;
```
pshapiro4broad: Is this done because performance was measured and you found that using `values().length` each time was too slow? It seems like in most cases this wouldn't have a measurable benefit.
If you keep it, it should be all caps as it's a static
```suggestion
    static final int NUM_GENOTYPE = Genotype.values().length;
```

src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromContaminatorSequence.java
```diff
@@ -53,10 +59,18 @@ public HaplotypeProbabilitiesFromContaminatorSequence(final HaplotypeBlock haplo
         this.contamination = contamination;
     }
 
-    /**
-     * Adds a base observation with the observed quality to the evidence for this haplotype
-     * based on the fact that the SNP is part of the haplotype.
-     */
+    public HaplotypeProbabilitiesFromContaminatorSequence(HaplotypeProbabilitiesFromContaminatorSequence other){
```
pshapiro4broad: whitespace
```suggestion
    public HaplotypeProbabilitiesFromContaminatorSequence(HaplotypeProbabilitiesFromContaminatorSequence other) {
```

src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromContaminatorSequence.java
```diff
@@ -90,16 +104,21 @@ public void addToProbs(final Snp snp, final byte base, final byte qual) {
...
             // p(a | g_c) = \sum_g_m { P(g_m) \prod_i P(a_i| g_m, g_c)}
-            ll[contGeno.v] = log10(MathUtil.sum(MathUtil.multiply(this.getPriorProbablities(), likelihoodMap[contGeno.v])));
+            ll[contGeno.v] = Math.log10(MathUtil.sum(MathUtil.multiply(this.getPriorProbablities(), likelihoodMap[contGeno.v])));
         }
         setLogLikelihoods(ll);
     }
 
     @Override
-    public void merge(final HaplotypeProbabilities other) {
+    public HaplotypeProbabilities deepCopy() {
```
pshapiro4broad: I would use covariant return types here so you can write
```suggestion
    public HaplotypeProbabilitiesFromContaminatorSequence deepCopy() {
```

src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromContaminatorSequence.java
```diff
@@ -90,16 +104,21 @@ public void addToProbs(final Snp snp, final byte base, final byte qual) {
...
+        return new HaplotypeProbabilitiesFromContaminatorSequence(this);
+    }
+
+    @Override
+    public HaplotypeProbabilities merge(final HaplotypeProbabilities other) {
```
pshapiro4broad: This can also use covariant return types

src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromGenotype.java
```diff
@@ -24,13 +24,26 @@
...
-    private final double[] likelihoods;
+    private double[] likelihoods;
+
+    @Override
+    public HaplotypeProbabilities deepCopy()  {
+        return new HaplotypeProbabilitiesFromGenotype(this);
+    }
+
+    public HaplotypeProbabilitiesFromGenotype(final HaplotypeProbabilitiesFromGenotype other){
+        super(other.getHaplotype());
```
pshapiro4broad: why not
```
why not
```
        this(snp, getHaplotype(), likelihoods[0], likelihoods[1], likelihoods[2]);
```

File: src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromGenotype.java [hunk 10]
```diff
@@ -24,13 +24,26 @@
...
+import java.util.Arrays;
+
 /**
  * Represents a set of HaplotypeProbabilities that were derived from a single SNP
  * genotype at a point in time.
  */
 public class HaplotypeProbabilitiesFromGenotype extends HaplotypeProbabilities {
     private final Snp snp;
-    private final double[] likelihoods;
+    private double[] likelihoods;
```
pshapiro4broad: It looks like this can be final

src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromGenotypeLikelihoods.java
```diff
@@ -54,11 +58,11 @@ public void addToLogLikelihoods(final Snp snp, final List<Allele> alleles, final
         assertSnpPartOfHaplotype(snp);
 
         // only allow biallelic snps
-        assert (logGenotypeLikelihoods.length == Genotype.values().length);
+        assert (logGenotypeLikelihoods.length == nGeno);
         assert (alleles.size() == 2);
 
         //make sure that alleles are comparable to SNPs
-        for (int i = 0; i < 2; i++) {
+        for (int i = 0; i < alleles.size(); i++) {
```
pshapiro4broad: why not
```suggestion
        for (allele : alleles) {
```

src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromGenotypeLikelihoods.java
```diff
@@ -54,11 +58,11 @@ public void addToLogLikelihoods(final Snp snp, final List<Allele> alleles, final
...
             assert (alleles.get(i).getBases().length == 1);
```
pshapiro4broad: As I said above I don't think this has any effect in practice.

src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromSequence.java
```diff
@@ -36,10 +35,21 @@
...
+        obsAllele2 = ob2;
+        obsAlleleOther = obOther;
+    }
+
     public HaplotypeProbabilitiesFromSequence(final HaplotypeBlock haplotypeBlock) {
         super(haplotypeBlock);
     }
 
+    public HaplotypeProbabilitiesFromSequence(final HaplotypeProbabilitiesFromSequence other) {
+        this(other.getHaplotype() , other.obsAllele1,other.obsAllele2,other.obsAlleleOther);
```
pshapiro4broad: whitespace
```suggestion
        this(other.getHaplotype(), other.obsAllele1, other.obsAllele2, other.obsAlleleOther);
```

src/main/java/picard/fingerprint/HaplotypeProbabilityOfNormalGivenTumor.java
```diff
@@ -51,12 +58,13 @@ public TransitionMatrix(double pLoH) {
...
 
-        public double[][] getTransitionMatrix() {
+        double[][] getTransitionMatrix() {
             return transitionMatrix;
         }
     }
 
     private final HaplotypeProbabilities hpOfTumor;
+    @SuppressWarnings("MismatchedQueryAndUpdateOfCollection")
     static private Map<Double, TransitionMatrix> transitionMatrixMap = new CollectionUtil.DefaultingMap<>(TransitionMatrix::new, true);
```
pshapiro4broad: This seems dangerous to me. For one, it's not thread safe. It also introduces state which is never cleared, e.g. a memory leak, if the code is ever used in a non-CLP situation.

src/test/java/picard/util/MiscTest.java
```diff
@@ -0,0 +1,26 @@
...
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+
+public class MiscTest {
+
+    @Test
+    public void canWriteToDevNull() throws IOException {
```
pshapiro4broad: What is this testing? It doesn't cover any code that we've written. It can't assert anything about the behavior of a system picard will be run on, only the behavior of the system where the test is run.

src/test/java/picard/util/MiscTest.java
```diff
@@ -0,0 +1,26 @@
...
+        File f = new File("/dev/null");
```
pshapiro4broad: If you really want to do this, it would be simpler to write
```Files.write(Paths.get("/dev/null"), "Just a test".getBytes());```

File: src/test/java/picard/fingerprint/FingerprintCheckerTest.java [hunk 17]
```diff
@@ -223,6 +229,145 @@ public void testQueryable(final File vcf, boolean expectedQueryable) {
...
+
+        for(HaplotypeBlock hb:fpA.keySet()){
+            Assert.assertEquals(fpA.get(hb).getLikelihoods(),fpB.get(hb).getLikelihoods());
+        }
+    }
+
+
+    @Test
+    public void testMergeIsSafeFromSequence() {
+        final Path na12891_r1 = new File(TEST_DATA_DIR, "NA12891.over.fingerprints.r1.sam").toPath();
```
pshapiro4broad: I think a more idiomatic way to write this is
```suggestion
        final Path na12891_r1 = Paths.get(TEST_DATA_DIR).resolve("NA12891.over.fingerprints.r1.sam");
```

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
```diff
@@ -223,6 +229,145 @@ public void testQueryable(final File vcf, boolean expectedQueryable) {
...
+
+        final Map<FingerprintIdDetails, Fingerprint> fingerprintIdDetailsFingerprintMap = checker.fingerprintFiles(listOfFiles, 1, 0, TimeUnit.DAYS);
+
+        final Fingerprint combinedFp = new Fingerprint("test", null, null);
+        fingerprintIdDetailsFingerprintMap.values().forEach(combinedFp::merge);
+
+        final Fingerprint combinedFp2 = new Fingerprint("test2", null, null);
+        fingerprintIdDetailsFingerprintMap.values().forEach(combinedFp2::merge);
+
+        for (final HaplotypeBlock block : combinedFp.keySet()) {
```
pshapiro4broad: In general, if you have an assert in a `for`, you need to assert that the thing you're looping over has at least one item, otherwise the test will pass when the `for` never executes.

src/main/java/picard/fingerprint/HaplotypeProbabilities.java
yfarjoun: not performance, just readability and writability

src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromContaminatorSequence.java
yfarjoun: I didn't know you can do that when `@Override`ing...thanks!

src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromContaminatorSequence.java
yfarjoun: :+1:

src/test/java/picard/util/MiscTest.java
yfarjoun: This was at the bottom of the Fingerprinting test for some reason...so I moved it out....I think that at some point in the past there were questions about the ability to write to /dev/null and so a test was added (possibly by me) in the wrong place...

yfarjoun: Ah, I see the problem..I forgot to remove it from where I found it..so it looks like I just added it out of the blue...

src/test/java/picard/util/MiscTest.java
yfarjoun: not new code.

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
yfarjoun: since TEST_DATA_DIR is a File, it's actually more lengthy as I need to stick a `getAbsoluteFileName()` in there as well....I'll do it if you prefer though.

yfarjoun:
Thanks for the comments @pshapiro4broad. back to you.

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
pshapiro4broad: Under what circumstances is `getAbsoluteFileName()` required? I would think that `Path` would work the same with or without an absolute filename.

pshapiro4broad: Oh, I thought it was `String`. If it's `File` then you can write:
```
        final Path na12891_r1 = TEST_DATA_DIR.toPath().resolve("NA12891.over.fingerprints.r1.sam");
```
I don't know why you'd need to use `getAbsoluteFileName()` in this code.

src/main/java/picard/fingerprint/HaplotypeProbabilities.java
```diff
@@ -63,7 +64,29 @@ public HaplotypeBlock getHaplotype() {
...
+
+        if (haplotypeBlock != null ? !haplotypeBlock.equals(that.haplotypeBlock) : that.haplotypeBlock != null) {
+            return false;
+        }
+        return Arrays.equals(getLikelihoods(), that.getLikelihoods());
+    }
+
+    @Override
+    public int hashCode() {
+        return haplotypeBlock != null ? haplotypeBlock.hashCode() : 0;
```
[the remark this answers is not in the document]
yfarjoun: yeah...this is on purpose since likelihood is mutable...I figured it's better to not use it in the hash, but I need it in the equals...

src/test/java/picard/fingerprint/CrosscheckReadGroupFingerprintsTest.java
```diff
@@ -102,7 +102,8 @@ public void testCrossCheckRGs(final File file1, final File file2, final boolean
         doTest(args, metrics, expectedRetVal, expectedNMetrics * expectedNMetrics , CrosscheckMetric.DataType.READGROUP, expectAllMatch);
     }
 
-    @DataProvider(name = "bamFilesLBs")
+
+    @DataProvider()
```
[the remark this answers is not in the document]
yfarjoun: without the "name" the default is the name of the function itself....so, yes it is providing data...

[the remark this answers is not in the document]
yfarjoun: I guess I can remove the parens though

yfarjoun:
thanks for the review. mostly fixed (not the equals/hash disconnect though)

src/main/java/picard/fingerprint/HaplotypeProbabilities.java
```diff
@@ -63,7 +64,29 @@ public HaplotypeBlock getHaplotype() {
      * Returns the posterior probabilities using the population frequency as a prior.
      */
     public double[] getPosteriorProbabilities() {
-        return pNormalizeVector(multiply(getLikelihoods(), getPriorProbablities()));
+        return MathUtil.pNormalizeVector(MathUtil.multiply(getLikelihoods(), getPriorProbablities()));
+    }
+
+    @Override
+    public boolean equals(final Object o) {
```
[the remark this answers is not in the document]
yfarjoun: hmmm. 

I use it for testing....I guess I could make a utility function that compares? would that be preferable since I only need it for testing?

yfarjoun:
I moved the equals to utility class...better?

src/main/java/picard/fingerprint/HaplotypeProbabilities.java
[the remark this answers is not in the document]
pshapiro4broad: OK, that's fine, it's true that you can hash two non-equal objects to the same value.

yfarjoun:
your thumb became stale because I added a commit...so can I please have another :+1: ? :-)

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
```diff
@@ -223,6 +230,173 @@ public void testQueryable(final File vcf, boolean expectedQueryable) {
...
+        final FingerprintChecker checker = new FingerprintChecker(SUBSETTED_HAPLOTYPE_DATABASE_FOR_TESTING);
+        final Map<FingerprintIdDetails, Fingerprint> fingerprintIdDetailsFingerprintMap = checker.fingerprintFiles(listOfFiles, 1, 0, TimeUnit.DAYS);
+
+        final Fingerprint combinedFp = new Fingerprint("test", null, null);
+        fingerprintIdDetailsFingerprintMap.values().forEach(combinedFp::merge);
+
+        final Fingerprint combinedFp2 = new Fingerprint("test2", null, null);
+        fingerprintIdDetailsFingerprintMap.values().forEach(combinedFp2::merge);
+
+        Assert.assertNotEquals(combinedFp.keySet().size(), 0);
```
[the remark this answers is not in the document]
pshapiro4broad: Now that you've refactored into `assertFingerPrintHPsAreEqual()` this assert is no longer necessary, as the refactored code has the assert:
```
        Assert.assertEquals(lhs.keySet().size(), rhs.keySet().size());
```

src/test/java/picard/fingerprint/FingerprintingTestUtils.java
```diff
@@ -0,0 +1,40 @@
...
+        Assert.assertTrue(areHaplotypeProbabilitiesEqual(lhs, rhs),
+                "Expected HaplotypeProbabilities to be equal, but they differ: " +
+                lhs + ", " + rhs);
+    }
+
+    public static void assertFingerPrintHPsAreEqual(final Fingerprint lhs, final Fingerprint rhs) {
+        Assert.assertEquals(lhs.keySet().size(), rhs.keySet().size());
+
+        for (final HaplotypeBlock block : lhs.keySet()) {
+        Assert.assertTrue(lhs.containsKey(block),"HaplotypeBlock was missing from lhs" + block);
```
pshapiro4broad: code should be indented, also missing whitespace
```suggestion
            Assert.assertTrue(lhs.containsKey(block), "HaplotypeBlock was missing from lhs" + block);
```

yfarjoun:
Thanks @pshapiro4broad for the attentive review.

yfarjoun:
I think I need to rebase....

yfarjoun:
ugh. the push requires a new approve...sorry about that @pshapiro4broad

Issue #1285, opened 2019-02-21
MarkDuplicates misses duplicates of Chimeric Reads


yfarjoun:
When a chimeric read pair has one end which aligns well and the other aligns with MappingQuality ==0 the algorithm breaks down and fails to mark duplicates of this template. This is because the aligner will randomly place read2 (the one with MQ0) and since both reads are marked as Mapped, MD considers the position of both reads. 

I propose changing this behavior so that a read which is mapped with MQ0 is considered unmapped for the purposes of marking duplicates. This will resolve the problem. 

Do folks see an issue with this? 
Do some aligners use MQ0 to mean something other than "read has more than one equivalent alignment"?
Is there value in keeping these (likely) duplicates-reads marked as non-duplicates?

@tfenne @nh13 @lbergelson @jamesemery 
I would like to hear your opinions on this.

tfenne:
@yfarjoun Where do you draw the line on this?  What if the mate mapping quality is low enough that one sequencing error could throw it to another position in the genome?

I would edit your title to say "duplicates of ... read pairs mapping quality 0 on one end".  This can just as easily happy when e.g. one end of a pair is in unique sequence and the other end is in a small tandem duplication or locally-repetitive region _within the expected insert size_.

Are you unconcerned about pairs where both ends are mapq=0 because the use case you're looking at will ignore those reads?  Because the same problem arises there?

In general I see the problem you're raising, and don't disagree that it should be addressed, but I think the scope is broader than you're suggesting.

yfarjoun:
I agree that the problem is perhaps greater than I am tackling in this issue...(changed the name to MQ0 , thanks) 

However...I think that the incremental model is appropriate. We deal with the problems that actually affect us and our work. as we are sequencing more and more FFPE samples and they have really bad chimeras compounded with PCR duplication, the mapped reads are not being marked as duplicates due to their MQ0 mates. I have not seen this issue with MQ=3 or other low value MQ, and so am suggesting to use MQ0 as a way to eliminate the large part of the problem. 

As you say, I am unconcerned by both mates having MQ0 since they will be filtered out and will not cause downstream problems. Just as we do not worry about marking completely unmapped pairs (where both reads are unmapped)...I'm sure that there are many duplicates of that form.

I think that it's relatively straightforward to resolve the single-ended MQ0 problem, while it will be more difficult to resolve the low MQ and the both-sides-MQ. So I propose to fix the lower-effort, higher-value part of the larger problem.

[...]

PR #1460, opened 2020-02-02
Consider "low" mapping quality reads to be unaligned for the purpose of Marking Duplicates.

### LINKED ISSUE #128: Duplicates missed when mate has low mapping quality
When a pair-end read has one read with good mapping quality and the second with low (e.g. 0) mapping quality, the second may be placed in several locations randomly according to the aligner (even with the same aligner). This means that a duplicated such fragment will be incorrectly _not_ marked as duplicate if the two version of the second read are aligned differently (note that this has nothing to do with secondary or supplemental alignments)

Perhaps the solution could be as simple as to add another condition to line 283 in MarkDuplicates.java (and perhaps a similar condition in MarkDuplicatesWithCigar?) that verifies that the mate is well mapped. I suspect that a MQ!=0 would suffice, though it depends on the aligner...so perhaps it could be an @Option.

yfarjoun:
Fixes #128 and #1285 (which are similar issues and about a factor of 10 difference in issue number...)

The main thrust here is to consider "low mapping quality" as an indication of the read being effectively unmapped, as its location is not well determined and thus two identical fragments where each have one read with low (=0, e.g) mapping quality should be considered to be duplicates, or not, based on the well-mapped reads, and not the semi-random low-mapping quality read. 

Since this uses the same mechanism of unmapped reads, it will also not mark the low-mapping quality read as duplicate when its mate is marked so. Unless the file is queryname sorted, in which case the unmapped and the low-mapping quality reads are marked like their well-aligned mate.

### Description

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/sam/markduplicates/MarkDuplicates.java
+++ b/src/main/java/picard/sam/markduplicates/MarkDuplicates.java
@@ -327,6 +327,7 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
 
                 DuplicationMetrics metrics = AbstractMarkDuplicatesCommandLineProgram.addReadToLibraryMetrics(rec, header, libraryIdGenerator);
 
+
                 // Now try and figure out the next duplicate index (if going by coordinate. if going by query name, only do this
                 // if the query name has changed.
                 nextDuplicateIndex = nextIndexIfNeeded(sortOrder, recordInFileIndex, nextDuplicateIndex, duplicateQueryName, rec, this.duplicateIndexes);
@@ -337,7 +338,6 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram {
 
                 if (isDuplicate) {
                     rec.setDuplicateReadFlag(true);
-
                     AbstractMarkDuplicatesCommandLineProgram.addDuplicateReadToMetrics(rec, metrics);
                 } else {
                     rec.setDuplicateReadFlag(false);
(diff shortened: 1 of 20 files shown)


fleharty:
@yfarjoun Just letting you know that I'm using this on some somatic data to determine if this affects the false positive rates.

yfarjoun:
great. thanks @fleharty

fleharty:
@yfarjoun 
Some preliminary results:
I ran 6 NA12878 samples in pairs against each other for a total of choose(6, 2) = 30, runs.

With the current MarkDuplicates
SNP FP = 30
SNP FP rate = 0.029 / Mb
Indel FP = 21
Indel FP rate = 0.020 / Mb

With "low" mapping quality reads unaligned.
SNP FP = 44
SNP FP rate = 0.428 / Mb
Indel FP = 17
Indel FP rate  = 0.017


It looks like the SNP false positive rate goes up with this new Mark Duplicates methods, but the indel false positive rate goes down.

fleharty:
@yfarjoun I would like to follow up on this to understand why this would increase FP rate in SNPs, it makes no sense to me.  I don't have time this week to look into that unfortunately.

Do you want me to pursue this further?

yfarjoun:
is there a typo in the new indel FP rate? Seems like a 0 is missing after the decimal.

yfarjoun:
if you can DM me the locations perhaps I can take a look. thanks for setting this up!

fleharty:
@yfarjoun 
I edited the comment to fix my typo you found.  And I send you e-mail for the locations.

jessicaway:
@fleharty Do you know if we want to proceed with this PR?

PR #1484, opened 2020-03-10
Hardclip adapter option in MergeBamAlignments

kachulis:
Adds option to hardclip adapter readthroughs in MergeBamAlignments. Adapters are identified by overlapping of reads combined with matching expected adapter bases.

@yfarjoun @fleharty still want to add a few more tests, but other than that this is ready for review.

#### Content

#### Review


src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -742,28 +756,55 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
+                                    CigarUtil.clip3PrimeEndOfRead(read1, clipFrom, CigarOperator.HARD_CLIP);
+                                    CigarUtil.clip3PrimeEndOfRead(read2, clipFrom, CigarOperator.HARD_CLIP);
+                                    return;
+                                }
+                            }
+                        }
+                    }
+
+
+                    //if we've gotten to here, try standard soft clipping.  Here we consider clipped position because we only are correcting where early sof-clipping isn't quite right
```
yfarjoun:
```suggestion
                    //if we've gotten to here, try standard soft clipping.  Here we consider clipped position because we only are correcting where early soft-clipping isn't quite right
```

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -742,28 +756,55 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
             if (read1.getReadNegativeStrandFlag() != read2.getReadNegativeStrandFlag()) {
                 final SAMRecord pos = (read1.getReadNegativeStrandFlag()) ? read2 : read1;
                 final SAMRecord neg = (read1.getReadNegativeStrandFlag()) ? read1 : read2;
 
                 // Innies only -- do we need to do anything else about jumping libraries?
                 if (pos.getAlignmentStart() < neg.getAlignmentEnd()) {
+
+                    if (clippingOperator.equals(CigarOperator.HARD_CLIP)) {
+
+                        //need to consider unclipped positions because often the readthrough bases have already been soft-clipped
```
yfarjoun: the "readthrough bases"? not clear what is meant by that....

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -742,28 +756,55 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
+                        final int posClipFrom = getReadPositionToClipFrom(pos, neg.getUnclippedEnd() + 1);
+                        final int negClipFrom = getReadPositionToClipFrom(neg, pos.getUnclippedStart() - 1);
+                        if (posClipFrom == negClipFrom && posClipFrom > 0) {
+                            final int clipFrom = posClipFrom;
+
```
yfarjoun: unclear what you are trying to do here....could you put a comment in the code?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -780,9 +821,154 @@ protected static void clipForOverlappingReads(final SAMRecord read1, final SAMRe
                         final int clipFrom = neg.getReadLength() - negDiff - clipped + 1;
                         CigarUtil.softClip3PrimeEndOfRead(neg, Math.min(neg.getReadLength(), clipFrom));
                     }
+
+                }
+            }
+        }
+    }
+
+    protected static int getReadPositionToClipFrom(final SAMRecord rec, final int refPosToClipFrom) {
```
yfarjoun: hmmm. this feels like code that already exists in htsjdk....basically you want to go from a reference position to a readoffset, right? how about htsjdk.samtools.SAMRecord#getReadPositionAtReferencePosition(htsjdk.samtools.SAMRecord, int, boolean) does that help?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -742,28 +756,55 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
+
+
     /**
-     * Checks to see whether the ends of the reads overlap and soft clips reads
-     * them if necessary.
+     * Checks to see whether the ends of the reads overlap and clips reads
+     * if necessary.
      */
-    protected static void clipForOverlappingReads(final SAMRecord read1, final SAMRecord read2) {
+    protected static void clipForOverlappingReads(final SAMRecord read1, final SAMRecord read2, final CigarOperator clippingOperator, final ReadStructure read1Structure, final ReadStructure read2Structure,
```
yfarjoun: the generalizability of having the clipping operator given as such seems like premature optimization...why not just a boolean hardClip ?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -964,8 +1150,18 @@ public boolean isClipOverlappingReads() {
         return clipOverlappingReads;
     }
 
-    public void setClipOverlappingReads(final boolean clipOverlappingReads) {
+    public void setClipOverlappingReads(final boolean clipOverlappingReads, final CigarOperator clipOverlappingReadsOperator,
+                                        final ReadStructure read1Structure, final ReadStructure read2Structure,
+                                        final List<AdapterPair> adapters) {
         this.clipOverlappingReads = clipOverlappingReads;
+        if (clipOverlappingReads && !clipOverlappingReadsOperator.isClipping()) {
+            throw new IllegalArgumentException("clippingOverlappingReadsOperator " + clipOverlappingReadsOperator + " is not a clipping operatore.");
```
yfarjoun:
```suggestion
            throw new IllegalArgumentException("clippingOverlappingReadsOperator " + clipOverlappingReadsOperator + " is not a clipping operator.");
```

src/main/java/picard/sam/MergeBamAlignment.java
```diff
@@ -280,6 +301,16 @@
             "the result will be an invalid record. To force no change use the DO_NOT_CHANGE_INVALID strategy.", optional = true)
     public AbstractAlignmentMerger.UnmappingReadStrategy UNMAPPED_READ_STRATEGY = AbstractAlignmentMerger.UnmappingReadStrategy.DO_NOT_CHANGE;
 
+    @Argument(doc = "Structure of read 1.  If this and READ_STRUCTURE both are unspecified, will be assumed that all bases in read 1 are template. Cannot be used with READ_STRUCTURE.", optional = true, mutex = {"READ_STRUCTURE"})
+    public ReadStructure READ_1_STRUCTURE;
```
yfarjoun: Why do we need two read structures? the read structure is something that breaks up the entire stream of cycles into various things: UMIs,  Sample BCs, Template reads, skips...it is not clear (to me!) how a single readStructure should be broken up into to.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -780,9 +821,154 @@ protected static void clipForOverlappingReads(final SAMRecord read1, final SAMRe
...
+
+        //if we haven't found the correct position, there is a bug
+        throw new PicardException("Could not find read position for reference position " + refPosToClipFrom + " which should be covered by read " + rec);
+    }
+
+    protected static void clipForOverlappingReads(final SAMRecord read1, final SAMRecord read2) {
+        clipForOverlappingReads(read1, read2, CigarOperator.SOFT_CLIP, null, null, Collections.emptyList());
+    }
+
+    protected static boolean basesToClipMatchAdapterPair(final SAMRecord read, final int posToClip, final AdapterPair adapterPair,
```
yfarjoun: This function is awfully complex....but it seems to be doing a combination of things for which other functions already exist...can you explain what it does? perhaps it needs to be broken up into a few smaller methods.

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -49,6 +77,272 @@
...
+                {"100M16H", true, 300, 410, -1},
+                {"100M16H", true, 300, 390, 26},
+                {"16H100M", false, 316, 310, -1},
+                {"16H100M", true, 316, 310, -1},
+                {"16H100M", true, 316, 350, 66}
+        };
+    }
+
+    @Test(dataProvider = "getReadPosToClipFromDataProvider")
+    public void testGetReadPositionToClipFrom(final String cigarString, final boolean negativeStrand, final int start, final int refPosToClipFrom, final int expectedReadPosToCLipFrom) {
```
fleharty: expectedReadPosToCLipFrom - funny capitalization in CLip

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -49,6 +77,272 @@
...
+                    final byte[] umi1 = Arrays.copyOfRange(templateBases, gapBeforeUMI, gapBeforeUMI + umiLength);
+                    final byte[] umi2 = Arrays.copyOfRange(templateBasesRC, gapBeforeUMI, gapBeforeUMI + umiLength);
+
+                    final String umi1String = StringUtil.bytesToString(umi1);
+                    final String umi2String = StringUtil.bytesToString(umi2);
+
+                    SequenceUtil.reverseComplement(umi1);
+                    SequenceUtil.reverseComplement(umi2);
+
+                    final byte[] templateBasesWithUMI = ArrayUtils.addAll(templateBasesClipped, umi1);
```
fleharty: I don't think templateBasesWithUMI or templateBasesRCWithUMI are used.

fleharty:
@yfarjoun Back to you, this is doing the simplified hard clipping we discussed, and also refactors tests to be more general purpose.

Thanks to @kachulis for pair programming!

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -60,6 +60,9 @@
     public static final int MAX_RECORDS_IN_RAM = 500000;
 
     private static final char[] RESERVED_ATTRIBUTE_STARTS = {'X', 'Y', 'Z'};
+    static final String HARD_CLIPPED_BASES_TAG = "XB";
```
yfarjoun: perhaps let's use non-X (but lowercase) tag, so that we can try to make it a real (upper-cased) tag in hts-specs later. look for a reasonable two-caracter tag name...perhaps cb and cq is not taken? also, need to think what to do if it's clipped from both ends....

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -751,54 +755,105 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
     }
 
     /**
-     * Checks to see whether the ends of the reads overlap and soft clips reads
-     * them if necessary.
+     * Checks to see whether the ends of the reads overlap and clips reads
+     * if necessary.
      */
-    protected static void clipForOverlappingReads(final SAMRecord read1, final SAMRecord read2) {
+    protected static void clipForOverlappingReads(final SAMRecord read1, final SAMRecord read2, final CigarOperator clippingOperator) {
```
yfarjoun: I think it's better to use a boolean as there are only 2 resaonable operators here....SOFT and HARD.

yfarjoun: (and then the sanity checks below will not be needed)

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -751,54 +755,105 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
-            final CigarElement elem = iterator.next();
-            if (elem.getOperator() != CigarOperator.SOFT_CLIP && elem.getOperator() != CigarOperator.HARD_CLIP) {
-                break;
+    private static void moveClippedBasesToTag(final SAMRecord rec, final int clipFrom) {
+
+        final byte[] bases = rec.getReadBases();
+        final byte[] baseQualities = rec.getBaseQualities();
+        final int readLength = rec.getReadLength();
+
+        if(rec.getReadNegativeStrandFlag()) {
```
yfarjoun: I think that the bases should be stored in original read order, so that in order to revert, you just need to stick these on the ends of the reverted records.

src/main/java/picard/sam/AbstractAlignmentMerger.java
yfarjoun: comment still stands

yfarjoun:
should this PR also modify RevertSamFile so that it sticks the clipped bases on the reads?

fleharty:
@yfarjoun Back to you.

Also FYI, the logic has changed a little bit because @kachulis noticed that there was a bug in the strange case of having a read pair with unequal read lengths.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -60,6 +60,9 @@
     public static final int MAX_RECORDS_IN_RAM = 500000;
 
     private static final char[] RESERVED_ATTRIBUTE_STARTS = {'X', 'Y', 'Z'};
+    static final String HARD_CLIPPED_BASES_TAG = "hB";
```
yfarjoun, line 63: since soft-clipped bases are not removed, I think that the "h" is superfluous. how about EB and EQ (for End Bases, and End Quality) since in theory, bases could also be (hard-)clipped off of the start of a read..

yfarjoun: also, as a //Todo item here to switch to using the htsjdk version if it gets into hts-spec and implemented

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -751,54 +755,121 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
-                    }
 
-                    if (negDiff > 0) {
-                        final int clipped = lengthOfSoftClipping(neg.getCigar().getCigarElements().iterator());
-                        final int clipFrom = neg.getReadLength() - negDiff - clipped + 1;
-                        CigarUtil.softClip3PrimeEndOfRead(neg, Math.min(neg.getReadLength(), clipFrom));
+                    // Need to consider unclipped positions because often the read through bases have already been soft-clipped
+                    final int posClipFrom = getDistanceFrom3PrimeEndToClipFrom(pos, neg.getUnclippedEnd() + 1);
+                    final int negClipFrom = getDistanceFrom3PrimeEndToClipFrom(neg, pos.getUnclippedStart() - 1);
+                    final CigarOperator clippingOperator = useHardClipping ? CigarOperator.HARD_CLIP : CigarOperator.SOFT_CLIP;
```
yfarjoun, line 774: is `clippingOperator` used elsewhere?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -751,54 +755,121 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
-            if (elem.getOperator() != CigarOperator.SOFT_CLIP && elem.getOperator() != CigarOperator.HARD_CLIP) {
-                break;
+    private static void clipRead(final SAMRecord rec, final int clipFrom, final boolean useHardClipping) {
+
+        // If we are using hard clips, add bases and qualities to SAM tag.
+        if (useHardClipping) {
+            final byte[] bases = rec.getReadBases();
+            final byte[] baseQualities = rec.getBaseQualities();
+            final int readLength = rec.getReadLength();
+
```
yfarjoun, line 794: add a sanity check that the tags are not already in use. at this point I think it's OK to break if they are in use, but losing data due to overwriting old tags would not be good.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -751,54 +755,121 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
+            if (rec.getReadNegativeStrandFlag()) {
```
yfarjoun, line 795: could you split this into three parts:
- extract the required bases/qualities
- optionally reverse/rev-comp the arrays
- set the attributes

while it takes up more lines of code, there's less code duplication.

yfarjoun: also, you defined a method to "move bases to tag" so perhaps you should be using it... :-)

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -751,54 +755,121 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
+            rec.setAttribute(HARD_CLIPPED_BASE_QUALITIES_TAG, new StringBuilder(SAMUtils.phredToFastq(Arrays.copyOf(baseQualities, baseQualities.length - clipFrom + 1))).reverse().toString());
+        } else {
+            rec.setAttribute(HARD_CLIPPED_BASES_TAG, StringUtil.bytesToString(Arrays.copyOfRange(bases, clipFrom - 1, readLength)));
+            rec.setAttribute(HARD_CLIPPED_BASE_QUALITIES_TAG, SAMUtils.phredToFastq(Arrays.copyOfRange(baseQualities,clipFrom - 1, readLength)));
+        }
+    }
+
+    protected static int getDistanceFrom3PrimeEndToClipFrom(final SAMRecord rec, final int refPosToClipFrom) {
+
+        if (refPosToClipFrom > rec.getUnclippedEnd() || refPosToClipFrom < rec.getUnclippedStart()) {
```
yfarjoun, line 827: does `refPosToClipFrom < rec.getUnclippedStart()` not mean that the entire read should be clipped?

src/test/java/picard/sam/MergeBamAlignmentTest.java
```diff
@@ -1206,6 +1206,50 @@ public void testShortFragmentClipping() throws Exception {
         result.close();
     }
 
+    @Test
+    public void testShortFragmentHardClipping() throws IOException {
+        final File output = File.createTempFile("testShortFragmentClipping", ".sam");
```
yfarjoun, line 1211: please add a test that has an expected sam file with the tags

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -751,54 +755,121 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
+            // Ensures that bases are reverse complemented and base qualities are reversed
+            rec.setAttribute(HARD_CLIPPED_BASES_TAG, SequenceUtil.reverseComplement(StringUtil.bytesToString(Arrays.copyOf(bases, bases.length - clipFrom + 1))));
+            rec.setAttribute(HARD_CLIPPED_BASE_QUALITIES_TAG, new StringBuilder(SAMUtils.phredToFastq(Arrays.copyOf(baseQualities, baseQualities.length - clipFrom + 1))).reverse().toString());
+        } else {
+            rec.setAttribute(HARD_CLIPPED_BASES_TAG, StringUtil.bytesToString(Arrays.copyOfRange(bases, clipFrom - 1, readLength)));
+            rec.setAttribute(HARD_CLIPPED_BASE_QUALITIES_TAG, SAMUtils.phredToFastq(Arrays.copyOfRange(baseQualities,clipFrom - 1, readLength)));
+        }
+    }
+
+    protected static int getDistanceFrom3PrimeEndToClipFrom(final SAMRecord rec, final int refPosToClipFrom) {
```
yfarjoun, line 825: This method seems overly complicated and doesn't make enough use of existing coordinate switching code like htsjdk.samtools.SAMRecord#getReadPositionAtReferencePosition(htsjdk.samtools.SAMRecord, int, boolean)

if you really need it, it needs to be heavily documented and tested.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -964,8 +1035,9 @@ public boolean isClipOverlappingReads() {
         return clipOverlappingReads;
     }
 
-    public void setClipOverlappingReads(final boolean clipOverlappingReads) {
+    public void setClipOverlappingReads(final boolean clipOverlappingReads, final boolean hardClipOverlappingReads) {
```
yfarjoun, line 1038: as a change in public API, this is annoying. perhaps you should leave the previous method with a default of false?

src/main/java/picard/sam/MergeBamAlignment.java
```diff
@@ -256,9 +257,12 @@
             "alignment is filtered out for some reason. For all strategies, ties are resolved arbitrarily.")
     public PrimaryAlignmentStrategy PRIMARY_ALIGNMENT_STRATEGY = PrimaryAlignmentStrategy.BestMapq;
 
-    @Argument(doc = "For paired reads, soft clip the 3' end of each read if necessary so that it does not extend past the 5' end of its mate.")
+    @Argument(doc = "For paired reads, clip the 3' end of each read if necessary so that it does not extend past the 5' end of its mate.  Clipping will be either soft or hard clipping, depending on CLIP_OVERLAPPING_READS_OPERATOR setting.")
     public boolean CLIP_OVERLAPPING_READS = true;
 
+    @Argument(doc = "If true, hard clipping will be applied to overlapping reads.  By default, soft clipping is used.")
```
yfarjoun, line 263: docuement where the clipped bases/qualities go to

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -1,15 +1,10 @@
...
-import htsjdk.samtools.SAMFileHeader;
-import htsjdk.samtools.SAMFileWriter;
-import htsjdk.samtools.SAMFileWriterFactory;
-import htsjdk.samtools.SAMRecord;
-import htsjdk.samtools.SAMRecordIterator;
-import htsjdk.samtools.SAMRecordSetBuilder;
-import htsjdk.samtools.SamReader;
-import htsjdk.samtools.SamReaderFactory;
-import htsjdk.samtools.ValidationStringency;
+import htsjdk.samtools.*;
```
yfarjoun, line 3: there seems to be a preference for explicit imports rather than wildcard. you can change your intelliJ settings for that.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -788,38 +797,42 @@ private static void clipRead(final SAMRecord rec, final int clipFrom, final bool
...
 
         final byte[] bases = rec.getReadBases();
         final byte[] baseQualities = rec.getBaseQualities();
         final int readLength = rec.getReadLength();
 
+        final int clipPositionFrom, clipPositionTo;
+        if (rec.getReadNegativeStrandFlag()) {
+            clipPositionFrom = 0;
+            clipPositionTo = bases.length - clipFrom + 1;
+        }
```
yfarjoun, line 820: single line `} else {`

yfarjoun:
what about the changes to RevertSamFile? shouldn't we code these up as well so that we can check that we can round-trip without losing data?

fleharty:
@yfarjoun Back to you.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -751,54 +765,103 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
-                break;
-            }
-            if (elem.getOperator() == CigarOperator.SOFT_CLIP) {
-                clipped = elem.getLength();
+    private static int getReadPositionAtReferencePositionIgnoreSoftClips(final SAMRecord rec, final int pos) {
+        final int readPosition;
+        final Cigar oldCigar = rec.getCigar();
+        final int oldStart = rec.getAlignmentStart();
+        final Cigar newCigar = new Cigar();
+        List<CigarElement> cigarElements = new ArrayList<>(oldCigar.getCigarElements()); //need to be modifiable
```
yfarjoun, line 800: `cigarElements` can still be final, right?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -751,54 +765,103 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
+        int posShift = 0;
+        boolean foundNonClip = false;
+        for (final CigarElement cigarElement : cigarElements) {
+            final CigarOperator op = cigarElement.getOperator();
+
+            if (op == CigarOperator.SOFT_CLIP) {
```
yfarjoun, line 806: I think some comments about what's happening here would be appreciated by a future developer...

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -751,54 +765,103 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
+            } else {
+                if (!op.isClipping()) {
+                    foundNonClip = true;
+                }
+                newCigar.add(new CigarElement(cigarElement.getLength(), op));
             }
         }
 
-        return clipped;
+        rec.setAlignmentStart(rec.getAlignmentStart() - posShift);
```
yfarjoun, line 819: what if this is negative?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -751,54 +765,103 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
+        rec.setCigar(newCigar);
+        readPosition = SAMRecord.getReadPositionAtReferencePosition(rec, pos, false);
+        rec.setCigar(oldCigar);
+        rec.setAlignmentStart(oldStart);
+        return readPosition;
+    }
+
+    private static void clipRead(final SAMRecord rec, final int clipFrom, final boolean useHardClipping) {
```
yfarjoun, line 827: I would include `3primeEnd` in the names of the methods here..

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -751,54 +765,103 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
+            moveClippedBasesToTag(rec, clipFrom);
+        }
+
+        // Actually clip the read
+        CigarUtil.clip3PrimeEndOfRead(rec, clipFrom, useHardClipping ? CigarOperator.HARD_CLIP : CigarOperator.SOFT_CLIP);
+    }
+
+    private static void moveClippedBasesToTag(final SAMRecord rec, final int clipFrom) {
+        if (rec.getAttribute(HARD_CLIPPED_BASES_TAG) != null || rec.getAttribute(HARD_CLIPPED_BASE_QUALITIES_TAG) != null) {
+            throw new PicardException("Record already contains tags for restoring hard clipped bases.  This operation will permanently erase information if it proceeds.");
```
yfarjoun, line 839: make sure you include the name of the offending record, and the names of the tags that you need...otherwise, debugging this might be tough.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -751,54 +765,103 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
+        } else {
+            clipPositionFrom = clipFrom - 1;
+            clipPositionTo = readLength;
+        }
+
+        String basesToKeepInTag = StringUtil.bytesToString(Arrays.copyOfRange(bases, clipPositionFrom, clipPositionTo));
+        String qualitiesToKeepInTag = SAMUtils.phredToFastq(Arrays.copyOfRange(baseQualities, clipPositionFrom, clipPositionTo));
+
+        if (rec.getReadNegativeStrandFlag()) {
+            // Ensures that bases are reverse complemented and base qualities are reversed
```
yfarjoun, line 859: "ensures that the qualities and bases in the tags are stored in their original order, as produced by the sequencer"

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -966,6 +1029,12 @@ public boolean isClipOverlappingReads() {
 
     public void setClipOverlappingReads(final boolean clipOverlappingReads) {
         this.clipOverlappingReads = clipOverlappingReads;
+        this.hardClipOverlappingReads = false;
```
yfarjoun, line 1032: I know I said this is better than changing a public API...but it's still not good. `set*` methods should only change the variable they are named for. why not add a `setClipWithHardClips` method and avoid changing two things in one method?

src/main/java/picard/sam/RevertSam.java
```diff
@@ -373,6 +366,10 @@ public void revertSamRecord(final SAMRecord rec) {
             rec.setDuplicateReadFlag(false);
         }
 
+        if (RESTORE_HARDCLIPS && !REMOVE_ALIGNMENT_INFORMATION) {
+            throw new PicardException("hi mom");
```
yfarjoun, line 370: hi mark

src/main/java/picard/sam/RevertSam.java
```diff
@@ -396,6 +393,20 @@ public void revertSamRecord(final SAMRecord rec) {
             rec.setMateReferenceIndex(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX);
             rec.setMateUnmappedFlag(rec.getReadPairedFlag());
 
+            if (RESTORE_HARDCLIPS) {
+                String hardClippedBases = rec.getStringAttribute(AbstractAlignmentMerger.HARD_CLIPPED_BASES_TAG);
+                String hardClippedQualities = rec.getStringAttribute(AbstractAlignmentMerger.HARD_CLIPPED_BASE_QUALITIES_TAG);
+                if (hardClippedBases != null && hardClippedQualities != null) {
+                    // Record has already been reverse complemented if this was on the negative strand
+                    rec.setReadString(rec.getReadString() + hardClippedBases);
```
yfarjoun, line 401: please verify your assumption (and throw otherwise)

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -538,4 +539,26 @@ public void testSanitizeAndDeduplicateRecords() throws Exception {
         Assert.assertEquals(runPicardCommandLine(args), 0);
         verifyPositiveResults(output, new RevertSam(), true, true, false, false, null, 8, null, null);
     }
+
+    @Test
+    public void testHardClippedRecovery() throws Exception {
+        final File outputFile = File.createTempFile("test-output-hard-clipped-recovery", ".sam");
+
+        // hardClippedSamToRevert is a sam file with the expected reverted reads and base qualities stored in the tB and tQ tags respectively
```
yfarjoun, line 547: update comment with actual tag name

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -538,4 +539,26 @@ public void testSanitizeAndDeduplicateRecords() throws Exception {
...
+        final String [] args = new String[]{
+                "I=" + hardClippedSamToRevert,
+                "RESTORE_HARDCLIPS=true",
+                "O=" + outputFile.getAbsolutePath()
+        };
+        Assert.assertEquals(runPicardCommandLine(args), 0);
+
+        // Ensure that the reverted reads and qualities match the tB and tQ tags
```
yfarjoun, line 555: comment

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -538,4 +539,26 @@ public void testSanitizeAndDeduplicateRecords() throws Exception {
         Assert.assertEquals(runPicardCommandLine(args), 0);
         verifyPositiveResults(output, new RevertSam(), true, true, false, false, null, 8, null, null);
     }
+
+    @Test
```
yfarjoun, line 543: Could you add around-tripping test, that merges and then restores?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -751,54 +765,104 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
+                newCigar.add(new CigarElement(cigarElement.getLength(), op));
             }
         }
 
-        return clipped;
+        // Temporarily use the newCigar that has SOFT_CLIPs replaced with MATCH_OR_MISMATCH to get read position at reference, but ignore existence of soft-clips
+        rec.setCigar(newCigar);
+        readPosition = SAMRecord.getReadPositionAtReferencePosition(rec, pos, false);
+        rec.setCigar(oldCigar);
+        return readPosition + posShift;
```
yfarjoun, line 825: `// instead of setting back the position of the read by posShift, which could create a negative start position, we add posShift the final position in the read.`

src/main/java/picard/sam/RevertSam.java
```diff
@@ -24,18 +24,8 @@
 
 package picard.sam;
 
-import htsjdk.samtools.BAMRecordCodec;
-import htsjdk.samtools.SAMFileHeader;
+import htsjdk.samtools.*;
```
yfarjoun, line 27: do explicit imports please.

src/main/java/picard/sam/RevertSam.java
yfarjoun: nm.

src/test/java/picard/sam/RevertSamTest.java
yfarjoun: tB and tQ tags?

src/test/java/picard/sam/MergeSamFilesTest.java
```diff
@@ -37,7 +37,8 @@
 
 import java.io.File;
 
-public class MergeSamFilesTest extends CommandLineProgramTest {
+public class
```
yfarjoun, line 40: why split this line?

fleharty:
@yfarjoun Thanks for the quick reviews!  Back to you.

src/test/java/picard/sam/RevertSamTest.java
yfarjoun: why does the comment keep saying `tB` and `tQ`?

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -538,4 +542,48 @@ public void testSanitizeAndDeduplicateRecords() throws Exception {
...
+        };
+        Assert.assertEquals(runPicardCommandLine("RevertSam", revertSamArgs), 0);
+
+        final SamReader revertedReader = SamReaderFactory.makeDefault().referenceSequence(referenceFasta).open(outputRevert);
+        final SamReader unmappedReader = SamReaderFactory.makeDefault().referenceSequence(referenceFasta).open(hardClippedUnmappedSam);
+        final SAMRecordIterator revertedIterator = revertedReader.iterator();
+        final SAMRecordIterator unmappedIterator = unmappedReader.iterator();
+
+        while (revertedIterator.hasNext() && unmappedIterator.hasNext()) {
+            SAMRecord reverted = revertedIterator.next();
```
yfarjoun, line 576: final

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -538,4 +542,48 @@ public void testSanitizeAndDeduplicateRecords() throws Exception {
...
+            SAMRecord unmapped = unmappedIterator.next();
```
yfarjoun, line 577: final

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -538,4 +542,48 @@ public void testSanitizeAndDeduplicateRecords() throws Exception {
...
+
+        final File outputRevert = File.createTempFile("test-output-hard-clipped-round-trip-reverted", ".sam");
+        final String [] revertSamArgs = new String[] {
+                "I=" + outputMBA.getAbsolutePath(),
+                "RESTORE_HARDCLIPS=true",
+                "O=" + outputRevert.getAbsolutePath()
+        };
+        Assert.assertEquals(runPicardCommandLine("RevertSam", revertSamArgs), 0);
+
+        final SamReader revertedReader = SamReaderFactory.makeDefault().referenceSequence(referenceFasta).open(outputRevert);
```
yfarjoun, line 570: convert to try-with-resources to assure closing the readers

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
yfarjoun: this.

src/test/java/picard/cmdline/CommandLineProgramTest.java
```diff
@@ -54,7 +59,13 @@ public int runPicardCommandLine(final String[] args) {
...
     }
 
+    public int runPicardCommandLine(final String programName, final String[] args) {
+        return new PicardCommandLine().instanceMain(makePicardCommandLineArgs(programName, Arrays.asList(args)));
+    }
+
     public int runPicardCommandLine(final Map<String, String> kwargs) {
         return new PicardCommandLine().instanceMain(makePicardCommandLineArgs(kwargs));
     }
+
```
yfarjoun, line 69: remove extra NLs

fleharty:
@yfarjoun Back to you again, sorry about the back and forth.

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -538,4 +542,45 @@ public void testSanitizeAndDeduplicateRecords() throws Exception {
         Assert.assertEquals(runPicardCommandLine(args), 0);
         verifyPositiveResults(output, new RevertSam(), true, true, false, false, null, 8, null, null);
     }
+
+    @Test
+    public void testHardClipRoundTrip() throws Exception {
+        // Runs sam files through MergeBamAlignment using hard clipping on overlapping reads.
+        // Tests to ensure that RevertSam can reconstruct the reads and base qualities from reads that have been hard clipped.
+
+        final File outputMBA = File.createTempFile("test-output-hard-clipped-round-trip-mba", ".sam");
```
yfarjoun, line 551: delete on exit

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -538,4 +542,45 @@ public void testSanitizeAndDeduplicateRecords() throws Exception {
...
+        final String [] mergeBamAlignmentsArgs = new String[] {
+           "UNMAPPED_BAM=" + hardClippedUnmappedSam.getAbsolutePath(),
+           "ALIGNED_BAM=" + hardClippedAlignedSam.getAbsolutePath(),
+           "OUTPUT=" + outputMBA.getAbsolutePath(),
+           "REFERENCE_SEQUENCE=" + hardClipFasta.getAbsolutePath(),
+           "HARD_CLIP_OVERLAPPING_READS=true"
+        };
+        Assert.assertEquals(runPicardCommandLine("MergeBamAlignment", mergeBamAlignmentsArgs), 0);
+
+        final File outputRevert = File.createTempFile("test-output-hard-clipped-round-trip-reverted", ".sam");
```
yfarjoun, line 561: delete on exit

yfarjoun:
ugh. you need to rebase again...

yfarjoun:
ah, no...just needed to update the branch.

Issue #1485, opened 2020-03-11
CheckIlluminaDirecotory fails with cbcl (NovaSeq) data if there are skips in the READ_STRUCTURE


nnordborg:
## Bug Report
We recently started to use NovaSeq for sequencing which produces cbcl basecall files. There seems to be a problem with the CheckIlluminaDirectory tool when the READ_STRUCTURE contains skips. We want to use 100T1S8B8B100T1S and have a pipeline that runs CheckIlluminaDirectory , ExtractIlluminaBarcodes and IlluminaBasecallsToFastq. The ExtractIlluminaBarcodes and IlluminaBasecallsToFastq tools work, but from CheckIlluminaDirectory  we get an error message:

File [.../Data/Intensities/BaseCalls/L001/C101.1/L001_2.cbcl] is not the expected size of 153723641 instead it is 148877983

The folder 'C101.1' contains data for the skipped cycle so it was surprising that it complained about this file. If I change the READ_STRUCTURE to 101T8B8B101T, CheckIlluminaDirectory works and I get the "SUCCEEDED!  All required files are present and non-empty." message.

I have implemented a possible fix for this problem here https://github.com/nnordborg/picard/commit/0e87c4c0ed04dda492151c28bb3a6de1be3ec17f

As far as i can tell the 'key' variable is only equal to the cycle if there are no skips in the READ_STRUCTURE. Since the "log.info()" is printing an array with actual cycle numbers I figured it would be safe to use that array for looking up the actual cycle number. But there may of course be other things to consider as well that I am not aware of. The change seems to fix the problem in our case.

### Affected tool(s)
CheckIlluminaDirectory 

### Affected version(s)
- [ 2.22.0 and 2.20.8] Latest public release version [version?]
- [ ] Latest development/master branch as of [date of test?]

### Description 
See above.

#### Steps to reproduce
cbcl files are needed and the READ_STRUCTURE must have skips in it.

I could not find an existing test case that seems to test cbcl files with CheckIlluminaDirectory, but there seems to be files in the 'testdata' directory and a test for ExtractIlluminaBarcodes. Unfortunately I do no know how to make this work with CheckIlluminaDirectory.

#### Expected behavior
CheckIlluminaDirectory should not fail.

#### Actual behavior
CheckIlluminaDirectory fails and print an error message:

File [.../Data/Intensities/BaseCalls/L001/C101.1/L001_2.cbcl] is not the expected size of 153723641 instead it is 148877983
----

PR #1489, opened 2020-03-20
Fixes CheckIlluminaDirectory problem with cbcl files and skips in the READ_STRUCTURE

nnordborg:
### Description

This is a possible fix for issue https://github.com/broadinstitute/picard/issues/1485

The problem is that the 'key' variable in the loop is not the same as the 'cycle' if there are skips in the READ_STRUCTURE. This causes the file size check to check files that belong to skipped cycles and the result is that CheckIlluminaDirectory fails with an error message. 

The fix uses the array returned by outputMapping.getOutputCycles() to lookup the correct 'cycle' number.

----

#### Content

#### Review


gbggrant:
Looks good @nnordborg - I've verified that this fixes similar behavior we've seen on our Novaseqs. 

Any chance you can add a test case for this?

nnordborg:
Thanks for looking into this. Regarding a test case, I am not sure if I am able to do that. I have looked at the code in CheckIlluminaDirectoryTest and if I understand it correctly it seems like it generates temporary fake files that are tested. The IlluminaFileUtilTest.makeFiles() doesn't have any option for Cbcl files.

There are files available in tetsdata/picard/illumina/151T8B8B151T_cbcl, so maybe it would be possible to use them. But I can't get them to validate when I test by manually running CheckIlluminaDirectory:

`java -jar picard.jar CheckIlluminaDirectory BASECALLS_DIR=testdata\picard\illumina\151T8B8B151T_cbcl\Data\Intensities\BaseCalls\ READ_STRUCTURE=1T LANES=1 DATA_TYPES=BaseCalls`

I get:
Exception in thread "main" picard.PicardException: File [testdata\picard\illumina\151T8B8B151T_cbcl\Data\Intensities\BaseCalls\L001\C1.1\L001_1.cbcl, testdata\picard\illumina\151T8B8B151T_cbcl\Data\Intensities\BaseCalls\L001\C1.1\L001_2.cbcl] is not the expected size of 68 instead it is 69

I have tried to understand why this is happening and have found that the expected and and actual sizes that are compared are calculated from the **file sizes**, the **header size** and the **compressed block size**. The CbclReader is reading the header and compressed block size from the cbcl files and both L001_1.cbcl and L001_2.cbcl have headerSize=65 and compressedBlockSize=34 which should mean that the file size must be 99 bytes. But L001_1.cbcl is 100 bytes so the comparison in CheckIlluminaDirectory fails. 

I have tested some of the other cycle directories as well, and in those where both cbcl files are 99 bytes (for example C2.1 and C4.1) CheckIlluminaDirectory reports SUCCEEDED!

So, it could be that some cbcl files have an extra byte that should be removed, or it could be incorrect header values, or there is some other header that should also be included in the calculations. I guess it would be good if someone with knowledge about the cbcl format could provide more information.

nnordborg:
I have tried to extract the bytes from the "compressed block" from C1.1\L001_1.cbcl and C1.1\L001_2.cbcl using the 'dd' command and then de-compress it with 'gunzip'. Using the 34 bytes as specified in the header caused it to fail for the L001_1.cbcl file:
`$ dd if=L001_1.cbcl ibs=1 skip=65 count=34 | gunzip > /dev/null
34+0 records in
0+1 records out
34 bytes (34 B) copied, 4.0866e-05 s, 832 kB/s

gzip: stdin: unexpected end of file`

Changing to 35 bytes for L001_1.cbcl and 34 bytes for L001_2.cbcl seems to work:
`$ dd if=L001_1.cbcl ibs=1 skip=65 count=35 | gunzip > /dev/null
35+0 records in
0+1 records out
35 bytes (35 B) copied, 4.4715e-05 s, 783 kB/s

$ dd if=L001_2.cbcl ibs=1 skip=65 count=34 | gunzip > /dev/null
34+0 records in
0+1 records out
34 bytes (34 B) copied, 4.3446e-05 s, 783 kB/s`

So, my guess is that the header information in the 100-byte cbcl files is incorrect. Note that I have only checked a few of the cbcl files and I don't know if there are files with different size or different value for the compressed block size header.

nnordborg:
I managed to add a simple test case that is using the existing data in the `testdata/picard/illumina/151T8B8B151T_cbcl` directory. I had to use READ_STRUCTURE=1S1T to get around the problem with incorrect value for the compressed block size in C1.1/L001_1.cbcl. In cycle 2 both cbcl files have correct headers. This is what I added to CheckIlluminaDirectoryTest:

```
private static final File CBCL_BASECALL_DIR = new File("testdata/picard/illumina/151T8B8B151T_cbcl/Data/Intensities/BaseCalls");

@Test
public void cbclWithSkipsInReadStructureTest()
{
    final int lane = 1;
    final IlluminaDataType[] dataTypes = new IlluminaDataType[]{BaseCalls};
    final String[] args = 
            makeCheckerArgs(CBCL_BASECALL_DIR, lane, "1S1T", dataTypes, new ArrayList<>(), false, false);
    Assert.assertEquals(runPicardCommandLine(args), 0);
}
```

What do you think? Is this enough?

gbggrant:
Thanks @nnordborg having that test is better than none at all.  Maybe just put a comment in the test explaining what you found about the incorrect value in the test cbcl.

gbggrant:
retest this please

PR #1497, opened 2020-04-13
GL-970.  Have CompareMetrics consider two NaN metrics equal

gbggrant:
### Description

Have CompareMetrics consider two NaN metrics equal

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/analysis/CompareMetrics.java
+++ b/src/main/java/picard/analysis/CompareMetrics.java
@@ -64,7 +64,7 @@ public class CompareMetrics extends CommandLineProgram {
     static final String USAGE_SUMMARY = "Compare two metrics files.";
     static final String USAGE_DETAIL = "This tool compares the metrics and histograms generated from metric tools to determine " +
             "if the generated results are identical.  Note that if there are differences in metric values, this tool describes those differences " +
-            "as the change of the first input metric relative to the second. " +
+            "as the change of the second input metric relative to the first. " +
             "<br /><br />  " +
             "<h4>Usage example:</h4>" +
             "<pre>" +
@@ -110,7 +110,7 @@ public class CompareMetrics extends CommandLineProgram {
 
     private static final Log log = Log.getInstance(CompareMetrics.class);
 
-    protected final static Map<String, Double> MetricToAllowableRelativeChange = new HashMap<>();
+    protected final Map<String, Double> MetricToAllowableRelativeChange = new HashMap<>();
 
     @Override
     protected int doWork() {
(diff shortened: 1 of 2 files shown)


src/main/java/picard/analysis/CompareMetrics.java
```diff
@@ -267,7 +267,11 @@ protected SimpleResult compareMetricValues(final Object value1, final Object val
             if (value1 instanceof Number) {
                 double numValue1 = ((Number) value1).doubleValue();
                 double numValue2 = ((Number) value2).doubleValue();
-                double relativeChange = (numValue1 == 0 && numValue2 == 0) ? 0 : (numValue1 == 0) ? Double.MAX_VALUE : (numValue1 - numValue2) / numValue2;
+                double relativeChange = 0;
+                if ((!Double.isNaN(numValue1)) || (!Double.isNaN(numValue2))) {
```
pshapiro4broad, line 271: Can use fewer parens
```suggestion
                if (!Double.isNaN(numValue1) || !Double.isNaN(numValue2)) {
```

src/main/java/picard/analysis/CompareMetrics.java
```diff
@@ -267,22 +267,26 @@ protected SimpleResult compareMetricValues(final Object value1, final Object val
...
                 double numValue1 = ((Number) value1).doubleValue();
                 double numValue2 = ((Number) value2).doubleValue();
-                double relativeChange = (numValue1 == 0 && numValue2 == 0) ? 0 : (numValue1 == 0) ? Double.MAX_VALUE : (numValue1 - numValue2) / numValue2;
-                if (relativeChange != 0) {
+                double absoluteChange = 0;
+                if (!Double.isNaN(numValue1) || !Double.isNaN(numValue2)) {
+                    absoluteChange = numValue2 - numValue1;
+                }
+                if (absoluteChange != 0) {
+                    double relativeChange = (numValue1 == 0 && numValue2 == 0) ? 0 : (numValue1 == 0) ? Double.MAX_VALUE : absoluteChange / numValue1;
```
pshapiro4broad, line 275: Under what conditions is it possible for `absoluteChange` to be non-zero and for `(numValue1 == 0 && numValue2 == 0)` to be true? Are this trying to avoid dividing by zero?

Also this code now divides by `numValue1` and before it divided by `numValue2`, is that a fix?

gbggrant:
@pshapiro4broad I've removed those comments and added absolute differences.

src/main/java/picard/analysis/CompareMetrics.java
gbggrant: Yes, that's to avoid the divide by zero.
I've changed the code to be comparison of the second metric, relative to the first metric, so that's why the denominator has changed.
change = (value2 - value1) / value1

pshapiro4broad: OK, so you only need to check `numValue1 == 0`. Dividing zero by anything (except NaN) is zero.

pshapiro4broad: That explains the second `numValue1 == 0` check. But I don't see how `(numValue1 == 0 && numValue2 == 0)` can be true if `absoluteChange` is not zero.

gbggrant: Probably true, the problem with doing lots of incremental changes.

gbggrant: And it's really not even a problem if the numerator is 0.

PR #1515, opened 2020-05-26
gl-916 add pipeline version to metrics

nikellepetrillo:
### Description

updated picard to add this new, nullable pipeline_version field 

----

#### Content

#### Review


src/main/java/picard/arrays/ArraysCallingMetricAccumulator.java
```diff
@@ -76,14 +77,15 @@
...
+                        detail.PIPELINE_VERSION = pipelineVersion;
                         return detail;
                     }, true);
 
     ArraysCallingMetricAccumulator(DbSnpBitSetUtil.DbSnpBitSets dbsnp) {
         this.dbsnp = dbsnp;
     }
 
-    public void setup(final VCFHeader vcfHeader) {
+    public void setup(final VCFHeader vcfHeader, final String pipelineVersion) {
```
pshapiro4broad, line 88: Given the way this class is used, I don't see why the code in `setup()` isn't done in the constructor directly.

gbggrant: The code has been rewritten to not pass pipelineVersion to CollectArraysVariantCallingMetrics, so this suggestion isn't germane to the PR (and probably should be done elsewhere).

testdata/picard/arrays/7991775143_R01C01_wo_optional_fields.vcf
```diff
@@ -179,6 +179,7 @@
 ##p95Green=4251
 ##p95Red=5479
 ##picardVersion=2.20.3-17-g8331bf3-SNAPSHOT
+##pipelineVersion=foo
```
gbggrant, line 182: Since this is an optional parameter, we should have one test without pipelineVersion set in the VCF.  It should allow that case.

src/main/java/picard/arrays/CollectArraysVariantCallingMetrics.java
```diff
@@ -43,10 +43,12 @@
                     "<h4>Usage example:</h4>" +
                     "<pre>" +
                     "java -jar picard.jar CollectArraysVariantCallingMetrics \\<br />" +
+                    // TODO - ADD example
```
pshapiro4broad, line 46: I don’t think you need to update the example. I would either add an example or remove the comment. Adding a TODO should be avoided if possible.

src/main/java/picard/arrays/GtcToVcf.java
```diff
@@ -538,6 +541,9 @@ private VCFHeader createVCFHeader(final Build37ExtendedIlluminaManifest manifest
         if (EXPECTED_GENDER != null) {
             lines.add(new VCFHeaderLine(InfiniumVcfFields.EXPECTED_GENDER, EXPECTED_GENDER));
         }
+        if (PIPELINE_VERSION != null) {
+            lines.add(new VCFHeaderLine(InfiniumVcfFields.PIPELINE_VERSION, PIPELINE_VERSION.toString()));
```
pshapiro4broad, line 545: Since `PIPELINE_VERSION` is already a String you don’t need to call `toString`
```suggestion
            lines.add(new VCFHeaderLine(InfiniumVcfFields.PIPELINE_VERSION, PIPELINE_VERSION);
```

PR #1529, opened 2020-06-22
Cleanup CreateSequenceDictionary and provide better error messaging

yfarjoun:
Helps (latter part of) #1527


src/main/java/picard/sam/CreateSequenceDictionary.java
```diff
@@ -270,30 +268,6 @@ private BufferedWriter makeWriter() throws FileNotFoundException {
         );
     }
 
-    /**
-<<<<<<< HEAD
```
lbergelson, line 274: this old merge marker is nice

yfarjoun: yeah.....we need a bot to clean that up....

Issue #1539, opened 2020-07-09
Better defaults for inflaters, deflaters and compression level lead  to better compute times.


rhpvorderman:
### Tool(s) involved
Picard Markduplicates

### Description

Picard Markduplicates (and other BAM writing and reading tools) can be faster if certain defaults are changed.

I tested version 2.23.1. To be specific, [the biocontainer](https://quay.io/biocontainers/picard)

I benchmarked several options using [hyperfine](https://github.com/sharkdp/hyperfine). 

### Benchmarks

I will use `VALIDATION_STRINGENCY=SILENT` troughout. This makes Picard MarkDuplicates faster. Whether this should be a default is not the point of this discussion. This is solely about compression. Using `VALIDATION_STRINGENCY=silent` will make the benchmarks run faster for now. Which matters because this benchmarking takes a lot of time.
I will also use `CREATE_INDEX=true` troughout as having a BAM index is necessary for follow up tasks. `CREATE_INDEX=true VALIDATION_STRINGENCY=silent` are our defaults in production and the benchmarks should reflect
production settings.

The test BAM is a 5 million read alignment (BWA) sorted by samtools on coordinates. The test bam file is compressed with level 1 and has a size of 706M.

I will use `-Xmx4G -XX:ParallelGCThreads=1`. 4G should be enough for the given BAM. Running one garbage collection thread is sufficient and it will prevent things like 1600% cpu usage on a single-threaded application (possible if 15 GCC threads are used).

I use hyperfine with 2 warmup runs (to make sure any files are cached to prevent inconsistencies), and 5 actual runs.  After each benchmark I measure the size of the resulting bam file.

Having said that, let's have a run with the (almost) default setting.

#### Default settings
```
hyperfine -w 2 -r 5 'singularity exec -eip docker://quay.io/biocontainers/picard:2.23.1--h37ae868_0 picard -Xmx4G -XX:ParallelGCThreads=1 MarkDuplicates INPUT=test.bam OUTPUT=markdup.bam CREATE_INDEX=true VALIDATION_STRINGENCY=SILENT METRICS_FILE=markdup.metrics' && du -h markdup.bam
[...]

PR #1540, opened 2020-07-09
Change default compression level for greater speed.

rhpvorderman:
### Description

As described in #1539 setting compression level to 2 has an effect of increasing the final file size with 6% but decreasing the compute time with 30+%. Also it is the default used in GATK. Therefore using level 2 seems preferable.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/cmdline/CommandLineProgram.java
+++ b/src/main/java/picard/cmdline/CommandLineProgram.java
@@ -111,7 +111,7 @@ public abstract class CommandLineProgram {
     public ValidationStringency VALIDATION_STRINGENCY = ValidationStringency.DEFAULT_STRINGENCY;
 
     @Argument(doc = "Compression level for all compressed files created (e.g. BAM and VCF).", common=true)
-    public int COMPRESSION_LEVEL = Defaults.COMPRESSION_LEVEL;
+    public int COMPRESSION_LEVEL = 2;  // Use GATK default level of 2 for much improved writing speed.
 
     @Argument(doc = "When writing files that need to be sorted, this will specify the number of records stored in RAM before spilling to disk. " +
             "Increasing this number reduces the number of file handles needed to sort the file, and increases the amount of RAM needed.", optional=true, common=true)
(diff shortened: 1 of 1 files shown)


lbergelson:
This seems like a good idea to me.  For intermediate results the trade is not worth it.  The last time we looked it seemed like 2/3 was the sweet spot with dramatically diminishing returns on compression performance vs compute. (That's why we changed it in gatk). For long term archiving 5 might be worth it, but definitely not for intermediate files.

rhpvorderman:
I have made a mistake in #1539 which caused me to mix up some Intel and jdk deflater result.

The jdk deflater performs best at level 1. The intel deflater gives a huge filesize at level 1. See discussion on #1541 for more context.

rhpvorderman:
@yfarjoun @lbergelson . I have read a bit more about the intel deflater. It uses its own optimized version of deflate on levels 1 and 2. It is much faster but also with a much larger filesize than the jdk deflater. On level 3 it uses an intel-optimized version of zlib. This gives the same filesizes as the JDK deflater (which presumably uses zlib or zlib-like implementations).

Setting it to compression level 3 will lead to consistent results in terms of filesize, regardless of deflater used. It is also faster than level 5. So I think it makes sense as a default.

pshapiro4broad:
This change replaces the htsjdk setting with a hardcoded setting for picard. If this is a generally a better setting, why not change this in htsjdk, not in picard? Then no change to picard would be necessary.

rhpvorderman:
@pshapiro4broad That's an excellent suggestion.

droazen:
@pshapiro4broad @rhpvorderman The next release of the GKL (which is due within the next ~month or so) is going to contain some significant updates to the compression libraries (updating otc zlib 1.2.8 to 1.2.11 and updating isa-l 2.21 to 2.28), and this is expected to have an effect on the performance of the Intel deflater/inflater. I would recommend waiting until this release can be benchmarked before changing any library-wide defaults in HTSJDK, as it's possible that it could affect the choice of the best default.

kbergin:
Hi @rhpvorderman @droazen @lbergelson - just going through open Picard PRs. Did this change get made into htsjdk? Should I close this PR?

rhpvorderman:
@kbergin, I have not proposed it yet. However, this issue can be closed as it is not in Picard scope.

PR #1541, opened 2020-07-09
Deprecate intel deflater and inflater

rhpvorderman:
### Description

As described in #1539. The speed differences with the the htsjdk inflater and deflater are marginal at best. There is no compelling reason to keep it as an extra dependency. 

I think one of the reasons for this is that many of the intel optimizations may have already landed in HTSJDK itself.

#### Content

#### Review


lbergelson:
@rhpvorderman I'm surprised by the results you've seen.  The last time we looked at it the intel deflator was substantially faster than the built in one. (~15 %)   (The 'htsjdk' deflater is just the default java deflater.).    

My understanding is that the intel deflator makes use of x86 processor features like SSE to be faster than the default java version.  It's very possible that 
A) the java version has improved in the last several years to incorporate similar features
B) there's something machine or configuration specific that's causing poor performance on some systems.  I don't really understand how singularity /docker interact with processor specific optimizations.
Could you provide some information about the system you're running on (What OS you're running and what your processor is.)

Of note, there's a new release of the intel library coming soon for the first time in several years which may bring speed improvements.

tfenne:
I would caution against this.  To make this kind of decision an analysis needs to be performed on a much broader set of parameters, including:
- Different intel and AMD CPUs
- Different OS versions (at least OS X and a recent linux)
- Different JDK versions (at least 8 and 10 or 11)
- In containers (i.e. docker) vs. on bare compute
- A broad range of data including e.g. deep coverage capture, light and deep coverage WGS
- Sorted and unsorted data

My testing is less rigorous than @rhpvorderman but shows quite different results.  On a ~500MB capture BAM with deep coverage on a mac with a fresh conda env created with `conda create -n picard -c bioconda picard` I see the following results:

```File                 Filesize  Runtime
sorted.intel.c1.bam  691M      0.28
sorted.intel.c2.bam  691M      0.29
sorted.intel.c3.bam  562M      0.56
sorted.intel.c4.bam  523M      0.66
sorted.intel.c5.bam  516M      0.71
sorted.jdk.c1.bam    607M      0.56
sorted.jdk.c2.bam    586M      0.62
sorted.jdk.c3.bam    565M      0.77
sorted.jdk.c4.bam    537M      0.77
sorted.jdk.c5.bam    515M      1.05
```

For any given compression level the inter deflator is substantially faster _in my environment_, though the resulting file sizes differ depending on the deflator.  Going from intel level 5 to JDK level 2 would barely speed things up for me and would inflate my file 13.5%.  Going to intel level 2 would be much faster, but increase the file size nearly 34%.

This is not to say that I'm against re-evaluating both the intel inflator/deflator and the default compression level.  But if that's to be done it needs to be based off a much broader range of testing.

rhpvorderman:
> My understanding is that the intel deflator makes use of x86 processor features like SSE to be faster than the default java version.

> I would caution against this. To make this kind of decision an analysis needs to be performed on a much broader set of parameters, including:

I agree. The difference might be that I was benchmarking on an AMD processor (Ryzen 5 3600).

If the intel libraries are provided by intel I fear the worst. Intel has a history of deliberately disabling features such as SSE if the processors are not made by Intel, despite the fact these features are supported on the processor of the competing brand. 

Since the difference is marginal on my system and seems to favour intel's deflater on another system it seems indeed good to be cautious.

What is strange is that the compression ratio's differ so much between the jdk deflater and the intel deflater.

>  I don't really understand how singularity /docker interact with processor specific optimizations.

Containerized process run on the same kernel as native processes. It is not virtualization of any kind. A colleague of mine once benchmarked conda environments vs singularity containers and found no difference in execution time. Most likely there is a small memory overhead for the container process.

> Could you provide some information about the system you're running on (What OS you're running and what your processor is.)

AMD ryzen 5 3600
32 GB DDR4-3200 CL16 RAM.
1 TB NVME SSD
Debian 10
Singularity version 3.4.2

I run picard in a biocontainer. Which means the java version comes packaged with it:

```
$ singularity shell docker://quay.io/biocontainers/picard:2.23.1--h37ae868_0
Singularity picard_2.23.1--h37ae868_0.sif:~> java --version
openjdk 11.0.1-internal 2018-10-16
OpenJDK Runtime Environment (build 11.0.1-internal+0-adhoc..src)
OpenJDK 64-Bit Server VM (build 11.0.1-internal+0-adhoc..src, mixed mode)
```

@tfenne can you check what happens when you run the benchmark with the same container? Docker could also be used instead of singularity. The image is `quay.io/biocontainers/picard:2.23.1--h37ae868_0`

I especially wonder what happens to the compression levels then, as for me there were no differences between the JDK deflater and the intel deflator for levels 2-5 in terms of filesize and speed (6 and above not tested).

rhpvorderman:
Some more benchmarks on a xeon system (e5-2660). 
I was curious to see how things fare on an old intel system (2012 processor introduced). 

The original bam file is 705 Mi, sorted with samtools sort. It is compressed with level 1.

```
Intel compression level 5
$ hyperfine -w 2 -r 5 'singularity exec -eip docker://quay.io/biocontainers/picard:2.23.1--h37ae868_0 picard -Xmx4G -XX:ParallelGCThreads=1 MarkDuplicates INPUT=test.bam OUTPUT=markdup.bam CREATE_INDEX=true VALIDATION_STRINGENCY=SILENT COMPRESSION_LEVEL=5 METRICS_FILE=markdup.metrics' && stat -c %s markdup.bam | numfmt --to=iec-i                                                                                            
Benchmark #1: singularity exec -eip docker://quay.io/biocontainers/picard:2.23.1--h37ae868_0 picard -Xmx4G -XX:ParallelGCThreads=1 MarkDuplicates INPUT=test.bam OUTPUT=markdup.bam CREATE_INDEX=true VALIDATION_STRINGENCY=SILENT COMPRESSION_LEVEL=5 METRICS_FILE=markdup.metrics
Time (mean ± σ):     202.168 s ±  6.038 s    [User: 241.791 s, System: 8.639 s]
  Range (min … max):   191.935 s … 207.536 s    5 runs
711Mi

Intel compression level 1
$ hyperfine -w 2 -r 5 'singularity exec -eip docker://quay.io/biocontainers/picard:2.23.1--h37ae868_0 picard -Xmx4G -XX:ParallelGCThreads=1 MarkDuplicates INPUT=test.bam OUTPUT=markdup.bam CREATE_INDEX=true VALIDATION_STRINGENCY=SILENT COMPRESSION_LEVEL=1 METRICS_FILE=markdup.metrics' && stat -c %s markdup.bam | numfmt --to=iec-i
Benchmark #1: singularity exec -eip docker://quay.io/biocontainers/picard:2.23.1--h37ae868_0 picard -Xmx4G -XX:ParallelGCThreads=1 MarkDuplicates INPUT=test.bam OUTPUT=markdup.bam CREATE_INDEX=true VALIDATION_STRINGENCY=SILENT COMPRESSION_LEVEL=1 METRICS_FILE=markdup.metrics
  Time (mean ± σ):     112.341 s ±  4.293 s    [User: 159.056 s, System: 7.862 s]
  Range (min … max):   106.777 s … 117.376 s    5 runs
960Mi

Intel compression level 2
$ hyperfine -w 2 -r 5 'singularity exec -eip docker://quay.io/biocontainers/picard:2.23.1--h37ae868_0 picard -Xmx4G -XX:ParallelGCThreads=1 MarkDuplicates INPUT=test.bam OUTPUT=markdup.bam CREATE_INDEX=true VALIDATION_STRINGENCY=SILENT COMPRESSION_LEVEL=2 METRICS_FILE=markdup.metrics' && stat -c %s markdup.bam | numfmt --to=iec-i
Benchmark #1: singularity exec -eip docker://quay.io/biocontainers/picard:2.23.1--h37ae868_0 picard -Xmx4G -XX:ParallelGCThreads=1 MarkDuplicates INPUT=test.bam OUTPUT=markdup.bam CREATE_INDEX=true VALIDATION_STRINGENCY=SILENT COMPRESSION_LEVEL=2 METRICS_FILE=markdup.metrics
  Time (mean ± σ):     124.673 s ±  6.990 s    [User: 167.482 s, System: 11.114 s]
  Range (min … max):   112.443 s … 129.351 s    5 runs
960Mi

Intel compression level 3
$ hyperfine -w 2 -r 5 'singularity exec -eip docker://quay.io/biocontainers/picard:2.23.1--h37ae868_0 picard -Xmx4G -XX:ParallelGCThreads=1 MarkDuplicates INPUT=test.bam OUTPUT=markdup.bam CREATE_INDEX=true VALIDATION_STRINGENCY=SILENT COMPRESSION_LEVEL=3 METRICS_FILE=markdup.metrics' && stat -c %s markdup.bam | numfmt --to=iec-i
Benchmark #1: singularity exec -eip docker://quay.io/biocontainers/picard:2.23.1--h37ae868_0 picard -Xmx4G -XX:ParallelGCThreads=1 MarkDuplicates INPUT=test.bam OUTPUT=markdup.bam CREATE_INDEX=true VALIDATION_STRINGENCY=SILENT COMPRESSION_LEVEL=3 METRICS_FILE=markdup.metrics
  Time (mean ± σ):     163.478 s ±  5.353 s    [User: 204.590 s, System: 8.764 s]
  Range (min … max):   160.655 s … 173.039 s    5 runs
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options. 
742Mi

JDK level 1
 hyperfine -w 2 -r 5 'singularity exec -eip docker://quay.io/biocontainers/picard:2.23.1--h37ae868_0 picard -Xmx4G -XX:ParallelGCThreads=1 MarkDuplicates INPUT=test.bam OUTPUT=markdup.bam CREATE_INDEX=true VALIDATION_STRINGENCY=SILENT COMPRESSION_LEVEL=1 METRICS_FILE=markdup.metrics USE_JDK_INFLATER=true USE_JDK_DEFLATER=true' && stat -c %s markdup.bam | numfmt --to=iec-i
Benchmark #1: singularity exec -eip docker://quay.io/biocontainers/picard:2.23.1--h37ae868_0 picard -Xmx4G -XX:ParallelGCThreads=1 MarkDuplicates INPUT=test.bam OUTPUT=markdup.bam CREATE_INDEX=true VALIDATION_STRINGENCY=SILENT COMPRESSION_LEVEL=1 METRICS_FILE=markdup.metrics USE_JDK_INFLATER=true USE_JDK_DEFLATER=true
  Time (mean ± σ):     147.671 s ±  5.456 s    [User: 194.152 s, System: 8.623 s]
  Range (min … max):   141.741 s … 156.001 s    5 runs
 
765Mi
```

Intel compression ratios are quite bad at level 1 and 2 on this system. Same as @tfenne's system. At compression level 3 it becomes acceptable. It uses somewhere between 160 and 173 seconds to process the BAM file on that ratio.

Benchmarks on this node were very slow. So I have not benchmarked intel compression ratio's above 3. But since results were similar to @tfenne's benchmarks I assume diminishing returns kick in.

The JDK deflater on level 1 creates an acceptable bam file size of 765M. It does so in 141-156 seconds.

On this particular system using the JDK deflater on level 1 seems to be the most attractive option.

One thing that is noticable from these benchmarks is that the JDK deflater consistently outputs a 765M bam file at level 1, whether it is on the AMD cpu or the intel CPU. The  intel deflater however seems to change algorithms based on the CPU architecture and has a wildly different level 2 depending on cpu architecture. In  my not so humble opinion this makes the intel deflater qualify as a **very** crappy piece of software. 

I will switch my defaults over to JDK inflaters and deflaters using compression level 1. It may not be the most performant choice in some edge cases, but it least it will be the most consistent across systems.

tfenne:
I would agree with @rhpvorderman that the fact that intel deflator yields more than nominally different file sizes for the same inputs depending on CPU architecture is concerning.  I would have expected the runtime to vary based on the CPU both due to available extended instructions and overall single-core speed, but to produce significantly different outputs is undesirable!

@lbergelson Perhaps this is worth mentioning to the collaborators at Intel if they are indeed working on a new version?

lbergelson:
@tfenne @rhpvorderman I also find that concerning that we get variable compression results on different systems. I suspect there is some attempt to identify processor features and switch between algorithms if they are present.  It sounds like maybe the feature detection isn't working quite right. 

I think there was a deliberate decision to swap compression performance for speed at level 1 in the intel deflater but I'm surprised that it seems to be changing algorithms depending on processor features ( especially on level 2 which I thought was using the same algorithm as 3+) I'll ask the intel team about it.  The next release coming is focusing on fixing bugs, but I believe it's pulling in the latest from the separate intel compression (ISA-L) acceleration project which should hopefully offer better compression/speed.  

Of note, @jkbonfield investigated a number of compression solutions for samtools and found that other implementations were more performant.  (http://www.htslib.org/benchmarks/zlib.html ).  No one on our team has had time to look into wrapping those in a way that we can use with htsjdk but it doing so would probably be a worthwhile project for anyone who is interested. 

I'm not certain that mark duplicates is the best benchmark as well.  It has the issue that is uses Snappy to compress intermediate output files.  Snappy is also a hardware optimized compression library which may be confounding results in hard to understand ways.  (If you want to use markduplicates for these test I would disable snappy by providing the following argument to the jvm `-Dsamjdk.snappy.disable=true`

rhpvorderman:
@tfenne @lbergelson  It appears I have made a mistake. The BAM file is the same size on an AMD system for level 2 (960M). Apparently I made some mistakes while benchmarking. I will reassess.

rhpvorderman:
After running the new benchmarks the results are as follows. Intel deflater level1: 960M, level2: 960M, level3: 742M. Jdk deflater level1: 765M, level2: 753M, level3: 742M.

The wall clock time for jdk deflator level 1 is 66 seconds. The intel deflater at level 1 takes 52 seconds. The intel deflater at level 3 takes 76 seconds.

If the jdk level 1 result is taken at 100% the results are as follows:

algorithm|  jdk level 1 | intel level 1 | intel level 3
--- | --- | --- | ---
size | 100% | 125% | 97%
time | 100% | 79% | 115%

Intel level 3 has a trade-off very much not worth it. Again, this may depend on the system used, but fact is that there are systems out there (mine) that do not perform well using these parameters, and hence it should not be a default.

It all comes down to if the increase in filesize is worth the decrause in compute time for the intel deflater at level 1. I do not think this is the case, given the multiple-GB files worked with in WGS. This is the difference between 108GB and 132 GB (measured with a real WGS sample). Storage constraints are a concern at this point.

In @tfenne's system intel level 3 is just as fast as jdk level 1. But not faster. So the intel deflater at best gets a 3% smaller filesize at the same speed, and at worst (on my system) it takes 15% longer. Consistency between systems is also important. I therefore feel that using the JDK deflater is still the best **default**. Maybe the intel deflator should be kept available, for systems where it is known to perform better.

jkbonfield:
If you can write a wrapper for libdeflate so it has the same Java bindings, then I really would recommend it.  For htslib comparing libdeflate vs vanilla zlib, libdeflate was both faster AND smaller for level 1 making it the optimal tool for temporary files.  Igzip is faster, but a bit specialist for BAM and I'm not sure how it'd pan out for other bgzf formats.  We didn't test it on vcf for example.  There's also SLZ, but it tends to be larger.

yfarjoun:
I think that this PR should be closed since the results it depends on are system specific...am I right?

rhpvorderman:
> I think that this PR should be closed since the results it depends on are system specific...am I right?

Well, the intel deflater shows inconsistent results over systems. The java deflater at compression level 1 gives good results consistently. So do we still want to use the intel deflater?

I think @jkbonfield makes a good suggestion by using to a different deflater alltogether. I added igzip in a PR to [xopen](https://github.com/marcelm/xopen) (a library which opens compressed files with an external program for python) and the results for cutadapt are quite good. A [33% reduction in compute time](https://github.com/marcelm/cutadapt/issues/425#issuecomment-658239834) for the whole application was measured.

Compression takes up a lot of cpu cycles. So I wonder if libdeflate with JNI bindings would help. Unfortunately I haven't been able to find a library.

rhpvorderman:
@yfarjoun closing this because it appears I am badly informed. I am sorry.

I checked the intel library (https://github.com/Intel-HLS/GKL) and it turns out it uses the isa-l libraries for level 1 and level 2 compression. 

The *current* iteration of the isa-l libraries provides igzip as well and it is better than gzip in any respect regardless of system. Unfortunately the GKL library uses a 3 year old version of isa-l. Once they update this, I expect better performance. And otherwise I believe the library could be improved with a few PRs. (For example, there is no difference between level 1 and level 2 at the moment when used in picard).

rhpvorderman:
I made this PR to integrate the newer versions of isa-l https://github.com/Intel-HLS/GKL/pull/109. igzip compression level 1 is both faster and smaller than gzip compression level 1 on my pc. So I hope this PR enables that for GKL as well.

PR #1542, opened 2020-07-09
Increase fraction of target coverage for panel metrics

hassanfa:
### Description

This PR adds extra steps for target coverage for fraction of bases for panel metrics. It adds extra steps to PCT_TARGET_BASES for 250, 500, 1000, 2500, 5000, 10000, 25000, 50000, and 100000.

Motivation for this PR is to provide extra coverage report: Targeted sequencing usually goes above 100X, and they can go at least up to 1000X.

----

#### Content

#### Review


yfarjoun:
hmmm. I am not pleased with the use of a headline metric as a defacto histogram. Since other targeted assays are interested in even higher coverages, this seems non-sustainable.

Perhaps it would be better to include the cumulative percent in the coverage histogram below the main metric?

hassanfa:
You are correct, this is not maintainable. Some panels can have as much as 3000X coverage, and some deepseq applications can go up to 20000-30000X. But it feels like those are special cases that coverage calculations should be done using other tools.

What do you think about adding just a few intervals instead? e.g. 250X, 500X, 750X, and 1000X.

I think 1000X is a good candidate to include. Reason is that at 1000X coverage, the assay has ~95% power to call variants at 1% allelic fraction. (I'm aware variant calling is one of the many use cases of targeted sequencing)

yfarjoun:
I would be on board with adding 300x, 1000x, 3000x, 10000x, 30000x, and 100000x and calling it a day. since this only gives an overall number, I don't see the point in having finer detail that that. I'm happy to ask others who deal with high depth assays

@fleharty @tfenne @nh13 @clintval

tfenne:
I mostly agree.  I think I would go log2 if anything, and add e.g. 250, 500, 1000, 2000, 4000 or something like that.

yfarjoun:
I'm not going split logs over this. 😆  As long as we have the information that is needed without adding extra fluff, I'm OK with it.

yfarjoun:
The problem with log2 is that you then "miss" the net power of 10. which is why I tried to do half-log10's....

yfarjoun:
How about: 1, 2.5, 5 and repeat that multiplied by powers of 10?
so: 
100, 250, 500,
1000, 2500, 5000,
10000, 25000, 50000,
100000

hassanfa:
1, 2.5, 5 sounds reasonable. Various clinical facilities on our side regularly sequence at least 3000X for small panels, 1000X-2000X for whole exome applications, and 20000X-30000X for deep sequencing (rare mutations and UMI applications).

tfenne:
Half log-10s works for me ;)

yfarjoun:
@hassanfa please modify your PR to add 250, 500, 1000, 2500, 5000, 10000, 25000, 50000, 100000

fleharty:
@hassanfa @yfarjoun 
250, 500, 1000, 2500, 5000, 10000, 25000, 50000, 100000
seems reasonable to me.

yfarjoun:
Great. Thanks for this PR. would you be able to add a test that shows that we are indeed able to reach these high coverage counts? I think you'd better use a SamRecordSetBuilder so that you dont' need to commit a file with 100000 records in it....

hassanfa:
Looked into SamRecordSetBuilder, and I think I can fix tests.

Ps: delay was due to vacation.

hassanfa:
@yfarjoun Could you please have a look at this when you've time? Thanks!

tfenne:
@yfarjoun Is this ready to be merged?

tfenne (approved):
LGTM.

yfarjoun:
There seems to be a problem with the branch test results not being reported. I suspect that it's because it from a fork.

@pshapiro4broad or @gbggrant do you know what needs to be done in order to get this merged?

gbggrant:
I agree about the branch tests results.  Can't you/we just merge as administrator?  @yfarjoun

yfarjoun:
I don't seem to be able to...if one of the admins would care to do the
honors, I :+1:

tfenne:
Thanks @gbggrant and @yfarjoun !

Issue #1546, opened 2020-07-14
GatherVcfs `--CREATE-INDEX` option doc refers to BAMs rather than VCFs.


cmnbroad:
This was originally reported in the [GATK repo](https://github.com/broadinstitute/gatk/issues/6697) by a user who noticed the issue in the gatk doc for GatherVcfs.

PR #1547, opened 2020-07-14
Update CREATE_INDEX arg javadoc.

cmnbroad:
Fixes https://github.com/broadinstitute/picard/issues/1546.


yfarjoun:
why not hook up the vcr writer to this argument instead of narrowing the scope?

cmnbroad:
@yfarjoun Its already hooked up. I expanded the doc to reflect the fact that its affects more than just BAM.

yfarjoun:
Oh I read it in reverse

PR #1549, opened 2020-07-21
Identify Contaminant Background Option and LoH handling

kachulis:
Adds an option to include a "background" sample to improve contaminant fingerprint extraction.  Also handles situation where contaminated sample is tumor with possible LoH and background sample is corresponding normal.  

----

#### Content

#### Review


testdata/picard/fingerprint/NA12878_chr1_fingerprint.vcf
```diff
@@ -0,0 +1,3399 @@
...
+##contig=<ID=HLA-DRB1*14:05:01,length=13933,assembly=38>
+##contig=<ID=HLA-DRB1*14:54:01,length=13936,assembly=38>
+##contig=<ID=HLA-DRB1*15:01:01:01,length=11080,assembly=38>
+##contig=<ID=HLA-DRB1*15:01:01:02,length=11571,assembly=38>
+##contig=<ID=HLA-DRB1*15:01:01:03,length=11056,assembly=38>
+##contig=<ID=HLA-DRB1*15:01:01:04,length=11056,assembly=38>
+##contig=<ID=HLA-DRB1*15:02:01,length=10313,assembly=38>
+##contig=<ID=HLA-DRB1*15:03:01:01,length=11567,assembly=38>
+##contig=<ID=HLA-DRB1*15:03:01:02,length=11569,assembly=38>
+##contig=<ID=HLA-DRB1*16:02:01,length=11005,assembly=38>
```
jessicaway, line 3370: Is it possible to remove header lines that are not specific to Chr1?

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
```diff
@@ -476,4 +480,108 @@ public void testWriteFingerprint() throws IOException {
 
         VcfTestUtils.assertVcfFilesAreEqual(vcfOutput, vcfExpected);
     }
+
+    @DataProvider(name = "identifyContaminantTestDataProvider")
+    public Object[][] identifyContaminantTestDataProvider() {
+        final Path NA12878_fingerprint = Paths.get(TEST_DATA_DIR.getAbsolutePath(), "NA12878_chr1_fingerprint.vcf");
```
jessicaway, line 486: To convert file to path you can use the following:
```suggestion
        final Path NA12878_fingerprint = TEST_DATA_DIR.toPath().resolve("NA12878_chr1_fingerprint.vcf");
```

src/main/java/picard/fingerprint/ExtractFingerprint.java
```diff
@@ -67,6 +76,16 @@
...
+            " this should be a sample from the same individual as the contaminated sample.  If we are extracting the fingerprint of the contaminated sample, this should be a sample from the same " +
+            "individual as the contaminating sample.  If the sample being removed is a tumor sample, this sample should optimally be a normal sample from the same individual.", optional = true)
+    public File BACKGROUND = null;
+
+    @Argument(doc = "Whether the sample being removed (not the sample whose fingerprint is being extracted) is a tumor sample.  This argument can only be set to true if a BACKGROUND sample is included." +
+            "  In this case, the BACKGROUND sample will be assumed to be a normal sample from the same individual.  In order to avoid potential confounding due to LoH, all fingerprinting sites which genotype " +
+            "as het in the BACKGROUND sample will be removed from the output fingerprint.  If this argument is set to true and no BACKGROUND sample is provided the tool will throw an exception.", optional = true)
+    public boolean REMOVED_IS_TUMOR = false;
```
jessicaway, line 87: would "BACKGROUND_IS_TUMOR" make sense here? Having trouble understanding the relationship between this arg and "BACKGROUND"

testdata/picard/fingerprint/NA12878_chr1_fingerprint.vcf
kachulis: good call, done

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
kachulis: 👍

src/main/java/picard/fingerprint/ExtractFingerprint.java
kachulis: I agree the naming here is confusing, but I don't think BACKGROUND_IS_TUMOR is quite right either.  What this flag indicates is that the sample which is mixed with the sample we are trying to fingerprint is a tumor.  The problem with "BACKGROUND_IS_TUMOR" is that the sample supplied in the BACKGROUND argument needs to be the associated normal, so I think that would make it more confusing.
Maybe "MIXED_WITH_TUMOR" would be clearer?

src/main/java/picard/fingerprint/ExtractFingerprint.java
```diff
@@ -67,6 +76,16 @@
             "If argument EXTRACT_CONTAMINATION=true the resulting samplename will be \"<SAMPLE>-contamination\" (if not provided).", optional = true)
     public String SAMPLE_ALIAS = null;
 
+    @Argument(doc = "SAM/BAM or VCF file of uncontaminated sample of the same individual whose information we are trying to remove.  If we are extracting the fingerprint of the contaminating sample," +
```
yfarjoun, line 79: I think that the use of "remove" and "extract" is very confusing here....this needs to be rethought and re-written, especially as there are currently two CLPs (ExtractFingerprint and IdentifyContaminant) that use 
this class.

src/main/java/picard/fingerprint/ExtractFingerprint.java
```diff
@@ -103,7 +126,9 @@ protected int doWork() {
             checker.setDefaultSampleID(SAMPLE_ALIAS);
         }
 
-        final Map<String, Fingerprint> fingerprintMap = checker.identifyContaminant(INPUT.toPath(), CONTAMINATION);
+
+        final Map<String, Fingerprint> fingerprintMap = checker.identifyContaminant(INPUT.toPath(), CONTAMINATION, BACKGROUND == null ? null : BACKGROUND.toPath(),
+                BACKGROUND == null ? null : getCorrespondingSampleMap(INPUT, BACKGROUND), REMOVED_IS_TUMOR);
```
yfarjoun, line 131: what if the background and the input contain data with the same samplename? will this break?

kachulis: It should be fine.  Can add a test though to make sure.

src/main/java/picard/fingerprint/FingerprintChecker.java
```diff
@@ -603,14 +612,73 @@ private FingerprintIdDetails createUnknownFP(final Path samFile, final SAMRecord
...
+                    Fingerprint.getFingerprintIdDetailsStringFunction(CrosscheckMetric.DataType.SAMPLE)).entrySet().stream()
+                    .collect(Collectors.toMap(e -> e.getKey().sample, Map.Entry::getValue));
+
+            //merge background fingerprint into contaminated fingerprint
+            for (final Map.Entry<String, Fingerprint> entry : fpIdDetailsBySampleBackground.entrySet()) {
+                final String sample = entry.getKey();
+                final Fingerprint backgroundFingerprint = entry.getValue();
+                final Fingerprint contaminatedFingerprint = fpIdDetailsBySample.get(sample);
+                for (final Map.Entry<HaplotypeBlock, HaplotypeProbabilities> fingerprintEntry : backgroundFingerprint.entrySet()) {
+                    if (contaminatedIsTumor && !isConfidentlyHom(fingerprintEntry.getValue())) {
```
yfarjoun, line 660: this will not work well in the cases where the fingerprints are not confident...is there a better way to do this that is more Bayesian?

src/main/java/picard/fingerprint/HaplotypeProbabilitiesFromContaminatorSequence.java
```diff
@@ -142,6 +142,24 @@ public HaplotypeProbabilitiesFromContaminatorSequence merge(final HaplotypeProba
         return this;
     }
 
+    public HaplotypeProbabilitiesFromContaminatorSequence mergeBackground(final HaplotypeProbabilities other) {
+        if (!this.getHaplotype().equals(other.getHaplotype())) {
+            throw new IllegalArgumentException("Mismatched haplotypes in call to HaplotypeProbabilities.merge(): " +
+                    getHaplotype() + ", " + other.getHaplotype());
+        }
+
+        final double[] backgroundLikelihoods = other.getLikelihoods();
```
yfarjoun, line 151: hmmm. can you update the math for this? the LaTeX file is here: `docs/fingerprinting/main.tex`

testdata/picard/fingerprint/NA12878_chr1_fingerprint.vcf
```diff
@@ -0,0 +1,31 @@
...
+chr1	150681174	.	T	C	.	.	.	AD:DP:PL	326,354:681:1186,0,1186
+chr1	150968149	.	T	G	.	.	.	AD:DP:PL	0,41:41:0,123,1220
+chr1	151399662	.	G	A	.	.	.	AD:DP:PL	46,0:46:1370,138,0
+chr1	154601491	.	T	C	.	.	.	AD:DP:PL	45,0:45:1340,135,0
+chr1	156232382	.	A	G	.	.	.	AD:DP:PL	39,0:39:1170,117,0
+chr1	156236330	.	G	A	.	.	.	AD:DP:PL	23,21:44:528,0,498
+chr1	189648927	.	C	T	.	.	.	AD:DP:PL	145,140:285:2375,0,2375
+chr1	214785415	.	C	T	.	.	.	AD:DP:PL	0,424:424:0,1274,3231
+chr1	230710048	.	A	G	.	.	.	AD:DP:PL	22,26:48:506,0,626
+chr1	236535170	.	G	A	.	.	.	AD:DP:PL	448,0:448:3230,1346,0
```
yfarjoun, line 31: in other tests we dont' check-in the index. this is possible here since reading this vcf without an index is fast enough.

yfarjoun: similarly, best to check-in a sam file and convert it to bam on the fly (or not) so that folks can see what is being checked in.

yfarjoun:
some priliminary comments. 

it seems that a rebase would be good since there are changes here that were already added in #1550 and #1556

src/main/java/picard/fingerprint/ExtractFingerprint.java
jessicaway: So if the input sample is a contaminated that includes a mix of tumor and normal, the normal would be passed as BACKGROUND and this flag would be set to "true"?

kbergin:
Hi @kachulis based on the PR comments I've added the On Hold label - seemed like there was a request to add tests. Feel free to remove if you don't think it's the appropriate label.

kachulis:
@kbergin yeah, this one has fallen off my plate, so On Hold seems best.  Hope to find a time to go back to it at some point.

Issue #1569, opened 2020-08-29
Inconsistency in MergeBamAlignment


yfarjoun:
A recent refactor in [htsjdk](https://github.com/samtools/htsjdk/pull/1461) and the changes in MergeBamAlignment that took advantage of said refactoring #1484, #1509, and #1513, result in occassionally inconsistent bam records:

for example:

> readname     99      chr3    93470586        60      **49M102S** =       93470586        **46**      * * MC:Z:102S46M**3S**  MQ:i:60 AS:i:71
 
Which has a read-length that is based on the **aligned** positions and clipping that is looking at the **unclipped** read start (note the 3S in the MateCigar) 

While this change in behavior was not intentional, it does raise the question, which is the more correct approach? 

1. Assume that the first **5' base** marks one end of the insert
2. Assume that the first **aligned 5' base** marks one end of the insert.

Since I think that there should be only one interpretation for the size of the insert, and that interpretation should be used for both TLEN and clipping overlapping reads, and I do not wish to change the definition of TLEN, I think that the right thing to do would be to revert the behaviour of MBA back to the original, i.e. use the aligned 5' ends to determine the ends of the Insert, but I'm open to hearing other arguments and options.

 

### Instructions

- Use a **concise** yet **descriptive** title;
- Determine whether your issue is a **bug report**, a **feature request**, or a **documentation request**;
- Choose the corresponding template block below and fill it in, replacing or deleting text in italics (surrounded by `_`) as appropriate;
- Delete the other template blocks and this header.

----

### Affected tool(s)
_Tool name(s), special parameters?_

### Affected version(s)
- [ ] Latest public release version [version?]
- [ ] Latest development/master branch as of [date of test?]

### Description 
_Describe the problem below. Provide **screenshots** , **stacktrace** , **logs** where appropriate._

#### Steps to reproduce
_Tell us how to reproduce this issue. If possible, include command lines that reproduce the problem and provide a minimal test case._

#### Expected behavior
_Tell us what should happen_

#### Actual behavior
_Tell us what happens instead_

----

### Tool(s) involved
_Tool name(s), special parameters?_

### Description
_Specify whether you want a modification of an existing behavior or addition of a new capability._
_Provide **examples**, **screenshots**, where appropriate._

----

### Tool(s) involved
_Tool name(s), parameters?_

### Description 
_Describe what needs to be added or modified._

----

yfarjoun:
[...]

PR #1571, opened 2020-09-01
Adjust MBA read overhang clipping to be consistent with past behavior

kachulis:
It has been noted in #1569 that #1484 changed the behavior of MBA when soft clipping overhanging reads.  This PR reverts the behavior when soft clipping so that it is the same as "old" (pre #1484) MBA.  The behavior when hard clipping is also adjusted slightly.  The behavior implemented here is as follows:
1. overlapping reads are soft clipped so that the 3' aligned end does not extend past the 5' aligned end of the mate. This reverts to the "old" MBA behavior when soft clipping
2. if using hard clipping, reads are additionally hard clipped so that the 3' unclipped end of the read does not extend past the 5' unclipped end of the mate. This implements the "new" MBA behavior when hard clipping.

The result of this behavior is that:
 - When running in soft clipping mode, the results are identical to the "old" MBA
 - When running in hard clipping mode, the hard clipped bases are the same as in the "new" MBA.  Additionally, any bases which would have been clipped by the "old" MBA but not by the "new" MBA are soft clipped.
 - Consistency between TLEN and the cigar is maintained for both situations, ie. we won't have reads with longer matches than the 5'-5' TLEN.

#### Content

#### Review

### DIFF
--- a/src/test/java/picard/sam/AbstractAlignmentMergerTest.java
+++ b/src/test/java/picard/sam/AbstractAlignmentMergerTest.java
@@ -60,11 +60,31 @@ public class AbstractAlignmentMergerTest extends CommandLineProgramTest {
                         default110LongR1Bases, default110LongR2Bases, default110LongR1Bases, default110LongR2Bases, null, null,
                         default110LongR1BaseQualities, default110LongR2BaseQualities, default110LongR1BaseQualities, default110LongR2BaseQualities, null, null}, // Non overlapping reads
 
+                {110, 100, 200, "110M", "110M", false, true, 100, 200, "110M", "110M", true,
+                        default110LongR1Bases, default110LongR2Bases, default110LongR1Bases, default110LongR2Bases, null, null,
+                        default110LongR1BaseQualities, default110LongR2BaseQualities, default110LongR1BaseQualities, default110LongR2BaseQualities, null, null},
+
+                {110, 100, 200, "110M", "110M", false, false, 100, 200, "110M", "110M", false,
+                        default110LongR1Bases, default110LongR2Bases, default110LongR1Bases, default110LongR2Bases, null, null,
+                        default110LongR1BaseQualities, default110LongR2BaseQualities, default110LongR1BaseQualities, default110LongR2BaseQualities, null, null}, // F1F2
+
+                {110, 100, 200, "110M", "110M", false, false, 100, 200, "110M", "110M", true,
+                        default110LongR1Bases, default110LongR2Bases, default110LongR1Bases, default110LongR2Bases, null, null,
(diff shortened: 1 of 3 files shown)


kachulis:
@yfarjoun I've added more tests, should be covering all cases for 5' softclips now.  Note that with respect to r1/r2, the tests are run once with the reads as presented from the DataProvider, and once with r1 and r2 swapped, so specifying a separate test with the reads swapped is unnecessary.

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -104,6 +138,67 @@
...
 
+                /*
+                                SSSSSSSSSSSMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM->
+                                     <-MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMSSSSSSSSS
+
+                                soft clipping becomes:
+                                SSSSSSSSSSSMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMSS->
+                                     <-SSSSMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMSSSSSSSSS
+
+                                hard clipping leaves as from soft clipping
```
yfarjoun, line 149: huh?

yfarjoun: Oh…I had a hard time parsing that sentance...I'd say, `hard-clipping results in the same as with soft-clipping in this case.`

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -104,6 +138,67 @@
...
+
+                /*
+                                SSSSSSSSSSSMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMSSSSS->
+                                     <-MMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMSSSSSSSSSSSSSSSSSSSSSS
+
+                                soft clipping becomes:
+                                SSSSSSSSSSSMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMSSSSSSSSSSSSSS->
+                                     <-SSSSMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMSSSSSSSSSSSSSSSSSSSSSS
+
+                                hard clipping leaves as from soft clipping
```
yfarjoun, line 169: same

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
kachulis: the point is that when running with hard clipping turned on the result is the same as when running with only soft clipping.  Maybe "running with hard clipping gives same result as running with soft clipping" would be clearer?

PR #1580, opened 2020-09-22
Include common (engine-level) args in help output by default.

cmnbroad:
Picard tools do not display `common` arguments (`TMP_DIR`, `VALIDATION_STRINGENCY`, or `COMPRESSION_LEVEL`, to name a few) in command line help output unless the user explicitly requests them using capital `-H`, or `--stdhelp`. This isn't obvious to users and it makes it hard to find these arguments (see https://github.com/broadinstitute/gatk/issues/6811).

This PR changes this behavior so that common args are displayed by default when the user asks for help or when a command line parsing error is detected. If approved, we'll need to make a corresponding change to the `usage` calls made by the parsers themselves so the behavior is the same when the user asks for help via -h.


droazen:
@yfarjoun Could we get a review on this one when you get a chance? (or nominate a different reviewer if you are too busy) It's needed to fix our issue https://github.com/broadinstitute/gatk/issues/6811. Thanks!

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -345,7 +345,7 @@ protected boolean parseArgs(final String[] argv) {
             ret = commandLineParser.parseArguments(System.err, argv);
         } catch (CommandLineException e) {
             // Barclay command line parser throws on parsing/argument errors
-            System.err.println(commandLineParser.usage(false,false));
+            System.err.println(commandLineParser.usage(true,false));
```
yfarjoun, line 348:
```suggestion
            System.err.println(commandLineParser.usage(true, false));
```

yfarjoun (commented):
hmmm. what the use-case for showing common args? they are either the default value (which shouldn't be a problem) or they should be provided by the user, (and then they should be aware of the value....) 

It seems that these changes make the "common" arguments a non-entity, meaning that we will no longer treat them differently from their non-common argument-relatives. since effort was expended to enable common/non-common argument, I'd like to understand why it is now no-longer important to distinguish between them. Alternatively, perhaps I'm mistaken and the distinction does occur in other places in the code....

cmnbroad:
@yfarjoun The problem is the common args are not easily discoverable, i.e., when the user encounters a problem and the solution involves use of a common arg (i.e., MarkDuplicates runs out of temp space and the solution is to use TMP_DIR); or when the user mis-spells the common arg (`--TMPDIR`). The common args don't show up by default in the output. 

As for where else common args treated specially, the generated online doc does list them separately. Not sure if there are others.

yfarjoun:
OK. so we should probably update the text in `org.broadinstitute.barclay.argparser.Argument#common` to reflect that it will be shown when parsing or validation errors occur.

yfarjoun:
though the barclay text. is strange, since barclay doesn't control when arguments are or are not shown.... :shrug:

yfarjoun:
👍

droazen:
@yfarjoun Can this one be merged? We'd like to get it in, but no one on our side has commit access. Thanks!

yfarjoun (approved):
OK. This works. though it adds to the clutter... 🤷

PR #1581, opened 2020-09-23
Fixed a bug in MergeBamAlignment

yfarjoun:
When there's a deletion in the base just beyond the 5' end of the mate, the code for clipping overhanging reads got confused and didn't clip anything. This commit fixes that.

Added tests that specifically cover this case.


src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -861,10 +860,21 @@ static int getReadPositionAtReferencePositionIgnoreSoftClips(final SAMRecord rec
...
 
-        return readPosition;
+        // if this returns zero, it means that there's a deletion at the position of the desired base,
+        // if the read is on the positive strand readPosition should be incremented by one to get the
+        // base __following__ the deletion rather than the one preceding it (which is what returnLastBaseIfDeleted argument in
+        // getReadPositionAtReferencePosition will do)
+        final int readPositionZeroOnDeletion = SAMRecord.getReadPositionAtReferencePosition(rec, pos + posShift, false);
+        final boolean increamentReadPosition = readPositionZeroOnDeletion == 0 &&
+                !rec.getReadNegativeStrandFlag() &&
+                readPosition < rec.getReadLength() - 1;
```
yfarjoun, line 874: this is to protect from a stupidly aligned read that has a deletion at the end somehow, so there are no bases __following__ that base.

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -199,6 +200,136 @@
...
+
+
+
+                //                  123456789.123456--789.123456789.123456789.123456789.123456789.123456789.123456789
+                //                                     MMMMMMMMMMMMMMMMMMsssssssss>
+                //                          <sssssMMiiMMMMMMMMMMMMMMMMMM
+                //                                  /\
+                //   with an insertion of two bases here should become
+                //                    123456789.123456789.123456789.123456789.123456789.123456789.123456789.123456789
+                //                                     MMMMMMMMMMMMMMMMMssssssssss>
```
fleharty, line 265: MMMMMMMMMMMMMMMMMssssssssss>

should be:

MMMMMMMMMMMMMMMMMSsssssssss>

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -199,6 +200,136 @@
...
+
+
+                {27, 14, 18, "7S20M", "16M2I2M7S", true, false, 18, 18, "11S16M", "16M11S", false,
+                        default27LongR1Bases, default27LongR2Bases, default27LongR1Bases, default27LongR2Bases, null, null,
+                        default27LongR1Qualities, default27LongR2Qualities, default27LongR1Qualities, default27LongR2Qualities, null, null},
+
+
+                //       123456789.123456789.123456789.1234--56789.123456789.123456789.123456789.123456789
+                //            <sssssssMMMMMMMMMMMMMMMMMMMM
+                //                        MMMMMMMMMMMMMMMMiMMMsssssss>
```
fleharty, line 306: The cigar for the second read is 16M2I2M7S so there should be two "i"s in the second read.

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -199,6 +200,136 @@
...
+                        default27LongR1Bases, default27LongR2Bases, default27LongR1Bases, default27LongR2Bases, null, null,
+                        default27LongR1Qualities, default27LongR2Qualities, default27LongR1Qualities, default27LongR2Qualities, null, null},
+
+                //       123456789.123456789.123456789.1234--56789.123456789.123456789.123456789.123456789
+                //            <sssssssMMMMMMMMMMMMMMMMMMMM
+                //                        MMMMMMMMMMMMMMMMMiMMsssssss>
+                //                                         /\
+                //         with an insertion of one bases here should become
+                //       123456789.123456789.123456789.123456789.123456789.123456789.123456789.123456789
+                //            <ssssssSSSSSMMMMMMMMMMMMMMMM
```
fleharty, line 324: <ssssssSSSSSMMMMMMMMMMMMMMMM
should be
<sssssssSSSSMMMMMMMMMMMMMMMM

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -199,6 +200,136 @@
...
+                //                        MMMMMMMMMMMMMMMMSSSSsssssss>
+
+
+                {27, 14, 18, "7S20M", "16M1I3M7S", true, false, 18, 18, "11S16M", "16M11S", false,
+                        default27LongR1Bases, default27LongR2Bases, default27LongR1Bases, default27LongR2Bases, null, null,
+                        default27LongR1Qualities, default27LongR2Qualities, default27LongR1Qualities, default27LongR2Qualities, null, null},
+
+                //       123456789.123456789.123456789.1234--56789.123456789.123456789.123456789.123456789
```
fleharty, line 318: 123456789.123456789.123456789.1234--56789.123456789.123456789.123456789.123456789
should be
123456789.123456789.123456789.1234-56789.123456789.123456789.123456789.123456789

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -199,6 +200,136 @@
...
+                //            <ssssssSSSSSMMMMMMMMMMMMMMMM
+                //                        MMMMMMMMMMMMMMMMSSSSsssssss>
+
+
+                {27, 14, 18, "7S20M", "16M2I2M7S", true, false, 18, 18, "11S16M", "16M11S", false,
+                        default27LongR1Bases, default27LongR2Bases, default27LongR1Bases, default27LongR2Bases, null, null,
+                        default27LongR1Qualities, default27LongR2Qualities, default27LongR1Qualities, default27LongR2Qualities, null, null},
+
+
+                //       123456789.123456789.123456789.1234--56789.123456789.123456789.123456789.123456789
```
fleharty, line 304: 123456789.123456789.123456789.1234--56789.123456789.123456789.123456789.123456789
should be
123456789.123456789.123456789.1234-56789.123456789.123456789.123456789.123456789

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -199,6 +200,136 @@
...
+                        default27LongR1Qualities, default27LongR2Qualities,default27LongR1Qualities, default27LongR2Qualities , null,null },
+
+
+                //       123456789.123456789.123456789.1234--56789.123456789.123456789.123456789.123456789
+                //            <sssssssMMMMMMMMMMMMMMMMMMMM
+                //                        MMMMMMMMMMMMMMMMiiMMsssssss>
+                //                                        /\
+                //        with an insertion of two bases here should become
+                //       123456789.123456789.123456789.123456789.123456789.123456789.123456789.123456789
+                //            <ssssssSSSSSMMMMMMMMMMMMMMMM
```
fleharty, line 295: //            <ssssssSSSSSMMMMMMMMMMMMMMMM
should be
                //            <sssssssSSSSMMMMMMMMMMMMMMMM

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -199,6 +200,136 @@
...
+                        default27LongR1Qualities, default27LongR2Qualities,default27LongR1Qualities, default27LongR2Qualities , null,null },
+
+
+                //       123456789.123456789.123456789.1234--56789.123456789.123456789.123456789.123456789
+                //            <sssssssMMMMMMMMMMMMMMMMMMMM
+                //                        MMMMMMMMMMMMMMMMMiiMsssssss>
+                //                                         /\
+                //         with an insertion of two bases here should become
+                //       123456789.123456789.123456789.123456789.123456789.123456789.123456789.123456789
+                //            <ssssssSSSSSMMMMMMMMMMMMMMMM
```
fleharty, line 280: //            <ssssssSSSSSMMMMMMMMMMMMMMMM
should be
                //            <sssssssSSSSMMMMMMMMMMMMMMMM

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -199,6 +200,136 @@
...
+                // 123456789.123456789.123456789.123456789.123456789.123456789.123456789.123456789
+                //                  MMMMMMMMMMMMMMMMMMsssssssss->
+                //     <ssssssssMMM-MMMMMMMMMMMMMMMM
+                //          should become
+                // 123456789.123456789.123456789.123456789.123456789.123456789.123456789.123456789
+                //                  MMMMMMMMMMMMMMMMSSsssssssss->
+                //      <ssssssssSSSMMMMMMMMMMMMMMMM
+
+
+                {27, 18, 14, "18M9S", "8S3M1D16M", false, true, 18, 18, "16M11S", "11S16M", false,
```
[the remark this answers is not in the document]
fleharty: could you also add a similar test with an insertion that spans the region to be clipped?

fleharty (approved):
I have no qualms with any of the code.
There are a few small typos in the figures for the comments.

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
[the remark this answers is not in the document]
yfarjoun: you mean a deletion? I'm not sure how to interpret this in the case of an insertion....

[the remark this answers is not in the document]
yfarjoun: I added the two-deletion test in line 290

fleharty:
@yfarjoun Looks good to me! 👍 
Thank you for your help on this!

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -192,24 +228,198 @@
...
+                // 123456789.123456789.123456789.123456789.123456789.123456789.123456789.123456789
+                //                  MMMMMMMMMMMMMMMMMMsssssssss>
+                //     <ssssssssMMM-MMMMMMMMMMMMMMMM
+                //          should become
+                // 123456789.123456789.123456789.123456789.123456789.123456789.123456789.123456789
+                //                  MMMMMMMMMMMMMMMMSSsssssssss>
+                //      <ssssssssSSSMMMMMMMMMMMMMMMM
+
+
+                {27, false, 18, 14, "18M9S", "8S3M1D16M", false, true, 18, 18, "16M11S", "11S16M",
```
kachulis, line 248: Could you add tests for the hardclipping case for these examples involving insertions and deletions as well?  Based on following the code and a simple test I tried out it seems to give an answer that I would considered "correct", but it gets there in a pretty unsafe round-about sort of way.  I suspect if it's not well tested the harclipping behavior in these cases could easily be unintentionally and unknowingly changed as a side effect of some future pr.

kachulis (commented):
Oh man, thanks for the fix @yfarjoun!

This looks like it works (and THANK you for the detailed comments, made it much easier to understand), but I think it could be done in a perhaps cleaner way.  Essentially, the problem is that `SAMRecord.getReadPositionAtReferencePosition` is only able to give us the base before a deletion, which in a forward read is the 3' most base not to be clipped, and in a reverse read is the 5' most base to be clipped.  But we are always asking for the 5' most base to be clipped when we call `getReadPositionAtReferencePositionIgnoreSoftClips`, which leads to needing to do these additional checks for deletion related issues.
Instead, we could adjust the way we call `getReadPositionAtReferencePositionIgnoreSoftClips` to work with what `SAMRecord.getReadPositionAtReferencePosition` is able to give us.  That is, for reverse reads we still ask for the position of the 5' most base to be clipped, but for forward reads we ask for the position of the 3' most base not to be clipped, and then increment in order to pass the position of the 5' most base to be clipped to `clip3PrimeEndOfRead`.  In this way we will always get the correct base position, and won't need to worry about figuring out whether there is a deletion adjacent to the clipping.

Thinking about this alternative solution, I was concerned initially that this would cause problems when there are insertions adjacent to the to be clipped portions of the read.  Based on your tests, however, I don't think this will be an issue since insertions end up being merged into adjacent soft-clips at the ends of the reads.  However, regardless of whether you take the above alternative solution, I think adding hard-clipping versions of the insertion/deletion tests would be a good idea, since I don't think the behavior in these edge cases is particularly robust against unintended consequences of future code changes.

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
yfarjoun: added.

yfarjoun:
I'm not sure about your proposal to change the logic. the problem is that the +1 needs to be added only in the case that the location in the reference is a deletion AND the read is on the negative strand...the +1 that is done after the call to getRefPos... is performed regardless of the presence of a deletion, so I'm not sure how they can be simplified away...but in any case, the tests are in so if you'd like to experiment, go ahead!.

kachulis (approved):

PR #1584, opened 2020-09-29
Safely run DownsampleSam multiple times

kachulis:
If DownsampleSam is run twice on the same file, the results can end up being incorrect.  This is because the ConstantMemory and Chained strategies both employ a hash based sampling which is initialized with a random seed.  If the seed used in two runs is the same, then the hashes associated with each read pair will be the same.   

As an example of the problem this causes, consider a case where we have Downsampled with the ConstantMemory strategy and a probability of 0.5.  What this means is that a hash is calculated for each read pair, and the hashes are (with appropriate normalization) uniformly distributed between 0 and 1.  All read pairs which hash to values less than 0.5 will pass the downsampling and be included in the output.  We now attempt to downsample the output a second time with ConstantMemory and probability 0.5, and the same seed as in the previous run.  Since the seed is the same, each read pair hashes to the same value, which is always less than 0.5, so all reads pass the downsampling.  The second downsampling has effectively done nothing.

If we are using the HighAccuracy strategy, there will be no issue, since this is a shuffle-and-cut strategy, and so compression of the hash-space causes no problems.

If we are using the Chained strategy, there will also be no issue.  The chained strategy first uses a hash based downsample to oversample, and then uses a shuffle-and-cut to get down to the correct fraction.  Since previous runs of downsampling can only serve to compress the hash space, the hash based part of this strategy will only end up oversampling.  Since the shuffle-and-cut part of this strategy uses the measured downsampled fraction (instead of the intended downsampled fraction) in its calculations, this strategy will still work, although it's memory footprint improvement over the HighAccuracy strategy will be reduced.

So, there are really three possibilities that need to be considered:
1) ConstantMemory now, ConstantMemory previously: in this case an earlier run with ConstantMemory strategy will have compressed the hash space by a factor of p_prev (ie, the previously downsampling probability).  We can therefore get the correct downsampling by internally using a value of p_prev * p_now to compare to the hashes.  Multiple previous runs with ConstantMemory strategy can be easily chained together, with each simply compressing the hash space by a factor of its downsampling probability.  Runs which used a different seed should have no effect on the hash space, so do not need to be considered when adjusting the internal probability.

2) ConstantMemory now, HighAccuracy previously:  since the HighAccuracy strategy has no effect on the hash space, this will work unchanged.

3) ConstantMemory now, Chained previously:  This is a slightly more complicated version of "ConstantMemory now, ConstantMemory previously".  The previous Chained run will have compressed the hash space, but the proportion is defined by `ChainedDownsamplingIterator.adjustProportion`.  We could technically adjust the internally used probability in the same way as "ConstantMemory now, ConstantMemory previously", but the fact that it depends on a private method in htsjdk makes me hesitant to do this.  For now, I am throwing an exception in this case.  Perhaps if `ChainedDownsamplingIterator.adjustProportion` is made public in the future then handling this case as well would be more reasonable.  Of course, when the seed is different there is no problem.

This PR implements all of the above, as well as adds tests of these types of repeated downsampling cases.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/sam/DownsampleSam.java
+++ b/src/main/java/picard/sam/DownsampleSam.java
@@ -235,12 +235,22 @@ public class DownsampleSam extends CommandLineProgram {
                                 new LegacyCommandLineArgumentParser(previousDownsample) : new CommandLineArgumentParser(previousDownsample);
                         previousParser.parseArguments(System.err, previousArgs);
 
-                        if (previousDownsample.STRATEGY == Strategy.ConstantMemory || previousDownsample.STRATEGY == Strategy.Chained) {
-                            //are the random seeds the same?
-                            if (RANDOM_SEED.equals(previousDownsample.RANDOM_SEED)) {
-                                throw new PicardException("Attempting to downsample using STRATEGY " + STRATEGY + " on data that has already been downsampled using STRATEGY " + previousDownsample.STRATEGY + " and same " +
-                                        "RANDOM_SEED  " + RANDOM_SEED + ".  In order to downsample this data further, you must either supply a different RANDOM_SEED, or use STRATEGY HighAccuracy or Chained");
-                            }
+                        if (previousDownsample.STRATEGY == Strategy.ConstantMemory && RANDOM_SEED.equals(previousDownsample.RANDOM_SEED)) {
+                            /*
+                            we need to decrease PROBABILITY by the value of the previously used PROBABILITY, since reads which passed
+                            the previous downsampling will hash to values uniformly distributed between 0 and previous PROBABILITY instead
+                            of between 0 and 1.
+                             */
+                            PROBABILITY *= previousDownsample.PROBABILITY;
+                        }
+
+                        if (previousDownsample.STRATEGY == Strategy.Chained && RANDOM_SEED.equals(previousDownsample.RANDOM_SEED)) {
+                            /*
+                            In this case we must fail, since the hash downsampling in the Chained strategy is adjusted dynamically, so we don't know how
+                            the hashes of the reads that previously passed are distributed.
+                             */
+                            throw new PicardException("Attempting to downsample using STRATEGY " + STRATEGY + " on data that has already been downsampled using STRATEGY " + previousDownsample.STRATEGY + " and same " +
+                                    "RANDOM_SEED  " + RANDOM_SEED + ".  In order to downsample this data further, you must either supply a different RANDOM_SEED, or use STRATEGY HighAccuracy or Chained");
                         }
                     }
                 }
--- a/src/main/java/picard/sam/DownsampleSam.java
+++ b/src/main/java/picard/sam/DownsampleSam.java
@@ -238,6 +238,7 @@ public class DownsampleSam extends CommandLineProgram {
 
         SAMProgramRecord pgRecord = getPGRecord(header);
         pgRecord.setAttribute(RANDOM_SEED_TAG, RANDOM_SEED.toString());
+        header.addProgramRecord(pgRecord);
         final SAMFileWriter out = new SAMFileWriterFactory().makeSAMOrBAMWriter(header, true, OUTPUT);
         final ProgressLogger progress = new ProgressLogger(log, (int) 1e7, "Wrote");
         final DownsamplingIterator iterator = DownsamplingIteratorFactory.make(in, STRATEGY, PROBABILITY, ACCURACY, RANDOM_SEED);
--- a/src/main/java/picard/sam/DownsampleSam.java
+++ b/src/main/java/picard/sam/DownsampleSam.java
@@ -239,7 +239,7 @@ public class DownsampleSam extends CommandLineProgram {
                             //are the random seeds the same?
                             if (RANDOM_SEED.equals(previousDownsample.RANDOM_SEED)) {
                                 throw new PicardException("Attempting to downsample using STRATEGY " + STRATEGY + " on data that has already been downsampled using STRATEGY " + previousDownsample.STRATEGY + " and same " +
-                                        "RANDOM_SEED  " + RANDOM_SEED + ".  In order to downsample this data further, you must either supply a different RANDOM_SEED, or use STRATEGY HighAccuracy");
+                                        "RANDOM_SEED  " + RANDOM_SEED + ".  In order to downsample this data further, you must either supply a different RANDOM_SEED, or use STRATEGY HighAccuracy or Chained");
                             }
                         }
                     }
--- a/src/main/java/picard/sam/DownsampleSam.java
+++ b/src/main/java/picard/sam/DownsampleSam.java
@@ -191,6 +191,8 @@ public class DownsampleSam extends CommandLineProgram {
 
     private final Log log = Log.getInstance(DownsampleSam.class);
 
+    public static final String RANDOM_SEED_TAG = "rs";
+
     @Override
     protected String[] customCommandLineValidation() {
         if (PROBABILITY < 0 || PROBABILITY > 1)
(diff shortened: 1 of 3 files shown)


kachulis:
@cmnbroad for this PR I need to parse a previous command line stored in the header of a bam.  Would appreciate your review of how I implemented this using barclay, and any guidance if there is a better way of doing it.  Thanks!

tfenne:
Apologies for the drive-by review, but this seems a rather complicated solution to a somewhat esoteric problem.  I'd suggest a couple of other things:

1. Change the default seed to be `null` such that the tool generates a unique seed each time by default and let users specify a constant seed if they want repeatable downsampling, or
2. Detect one or more PG line(s) from previous run(s) and if any run has the same seed as the current run, modify the seed in some deterministic way (e.g. doubling it until it is unique)

kachulis:
Thanks for the suggestions @tfenne, I like your second proposal, so have switched to implementing that.  I have also switched to adding a tag for the random seed used to the pg record in the header instead of parsing the previous command line, which simplifies the code quite a bit.

src/main/java/picard/sam/DownsampleSam.java
```diff
@@ -212,7 +218,28 @@ protected int doWork() {
...
-        final SAMFileWriter out = new SAMFileWriterFactory().makeSAMOrBAMWriter(in.getFileHeader(), true, OUTPUT);
+        final SAMFileHeader header = in.getFileHeader().clone();
+
+        if (STRATEGY == Strategy.ConstantMemory || STRATEGY == Strategy.Chained) {
+            //if running using ConstantMemory or Chained strategy, need to check if we have previously run with the same random seed
+            //collect previously used seeds
+            final Set<Integer> previousSeeds = new HashSet<>();
+            for (final SAMProgramRecord pg : header.getProgramRecords()) {
+                if (pg.getProgramName() != null && pg.getProgramName().equals(PG_PROGRAM_NAME)) {
+                    final int previousSeed = Integer.parseInt(pg.getAttribute(RANDOM_SEED_TAG));
```
yfarjoun, line 229: you might come across a PG that will not have the `rs` tag...should protect against that.

src/main/java/picard/sam/DownsampleSam.java
```diff
@@ -212,7 +218,28 @@ protected int doWork() {
...
+                    previousSeeds.add(previousSeed);
+                }
+            }
+
+            while (previousSeeds.contains(RANDOM_SEED)) {
+                RANDOM_SEED *= 2;
```
yfarjoun, line 235: This is too small a space to traverse. I would pull out random seed from a RNG that itself is initialized with the provided seed.

yfarjoun: also, if you modified the seed, put out a warning that the provided random seed had already been used and that you are using a different seed (and state what it is)

src/test/java/picard/sam/DownsampleSamTest.java
```diff
@@ -148,5 +151,52 @@ private void testDownsampleWorker(final File samFile, final double fraction, fin
...
         }
+        return downsampled;
+    }
+
+
+
+    @DataProvider(name = "RepeatedDownsamplingProvider")
+    public Object[][] repeatedDownsamplingProvider() {
+        final List<Object[]> rets = new ArrayList<>();
+        rets.add(new Object[]{Arrays.asList(DownsamplingIteratorFactory.Strategy.ConstantMemory, ConstantMemory), Arrays.asList(2,1)});
```
yfarjoun, line 162: use 0 and MAX_INT as two options.

src/main/java/picard/sam/DownsampleSam.java
```diff
@@ -183,6 +187,8 @@
 
     private final Log log = Log.getInstance(DownsampleSam.class);
 
+    public static final String RANDOM_SEED_TAG = "rs";
```
yfarjoun, line 190: update the doc for the seed argument to reflect some of the logic here.

src/test/java/picard/sam/DownsampleSamTest.java
```diff
@@ -148,5 +151,52 @@ private void testDownsampleWorker(final File samFile, final double fraction, fin
...
+                seeds.add(seed);
+                strategies.add(strategy);
+            }
+            rets.add(new Object[]{strategies, seeds});
+        }
+
+        return rets.toArray(new Object[0][]);
+    }
+
+    @Test(dataProvider = "RepeatedDownsamplingProvider")
```
yfarjoun, line 194: please add a test that shows that downsampling twice with 50% actually yields 25%.

yfarjoun (changes requested):
I agree that this is much better.

some ideas for further improvment.

src/main/java/picard/sam/DownsampleSam.java
kachulis: done

src/test/java/picard/sam/DownsampleSamTest.java
kachulis: `testDownsampleWorker` does this in some sense, in that it tests that each downsampling step actually downsamples the data as intended.  I have also added code to test the the cumulative downsampling is as expected.

kachulis:
Thanks for the review @yfarjoun, back to you

yfarjoun (commented):
more comments

src/main/java/picard/sam/DownsampleSam.java
```diff
@@ -212,7 +220,45 @@ protected int doWork() {
         }
 
         final SamReader in = SamReaderFactory.makeDefault().referenceSequence(REFERENCE_SEQUENCE).open(SamInputResource.of(INPUT));
-        final SAMFileWriter out = new SAMFileWriterFactory().makeSAMOrBAMWriter(in.getFileHeader(), true, OUTPUT);
+        final SAMFileHeader header = in.getFileHeader().clone();
+
+        if (STRATEGY == Strategy.ConstantMemory || STRATEGY == Strategy.Chained) {
+            //if running using ConstantMemory or Chained strategy, need to check if we have previously run with the same random seed
+            //collect previously used seeds
+            final int USER_SEED = RANDOM_SEED;
```
yfarjoun, line 228: RANDOM_SEED may be null.

src/main/java/picard/sam/DownsampleSam.java
```diff
@@ -212,7 +220,45 @@ protected int doWork() {
...
+                    final int previousSeed = Integer.parseInt(previousSeedString);
+                    previousSeeds.add(previousSeed);
+                }
+            }
+
+            final Random rnd = new Random(RANDOM_SEED);
+            while (previousSeeds.contains(RANDOM_SEED)) {
+                final int previousSeed = RANDOM_SEED;
+                RANDOM_SEED = rnd.nextInt();
+                log.warn("DownsampleSam has been run before on this data with the seed .  The random seed will be changed to avoid using the " +
```
yfarjoun, line 250: extra space before period...were you going to put a more informative error message here?

src/main/java/picard/sam/DownsampleSam.java
yfarjoun: also, no reason to ALL_CAPS the userSeed variable. We usually reserve that for input arguments and statics

src/test/java/picard/sam/DownsampleSamTest.java
yfarjoun: 👍

src/test/java/picard/sam/DownsampleSamTest.java
```diff
@@ -148,5 +153,63 @@ private void testDownsampleWorker(final File samFile, final double fraction, fin
...
+        rets.add(new Object[]{Arrays.asList(ConstantMemory, HighAccuracy), Arrays.asList(1,1)});
+        rets.add(new Object[]{Arrays.asList(Chained, Chained), Arrays.asList(1,1)});
+        rets.add(new Object[]{Arrays.asList(HighAccuracy, Chained), Arrays.asList(1,1)});
+        rets.add(new Object[]{Arrays.asList(HighAccuracy, HighAccuracy), Arrays.asList(1,1)});
+        rets.add(new Object[]{Arrays.asList(Chained, HighAccuracy), Arrays.asList(1,1)});
+
+        //randomly generate some sequences to test out
+        final Strategy[] availableStratagies = Strategy.values();
+        final Random random = new Random(12345);
+        for (int i =0; i<20; i++) {
```
yfarjoun, line 184: spaces

src/test/java/picard/sam/DownsampleSamTest.java
```diff
@@ -148,5 +153,63 @@ private void testDownsampleWorker(final File samFile, final double fraction, fin
             TestNGUtil.assertGreaterThan(SamTestUtil.countSamTotalRecord(downsampled), fraction * .8 * SamTestUtil.countSamTotalRecord(samFile));
             TestNGUtil.assertLessThan(SamTestUtil.countSamTotalRecord(downsampled), fraction * 1.2 * SamTestUtil.countSamTotalRecord(samFile));
         }
+        return downsampled;
+    }
+
```
yfarjoun, line 158: too many spaces

yfarjoun (commented):
a few straggling nits.

src/main/java/picard/sam/DownsampleSam.java
kachulis: null RANDOM_SEED is checked for earlier, and at this point in the code it can't be null.

src/main/java/picard/sam/DownsampleSam.java
```diff
@@ -212,7 +220,45 @@ protected int doWork() {
...
+            final Integer userSeed = RANDOM_SEED;
+            final Set<Integer> previousSeeds = new HashSet<>();
+            for (final SAMProgramRecord pg : header.getProgramRecords()) {
+                if (pg.getProgramName() != null && pg.getProgramName().equals(PG_PROGRAM_NAME)) {
+                    final String previousSeedString = pg.getAttribute(RANDOM_SEED_TAG);
+                    if (previousSeedString == null) {
+                        /* The previous seed was not recorded.  In this case, the current seed may be the same as the previous seed,
+                        so we will change it to a randomly selected seed, which is very likely to be unique
+                         */
+                        RANDOM_SEED = new Random(pg.hashCode()).nextInt();
```
yfarjoun, line 237: This hash is effectively random since it's the Object.hashCode(), right? but it gives the illusion of being deterministic. 

I think it should be as deterministic as possible, but am not sure what here can be used as a seed. perhaps the collection of PU fields from the RG tags? not sure how you'd like to resolve this, 
but I don't like hidden randomness.

src/main/java/picard/sam/DownsampleSam.java
```diff
@@ -212,7 +220,45 @@ protected int doWork() {
...
+                        log.warn("DownsampleSam has been run before on this data, but the previous seed was not recorded.  The used seed will be changed to minimize the chance of using the" +
+                                " same seed as in a previous run.");
+                    }
+                    final int previousSeed = Integer.parseInt(previousSeedString);
```
yfarjoun, line 241: no need to split this into two lines.

src/main/java/picard/sam/DownsampleSam.java
```diff
@@ -212,7 +220,45 @@ protected int doWork() {
...
+                RANDOM_SEED = rnd.nextInt();
+                log.warn("DownsampleSam has been run before on this data with the seed " + RANDOM_SEED + ".  The random seed will be changed to avoid using the " +
+                        "same seed as previously.");
+            }
+            if (!userSeed.equals(RANDOM_SEED)) {
+                log.warn("RANDOM_SEED has been changed to " + RANDOM_SEED + ".");
+            }
+        }
+
+        SAMProgramRecord pgRecord = getPGRecord(header);
```
yfarjoun, line 258: final

yfarjoun (changes requested):
Sorry about the staggered review...but I noticed more things.

src/main/java/picard/sam/DownsampleSam.java
yfarjoun: I meant the extra variable `previousSeed`

kachulis: that makes more sense

src/main/java/picard/sam/DownsampleSam.java
kachulis: `SAMProgramRecord` has a hashCode method that avoids dropping into `Object.hashCode()`, so I believe this should be deterministic.

src/main/java/picard/sam/DownsampleSam.java
```diff
@@ -212,7 +220,42 @@ protected int doWork() {
...
+                if (pg.getProgramName() != null && pg.getProgramName().equals(PG_PROGRAM_NAME)) {
+                    final String previousSeedString = pg.getAttribute(RANDOM_SEED_TAG);
+                    if (previousSeedString == null) {
+                        /* The previous seed was not recorded.  In this case, the current seed may be the same as the previous seed,
+                        so we will change it to a randomly selected seed, which is very likely to be unique
+                         */
+                        RANDOM_SEED = new Random().nextInt();
+                        log.warn("DownsampleSam has been run before on this data, but the previous seed was not recorded.  The current seed has been set to " + RANDOM_SEED + " to avoid using the " +
+                                "same seed as previously.");
+                        break;
```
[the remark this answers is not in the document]
yfarjoun: sorry, I missed that.

src/main/java/picard/sam/DownsampleSam.java
yfarjoun: thanks for pointing that out.

yfarjoun (approved):

PR #1589, opened 2020-10-01
MBA deletion edge case alternate solution

kachulis:
This is the alternate solution discuss in #1581.  I think it makes the code a bit cleaner and more understandable.

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/sam/AbstractAlignmentMerger.java
+++ b/src/main/java/picard/sam/AbstractAlignmentMerger.java
@@ -829,14 +829,14 @@ public abstract class AbstractAlignmentMerger {
          the queried base when the queried base is in a deletion on a negative strand read
          */
 
-        int negClipFrom = getReadPositionAtReferencePositionIgnoreSoftClips(neg, posStart - 1);
-        negClipFrom = negClipFrom > 0 ? (neg.getReadLength() + 1) - negClipFrom : 0;
+        int negFirstBaseToClip = getReadPositionAtReferencePositionIgnoreSoftClips(neg, posStart - 1);
+        negFirstBaseToClip = negFirstBaseToClip > 0 ? (neg.getReadLength() + 1) - negFirstBaseToClip : 0;
 
         if(posLastUnclipped > 0 && posLastUnclipped < pos.getReadLength()) {
             clip3PrimeEndOfRead(pos, posLastUnclipped + 1, hardClipReads);
         }
-        if(negClipFrom > 0) {
-            clip3PrimeEndOfRead(neg, negClipFrom, hardClipReads);
+        if(negFirstBaseToClip > 0) {
+            clip3PrimeEndOfRead(neg, negFirstBaseToClip, hardClipReads);
         }
     }
 
--- a/src/main/java/picard/sam/AbstractAlignmentMerger.java
+++ b/src/main/java/picard/sam/AbstractAlignmentMerger.java
@@ -820,6 +820,13 @@ public abstract class AbstractAlignmentMerger {
           For the positive strand, we ask for the position of the 3' most base which will not be clipped, and then increment to find the 5' most base to clip.
           We do this because getReadPositionAtReferencePositionIgnoreSoftClips will return the base before a deletion, so will return the 3' most base before
           the queried base when the queried base is in a deletion on a positive strand read
+
+            3'<SSSSSSSSMMMMMMMMMMMM5'
+                     5'MMMMMMMMMMMMSSSSSSSS>3'
+                                  ||
+                                  ||---> 5' most base to clip
+                                  |
+                                  |---> 3' most base not to clip
          */
 
         final int posLastUnclipped = getReadPositionAtReferencePositionIgnoreSoftClips(pos, negEnd);
--- a/src/main/java/picard/sam/AbstractAlignmentMerger.java
+++ b/src/main/java/picard/sam/AbstractAlignmentMerger.java
@@ -796,36 +796,36 @@ public abstract class AbstractAlignmentMerger {
      * starts and ends of each read are the same.
      */
     protected static void clipForOverlappingReads(final SAMRecord read1, final SAMRecord read2, final boolean useHardClipping) {
-        // If both reads are mapped, see if we need to clip the ends due to small insert size
-        if (!(read1.getReadUnmappedFlag() || read2.getReadUnmappedFlag())) {
-            if (read1.getReadNegativeStrandFlag() != read2.getReadNegativeStrandFlag()) {
-                final SAMRecord pos = (read1.getReadNegativeStrandFlag()) ? read2 : read1;
-                final SAMRecord neg = (read1.getReadNegativeStrandFlag()) ? read1 : read2;
+        // Only clip if both reads are mapped, on opposite strands, overlapping
+        if (!read1.getReadUnmappedFlag() && !read2.getReadUnmappedFlag() &&
+                read1.getReadNegativeStrandFlag() != read2.getReadNegativeStrandFlag() &&
+                read1.overlaps(read2)) {
+            final SAMRecord pos = (read1.getReadNegativeStrandFlag()) ? read2 : read1;
+            final SAMRecord neg = (read1.getReadNegativeStrandFlag()) ? read1 : read2;
 
-                // Innies only -- do we need to do anything else about jumping libraries?
-                if (pos.getAlignmentStart() < neg.getAlignmentEnd()) {
-                    // first we softclip the 3' end of each read so that its 3' aligned end does not extends past the 5' aligned start of it's mate
-                    clip3primeEndsTo5primeEnds(pos, neg, false, false);
+            // first we softclip the 3' end of each read so that its 3' aligned end does not extends past the 5' aligned start of it's mate
+            clip3primeEndsTo5primeEnds(pos, neg, false, false);
 
-                    if (useHardClipping) {
-                        // if we want to hardclip, we additionally hardclip the 3' end of each read so that its 3' unclipped end does not extend past the 5' unclipped start of its mate
-                        clip3primeEndsTo5primeEnds(pos, neg, true, true);
-                    }
-                }
+            if (useHardClipping) {
+                // if we want to hardclip, we additionally hardclip the 3' end of each read so that its 3' unclipped end does not extend past the 5' unclipped start of its mate
+                clip3primeEndsTo5primeEnds(pos, neg, true, true);
             }
         }
     }
+
 
     private static void clip3primeEndsTo5primeEnds(final SAMRecord pos, final SAMRecord neg, final boolean hardClipReads, final boolean useUnclippedEnds) {
         final int negEnd = useUnclippedEnds? neg.getUnclippedEnd() : neg.getEnd();
         final int posStart = useUnclippedEnds? pos.getUnclippedStart() : pos.getStart();
         //for positive strand reads, we ask for the position of the 3' most base which will not be clipped, and then increment to find the 5' most base to
-        final int posClipFrom = getReadPositionAtReferencePositionIgnoreSoftClips(pos, negEnd) + 1;
+        final int posLastUnclipped = getReadPositionAtReferencePositionIgnoreSoftClips(pos, negEnd);
+
         int negClipFrom = getReadPositionAtReferencePositionIgnoreSoftClips(neg, posStart - 1);
         negClipFrom = negClipFrom > 0 ? (neg.getReadLength() + 1) - negClipFrom : 0;
 
-        if(posClipFrom > 0) {
-            clip3PrimeEndOfRead(pos, posClipFrom, hardClipReads);
+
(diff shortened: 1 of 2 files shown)


src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -796,41 +796,57 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
-        if(negClipFrom > 0) {
-            clip3PrimeEndOfRead(neg, negClipFrom, hardClipReads);
+        if(negFirstBaseToClip > 0) {
+            clip3PrimeEndOfRead(neg, negFirstBaseToClip, hardClipReads);
         }
     }
 
+    /**
+     * Gets the read position that corresponds to a particular position on the reference.  If the position on the reference
+     * falls in a deletion in the alignment of the read, the position before the deletion will be returned.
```
yfarjoun, line 845: 1-based position,
0 if reference position doesn't overlap with read even with softclips converted to matches.

yfarjoun (changes requested):
it seems to concentrate all the shenanigans in the same place...so that's an improvement. still lots of +1 and -1 required. perhaps you can add some comment with read illustrations to clarify the reasoning?

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -796,41 +796,57 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
-
-                    if (useHardClipping) {
-                        // if we want to hardclip, we additionally hardclip the 3' end of each read so that its 3' unclipped end does not extend past the 5' unclipped start of its mate
-                        clip3primeEndsTo5primeEnds(pos, neg, true, true);
-                    }
-                }
+        // Only clip if both reads are mapped, on opposite strands, overlapping
+        if (!read1.getReadUnmappedFlag() && !read2.getReadUnmappedFlag() &&
+                read1.getReadNegativeStrandFlag() != read2.getReadNegativeStrandFlag() &&
+                read1.overlaps(read2)) {
```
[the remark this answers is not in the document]
kachulis: The explicit "innie" requirement is dropped, but this still will not clip outies, because I believe by definition outies cannot be overlapping with reads on opposite strands.  I don't think innie and outie really have meaning for overlapping reads, though I may be confused about this.  There is a test commented as `Non overlapping "outies"` which tests outies.

Previously, the "innie" requirement was enforced with the logic `pos.getAlignmentStart() < neg.getAlignmentEnd()`. Combining this with `posClipFrom > 0`/ `negClipFrom > 0` is _almost_ equivalent to the overlapping requirement, except for when the two reads overlap by only a single base.  In this case, the new check of `read1.overlaps(read2)` will pass, while the old check of `pos.getAlignmentStart() < neg.getAlignmentEnd()` will not, because instead `pos.getAlignmentStart() == neg.getAlignmentEnd()`.  This is not a behavior change I was intending, as I wasn't aware of this edge case difference until just now.  But I believe the old behavior was actually a bug, since the new behavior is more consistent with my understanding of what is intended.  It doesn't make sense to me that we would clip when two or more bases overlap, but not when only one base overlaps, so I'm guessing that was not intentional.  I have added a test demonstrating this edge case (commented as "overlap by one base only"), and have also added the same test on the branch [ck_mba_single_overlap_test_example](https://github.com/broadinstitute/picard/tree/ck_mba_single_overlap_test_example) to demonstrate the previous behavior.

For your softclipped example, these reads will not go through any additional clipping since they are not overlapping. 
 This is consistent with past behavior, since they would not pass the `pos.getAlignmentStart() < neg.getAlignmentEnd()` check.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -796,41 +796,57 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
     }
 
     private static void clip3primeEndsTo5primeEnds(final SAMRecord pos, final SAMRecord neg, final boolean hardClipReads, final boolean useUnclippedEnds) {
         final int negEnd = useUnclippedEnds? neg.getUnclippedEnd() : neg.getEnd();
         final int posStart = useUnclippedEnds? pos.getUnclippedStart() : pos.getStart();
-        final int posClipFrom = getReadPositionAtReferencePositionIgnoreSoftClips(pos, negEnd + 1);
-        int negClipFrom = getReadPositionAtReferencePositionIgnoreSoftClips(neg, posStart - 1);
-        negClipFrom = negClipFrom > 0 ? (neg.getReadLength() + 1) - negClipFrom : 0;
+        /*
+          For the positive strand, we ask for the position of the 3' most base which will not be clipped, and then increment to find the 5' most base to clip.
```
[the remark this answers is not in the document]
kachulis: I think this is correct.  On a forward read, the first base clipped is the 5' most base to clip, and the last base not clipped is the 3' most base not to be clipped.  

```
3'<SSSSSSSSMMMMMMMMMMMM5'
         5'MMMMMMMMMMMMSSSSSSSS>3'
                      ||
                      ||---> 5' most base to clip
                      |
                      |---> 3' most base not to clip
```

I agree this is pretty confusing to parse, so I will add this ascii art to the comment in the code to help clarify the language

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -796,41 +796,57 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
+        final int posLastUnclipped = getReadPositionAtReferencePositionIgnoreSoftClips(pos, negEnd);
 
-        if(posClipFrom > 0) {
-            clip3PrimeEndOfRead(pos, posClipFrom, hardClipReads);
+        /*
+        For the negative strand, we ask for the position of the 5' most base to clip.  getReadPositionAtReferencePositionIgnoreSoftClips will return the 5' most base before
+         the queried base when the queried base is in a deletion on a negative strand read
+         */
+
+        int negFirstBaseToClip = getReadPositionAtReferencePositionIgnoreSoftClips(neg, posStart - 1);
```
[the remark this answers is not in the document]
kachulis: Agreed these variables could be names more clearly (though I think this should be `negFirstBaseFrom5PrimeEndToClip`).  I've tried to improve the naming, and added some comments to clarify what all these variables are.

src/main/java/picard/sam/AbstractAlignmentMerger.java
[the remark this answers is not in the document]
yfarjoun: 👍 yeah, thanks.

src/main/java/picard/sam/AbstractAlignmentMerger.java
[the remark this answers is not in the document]
yfarjoun: I think that part of the problem is that `pos` is overloaded (both POSition and POSitive).... 🤷

kachulis:
Thanks for the review @yfarjoun.  I believe I've responded to your comments, so back to you.  I have also realized there is actually a behavior change here I wasn't intending.  I think it qualifies as a bug fix, so I think this is ok.  But if the behavior change is included, this PR is no longer just a refactor as was initially intended.

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -796,41 +796,69 @@ private void transferAlignmentInfoToPairedRead(final SAMRecord firstUnaligned, f
...
+         the queried base when the queried base is in a deletion on a negative strand read.
+         */
+
+        //this is the position counting from the aligned start of the read
+        final int neg5PrimeMostBaseToClipPositionFromStart = getReadPositionAtReferencePositionIgnoreSoftClips(neg, posStart - 1);
+
+        //this is the position counting from the 5' end of the read
+        final int negFirstBaseFrom5PrimeEndToClip = neg5PrimeMostBaseToClipPositionFromStart > 0 ? (neg.getReadLength() + 1) - neg5PrimeMostBaseToClipPositionFromStart : 0;
+
+        if(negFirstBaseFrom5PrimeEndToClip > 0) {
```
yfarjoun, line 849:
```suggestion
        if (negFirstBaseFrom5PrimeEndToClip > 0) {
```

src/main/java/picard/sam/AbstractAlignmentMerger.java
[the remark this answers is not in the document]
yfarjoun: 👍 thanks for the explanation

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -820,29 +820,41 @@ private static void clip3primeEndsTo5primeEnds(final SAMRecord pos, final SAMRec
...
 
         /*
         For the negative strand, we ask for the position of the 5' most base to clip.  getReadPositionAtReferencePositionIgnoreSoftClips will return the 5' most base before
-         the queried base when the queried base is in a deletion on a negative strand read
+         the queried base when the queried base is in a deletion on a negative strand read.
          */
 
-        int negFirstBaseToClip = getReadPositionAtReferencePositionIgnoreSoftClips(neg, posStart - 1);
-        negFirstBaseToClip = negFirstBaseToClip > 0 ? (neg.getReadLength() + 1) - negFirstBaseToClip : 0;
+        //this is the position counting from the aligned start of the read
```
yfarjoun, line 843:
```suggestion
        // this is the position counting from the aligned start of the read
```

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -820,29 +820,41 @@ private static void clip3primeEndsTo5primeEnds(final SAMRecord pos, final SAMRec
...
+        final int neg5PrimeMostBaseToClipPositionFromStart = getReadPositionAtReferencePositionIgnoreSoftClips(neg, posStart - 1);
 
-        if(posLastUnclipped > 0 && posLastUnclipped < pos.getReadLength()) {
-            clip3PrimeEndOfRead(pos, posLastUnclipped + 1, hardClipReads);
-        }
-        if(negFirstBaseToClip > 0) {
-            clip3PrimeEndOfRead(neg, negFirstBaseToClip, hardClipReads);
+        //this is the position counting from the 5' end of the read
```
yfarjoun, line 846:
```suggestion
        // this is the position counting from the 5' end of the read
```

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -820,29 +820,41 @@ private static void clip3primeEndsTo5primeEnds(final SAMRecord pos, final SAMRec
...
+                     5'MMMMMMMMMMMMSSSSSSSS>3'
+                                  ||
+                                  ||---> 5' most base to clip
+                                  |
+                                  |---> 3' most base not to clip
          */
 
-        final int posLastUnclipped = getReadPositionAtReferencePositionIgnoreSoftClips(pos, negEnd);
+        final int pos3PrimeMostUnclipped = getReadPositionAtReferencePositionIgnoreSoftClips(pos, negEnd);
+        if(pos3PrimeMostUnclipped > 0 && pos3PrimeMostUnclipped < pos.getReadLength()) {
```
yfarjoun, line 833:
```suggestion
        if (pos3PrimeMostUnclipped > 0 && pos3PrimeMostUnclipped < pos.getReadLength()) {
```

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -820,29 +820,41 @@ private static void clip3primeEndsTo5primeEnds(final SAMRecord pos, final SAMRec
           For the positive strand, we ask for the position of the 3' most base which will not be clipped, and then increment to find the 5' most base to clip.
           We do this because getReadPositionAtReferencePositionIgnoreSoftClips will return the base before a deletion, so will return the 3' most base before
           the queried base when the queried base is in a deletion on a positive strand read
+
+            3'<SSSSSSSSMMMMMMMMMMMM5'
```
yfarjoun, line 824: the 5 looks like an S...
```suggestion
            3' <SSSSSSSSMMMMMMMMMMMM 5'
```

src/main/java/picard/sam/AbstractAlignmentMerger.java
```diff
@@ -820,29 +820,41 @@ private static void clip3primeEndsTo5primeEnds(final SAMRecord pos, final SAMRec
...
+                     5'MMMMMMMMMMMMSSSSSSSS>3'
```
yfarjoun, line 825:
```suggestion
                     5' MMMMMMMMMMMMSSSSSSSS> 3'
```

yfarjoun (approved):
a few space requests, and you're all set. 

Thanks for this I don't think that the tiny difference in the definition of outie will upset anyone...so this should be fine!

PR #1596, opened 2020-10-22
Yf get reference properly threaded through

yfarjoun:
not sure how this happened...but it seems that the code was looking at the wrong version of the reference...

### DIFF
--- a/src/main/java/picard/fingerprint/ExtractFingerprint.java
+++ b/src/main/java/picard/fingerprint/ExtractFingerprint.java
@@ -97,7 +97,7 @@ public class ExtractFingerprint extends CommandLineProgram {
         }
 
         checker.setLocusMaxReads(LOCUS_MAX_READS);
-        checker.setReferenceFasta(REFERENCE_SEQUENCE);
+        checker.setReferenceFasta(referenceSequence.getReferenceFile());
         checker.setValidationStringency(VALIDATION_STRINGENCY);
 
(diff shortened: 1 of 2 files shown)


src/main/java/picard/fingerprint/ExtractFingerprint.java
```diff
@@ -97,7 +97,7 @@ protected int doWork() {
         }
 
         checker.setLocusMaxReads(LOCUS_MAX_READS);
-        checker.setReferenceFasta(REFERENCE_SEQUENCE);
+        checker.setReferenceFasta(referenceSequence.getReferenceFile());
```
gbggrant, line 100: Isn't parseArgs doing this?https://github.com/broadinstitute/picard/blob/4c937e5bb8b790a5b7096c1e61564633e83298f0/src/main/java/picard/cmdline/CommandLineProgram.java#L359
(or are you removing deprecated usage?)

gbggrant (approved):
LGTM.  Not clear why this change is needed.

src/main/java/picard/fingerprint/ExtractFingerprint.java
yfarjoun: I got a report that without this fix the tool was unable to open cram files. I didn't verify this claim though...so I should probably look at the tests....

yfarjoun:
yeah....this works in master...sorry for the trouble.

PR #1606, opened 2020-11-14
Added ref to software catalogs to citation section

smoe:


README.md
```diff
@@ -149,6 +149,7 @@ Please cite this repository when using Picard tools for your publications.
 }
 ```
 
+Identifiers from software registries are increasingly accepted by journals, as in (bio.tools:Picard_tools) or (RRID:SCR_006525).
```
yfarjoun, line 152: this is cool. there was also an idea to setup a zendo link...but no-one did that in the end.... @gbggrant @tfenne what do you guys think?

gbggrant:
@ekiernan do you have any thoughts about this?

smoe:
But bio.tools is now good at strictly enforcing lower-case only, so you now find this as https://bio.tools/picard_tools and from a look at that page, the dot is omitted for references: biotools:picard_tools.

README.md
ekiernan: > @ekiernan do you have any thoughts about this?



> @ekiernan do you have any thoughts about this?

@gbggrant We recently had this conversation for WARP and the conclusion was that DOIs from Zenodo might be hard to maintain over time; we opted for trying to create an Application Note that we could submit to a journal and have people cite. 

I think since the bio.tools and RRID is already set up, that's probably a good solution for now. Looping in @kbergin

kbergin: Agreed sounds good to point to what already exists and we can evaluate if we want to put onto zenodo later. I don't know how to keep bio.tools up to date, it seems to reference version 2.17.x right now, so we may want to look into that too.

jessicaway:
@kbergin do we want to move forward with this PR or close it out?

ekiernan:
> @kbergin do we want to move forward with this PR or close it out?

@jessicaway @kbergin The RRIS and Bio.tools ID are now correct and this can be merged.

ekiernan (approved):

README.md
matuskalas: Sorry I missed this one 😥 What @smoe suggest is indeed the ultimate idea, but to my understanding _bio.tools_ isn't there yet. Please update @hansioan @hmenager

Issue #1499, opened 2020-04-17
Add Support To Merge Only Overlapping Intervals


rklein2:
### Tool(s) involved
IntervalListTools UNIQUE=true

### Description
Many GATK tools internally merge only overlapping intervals without merging adjacent intervals. From what I can tell, no tool exists to reproduce these internally used intervals. The obvious first place to look is picard where we can use the option UNIQUE=true. This option, however, merges both adjacent and overlapping intervals. The feature I am requesting is meant to solve this loophole by allowing a separate option to only merge overlapping intervals.

PR #1608, opened 2020-11-16
Add functionality to not combine abutting intervals

tmelman:
### Description

The behavior up until now in `IntervalListTools` for dealing with combining interval lists has been to combine abutting and overlapping interval lists. However, GATK internally often keeps abutting intervals separate, as detailed in this issue: https://github.com/broadinstitute/picard/issues/1499. 

Here I've added a functionality to create the interval list as requested, keeping abutting interval lists separate. 

Changes made: 
* Added an option for `DONT_COMBINE_ABUTTING` to `IntervalListTools`. Default is `false` to keep current behavior the same. 
* When `DONT_COMBINE_ABUTTING=true` and `UNIQUE=true` and there is no action, this will create interval lists merging ONLY overlapping intervals.
* Throw an error when `DONT_COMBINE_ABUTTING=true` but `ACTION=UNION`, as this is unsupported.
* Added tests and test data. 

----

#### Content

#### Review


src/main/java/picard/util/IntervalListTools.java
```diff
@@ -402,15 +405,18 @@ protected int doWork() {
...
+            SORT = false; // no need to sort, since uniqued() output will be sorted by definition.
             UNIQUE = true;
         }
 
         final IntervalList possiblySortedResult = SORT ? result.sorted() : result;
         final IntervalList possiblyInvertedResult = INVERT ? IntervalList.invert(possiblySortedResult) : possiblySortedResult;
 
         //only get unique if this has been asked unless inverting (since the invert will return a unique list)
-        List<Interval> finalIntervals = UNIQUE ? possiblyInvertedResult.uniqued().getIntervals() : possiblyInvertedResult.getIntervals();
+//        List<Interval> finalIntervals = UNIQUE ? possiblyInvertedResult.uniqued().getIntervals() : possiblyInvertedResult.getIntervals();
```
kachulis, line 416: remove commented out old code

src/test/java/picard/util/IntervalListToolsTest.java
```diff
@@ -406,4 +409,63 @@ public void testLargeScatters() throws IOException {
 
         Assert.assertEquals(gather, original);
     }
+
+    @Test(timeOut = 40_000)
```
kachulis, line 413: don't need the `timeOut` since you're testing on very small files.

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -513,6 +519,9 @@ private IntervalList openIntervalLists(final List<File> files, BinaryOperator<In
         if (COUNT_OUTPUT != null && OUTPUT_VALUE == Output.NONE) {
             errorMsgs.add("COUNT_OUTPUT was provided but OUTPUT_VALUE is set to NONE.");
         }
+        if (ACTION == Action.UNION && DONT_COMBINE_ABUTTING) {
+            errorMsgs.add("ACTION=UNION action combines abutting intervals by default, so it is incompatible with DONT_COMBINE_ABUTTING=true.");
```
yfarjoun, line 523: I'm not sure I understand why this is inherently incompatible with UNION. I get that _currently_ the htsjdk `uniqued` method combines abbutting intervals, but I don't see why it has to be that way....it seems plausible, to work around it by using the lower-level `getUniqueIntervals` directly.

src/test/java/picard/util/IntervalListToolsTest.java
```diff
@@ -406,4 +409,63 @@ public void testLargeScatters() throws IOException {
...
+    public void testCombineAbuttingIntervals() throws IOException {
+        // Test the default behavior of UNION, which is to combine abutting and overlapping intervals.
+        //gather
+        final File ilOut = File.createTempFile("IntervalListTools", ".interval_list");
+        ilOut.deleteOnExit();
+        {
```
kachulis, line 419: unneeded brackets (maybe this used to be inside a try block?)

yfarjoun (changes requested):
Thanks for writing up all these nice tests!

I think you should be able to enable this to work with UNION as well...I have faith in you!

src/test/java/picard/util/IntervalListToolsTest.java
```diff
@@ -406,4 +409,63 @@ public void testLargeScatters() throws IOException {
...
+            args.add("ACTION=UNION"); // Note: ACTION=UNION is equivalent to UNIQUE=true and DONT_COMBINE_ABUTTING=false, or defaults
+            Assert.assertEquals(runPicardCommandLine(args), 0);
+        }
+        final IntervalList gather = IntervalList.fromFile(ilOut);
+        final IntervalList original = IntervalList.fromFile(abutting_combined);
+
+        Assert.assertEquals(gather, original); // equal to expected output
+    }
+    @Test(timeOut = 40_000)
+    public void testDontCombineAbuttingIntervals() throws IOException {
```
kachulis, line 432: Would like to see some tests of the other actions as well.  And I think you can combine this with `testCombineAbuttingIntervals` and use a DataProvider

kachulis (commented):
nice!  a few comments, and I think the tests could be condensed and expanded with a DataProvider.

src/main/java/picard/util/IntervalListTools.java
tmelman: There is an additional wrinkle, that `getUniqueIntervals` returns `List<Intervals>` whereas `uniqued()` returns `IntervalList`, and there is no constructor that takes in a `List<Intervals>` and turns it into an `IntervalList`. I can see if there's a way to do this with `addAll()` and removing the old intervals. If you have any ideas let me know.

src/test/java/picard/util/IntervalListToolsTest.java
tmelman: I was wondering about this, I copied and pasted from another test. Happy to remove! :)

src/main/java/picard/util/IntervalListTools.java
yfarjoun: you can replicate the htsjdk code to make a private method that also allows you to state "mergeAbutting", right? eventually perhaps someone will push that down to htsdjk.

tmelman: Fixed @yfarjoun

src/test/java/picard/util/IntervalListToolsTest.java
tmelman: Fixed @kachulis

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -501,6 +514,22 @@ private IntervalList openIntervalLists(final List<File> files, BinaryOperator<In
                 .orElse(null);
     }
 
+    private static IntervalList uniqued(IntervalList non_unique, boolean merge_abutting) {
```
kachulis, line 517: why does this need to return an IntervalList? can't you let it return a list of intervals instead?  You immediately call `getIntervals()` on it the only place I see it used, so it seems like a lot of unneeded converting back and forth between lists of intervals and interval lists.

tmelman: Because I'm reproducing the functionality of "uniqued" in htsjdk's IntervalList. The idea is to have a "uniqued" method that we can one day add to their code base in a PR. Alternatively they have a "getUniqueIntervals()" method that rerturns a List of Intervals. I could use that but this makes for cleaner code with all the defaults. 

tl;dr: this was purely a design decision, I don't have a strong preference on which way to do it, and I'm happy to change it to according to your preferences, as you will also be interacting with it.

kachulis: I see, that seems reasonable then

kachulis (approved):
looks good!

testdata/picard/util/abutting.interval_list
```diff
@@ -0,0 +1,6 @@
+@HD	VN:1.6	SO:coordinate
+@SQ	SN:1	LN:249250621	AS:GRCh37	UR:http://www.broadinstitute.org/ftp/pub/seq/references/Homo_sapiens_assembly19.fasta	M5:1b22b98cdeb4a9304cb5d48026a85128	SP:Homo Sapiens
+1	30000	30001	+	target_1
+1	30002	30003	+	target_2
```
yfarjoun, line 4: perhaps use the name of the interval to comment about the properties of each interval:

e.g 

```
1	30000	30001	+	target_1_abutting_the_next
1	30002	30003	+	target_2_abutting_the_previous_but_not_the_next
1	30005	30007	+	target_3_overlapping_the_next_but_not_abutting_the_previous
1	30006	30008	+	target_4_overlapping_the_previous
```

testdata/picard/util/abutting_combined.interval_list
```diff
@@ -0,0 +1,4 @@
+@HD	VN:1.6	SO:coordinate
+@SQ	SN:1	LN:249250621	AS:GRCh37	UR:http://www.broadinstitute.org/ftp/pub/seq/references/Homo_sapiens_assembly19.fasta	M5:1b22b98cdeb4a9304cb5d48026a85128	SP:Homo Sapiens
+1	30000	30003	+	target_1
```
yfarjoun, line 3: if you name the intervals in a meaningful way, you can also have the tests "CONTATENATE=true" and the resulting names here will show which intervals got merged.

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -495,6 +514,22 @@ private IntervalList openIntervalLists(final List<File> files, BinaryOperator<In
...
```
yfarjoun, line 517: I'd add a "//TODO:'' item to move this to htsjdk and use that once we can. (possibly adding issues that can be referenced)

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -402,15 +416,20 @@ protected int doWork() {
...
             UNIQUE = true;
         }
 
         final IntervalList possiblySortedResult = SORT ? result.sorted() : result;
         final IntervalList possiblyInvertedResult = INVERT ? IntervalList.invert(possiblySortedResult) : possiblySortedResult;
 
-        //only get unique if this has been asked unless inverting (since the invert will return a unique list)
-        List<Interval> finalIntervals = UNIQUE ? possiblyInvertedResult.uniqued().getIntervals() : possiblyInvertedResult.getIntervals();
+        //only get unique if this has been asked OR if action is UNION, unless inverting (since the invert will return a unique list)
+        List<Interval> finalIntervals = UNIQUE? IntervalListTools.uniqued(possiblyInvertedResult, !DONT_MERGE_ABUTTING).getIntervals() : possiblyInvertedResult.getIntervals();
```
yfarjoun, line 432:
```suggestion
        List<Interval> finalIntervals = UNIQUE ? uniqued(possiblyInvertedResult, !DONT_MERGE_ABUTTING).getIntervals() : possiblyInvertedResult.getIntervals();
```

src/main/java/picard/util/IntervalListTools.java
yfarjoun, line 517: We tend to use camelCase for variables and methods
```suggestion
    private static IntervalList uniqued(IntervalList nonUnique, boolean mergeAbutting) {
```

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -495,6 +514,22 @@ private IntervalList openIntervalLists(final List<File> files, BinaryOperator<In
...
 
+    private static IntervalList uniqued(IntervalList non_unique, boolean merge_abutting) {
+        // A subroutine to replace htsjdk's IntervalList "unique()" method, which combines abutting intervals by default.
+        // Returns an IntervalList.
+        // Inputs:
+        //  - non_unique: an IntervalList to be operated on
+        //  - merge_abutting: a boolean to combine abutting intervals (if true) or not (if false).
+        // Outputs:
+        //  - an IntervalList that is uniqued with overlapping intervals merged, and abutting intervals handled per the boolean flag.
+        final boolean CONCATENATE_NAMES = true;
```
yfarjoun, line 525: why are these in variables? you can just put `true,false` in the method call, no?

src/main/java/picard/util/IntervalListTools.java
```diff
@@ -402,15 +416,20 @@ protected int doWork() {
         final IntervalList result = ACTION.act(lists, secondLists);
 
         if (INVERT) {
-            SORT = false; // no need to sort, since return will be sorted by definition.
+            SORT = false; // no need to sort, since uniqued() output will be sorted by definition.
+            UNIQUE = true;
+        }
+
+        if (ACTION == Action.UNION) { // UNION is basically Action.CONCAT with SORT and UNIQUE
+            SORT = true;
```
yfarjoun, line 424: if UNIQUE sorts by default, shouldn't this be `SORT=false`?

yfarjoun (changes requested):
Thanks for the srcond round!

a few more small nit-picks!

src/main/java/picard/util/IntervalListTools.java
tmelman: I definitely can, it just makes for more readable code. Purely design decision, happy to change it

src/main/java/picard/util/IntervalListTools.java
tmelman: per the previous implementation, UNION assumes sort and unique are true. so if action is union we have to reproduce that behavior.

src/main/java/picard/util/IntervalListTools.java
yfarjoun: OK. So in that case, just change the variable names to camelCase

src/main/java/picard/util/IntervalListTools.java
yfarjoun: I guess I'm confused by why the INVERT==true branch claims (in the comments) that theres no need to sort, and here there seems to be a need to sort again.

tmelman: theres a bit of a loop of flags and actions overlapping lol… but I don't know what "UNION" would be with "UNIQUE" false.

src/main/java/picard/util/IntervalListTools.java
tmelman: done 👍 :)

tmelman:
@yfarjoun round 2 changes resolved, can you take a look?

yfarjoun (approved):

yfarjoun:
merge away! Thanks for this!!

🎉

yfarjoun:
you can delete your branch now @tmelman

Issue #1290, opened 2019-03-01
Fast fail when Rscript isn't installed


bartgrantham:
### Tool(s) involved
Subcommands: CollectInsertSizeMetrics, QualityScoreDistribution, MeanQualityByCycle, CollectBaseDistributionByCycle, CollectGcBiasMetrics, CollectRnaSeqMetrics, CollectRrbsMetrics, CollectMultipleMetrics, CollectWgsMetricsWithNonZeroCoverage (and probably more)

### Description
These tools should fail immediately if R/Rscript is not installed with the required dependencies.  It's very frustrating to have a tool like CollectMultipleMetrics crank away at a file for hours only to have it fail at the end because R isn't installed and therefore it can't generate plots.

(I shudder to think of all the time and electricity has been burned by people using picard for these purposes only to have it crash at the end when it could have warned them at the start)

tfenne:
I would suggest a different solution.  I think Picard tools should detect whether R is present or not (and has the requires packages installed or not) and then: a) log a warning if those dependencies aren't met and b) only attempt plot generation if they are not.

bartgrantham:
I disagree, that's even worse.  At least as it is now if someone was expecting a graphical result they know that it didn't work _at runtime_.  If the user is expecting a plot to be produced, this kind of silent failure wouldn't be known until potentially much later, after they've run a huge batch.  If picard had the behavior you suggest I would have burned at least a few hundred dollars in cloud computing just to find out that it didn't actually do what I expected.

Please take into account that many people use these tools in larger workflows in batch processing and simply cannot watch every single line of output.  Even for users who are running these tools manually and serially picard is so chatty it would be very easy to miss that warning among all the other output (gatk has the same problem).

Perhaps the creation of plots should be a command line switch (or maybe _not_ generating plots is the switch?), and the error happens if plots are expected and R isn't available.

DarioS:
This happened to me today. It is disappointing that the prerequisite isn't checked at the beginning of the analysis. The documentation of `CollectInsertSizeMetrics` also doesn't mention it at all.

```
INFO    2019-03-15 16:08:54     SinglePassSamProgram    Processed   863,000,000 records.  Elapsed time: 00:33:09s.  Time for last 1,000,000:    2s.  Last read position: chrUn_GL000216v2:160,617
Caused by: java.io.IOException: Cannot run program "Rscript": error=2, No such file or directory
```

[...]

PR #1613, opened 2020-11-25
Fast fail if R is not installed in tools that use R

sophiacrennan:
### Description

Many metrics collectors in Picard rely on R to generate a plot at the end of the tool. Unfortunately, if a user does not have R installed, a tool could fail hours after running as mentioned in this issue: #1290. This PR throws an error at the beginning of a run, if R is not installed and the tool uses an R script.

Changes Made:

Create a method in `CommandLineProgram` that checks if R is installed when a chart is expected as an output

Call the method in `customCommandLineValidation` for every tool that uses an R script

----

#### Content

#### Review


src/main/java/picard/analysis/CollectMultipleMetrics.java
```diff
@@ -669,6 +669,8 @@ public String getHelpDoc() {
         }
         programsToRun = new LinkedHashSet<>(PROGRAM);
 
+        checkRInstallation(true);
```
khajoue2, line 672: Why is this true?

khajoue2 (approved):

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -497,4 +498,18 @@ public static boolean hasWebDocumentation(final Class<?> clazz){
     public static String getFaqLink() {
         return "To get help, see http://broadinstitute.github.io/picard/index.html#GettingHelp";
     }
+
+    /**
+     * Check if R is installed
+     * @param has_chart_output
+     */
+    public static void checkRInstallation(final boolean has_chart_output) {
+        if(has_chart_output) {
```
gbggrant, line 507: space between `if` and `(`

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -497,4 +498,18 @@ public static boolean hasWebDocumentation(final Class<?> clazz){
...
+            try {
+                RExecutor.executeFromClasspath("picard/analysis/checkRInstallation.R");
+            } catch (htsjdk.samtools.SAMException e) {
+                throw new PicardException("R is not installed on this machine. It is required for creating the chart.");
```
gbggrant, line 511: Rather than throwing an exception here, could you return the error string (or just a true / false) from this class if R is not installed?
And then in the calling program's CustomCommandLineValidator code you could just check if checkRInstallation failed and return a string array like:

`         if (!checkRInstallation(CHART_OUTPUT != null)) {
             return new String[]{"R is not installed on this machine. It is required for creating the chart."};
         }`

This is more in line with how CustomCommandLine validation should work.  
I think the usage in CollectInsertSizeMetrics.java is more correct than that in CollectRnaSeqMetrics.java

gbggrant (changes requested):
Looks good (and works fine) - I suggest you don't throw the exception, rather pass up an error message for the CLP to display along with the usage.

src/main/java/picard/analysis/CollectMultipleMetrics.java
sophiacrennan: Almost all of the metrics within CollectMultipleMetrics are given a CHART_OUTPUT/Histogram_FILE which means the rscript will be run. CollectMultipleMetrics itself doesn't have those as an argument. Since those metrics will run the rscript anyways, I just put true.

gbggrant (approved):

Issue #805, opened 2017-04-27
GTF to Refflat tool for use with CollectRnaSeqMetrics


nh13:
Since most folks use GTFs for RNASeq, but `CollectRnaSeqMetrics` uses a refflat file, it would be nice to have a tool that can convert GTF to refflat, and then we can update `CollectRnaSeqMetrics` to take in either a refflat or GTF file.

This an example method (requires Jim Kent's tools):
```
 gtfToGenePred \
    -genePredExt \
    -geneNameAsName2 \
    -ignoreGroupsWithoutExons \
    !{params.gtf} \
    /dev/stdout | \
    awk 'BEGIN { OFS="\t"} {print $12, $1, $2, $3, $4, $5, $6, $7, $8, $9, $10}' 
```

Also, it wouldn't be hard to pull out the records with `gene_type` equal to `rRNA` to create the requisite ribosomal intervals.

sooheelee:
Commented prematurely. When you say tool, you mean something under the hood that CollectRnaSeqMetrics can call upon, not a outward facing tool.

nh13:
@sooheelee yes, making `CollectRnaSeqMetrics` accept a refflat file would be great; I think I mentioned that above: "then we can update CollectRnaSeqMetrics to take in either a refflat or GTF file."  Nonetheless, if we are going to support a GTF as input to `CollectRnaSeqMetrics`, then 95% of the work would be done, and we could _also_ have a really useful tool.

nh13:
@yfarjoun thoughts?

yfarjoun:
are you volunteering? I have no objections to having the tool accept GTF as input.

nh13:
@yfarjoun maybe, but not at the moment, I think it is very mechanical and would be a great exercise for a student project if you have any students?

yfarjoun:
interesting idea. I'll run it by folks that might want to cut their teeth on some Picard stuff.

royfrancis:
Just checking up on this 7 years later. So, can `CollectRnaSeqMetrics` take a GTF file?

serge2016:
Same question here. Any progress?

PR #1620, opened 2020-12-22
Allow CollectRnaSeqMetrics to take in a GTF file

sophiacrennan:
### Description

Most people use GTF files for RNA gene location information. CollectRNASeqMetrics tool accepts a refflat file which is not as common of a format anymore. It was requested to have the tool accept either a refflat or a GTF #805. 
CollectRNASeqMetrics now accepts either a refflat or a GTF file. If a GTF is accepted, then it is converted to a refflat using the class GtfToRefflat that I created. I have added unit tests with examples of GTF files and their respective refflat file.

----

#### Content

#### Review


src/main/java/picard/analysis/GtfToRefflat.java
```diff
@@ -0,0 +1,266 @@
...
+                    gene_id = feature.getID();
+                    chromosome = feature.getContig();
+                    strand = feature.getStrand();
+                    type = feature.getType().toLowerCase();
+
+                    if (type.equals("exon")) {
+                        exonStarts.add(feature.getStart() - 1);
+                        exonEnds.add(feature.getEnd());
+                        hasExon = true;
+                    } else if (hasExon == false) {
```
kishorikonwar, line 95: Can CDS be explicit?

gbggrant: @kishorikonwar I have inherited this PR from Sophie and don't know the domain that well.  Can you answer your own question here?  If you're not sure, I'm inclined to leave the code as is.  Thanks.

kishorikonwar: @gbggrant This is fine now and we have resolved it already, therefore, you can safely ignore it.

gbggrant:
Assigning to @kachulis for questions about implementation detail

kachulis (changes requested):
Haven't reviewed in detail yet, but after a quick initial look, it seems there is a lot of intermediate file creation being done to support reading in a new format.  I would propose instead a few different options:
1) Instead of making the tool work with GTF, just make it work with GFF3 instead.
2) If it is important to handle GTF, create a separate GTF codec, which can return either a new `GTFFeature` type or `Gff3Feature`. 
Whichever option is chosen, the main thing you need to extract from the input file is an `OverlapDetector<Gene>` object, which is then passed to RnaSeqMetricsCollector contstructor.  So all you really need is a `GtfReader` class to correspond to `RefFlatReader`.  That reader basically just needs to have a method to convert from a Gff3Feature to Gene, I think.

gbggrant:
Putting this PR 'on hold'

Discussed with @kachulis  - he has concerns that the implementation itself may not be correct (so he is not okay with the PR being merged in its current state)

He would prefer that it be implemented using a specific codec for the GTC 'use a separate GTF codec, which can return either a new GTFFeature type or Gff3Feature'

He (Chris Kachulis) has a Picard branch in which he created a GFF3 codec, but it was not merged as he didn’t think it was feature complete: “the branch I had for this is ck_Gff3_RnaSeqMetrics .  It looks like I actually made an attempt at the ontology on that branch, which is probably a bit too far for now.  But you could rip that out and hard code in the feature types you care about.  And it looks like that branch still uses a local tag of htsjdk, but if you just rebase and bump the tag, it might just work now, or if not I don’t think it would take too much change to get it working.”

PR #1633, opened 2021-01-25
Parse version directly from tile metrics instead of using a clp flag.

jacarey:
### Description

The currently code for switching between version 2 and version 3 of the Illumina tile metrics files relies on a flag set in `CollectIlluminaLaneMetrics` called `IS_NOVASEQ` 

This flag is both unnecessary and a bit confusing since other machine types can output version 3 of the tile metrics output files.

This PR parses the first byte of the header of the file to determine which version it is and then uses that as the switch to determine which set of parsing functions to use.

----

#### Content

#### Review


src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
```diff
@@ -176,29 +178,34 @@ public static void collectLaneMetrics(final File runDirectory, final File output
...
+            int tileMetricsVersion = determineTileMetricsVersion(runDirectory, readStructure);
+            final Map<Integer, ? extends Collection<Tile>> laneTiles = readLaneTiles(runDirectory, readStructure, validationStringency, tileMetricsVersion);
             writeLaneMetrics(laneTiles, outputDirectory, outputPrefix, laneMetricsFile, fileExtension);
-            writePhasingMetrics(laneTiles, outputDirectory, outputPrefix, phasingMetricsFile, fileExtension, isNovaSeq);
+            writePhasingMetrics(laneTiles, outputDirectory, outputPrefix, phasingMetricsFile, fileExtension, tileMetricsVersion);
         }
 
-        public static File writePhasingMetrics(final Map<Integer, ? extends Collection<Tile>> laneTiles, final File outputDirectory,
+        private static int determineTileMetricsVersion(File illuminaRunDirectory, ReadStructure readStructure) {
+            final List<File> tileMetricsOutFiles = TileMetricsUtil.findTileMetricsFiles(illuminaRunDirectory, readStructure.totalCycles);
```
gbggrant, line 189: Do you need to find ALL the tileMetricsFiles for all the cycles, since you are only retrieving element 0 below?   I'm assuming the version in element 0 will be the same as in all tileMetricsOutFiles?  (possible I am not understanding cycles vs. tiles here).

src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
```diff
@@ -176,29 +178,34 @@ public static void collectLaneMetrics(final File runDirectory, final File output
...
+            return new TileMetricsOutReader(tileMetricsOutFiles.get(0)).getVersion();
+        }
+
+        public static void writePhasingMetrics(final Map<Integer, ? extends Collection<Tile>> laneTiles, final File outputDirectory,
```
gbggrant, line 193: Can this be private?  I don't see it being used outside of this class.

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -137,7 +126,7 @@ public static File renderTileMetricsFileFromBasecallingDirectory(final File illu
     ) throws FileNotFoundException {
         final Map<Integer, Map<Integer, Collection<TilePhasingValue>>> phasingValues = getTilePhasingValues(phasingMetricsFiles, readStructure);
         for (File tileMetricsOutFile : tileMetricsOutFiles) {
-            final TileMetricsOutReader tileMetricsIterator = new TileMetricsOutReader(tileMetricsOutFile, TileMetricsOutReader.TileMetricsVersion.THREE);
+            final TileMetricsOutReader tileMetricsIterator = new TileMetricsOutReader(tileMetricsOutFile);
```
gbggrant, line 129: Do you still need the V3 in the method name here?  That is - will this now work with V2 and V3?

src/main/java/picard/illumina/parser/TileMetricsUtil.java
```diff
@@ -163,7 +152,7 @@ public static File renderTileMetricsFileFromBasecallingDirectory(final File illu
                                                     final ValidationStringency validationStringency)
             throws FileNotFoundException {
```
gbggrant, line 153: Not yours, but I don't think this method throws FileNotFoundException, nor does it use validationStringency.

src/test/java/picard/illumina/IlluminaLaneMetricsCollectorTest.java
```diff
@@ -50,14 +50,13 @@ public void testWriteLaneMetrics(final String testRun) throws Exception {
...
+    public void testCollectIlluminaLaneMetrics(final String testRun, final ReadStructure readStructure) throws Exception {
         for (final boolean useReadStructure : Arrays.asList(true, false)) {
             final File runDirectory = new File(TILE_RUN_DIRECTORY, testRun);
             final CollectIlluminaLaneMetrics clp = new CollectIlluminaLaneMetrics();
             clp.OUTPUT_DIRECTORY = IOUtil.createTempDir("illuminaLaneMetricsCollectorTest", null);
             clp.RUN_DIRECTORY = runDirectory;
             clp.OUTPUT_PREFIX = "test";
-            clp.IS_NOVASEQ = isNovaSeq;
             if (useReadStructure) clp.READ_STRUCTURE = readStructure;
             clp.doWork();
```
gbggrant, line 61: Any concern that the metrics generated would be different if the parser misidentified the file as V2 vs V3?

gbggrant (approved):

src/main/java/picard/illumina/CollectIlluminaLaneMetrics.java
jacarey: Yeah I was leveraging a method that already exists to find the TileMetricsFiles even though I really only need to check one. I could run through them all and assert that they are all ==

src/test/java/picard/illumina/IlluminaLaneMetricsCollectorTest.java
jacarey: If they are misidentified then the parsing with throw an error. They are significantly different in structure.

gbggrant:
Looks good @jacarey if you want to update to master, it's good to merge.

PR #1643, opened 2021-02-16
Add sample -> individual map capability to CrosscheckFingerprints

kachulis:
Currently, when performing an all sample crosscheck on a large cohort, multiple different samples from the same individual may be present, but Crosscheck will not expect these samples to match, because matching expectations are based solely the equality (or not) of the sample name.  

This PR adds the option of passing a sample -> individual map, which associates an individual id with each sample id.  Matching expectations are then based on the individual ids; samples associated with the same individual id will be expected to match, those associated with different individual ids will be expected not to match.

----

#### Content

#### Review


gbggrant:
@yfarjoun are you able to look at this?

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
```diff
@@ -289,6 +289,11 @@
             "Should only be used with SECOND_INPUT. ", optional = true)
     public File SECOND_INPUT_SAMPLE_MAP;
 
+    @Argument(doc = "A tsv with two columns representing the individual with which each sample is associated.  When this sample is specified, expectations for matches will be based " +
+            "on the equality or inequality of the individual associated with two samples, as opposed to the sample ids themselves.  Samples which are not included in this file " +
+            "will have their sample id used as their individual id", optional = true)
```
yfarjoun, line 294: - what happens if this points one sample onto another sample that isn't mapped? 
- is the map the same for samples in INPUT and SECOND_INPUT?

yfarjoun: Be more specific about the format of the file, since it currently isn't specified which column is which and there are no column headers mentioned.

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
```diff
@@ -787,7 +797,9 @@ private int crossCheckFingerprints(final Map<FingerprintIdDetails, Fingerprint>
 
             for (int col = 0; col < rhsFingerprintIdDetails.size(); col++) {
                 final FingerprintIdDetails rhsId = rhsFingerprintIdDetails.get(col);
-                final boolean expectedToMatch = EXPECT_ALL_GROUPS_TO_MATCH || lhsId.sample.equals(rhsId.sample);
+                final String lhsMatchId = sampleIndividualMap == null ? lhsId.sample : sampleIndividualMap.getOrDefault(lhsId.sample, lhsId.sample);
```
yfarjoun, line 800: this should be a private method `resolveSampleIfPossible` (or somesuch)

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
```diff
@@ -432,6 +438,10 @@ protected int doWork() {
         if (SECOND_INPUT_SAMPLE_MAP != null) {
             IOUtil.assertFileIsReadable(SECOND_INPUT_SAMPLE_MAP);
         }
+        if (SAMPLE_INDIVIDUAL_MAP != null) {
+            IOUtil.assertFileIsReadable(SAMPLE_INDIVIDUAL_MAP);
+            sampleIndividualMap = getStringStringMap(SAMPLE_INDIVIDUAL_MAP, "SAMPLE_INDIVIDUAL_MAP");
```
yfarjoun, line 443: I would emit a warning if any of the individuals are also unmapped samples, as that might not be intended.

src/test/java/picard/fingerprint/CrosscheckFingerprintsTest.java
```diff
@@ -432,6 +435,42 @@ public void testCrossCheckSMs(final File file1, final File file2, final int expe
...
+                "CROSSCHECK_BY=SAMPLE",
+                "SAMPLE_INDIVIDUAL_MAP=" + individualsMap.getAbsolutePath()
+        };
+        doTest(args, metrics, expectedRetVal, numberOfSamples * numberOfSamples  , CrosscheckMetric.DataType.SAMPLE);
+
+        TabbedTextFileWithHeaderParser matrixParser = new TabbedTextFileWithHeaderParser(matrix);
+        Assert.assertEquals(matrixParser.columnLabelsList().size(), numberOfSamples + 1 );
+    }
+
+    @Test()
```
yfarjoun, line 472: this a test?

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
yfarjoun: Also, perhaps emit a warning if any of the samples in the map are not samples in the input.

yfarjoun (changes requested):

src/test/java/picard/fingerprint/CrosscheckFingerprintsTest.java
kachulis: ha, it would not seem so.

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
kachulis: Not sure I understand the first question.  This maps samples to individuals, not samples to other samples.  So if only one sample maps to a particular individual, that is not a problem, it just means that we will never expect this sample to match when checked against any of the other samples.

Yes, the same map is used for both INPUT and SECOND_INPUT.   Since this map represents the relationship between samples and actual individuals, there is no need to distinguish between samples in INPUT and SECOND_INPUT when enumerating these.  If the same sample id occurs in both INPUT and SECOND_INPUT, it would be associated to the same individual.  Note that we can actually use INPUT_SAMPLE_MAP/SECOND_INPUT_SAMPLE_MAP and SAMPLE_INDIVIDUAL_MAP together, so that we can actually map sample --> sample alias --> individual if we want/need to. 

And added more specifics on the file format and details on the usage.

yfarjoun: I guess I meant what if the map points a sample at an individual that is also another sample? these are all strings...so you can have
```
NA12878 -> GM12878 (in the map)
GM12878 (a sample, not in the map) 
```
it should be made clear that in this case, NA12878 and GM12878 will be considered to be the same "individual".

yfarjoun: Also, should make sure that it is clear that "GROUP_BY=SAMPLE" means AFTER mapping.

yfarjoun: Actually....it doesn't seem that you did that. Should there also be a GROUP_BY=INDIVIDUAL that enables this?

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
kachulis: done

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
kachulis: Ahh, I see.  Added some documentation to clarify that this is the case, and logging a warning when this happens

kachulis: Right, this has no effect on the group_by operation currently, just on the match/no match expectations.  A `CROSSCHECK_BY=INDIVIDUAL` option could be added, but I think it would be a little more involved, for example I think a new `individual` field would need to be added to the `FingerprintIdDetails` class, and some handling of setting that field in different scenarios.  I also think that use case is much rarer than the one this PR is trying to solve, it seems rather unusual to me to want to actually combine all the data from different samples which are from the same individual into a single fingerprint, and then check that against other "combined" fingerprints composed of multiple different samples from other individuals.  In fact, I kind of think that option would be more often invoked incorrectly by mistake than it would be used intentionally correctly.  So I'm not sure that it is worth adding here.

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
kachulis: added warnings

kachulis:
@yfarjoun thanks for the review!  I've addressed most of your comments I believe.  Not sure how I feel about adding a `CROSSCHECK_BY=INDIVIDUAL` option also, see response to your comment in the code.

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
```diff
@@ -289,6 +289,15 @@
...
+            "column is the associated individual id.  Values in the first column must be unique.  If INPUT_SAMPLE_MAP or SECOND_INPUT_SAMPLE_MAP is also specified, " +
+            "then the values in the first column of this file should be the sample aliases specified in the second columns of INPUT_SAMPLE_MAP and SECOND_INPUT_SAMPLE_MAP, " +
+            "respectively.  When this input is specified, expectations for matches will be based on the equality or inequality of the individual ids associated with two " +
+            "samples, as opposed to the sample ids themselves.  Samples which are not listed in this file will have their sample id used as their individual id, for the " +
+            "purposes of match expectations.  This means that one sample id could be used as the individual id for another sample, but not included in the map itself, and " +
+            "these two samples would be considered to have come from the same individual.", optional = true)
+    public File SAMPLE_INDIVIDUAL_MAP;
+
     @Argument(doc = "An argument that controls how crosschecking with both INPUT and SECOND_INPUT should occur. ")
```
yfarjoun, line 301: so if you are not going to modify the GROUP_BY=SAMPLE argument, could you add a note there (or in the enum) indicating that the value used for grouping will be the original, unmapped sample id?

yfarjoun (changes requested):

src/main/java/picard/fingerprint/CrosscheckFingerprints.java
kachulis: added a note in the documentation to clarify this.

kachulis:
@yfarjoun back to you

yfarjoun:
I think that not mapping all the samples to individuals is more-or-less expected, and so a warning there will mask other warnings that might happen...that's why I thought it makes sense to make it an INFO. 

That said, I don't care too much, and so 👍

yfarjoun (approved):
I prefer the not-all-samples-were mapped warning were an INFO, but I'm not going to hold up this PR for it.

PR #1654, opened 2021-03-11
Change docker base image for security reasons.

gbggrant:
### Description

The trivvy security check failed, flagging issues with the openjdk:8 docker image we are using for a base.  This PR updates this to adoptopenjdk:8

----

#### Content

#### Review

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:8
+FROM adoptopenjdk:8
 MAINTAINER Broad Institute DSDE <dsde-engineering@broadinstitute.org>
 
 ARG build_command=shadowJar


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openjdk:8
+FROM adoptopenjdk:8
```
kachulis, line 1: do we want to use a digest instead of a tag so that the base image doesn't change out from under us again in the future?

kachulis (approved):
Looks good.  Only question is whether we should use a digest instead of a tag for the base image.

Dockerfile
gbggrant: Not sure - It's nice getting the latest in this case as they might fix bugs / improve security?  Did @dinvlad  have anything to say about this?

kachulis: Not in our earlier discussion.  @dinvlad any guidance on this question?

dinvlad: Yes, @gbggrant this was my recommendation! As long as it doesn't break your build or is easy to amend the Dockerfile to unbreak it, please use this image instead - it has much fewer findings.

dinvlad: (any image or tag under `adoptopenjdk` umbrella is also fine - feel free to use any from https://hub.docker.com/_/adoptopenjdk or https://hub.docker.com/u/adoptopenjdk)

dinvlad: That being said, getting the latest (`8` in this case) tag is great because you're assured of the up-to-date fixes.

Issue #1217, opened 2018-08-24
CollectInsertSizeMetrics-make HISTOGRAM_FILE optional


joshua-gould:
It would be useful to turn off creating the pdf output for CollectInsertSizeMetrics because it takes extra time to make a sometimes un-needed plot and required R.

yfarjoun:
Sounds like a nice I idea. can you submit a pr?

Poshi:
Here we are also having some issues with some weird data where most of the alignments are chimeras. The R script is taking eons to finish (10 minutes to process the input, more than one day to generate the PDF) and we never look at the PDF, we are only interested in the data itself. It would be really interesting to avoid the PDF generation in all tools if not explicitly requested. This would generate a few files less, it could effectively remove the dependency on R and it will greatly speedup some analysis like the one I just described.

yfarjoun:
I think that adding an argument `SKIP_PLOT_GENERATION` (or similar) and wrapping the final `RExecutor.executeFromClasspath...` step with a

```
if(!SKIP_PLOT_GENERATION) {
...
}
```
would be an improvment. Feel free to submit a PR to that effect.

Poshi:
I think it's easier: don't make a parameter mandatory. It won't make sense to have to give an output file name for the PDF and then add another parameter to say that it has not to be generated.
In any case, I don't have a clue how Picard is organized and I don't have the proper environment for development and testing of a patch. For sure you don't want a PR that has not been tested, even compiled. For my specific needs, it would probably be easier to just fake the Rscript command so it is apparently run but does nothing. But that would be a very ugly solution and won't help anyone else besides me :-(
Take this idea as a possible improvement. Improving the R script used under the hood to run faster when these conditions apply would also be a good general improvement and would solve the most relevant part of the issue.

hansolo-bioinfo:
I have the same issue. Do we have a solution yet?

Poshi:
Hey! It looks like there was an almost finished PR for that, besides the fact that I could not properly tackle the main issue here.
But I found a workaround to our main issue: the graphic generation taking hours. By optimizing the R script that generates the graphic I managed to cut the time from several hours to 13 seconds. A change that is worth having in any case.

I will try to generate a PR for you, but this will be my first time, so sorry if I screw something in the process (I mean posting the PR to the wrong branch or anything like that). I'm not sure if I should link that PR to any issue, as it is not adressing this specific issue.

Let me know if I should take anything into account for the PR.

PR #1659, opened 2021-03-26
Fix: getCumulative R method optimized

Poshi:
### Description

In this PR I optimized the R method of the script where most of the time was used.

Originally, the cumulative sum of the tails was computed over and over with the whole dataset. This lead to very long times (in the order of hours) in the calculus of an aparently quick calculus. I rewrote the method in order to do the cumulative sum by using the computed sum up to the last position, wich lowers the calculus from a quadratic approach to a linear one, drastically reducing runtime to a fes seconds when the list is large.

This serves as a workaround for a comment in [issue 1217](https://github.com/broadinstitute/picard/issues/1217). This is not the best solution for our problem, but solves the problem at hand.

As this is an optimization, with no functionality change, no tests have been neither modified nor added. Current tests should work. There are no changes in the documentation either. I'm not sure if I should check the marks in the section below...

----

#### Content

#### Review


Poshi:
I got some timings between original and new version with one of my problemmatics datasets:

Version | Real time | System time | User time
------- | --------- | ----------- | ---------
Original | 371m22.927s | 0m19.365s | 370m53.525s
New | 0m12.671s | 0m0.230s | 0m12.387s

gbggrant:
@khajoue2 or @kachulis could one of you take a look at this small PR?  I'm not familiar with R.

kachulis (approved):
that is quite the speedup, thanks for this @Poshi!

Poshi:
You are welcome!
The speedup usually is not so high. The original code is slow, but the common datasets contain a reduced number of entries. The issue comes when you feed it with the results of some kind of experiment that causes the insert sizes vary wildly, so you don't have a small range of sizes (few hundreds), but a broad one (millions). Then, the original quadratic approach shows up.

We should not feed this kind of data on the first place, I know. But this statistic is computed by default for all the experiments and it is not straightforward to get rid of it in specific cases.

Also, the real solution for us would be not having to generate the graph, as we don't use it for anything. But as long as the graph generation is not a bottleneck, we are fine :-) And this optimization, for sure, helps a lot :-)

khajoue2 (approved):
Looks good!

Poshi:
Do I need to do something else?

PR #1672, opened 2021-04-28
[bugfix] Fix an issue with cbcl parsing when barcodes files are in a different directory from basecalls.

jacarey:
The `NewIlluminaDataProvider` was building the barcode writer map based on barcode files it found in the `basecallsDirectory`. If a separate directory was specified then the map was empty and all read were being classified as "unmatched"

This PR changes the `NewIlluminaDataProvider` to search for barcodes files in the proper `barcodesDirectory` when specified.

#### Content

#### Review


src/main/java/picard/illumina/parser/NewIlluminaDataProvider.java
```diff
@@ -66,7 +69,7 @@
         //barcodes
         final Pattern barcodeRegex = Pattern.compile(ParameterizedFileUtil.escapePeriods(
                 ParameterizedFileUtil.makeBarcodeRegex(lane)));
-        final File[] barcodeFiles = getTiledFiles(basecallDirectory, barcodeRegex);
+        final File[] barcodeFiles = getTiledFiles(barcodesDirectory, barcodeRegex);
```
gbggrant, line 72: Would it make sense to throw an exception here if barcodeFiles is empty?
Also add a ` IOUtil.assertFilesAreReadable(Arrays.asList(barcodeFiles));'
as is done for the locs and filter files.

gbggrant (approved):
One suggestion.

src/main/java/picard/illumina/parser/NewIlluminaDataProvider.java
jacarey: Done

src/main/java/picard/illumina/parser/NewIlluminaDataProvider.java
```diff
@@ -66,7 +69,13 @@
         //barcodes
         final Pattern barcodeRegex = Pattern.compile(ParameterizedFileUtil.escapePeriods(
                 ParameterizedFileUtil.makeBarcodeRegex(lane)));
-        final File[] barcodeFiles = getTiledFiles(basecallDirectory, barcodeRegex);
+
+        final File[] barcodeFiles = getTiledFiles(barcodesDirectory, barcodeRegex);
+        if (barcodeFiles.length == 0) {
```
gbggrant, line 74: Probably should use StringUtils.isEmpty here, since (it looks like) directory.listFiles could return null.

src/main/java/picard/illumina/parser/NewIlluminaDataProvider.java
```diff
@@ -66,7 +69,13 @@
...
+            throw new PicardException("No barcodes files found in the barcodesDirectory " + barcodesDirectory.getAbsolutePath());
```
gbggrant, line 75: Is it "No barcode files"?

gbggrant (approved):
more picky.

src/main/java/picard/illumina/parser/NewIlluminaDataProvider.java
gbggrant: Actually that would be ArrayUtils.  Up to you.

jacarey: Probably easier to check `nonNull` vs converting File to String.

src/main/java/picard/illumina/parser/IlluminaDataProviderFactory.java
```diff
@@ -32,17 +32,7 @@
...
-import java.util.Collections;
-import java.util.EnumMap;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.SortedSet;
-import java.util.TreeSet;
+import java.util.*;
```
jessicaway, line 35: Usually we prefer the the explicit import statements rather than *

src/test/java/picard/illumina/IlluminaBasecallsToSamTest.java
```diff
@@ -176,30 +177,32 @@ public void testMolecularBarcodes(final String readStructure, final String[] umi
     @Test(dataProvider = "multiplexedData")
     public void testMultiplexed(final boolean includeBcInHeader, final ClusterDataToSamConverter.PopulateBarcode populateBarcode,
                                 final boolean includeBarcodeQuality, final File testDataDir) throws Exception {
-        runStandardTest(new int[]{1}, "multiplexedBarcode.", "library.params", 1, "25T8B25T", BASECALLS_DIR, testDataDir, null, includeBcInHeader, populateBarcode, includeBarcodeQuality);
+        runStandardTest(new int[]{1}, "multiplexedBarcode.", "library.params", 1, "25T8B25T", BASECALLS_DIR, BASECALLS_DIR, testDataDir, null, includeBcInHeader, populateBarcode, includeBarcodeQuality);
```
jessicaway, line 180: Would it be a lot of work to add a separate directory for the Barcodes rather than storing them in the same directory as the basecalls? I am concerned that this would test would not catch the but that is fixed here

jacarey: There is one test that does just that. All of the other test cases use the default case of using the same directory for both.

jessicaway (approved):

PR #1680, opened 2021-05-13
Fixing Integer comparison for chimeric sequences for ribosomal interval overlap

thomas-watchmaker:
### Description

The current branch uses "==" as Integer comparison to compare reference ids which masks a bug as the Integer gets larger.

As described here: https://stackoverflow.com/questions/3637936/java-integer-equals-vs

This bug can yield an artificially deflated rrna base count if intervals map to contigs with a reference id out of the range given above.

----

#### Content

#### Review


yfarjoun:
this looks fine. I only wonder how many other Integer `==` we have lurking in the code....

jmarshall:
A few minutes with grep suggests there are ~4 instances of this in the HTSJDK code but happily perhaps no more other than this one in the Picard code. (This is checking only code involving `getReferenceIndex` or `getMateReferenceIndex`.) Perhaps there is a linter that will warn about this problem?

jessicaway (approved):

gbggrant (approved):

PR #1690, opened 2021-06-08
Jc fix small data race condition

jacarey:
Thread pool executors often say they are done when in fact there is lingering processing being done in a worker thread. This causes issues with race conditions when tile data is tiny (unit tests)

The performance hit is negligible for using a different pool for each tile and then shutting it down and cleaning it up to ensure that a race condition can not occur.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/illumina/SortedBasecallsConverter.java
+++ b/src/main/java/picard/illumina/SortedBasecallsConverter.java
@@ -104,7 +104,7 @@ public class SortedBasecallsConverter<CLUSTER_OUTPUT_RECORD> extends BasecallsCo
         this.codecPrototype = codecPrototype;
         this.outputRecordComparator = outputRecordComparator;
         this.outputRecordClass = outputRecordClass;
-        tileWriteExecutor = new ThreadPoolExecutorWithExceptions(barcodeRecordWriterMap.keySet().size());
+        tileWriteExecutor = new ThreadPoolExecutorWithExceptions(numThreads);
         tileReadExecutor = new ThreadPoolExecutorWithExceptions(numThreads);
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/picard/illumina/SortedBasecallsConverter.java
```diff
@@ -234,38 +224,33 @@ private synchronized void addRecord(final String barcode, final CLUSTER_OUTPUT_R
...
+        awaitExecutor(tileReadExecutor);
 
         int tileProcessingIndex = 0;
-
+        ThreadPoolExecutorWithExceptions tileWriteExecutor = null;
         while (tileProcessingIndex < tiles.size()) {
-            if (tileWriteJobs.get() == 0) {
-                completedWork.get(tiles.get(tileProcessingIndex)).forEach(tileWriteExecutor::submit);
+                awaitExecutor(tileWriteExecutor);
+                ThreadPoolExecutorWithExceptions finalTileWriters = new ThreadPoolExecutorWithExceptions(numThreads);
```
gbggrant, line 233: Can this be final?

src/main/java/picard/util/ThreadPoolExecutorWithExceptions.java
```diff
@@ -61,4 +61,13 @@ protected void beforeExecute(Thread t, Runnable r) {
     public boolean hasError() {
         return exception != null;
     }
+
+    /**
+     * Calls `shutdownNow, adjusts the core size to 0 and low timeout to ensure threads get killed an GCd
```
gbggrant, line 66:
```suggestion
     * Calls `shutdownNow, adjusts the core size to 0 and low timeout to ensure threads get killed and garbage collected.
```

gbggrant (approved):

src/main/java/picard/illumina/SortedBasecallsConverter.java
jacarey: Actually the intermediate variable isn't even needed since we aren't using it in a lambda

jessicaway (approved):
Is there any changes we could make to the tests so that this issue is caught in the future?

jacarey:
> Is there any changes we could make to the tests so that this issue is caught in the future?

Unfortunately the nature of the race condition makes it difficult to devise a test that would attempt to recreate it.

PR #1694, opened 2021-06-25
bugfix and nio access in Fingerprinting code (no need to panic)

yfarjoun:
feat: Enable NIO access for three other CLPs (ExtractFingerprint, CalculateContamination, and CalculateFingerprintMetrics).
fix: A bug in ExtractFingerprint & CalculateContamination were in the resulting vcf the REF and ALT allele counts in AD were swaped (doesn't affect LOD scores since the downstream tools only looked at PL) 😌 

(https://www.conventionalcommits.org/en/v1.0.0/)


jessicaway (approved):
Looks good, thanks @yfarjoun!

PR #1697, opened 2021-07-09
fix: underflow in HaplotypeProbabilitiesFromContaminatorSequence

yfarjoun:
This PR fixes an underflow (reported to me in private by @sidwekhande) which affects both IdentifyContamination and ExtractFingerprint in deep coverage. This was leading to non-informative sites in RNA and similar sequencing data.

The problem was that internally the code was using linear likelihood to collect the information about the possible combinations of contamination and this quickly led to underflows. 

Here I added tests (that initially failed!) and moved the entire calculation to log10 space.

### DIFF
--- a/src/test/java/picard/fingerprint/FingerprintCheckerTest.java
+++ b/src/test/java/picard/fingerprint/FingerprintCheckerTest.java
@@ -1,13 +1,17 @@
 package picard.fingerprint;
 
 import htsjdk.samtools.metrics.MetricsFile;
+import htsjdk.samtools.reference.ReferenceSequenceFile;
+import htsjdk.samtools.reference.ReferenceSequenceFileFactory;
 import htsjdk.samtools.util.CollectionUtil;
+import htsjdk.variant.variantcontext.VariantContext;
 import htsjdk.variant.vcf.VCFFileReader;
 import org.testng.Assert;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 import picard.PicardException;
+import picard.util.MathUtil;
 import picard.vcf.VcfTestUtils;
 
 import java.io.File;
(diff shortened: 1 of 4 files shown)


src/test/java/picard/fingerprint/FingerprintCheckerTest.java
```diff
@@ -316,8 +320,45 @@ public void testQueryable(final File vcf, boolean expectedQueryable) {
...
+        addObservation(hp1, hb, counts, hb.getFirstSnp().getAllele2());
+
+        final double[] logLikelihoods = hp1.getLogLikelihoods();
+        Assert.assertTrue( MathUtil.min(logLikelihoods) < 0 );
+    }
+
+    @Test(dataProvider = "deepDataProvider")
+    void testCanHandleDeepDataHPFromContamination(int counts) throws IOException {
+        final double maf = 0.4;
+        final Snp snp = new Snp("test", "chr1", 1, (byte) 'A', (byte) 'T', maf, Collections.singletonList("dummy"));
```
kachulis, line 341: What about changing the class field snp so that one of the alleles matches the reference, so you can then use the shared `hb`, instead of making a nearly identical local version here?

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
```diff
@@ -316,8 +320,45 @@ public void testQueryable(final File vcf, boolean expectedQueryable) {
...
+        final HaplotypeBlock hb = new HaplotypeBlock(maf);
+        hb.addSnp(snp);
+
+        final HaplotypeProbabilitiesFromSequence hp1 = new HaplotypeProbabilitiesFromContaminatorSequence(hb,.999);
+        addObservation(hp1, hb, counts, hb.getFirstSnp().getAllele1());
+        addObservation(hp1, hb, counts, hb.getFirstSnp().getAllele2());
+
+        final double[] logLikelihoods = hp1.getLogLikelihoods();
+        Assert.assertTrue( MathUtil.min(logLikelihoods) < 0);
```
kachulis, line 350: is there any reason not to check that the logLikelihoods match the actual values you expect, instead of just checking that they haven't underflowed?

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
```diff
@@ -316,8 +320,45 @@ public void testQueryable(final File vcf, boolean expectedQueryable) {
...
+
+        final File fasta = new File(TEST_DATA_DIR, "reference.fasta");
+
+        try (final ReferenceSequenceFile ref = ReferenceSequenceFileFactory.getReferenceSequenceFile(fasta)) {
```
kachulis, line 354: Why is this also being checked here?  Isn't this redundant with just checking the logLikelihoods of the `HaplotypeProbablity`?  Or if not, why is it only in testCanHandleDeepDataHPFromContamination, but not in testCanHandleDeepData?

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
```diff
@@ -316,8 +320,45 @@ public void testQueryable(final File vcf, boolean expectedQueryable) {
         }
     }
 
+    @DataProvider
+    Object[][] deepDataProvider() {
+        return new Object[][]{{10}, {50}, {100}, {500}, {1000}, {5000}, {10000}, {50000}};
+    }
+
+    @Test(dataProvider = "deepDataProvider")
+    void testCanHandleDeepData(int counts) {
```
kachulis, line 329: I feel like `testCanHandleDeepData` and `testCanHandleDeepDataHPFromContamination` could be refactored slightly more, considering they should be basically doing exactly the same thing.

kachulis (changes requested):
Looks good, just a few minor comments about the tests.

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
yfarjoun: With a data provider making this many tests, I didn't feel like calculating the expectation and this was enough to catch the underflow, which was the purpose of this PR...

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
yfarjoun: :+1:

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
yfarjoun: I wasn't sure where the problem was....since there's a transformation between fingerprinting and the final VCF, I wanted to make sure I look at the VCF....I can refactor as you suggested above.

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
yfarjoun: 👍

yfarjoun:
Thanks for the review @kachulis. back to you.

src/test/java/picard/fingerprint/FingerprintCheckerTest.java
```diff
@@ -320,43 +321,37 @@ public void testQueryable(final File vcf, boolean expectedQueryable) {
...
         final File fasta = new File(TEST_DATA_DIR, "reference.fasta");
 
         try (final ReferenceSequenceFile ref = ReferenceSequenceFileFactory.getReferenceSequenceFile(fasta)) {
-            final VariantContext vc = FingerprintUtils.getVariantContext(ref, "test", hp1);
+            final VariantContext vc = FingerprintUtils.getVariantContext(ref, "test", hp);
             Assert.assertTrue(MathUtil.max(MathUtil.promote(vc.getGenotype(0).getPL())) > 0);
         }
     }
 
+
```
kachulis, line 354: extra newline

kachulis (approved):
Looks good, thanks for the fix @yfarjoun!

PR #1700, opened 2021-07-28
Adjust handling of missing data situations in checkFingerprints

kachulis:
In checkFingerprints, if all lod scores are 0, the tool returns an exit code of 1, since there were no valid checks preformed.  There are two dominant ways this situation can occur.  The first is when the sample being checked is not present in the genotypes file.  The second is when the sample being checked is found, but there is no data covering the fingerprinting sites (or no overlap between the fingerprinting sites which are covered).  This PR separates the code paths of these two cases, since in some situations it may be desirable to handle them differently.  Also makes the exit codes returned in these cases parametrizable via the command line (as is already the case in crosscheckfingerprints).


src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -408,4 +380,49 @@ static boolean fileContainsReads(final Path p) {
                 p.toUri().getRawPath().endsWith(SamReader.Type.SAM_TYPE.fileExtension()) ||
                 p.toUri().getRawPath().endsWith(SamReader.Type.CRAM_TYPE.fileExtension()));
     }
+
+    private boolean containsSample(final Path p, final String sample) {
```
gbggrant, line 384: Can you add a unit test for this method?

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -408,4 +380,49 @@ static boolean fileContainsReads(final Path p) {
...
+        try (final VCFFileReader fileReader = new VCFFileReader(p, false)) {
+            final VCFHeader fileHeader = fileReader.getFileHeader();
+            return fileHeader.getGenotypeSamples().contains(sample);
+        }
+    }
+
+    private String extractObservedSampleName() {
```
gbggrant, line 391: Similarly, a unit test for this would be good.

src/main/java/picard/fingerprint/CheckFingerprint.java
```diff
@@ -208,6 +208,12 @@
             "ignoring RG annotations, and producing a single fingerprint metric for the entire BAM.")
     public boolean IGNORE_READ_GROUPS = false;
 
+    @Argument(doc = "When the expected fingerprint sample is not found in the genotypes file, this exit code is returned.")
+    public int EXIT_CODE_WHEN_EXPECTED_SAMPLE_NOT_FOUND = 1;
+
+    @Argument(doc = "When all LOD score are zero, exit with this value.")
```
gbggrant, line 214: Do you think it makes sense to have these default to different return codes?  I can't imagine there's a lot of legacy code checking specifically for 1 vs !0.

src/test/java/picard/fingerprint/CheckFingerprintTest.java
```diff
@@ -125,6 +125,20 @@ public void testMismatchingSamples2() {
         Assert.assertEquals(runPicardCommandLine(args), 1);
     }
 
+    @Test
```
gbggrant, line 128: Shouldn't you have tests to validate that you get the two different exit codes for the two different code paths you have added?

gbggrant (changes requested):
This looks really good, just think it would be nice to have tests to cover the new possible exit values.

src/main/java/picard/fingerprint/CheckFingerprint.java
kachulis: yeah, I always hesitant to change return code behavior like this, but I think in this case you are right the risk is very low.

gbggrant (approved):

gbggrant:
Looks good @kachulis

PR #1760, opened 2021-12-13
Upgrade Barclay.

cmnbroad:
Looks like we never upgraded Picard to Barclay 4.0. This change upgrades to Barclay 4.02, which version uses the log4j 2.15.


gbggrant (approved):

gbggrant:
Thank you!

PR #1782, opened 2022-03-08
Add support for documentation generation for metrics classes.

cmnbroad:
@derekca @gbrandt6 @droazen @lbergelson FYI.

This implements documentation generation for metrics classes. It uses the existing `@DocumentedFeature` annotation, along with a custom doc template (which is easier than overloading the existing template), and a few lines of custom doclet code that selects that template for metrics classes and displays all public fields (in all super classes in the class hierarchy of the annotated metrics class) in a table.

There are two commits; one with the core code, and one that add the annotations to the various metrics classes. A separate sweep should probably be done to add metrics-specific summaries (using the `@DocumentedFeature` `summary` attribute) for each class, and also some of the metrics source classes seem to have no documentation at all.

Alternative implementations would be:
- add a new class-level annotation to Barclay, analogous to `@DocumentedFeature`, and interpret it to mean "document all  for non-@Argument fields in a table", or
- Add a new explicit field-level annotation to Barclay, analogous to `@Argument`, but that would require field-level annotation of each metric variable in each class

But the solution in this PR requires no Barclay changes, and given that the behavior seems pretty specific to metrics, I think its the simplest solution - its basically all custom code in Picard that specializes/overrides the default Barclay doclet. The one caveat is that if we want the GATK doc build to pick up the metrics doc as well, we'll need to propagate the new template and the few lines of custom doclet code to the GATK repo, since GATK uses it's own custom doclet and templates.

Below is some sample output (I don't have all the correct style sheets locally so these are not quite "publish-ready).

**Index Page:** (note that only AlignmentSummaryMetrics has a n actual summary, because its the only one I wrote - as mentioned above, all of the classes need to have a short summary line to be added):
Screen Shot 2022-03-08 at 10 56 29 AM

**Detail Page:**
Screen Shot 2022-03-08 at 10 56 41 AM


src/main/resources/picard/helpTemplates/metrics.template.html
```diff
@@ -0,0 +1,79 @@
...
+		<#if group?? >
+			<h3>Category
+				<small> ${group}</small>
+			</h3>
+		</#if>
+		<hr>
+		<h2>Overview</h2>
+		${description}
+
+			<#-- Create the argument summary -->
```
cmnbroad, line 44: These are not arguments, their metrics....

src/main/resources/picard/helpTemplates/metrics.template.html
```diff
@@ -0,0 +1,79 @@
...
+
+			<#-- List all of the things -->
+			<#if arguments.all?size != 0>
+				<#-- Create the argument details -->
+					<h3>Argument details</h3>
+					<p>Arguments in this list are specific to this tool. Keep in mind that other arguments are available that are shared with other tools (e.g. command-line GATK arguments); see Inherited arguments above.</p>
+					<#list arguments.all as arg>
+						<@argumentDetails arg=arg/>
+					</#list>
+			</#if>
```
cmnbroad, line 72: This if block and the comment are irrelevant and should be removed..

src/main/java/picard/util/help/PicardHelpDocWorkUnitHandler.java
```diff
@@ -50,10 +71,23 @@ protected void addCustomBindings(final DocWorkUnit currentWorkUnit) {
...
+            currentWorkUnit.setProperty(WORK_UNIT_SUMMARY_KEY, currentWorkUnit.getSummary());
+            final List<Map<String, String>> metricsFields = new ArrayList<>();
+            currentWorkUnit.setProperty(METRICS_MAP_ENTRY_KEY, metricsFields);
+            final FieldDoc[] fieldDocs = currentWorkUnit.getClassDoc().fields(false);
+            for (final FieldDoc fd : fieldDocs) {
+                if (fd.isPublic()) {
+                    final Map<String, String> metricsField = new HashMap();
+                    metricsField.put(METRICS_MAP_NAME_KEY, fd.name());
+                    metricsField.put(METRICS_MAP_SUMMARY_KEY, fd.getRawCommentText());
+                    metricsFields.add(metricsField);
```
cmnbroad, line 88: Hm. Reviewing you own code days later is interesting. `metricsFields` -> `metricsList` and `metricsField` -> `metricsFields`.

derekca:
@cmnbroad I'm not familiar enough with how the backend works, but what you're describing here is basically what we had in mind. I think that it should help us in migrating over the bulk of the information that we need onto the GATK repo.
Whatever else is leftover can be done manually by us in User Ed, if there is anything left over on the deprecated site.

src/main/java/picard/util/help/HelpConstants.java
```diff
@@ -43,6 +43,9 @@ private HelpConstants() {}
     public final static String DOC_CAT_GENOTYPING_ARRAYS_MANIPULATION = "Genotyping Arrays Manipulation";
     public final static String DOC_CAT_GENOTYPING_ARRAYS_MANIPULATION_SUMMARY = "Tools that manipulate data generated by Genotyping arrays";
 
+    public final static String DOC_CAT_METRICS = "Metrics";
+    public final static String DOC_CAT_METRICS_SUMMARY = "Metrics ";
```
gbggrant, line 47: Is the trailing space on "Metrics " intentional?

gbggrant (approved):
LGTM. One very minor question.

src/main/resources/picard/helpTemplates/metrics.template.html
cmnbroad: Fixed.

src/main/java/picard/util/help/PicardHelpDocWorkUnitHandler.java
cmnbroad: Fixed.

derekca:
> _Is the plan to restart the publishing of the Picard doc along with Picard releases (I assume we're still not doing that)_

Basically, we will be shutting down the [Picard](https://broadinstitute.github.io/picard/) site. There is no set date for the site to be shut down, but it will happen as soon as we finish migrating all the information from that site into GATK documentation as part of this ticket.

So, any new Picard documentation or updates should go directly to the GATK docs, and not to the site.

> _or to just continue to publish the output of the GATK doc build ?_

Yes, exactly. The GATK docs will continue to get built, but now they can pull the [missing] Picard metrics information as well.

> _In order for the metrics doc to be produced by the GATK doc build, a separate PR (a subset of this one) will be required in GATK to enable these there. Once this is merged, I'll add that PR to GATK._

Okay, if that makes the most sense, then let's do that!

cmnbroad:
Oh I see [derekca](https://github.com/derekca) - that sounds good. I'll need to rebase this now since its out-of-date and then we can merge it so I can make the GATK PR.

gbggrant:
@cmnbroad is this PR ready to merge?

src/main/java/picard/util/help/PicardHelpDocWorkUnitHandler.java
```diff
@@ -50,10 +71,23 @@ protected void addCustomBindings(final DocWorkUnit currentWorkUnit) {
...
-            currentWorkUnit.setProperty("summary", clpProperties.summary());
+            currentWorkUnit.setProperty(WORK_UNIT_SUMMARY_KEY, clpProperties.summary());
+        } else if (MetricBase.class.isAssignableFrom(clazz)) {
+            currentWorkUnit.setProperty(WORK_UNIT_SUMMARY_KEY, currentWorkUnit.getSummary());
+            final List<Map<String, String>> workUnitMetricsList = new ArrayList<>();
+            currentWorkUnit.setProperty(METRICS_MAP_ENTRY_KEY, workUnitMetricsList);
+            final FieldDoc[] fieldDocs = currentWorkUnit.getClassDoc().fields(false);
+            for (final FieldDoc fd : fieldDocs) {
+                if (fd.isPublic()) {
+                    final Map<String, String> metricsFields = new HashMap();
```
cmnbroad, line 85: The stricter GATK build complains about this raw type. Fixed.

cmnbroad:
I updated one line due to fix usage of a raw generic type that was detected by the GATK build when I made the GATK pr to reflect these changes. This is ready once tests pass.

Issue #1784, opened 2022-03-11
mixed command line argument syntax throws too unspecific error message


dlaehnemann:
### Affected tool(s)

I assume all tools, but I have tested this with `MarkDuplicates`.

### Affected version(s)

- [x] Latest public release version [2.26.11 installed via bioconda]
- [ ] Latest development/master branch as of [date of test?]

### Description 

Because we construct the command of `picard MarkDuplicates` from various sources, a recent transition to using the new command line argument syntax accidentally resulted in a command with mixed syntax usage:

```
picard MarkDuplicates VALIDATION_STRINGENCY=LENIENT   --INPUT path/to/some.sorted.bam --OUTPUT path/to/some.marked.sorted.bam --METRICS_FILE path/to/some.marked.metrics.txt
```
This throws an error about the `--INPUT` argument being invalid, although it is perfectly valid under the new syntax. So the error message is too unspecific to quickly find the problem, in my opinion -- at least it took me quite a while to spot the problem.

#### Steps to reproduce

Run the above command.

#### Expected behavior

Throw an error message that says something along the lines of:
```
ERROR: You are mixing old and new command line argument syntax, but have to stick to one. Compare 'VALIDATION_STRINGENCY=LENIENT' to '--INPUT path/to/some.sorted.bam'. More info here: https://github.com/broadinstitute/picard/wiki/Command-Line-Syntax-Transition-For-Users-(Pre-Transition)
```

#### Actual behavior

It throws an error that does not point to the actual problem:
```
ERROR: Invalid argument '--INPUT'.
```

cmnbroad:
Agreed, the failure mode/error message should be better.

PR #1785, opened 2022-03-15
Update command line parser argument style detection, add tests.

cmnbroad:
Fixes https://github.com/broadinstitute/picard/issues/1784 (along with an incidental issue uncovered by that ticket).

Edit: Its worth noting that in addition to improving the error reporting in the case of an ambiguity (a putative mix of legacy and posix), this PR also changes the default i from "legacy" to "posix" for that case. See inline comments for more detail.


mjhipp:
Any updates on this? We also have a use case where we need to use a quoted string with an `=` along with the new CLI syntax

clintval:
@cmnbroad would you be willing to merge so we can rely on this patch? Thank-you!

PR #1787, opened 2022-03-15
Simplify handling of paths for input in BuildBamIndex.java

markjschreiber:
### Description

Changed to treat all input as a `Path` and use `IOUtil` to choose the path type so that paths with schemes handled by SPI providers are correctly handled as well as paths that have no scheme and are local files.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/vcf/AccumulateVariantCallingMetrics.java
+++ b/src/main/java/picard/vcf/AccumulateVariantCallingMetrics.java
@@ -24,9 +24,9 @@
 package picard.vcf;
 
 import htsjdk.samtools.metrics.MetricsFile;
-import htsjdk.samtools.util.*;
-import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
+import htsjdk.samtools.util.IOUtil;
 import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
 import org.broadinstitute.barclay.help.DocumentedFeature;
 import picard.PicardException;
 import picard.cmdline.CommandLineProgram;
@@ -34,9 +34,13 @@ import picard.cmdline.StandardOptionDefinitions;
 import picard.cmdline.programgroups.DiagnosticsAndQCProgramGroup;
 
 import java.io.File;
-import java.io.FileReader;
 import java.io.IOException;
-import java.util.*;
+import java.nio.file.FileSystem;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 /**
  * Combines multiple Variant Calling Metrics files into a single file.
@@ -53,7 +57,7 @@ import java.util.*;
 public class AccumulateVariantCallingMetrics extends CommandLineProgram {
 
     @Argument(shortName= StandardOptionDefinitions.INPUT_SHORT_NAME, doc="Paths (except for the file extensions) of Variant Calling Metrics files to read and merge.", minElements=1)
-    public List<File> INPUT;
+    public List<String> INPUT;
 
     @Argument(shortName = StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "Path (except for the file extension) of output metrics files to write.")
     public File OUTPUT;
@@ -71,15 +75,17 @@ public class AccumulateVariantCallingMetrics extends CommandLineProgram {
         final Map<String, CollectVariantCallingMetrics.VariantCallingDetailMetrics> collapsedSampleDetailsMap = new HashMap<>();
         final CollectVariantCallingMetrics.VariantCallingSummaryMetrics collapsedSummary = new CollectVariantCallingMetrics.VariantCallingSummaryMetrics();
 
-        for (final File file : INPUT) {
-            final String inputPrefix = file.getAbsolutePath() + ".";
+        List<Path> inputPaths = IOUtil.getPaths(INPUT);
+        for (final Path path : inputPaths) {
+            final FileSystem fs = path.getFileSystem();
+            final String inputPrefix = path.toAbsolutePath()+ ".";
 
             try {
                 // read in the detailed metrics file
-                final File detail = new File(inputPrefix + CollectVariantCallingMetrics.VariantCallingDetailMetrics.getFileExtension());
+                final Path detail = fs.getPath(inputPrefix + CollectVariantCallingMetrics.VariantCallingDetailMetrics.getFileExtension());
                 IOUtil.assertFileIsReadable(detail);
                 MetricsFile<CollectVariantCallingMetrics.VariantCallingDetailMetrics, ?> detailedMetricsFile = getMetricsFile();
-                detailedMetricsFile.read(new FileReader(detail));
+                detailedMetricsFile.read(Files.newBufferedReader(detail));
 
                 // for each sample in the detailed metrics...
                 long totalHetDepth = 0L;
@@ -94,10 +100,10 @@ public class AccumulateVariantCallingMetrics extends CommandLineProgram {
                 }
 
                 // next, read in the new summary metrics
-                final File summary = new File(inputPrefix + CollectVariantCallingMetrics.VariantCallingSummaryMetrics.getFileExtension());
+                final Path summary = fs.getPath(inputPrefix + CollectVariantCallingMetrics.VariantCallingSummaryMetrics.getFileExtension());
                 IOUtil.assertFileIsReadable(summary);
                 MetricsFile<CollectVariantCallingMetrics.VariantCallingSummaryMetrics, ?> summaryMetricsFile = getMetricsFile();
-                summaryMetricsFile.read(new FileReader(summary));
+                summaryMetricsFile.read(Files.newBufferedReader(summary));
                 if (summaryMetricsFile.getMetrics().size() != 1) {
                     throw new PicardException(String.format("Expected 1 row in the summary metrics file but saw %d", summaryMetricsFile.getMetrics().size()));
                 }
--- a/src/main/java/picard/vcf/AccumulateVariantCallingMetrics.java
+++ b/src/main/java/picard/vcf/AccumulateVariantCallingMetrics.java
@@ -32,6 +32,7 @@ import picard.PicardException;
 import picard.cmdline.CommandLineProgram;
 import picard.cmdline.StandardOptionDefinitions;
 import picard.cmdline.programgroups.DiagnosticsAndQCProgramGroup;
+import picard.nio.PicardHtsPath;
 
 import java.io.File;
 import java.io.IOException;
@@ -41,6 +42,7 @@ import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * Combines multiple Variant Calling Metrics files into a single file.
@@ -57,7 +59,7 @@ import java.util.Map;
 public class AccumulateVariantCallingMetrics extends CommandLineProgram {
 
     @Argument(shortName= StandardOptionDefinitions.INPUT_SHORT_NAME, doc="Paths (except for the file extensions) of Variant Calling Metrics files to read and merge.", minElements=1)
-    public List<String> INPUT;
+    public List<PicardHtsPath> INPUT;
 
     @Argument(shortName = StandardOptionDefinitions.OUTPUT_SHORT_NAME, doc = "Path (except for the file extension) of output metrics files to write.")
     public File OUTPUT;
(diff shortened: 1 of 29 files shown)


src/main/java/picard/vcf/SplitVcfs.java
```diff
@@ -73,16 +73,26 @@ public SplitVcfs() {
 
     @Override
     protected int doWork() {
-        IOUtil.assertFileIsReadable(INPUT);
+        Path inputPath;
```
yfarjoun, line 76: final

src/main/java/picard/vcf/SplitVcfs.java
```diff
@@ -73,16 +73,26 @@ public SplitVcfs() {
...
+        try {
+            inputPath = IOUtil.getPath(INPUT);
+        } catch (IOException e) {
+            throw new RuntimeIOException(e);
+        }
+        IOUtil.assertPathsAreReadable(Stream.of(inputPath).collect(Collectors.toList()));
```
yfarjoun, line 82: there's only one path here, no? why the stream and list?

src/main/java/picard/vcf/processor/VariantIteratorProducer.java
```diff
@@ -68,17 +60,30 @@
      * filtering of {@link VariantContext}
      */
     public static VariantIteratorProducer byHundredMegabaseChunksWithOnTheFlyFilteringByInterval(final List<File> vcfs, final IntervalList intervalList) {
-        return new Threadsafe(VcfFileSegmentGenerator.byWholeContigSubdividingWithWidth(ONE_HUNDRED_MILLION), vcfs, intervalList);
+        return new Threadsafe(VcfPathSegmentGenerator.byWholeContigSubdividingWithWidth(ONE_HUNDRED_MILLION), vcfs.stream().map(File::toPath).collect(Collectors.toList()), intervalList);
```
yfarjoun, line 63: htsjdk.samtools.util.IOUtil#filesToPaths

yfarjoun (changes requested):
Thanks for this long overdue transition from file to path. 

Have you evaluated the option of using HtsPath? I know that we could change to that later on, but since this PR is opening many public APIs, I don't want to have to be in a situation of either maintaining or deprecating too many end points....

@cmnbroad perhaps you can opine since you are more aware of the readiness status of HtsPath (perhaps)....

markjschreiber:
> 

Thanks for the tip about `HtsPath`, that is more elegant. It seems `HtsPath.toPath()` already exists and returns an nio Path so I can use that. It works for paths with schemes (URIs) like `s3://foo/baa` and local files like `/path/to/file.bam`. Is there a need to override it?

cmnbroad:
Yes, HtsPath is what we use in GATK, since its always backed by a URI, and handles protocol schemes, query params, , absolute and relative local paths, url-encoding for things like embedded spaces, etc. We'd prefer to subclass it since it allows us to implement provider-specific functionality when loading file systems (we have needed to do that for GCS in GATK, and that can't reside in htsjdk since it would confer unwanted dependencies there). But I don't want to increase the tax too much here - if you can add the subclass, great, if not, just use HtsPath and I'll do the rest in a separate pass.

markjschreiber:
> Yes, HtsPath is what we use in GATK, since its always backed by a URI, and handles protocol schemes, query params, absolute and relative local paths, url-encoding for things like embedded spaces, etc. We'd prefer to subclass it since it allows us to implement provider-specific functionality when loading file systems (we have needed to do that for GCS in GATK, and that can't reside in htsjdk since it would confer unwanted dependencies there). But I don't want to increase the tax too much here - if you can add the subclass, great, if not, just use HtsPath and I'll do the rest in a separate pass.

Happy to subclass it but I can't see what is needed in the overloaded `toPath()` method. I had thought that NIO SPI implementations were not supposed to need special handling when loading other than being on the class path or the in the JDK extension directory?  The s3 nio spi that I have been developing seems to be automatically loaded by the JVM when it sees a URI with the `s3` scheme.

src/main/java/picard/vcf/processor/VcfPathSegmentGenerator.java
```diff
@@ -0,0 +1,189 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+/**
+ * Describes a mechanism for producing {@link VcfPathSegment}s from a VCF file path.
+ *
+ */
+public abstract class VcfPathSegmentGenerator {
```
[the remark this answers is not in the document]
markjschreiber: True, although I think the duplication is with code from deprecated files which I assume would eventually be removed.

src/main/java/picard/vcf/SplitVcfs.java
markjschreiber: There is no `assertPathIsReadable(Path p)` in `IOUtils` so I have to make a `List`, however I can use the more concise `Collections.singletonList(inputPath)`

markjschreiber:
I have updated the code based on PR feedback including extending `HTSPath`. Let me know if you want additional changes.

src/main/java/picard/vcf/processor/VcfPathSegmentGenerator.java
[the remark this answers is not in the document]
yfarjoun: ah yes. I realized that eventually, and meant to remove my comment...sorry.

yfarjoun (changes requested):
Thanks for the fast turnaround. I think that you were a little overzealous with the use of HtsPath. My understanding is that it should be used as a way of inputing commandline Strings into path-like objects, but that the code itself should use regular Path. @cmnbroad do yuo agree?

cmnbroad:
@yfarjoun @markjschreiber In my mind, the ultimate goal is to use the `HtsPath`-derived class as currency wherever possible, eliminating any use of, and interconversion between `File` or `String` (again, wherever possible); and converting just-in-time to `Path` at any API boundary that requires it (except htsjdk, as it becomes more ubiquitous there.) Part of the reason is related to another question @markjschreiber asked, which is why we need to subclass `HtsPath` in the first place. In GATK  we have had to put provider-specific functionality into the `HtsPath`-derived class for GCS and HDFS, to support things like file system loading, getting an input stream, etc. Ideally this shouldn't be required, but in practice it has been necessary, and it's so much easier if we have a central location to enable that kind of thing. Also,`HtsPath` already has some useful utilities such as `isCram`/'isBam`, etc., that I expect that to continue to evolve.

But ubiquitous `HtsPath` is a longer term goal - for now we don't have all the utility support we need to do that (i.e, assertIsReadable).  So we won't get there in this PR, but I think this is a big step forward in the right direction.

src/main/java/picard/vcf/AccumulateVariantCallingMetrics.java
```diff
@@ -75,7 +77,7 @@ protected int doWork() {
         final Map<String, CollectVariantCallingMetrics.VariantCallingDetailMetrics> collapsedSampleDetailsMap = new HashMap<>();
         final CollectVariantCallingMetrics.VariantCallingSummaryMetrics collapsedSummary = new CollectVariantCallingMetrics.VariantCallingSummaryMetrics();
 
-        List<Path> inputPaths = IOUtil.getPaths(INPUT);
+        final List<Path> inputPaths = INPUT.stream().map(PicardHtsPath::toPath).collect(Collectors.toList());
```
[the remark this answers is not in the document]
markjschreiber: Done

markjschreiber:
> @yfarjoun @markjschreiber In my mind, the ultimate goal is to use the `HtsPath`-derived class as currency wherever possible, eliminating any use of, and interconversion between `File` or `String` (again, wherever possible); and converting just-in-time to `Path` at any API boundary that requires it (except htsjdk, as it becomes more ubiquitous there.) Part of the reason is related to another question @markjschreiber asked, which is why we need to subclass `HtsPath` in the first place. In GATK we have had to put provider-specific functionality into the `HtsPath`-derived class for GCS and HDFS, to support things like file system loading, getting an input stream, etc. Ideally this shouldn't be required, but in practice it has been necessary, and it's so much easier if we have a central location to enable that kind of thing. Also,`HtsPath` already has some useful utilities such as `isCram`/'isBam`, etc., that I expect that to continue to evolve.
> 
> But ubiquitous `HtsPath` is a longer term goal - for now we don't have all the utility support we need to do that (i.e, assertIsReadable). So we won't get there in this PR, but I think this is a big step forward in the right direction.

In my latest commit, I have reverted the use of `PicardHtsPath` for already existing public APIs but I have used them in new classes I have added to push further into a future where this becomes the standard Path type for Picard.

If there is interest I can continue to convert further Picard applications with the new pattern.

markjschreiber:
Hi @yfarjoun and @cmnbroad, I wasn't totally sure were you landed on the use of `Path` vs `PicardHtsPath` in the library but I think my last commit is in line with what you suggested. Can you review and let me know? 

If possible I'd like to merge soon and continue updating additional Picard applications to be able to use URIs for input.

src/test/java/picard/vcf/AccumulateVariantCallingMetricsTest.java
```diff
@@ -57,15 +61,15 @@
...
+    public void testMerge(final List<String> inputs) throws IOException {
         final File mergedFilePrefix = new File(TEST_DATA_DIR + "mergeTest");
         final File mergedSummaryFile = new File(mergedFilePrefix.getAbsolutePath() + ".variant_calling_summary_metrics");
         final File mergedDetailFile = new File(mergedFilePrefix.getAbsolutePath() + ".variant_calling_detail_metrics");
         mergedSummaryFile.deleteOnExit();
         mergedDetailFile.deleteOnExit();
 
         final AccumulateVariantCallingMetrics program = new AccumulateVariantCallingMetrics();
-        program.INPUT = inputs;
+        program.INPUT = inputs.stream().map(PicardHtsPath::new).collect(Collectors.toList());
```
yfarjoun, line 72: this should be a util method

yfarjoun (approved):
looks good to me (barring one small request). 

@cmnbroad want to give a thumb regarding the use of path vs. htsjdkpath?

src/test/java/picard/vcf/AccumulateVariantCallingMetricsTest.java
markjschreiber: Done

markjschreiber:
@yfarjoun, it looks like you need to approve running of the DSP AppSec check as well. As a first time contributor I can't do this.

src/main/java/picard/vcf/SplitVcfs.java
```diff
@@ -60,7 +65,7 @@
     public File INDEL_OUTPUT;
 
     @Argument(shortName = "D", doc = "The index sequence dictionary to use instead of the sequence dictionaries in the input files", optional = true)
-    public File SEQUENCE_DICTIONARY;
+    public String SEQUENCE_DICTIONARY;
```
cmnbroad, line 68: Is there a reason not to use `PicardHtsPath` here ? Then I think you wouldn't need to add the try/catch block where it's used below to get the sequence dictionary.

src/test/java/picard/vcf/GatherVcfsTest.java
```diff
@@ -61,18 +64,18 @@ public void setup() throws IOException {
...
     @Test(dataProvider = "vcfshards")
-    public void TestGatherFiles(final List<File> inputFiles, final File expectedOutput, final int expectedRetVal, boolean reorder) throws IOException {
+    public void TestGatherFiles(final List<HtsPath> inputFiles, final File expectedOutput, final int expectedRetVal, boolean reorder) throws IOException {
         final String comment1 = "This is a comment";
         final List<String> args = new ArrayList<>();
 
         final File output = VcfTestUtils.createTemporaryIndexedFile("result", expectedOutput.getAbsolutePath().endsWith(".vcf") ? ".vcf" : ".vcf.gz");
 
-        inputFiles.forEach(f -> args.add("INPUT=" + f.getAbsolutePath()));
+        inputFiles.forEach(f -> args.add("INPUT=" + f.toPath().toUri()));
```
cmnbroad, line 73:
```suggestion
        inputFiles.forEach(f -> args.add("INPUT=" + f.getURI()));
```

cmnbroad (approved):
Two very minor optional changes. Otherwise the HtsPath usage looks good to me.

src/main/java/picard/vcf/SplitVcfs.java
markjschreiber: Good catch

src/test/java/picard/vcf/GatherVcfsTest.java
markjschreiber: Done

markjschreiber:
@yfarjoun @cmnbroad , thanks for your approvals. I can't merge the PR as I don't have permission to run "DSP AppSec Trivy check" as a first time contributor. Could you run this check/ and or merge the PR? Thanks.

cmnbroad:
@markjschreiber Thanks so much for doing this!

markjschreiber:
Thanks for adding these changes!

Issue #1810, opened 2022-05-25
FastqToSam now breaks with gzipped FASTQ input


mjhipp:
## Bug Report

After #1794, `FastqToSam` no longer works with gzipped input.

### Affected tool(s)

- `FastqToSam`

### Affected version(s)
- [ ] Latest public release version 
- [x] Latest development/master branch as of 5/25/2022
- [x] d784ca32864db72c821d38875fed7bea0524ab87 and after

Not affected: 9eafe4ea7b450d6d545ad73512b6e3cee3fb48a5 and before

#### Steps to reproduce

Run `FastqToSam` with gzipped input.

```
❯ java -jar picard.jar FastqToSam \
   FASTQ= sample.R1.fastq.gz \
   FASTQ2= sample.R2.fastq.gz \
   SO=queryname \
   O= out.bam \
   SAMPLE_NAME=sample
```

If you unzip and run again with unzipped fastqs, it works as expected.

#### Expected behavior

Gzipped FASTQs should be converted to SAM/BAM as in previous versions.

Even after updates, the help string says that gzipped input is supported:

```
FASTQ=PicardHtsPath
F1=PicardHtsPath              Input fastq file (optionally gzipped) for single end data, or first read in paired end
                              data.  Required.

FASTQ2=PicardHtsPath
F2=PicardHtsPath              Input fastq file (optionally gzipped) for the second read of paired end data.  Default
                              value: null.
```

#### Actual behavior

Failure when creating/referencing the `FastqReader`.

```
[Wed May 25 13:32:52 PDT 2022] FastqToSam FASTQ=sample.R1.fastq.gz FASTQ2=sample.R2.fastq.gz OUTPUT=out.bam SAMPLE_NAME=sample SORT_ORDER=queryname    USE_SEQUENTIAL_FASTQS=false READ_GROUP_NAME=A MIN_Q=0 MAX_Q=93 STRIP_UNPAIRED_MATE_NUMBER=false ALLOW_AND_IGNORE_EMPTY_LINES=false VERBOSITY=INFO QUIET=false VALIDATION_STRINGENCY=STRICT COMPRESSION_LEVEL=5 MAX_RECORDS_IN_RAM=500000 CREATE_INDEX=false CREATE_MD5_FILE=false GA4GH_CLIENT_SECRETS=client_secrets.json USE_JDK_DEFLATER=false USE_JDK_INFLATER=false
[...]

PR #1811, opened 2022-05-31
Fix FastqToSam regression.

cmnbroad:
Fixes https://github.com/broadinstitute/picard/issues/1810.


mjhipp:
Thanks! I tried with test case from #1810 and it works as expected!

cmnbroad:
Thanks for confirming, @mjhipp.

gbggrant (approved):

PR #1825, opened 2022-08-23
Enables setting the google-project that will be charged for requests …

yfarjoun:
…that have a "requester pays" component.


src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -237,6 +240,13 @@ public int instanceMain(final String[] argv) {
         if (System.getProperty("ga4gh.client_secrets") == null) {
           System.setProperty("ga4gh.client_secrets", GA4GH_CLIENT_SECRETS);
         }
+
+        if (System.getProperty("google_project_requester_pays") == null && REQUESTER_PAYS_PROJECT != null) {
+            System.setProperty("google_project_requester_pays", REQUESTER_PAYS_PROJECT);
+            Log.getInstance(this.getClass()).info(
```
lbergelson, line 246: Should this only be output if the google libraries are loaded?

src/main/java/picard/cmdline/CommandLineProgram.java
```diff
@@ -237,6 +240,13 @@ public int instanceMain(final String[] argv) {
         if (System.getProperty("ga4gh.client_secrets") == null) {
           System.setProperty("ga4gh.client_secrets", GA4GH_CLIENT_SECRETS);
         }
+
+        if (System.getProperty("google_project_requester_pays") == null && REQUESTER_PAYS_PROJECT != null) {
```
lbergelson, line 244: There's a way to get the default project from the environment I think.  I don't know if you want to get that by default?

src/main/java/picard/nio/GoogleStorageUtils.java
```diff
@@ -51,18 +50,20 @@ class GoogleStorageUtils {
 
     public static void initialize() {
         // requester pays support is currently not configured
```
lbergelson, line 52: Comment is out of date now.

src/main/java/picard/cmdline/CommandLineProgram.java
yfarjoun: does GATK do that?

src/main/java/picard/nio/GoogleStorageUtils.java
yfarjoun: 👍

src/main/java/picard/nio/GoogleStorageUtils.java
```diff
@@ -50,19 +49,20 @@
 class GoogleStorageUtils {
 
     public static void initialize() {
-        // requester pays support is currently not configured
-        CloudStorageFileSystemProvider.setDefaultCloudStorageConfiguration(GoogleStorageUtils.getCloudStorageConfiguration(20, null));
+        final String google_project = System.getProperty("google_project_requester_pays");
```
lbergelson, line 52: It seems like this is a bit load-order dependent.  Have you checked that the commandline argument overrides the setting in the way you want?  I think it's also possible to get the default google project so it's not necessary to specify at all.  In gatk we instantiate a preferences object and check if it has a value for the project I think, but I'm not sure if it works or has any weirdness.

lbergelson (commented):
@yfarjoun Looks good I think, although the value looks like it depends on the order things get loaded so I would check that it's working as intended when overriding a system property.

yfarjoun:
@kachulis what's the protocol now? who needs to give this a 👍  so that it can be merged?

src/main/java/picard/nio/GoogleStorageUtils.java
yfarjoun: how about we get this simple version to work and if people need something else, it can be added later?

src/main/java/picard/cmdline/CommandLineProgram.java
yfarjoun: if the libraries are not loaded, what's the meaning of having "requestor pays"? is there something you'd like it to output in that case?

droazen:
@yfarjoun @lbergelson This PR can't be merged as-is because:

- The argument looks like it won't work reliably, since the `PathProvider` enum will get loaded before the argument value is propagated to the system property
- This entire approach is obsolete anyway since there's no longer a separate Picard cloud jar, and we can just unconditionally set the requester pays project in `instanceMain()`

Issue #1827, opened 2022-09-09
BuildBamIndex creates index file in the current directory


regnveig:
Dear colleagues,

I discovered a thing which seems invalid. `BuildBamIndex` creates index file in the current directory, not in the directory which contains bam file. I'm absolutely sure that it behaved as I expected in previous versions. 

Example:

```bash
source ~/miniconda3/bin/activate gatk
cd /some/path
gatk BuildBamIndex -I /path/to/bam/my_file.bam
```

Expected output:

```bash
/path/to/bam/my_file.bai
```

Observed:

```bash
/some/path/my_file.bai
```

**OS:** Ubuntu 22.04
**GATK version:** 4.2.6.1
**Environment:** miniconda3

Could you check that, please?

cmnbroad:
This was originally reported in GATK, but the issue is in BuildBamIndex.

cmnbroad:
@regnveig To work around this for now, I think it will work correctly if you specify the full path of the expected output using `-O`.

regnveig:
@cmnbroad thank you for your suggestion! Looking forward to the fixed version!

kachulis:
will be fixed by #1852

rickymagner:
Looks like this was fixed by #1852.

cmnbroad:
I think there was a lot of confusion around all the PRs related to cloud-enabling. This *will* be fixed by cloud-enabling BuildBamIndex but I don't see any evidence that supports that having been done in #1852.

kockan:
@cmnbroad Is this intended to be part of #1804 ? If not, would it be a simple matter of updating the input/output paths from `File` to `PicardHtsPath` (and the corresponding path conversions in the `BuildBamIndex` source), mirroring how @takutosato seems to have been doing it? I could give it a try, or just wait until #1804 is merged if preferred.

cmnbroad:
@kockan Ah, this one has had a bit of a tortured lifecycle. `BuildBamIndex` already uses `PicardHtsPath` for inputs, but the outputs weren't done simultaneously, and they really should have been, because in some code paths the tool has to derive the output filename from the input filename.

[...]

PR #1838, opened 2022-10-14
Fix bai location when output not specified for BuildBamIndex

kachulis:
Addresses #1827

### DIFF
--- a/src/main/java/picard/sam/BuildBamIndex.java
+++ b/src/main/java/picard/sam/BuildBamIndex.java
@@ -88,7 +88,7 @@ public class BuildBamIndex extends CommandLineProgram {
         // set default output file - input-file.bai
         if (OUTPUT == null) {
 
-            final String baseFileName = inputPath.getFileName().toString();
+            final String baseFileName = INPUT.hasFileSystemProvider() ? inputPath.toAbsolutePath().toString() : inputPath.getFileName().toString();
 
             // only BAI indices can be created for now, although CSI indices can be read as well
             if (baseFileName.endsWith(FileExtensions.BAM)) {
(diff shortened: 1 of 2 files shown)


src/main/java/picard/sam/BuildBamIndex.java
```diff
@@ -88,7 +88,7 @@ protected int doWork() {
         // set default output file - input-file.bai
         if (OUTPUT == null) {
 
-            final String baseFileName = inputPath.getFileName().toString();
+            final String baseFileName = INPUT.hasFileSystemProvider() ? inputPath.toAbsolutePath().toString() : inputPath.getFileName().toString();
```
cmnbroad, line 91: The best fix for this would be to change `OUTPUT` to have the same type (`PicardHtsPath`) as `INPUT`, so that they have the same semantics. Keeping `OUTPUT` as `File` while `INPUT` is a `PicardHtsPath` will most likely cause problems with cloud inputs.

src/test/java/picard/sam/BuildBamIndexTest.java
```diff
@@ -41,6 +42,20 @@ public void testBuildBamIndexOK() throws IOException {
         Assert.assertEquals(FileUtils.readFileToByteArray(OUTPUT_INDEX_FILE),  FileUtils.readFileToByteArray(EXPECTED_BAI_FILE));
     }
 
+    @Test
+    public void testBuildBamIndexNoOutputSpecified() throws IOException {
+        final List<String> args = new ArrayList<>();
+        /* First sort, before indexing */
+        new SortSam().instanceMain(new String[]{
+                "I=" + INPUT_FILE,
+                "O=" + OUTPUT_SORTED_FILE,
```
cmnbroad, line 51: The test is called "NoOutputSpecified", but it specifies an output .

cmnbroad: The existing tests in this file are already a bit sketchy, since they write their outputs into the testdata directory. Its not clear if they're actually overwriting test data or not (I don't think they are, but it LOOKS like they are).

cmnbroad (commented):
@kachulis I have a branch with the fix I proposed below, and a test, but I haven't yet fixed the test issue I mentioned below with the existing tests (writing into the testdata directory). I'll finish that and get it submitted.

src/main/java/picard/sam/BuildBamIndex.java
kachulis: Awesome, I was hoping you might have a better solution to this :)

lbergelson:
@kachulis What happened to this branch?  Was it abandoned on purpose or by accident?

kachulis:
Closed on purpose because @cmnbroad said he had a better solution.  Not sure if that ever got implemented.

lbergelson:
We're closing this and adding it to the list of things to be cloudified.

PR #1839, opened 2022-10-28
Improvements to CompareMetrics

kachulis:
Adds the following additional features to CompareMetrics:

1. Command line parameter `METRICS_NOT_REQUIRED` to specify metrics which may be missing from either input.
2. Command line parameter `KEY` to specify a metric column to use as a key for metrics comparisons.  This is useful for checking if the only difference between two metrics files is the order of the rows.
3. Option to output a tsv table of metrics differences.  This is much easier to evaluate than the current text output.


meganshand (approved):
👍

nikellepetrillo (approved):
This looks good to me, thank you!

PR #1843, opened 2022-11-16
Fix Duplicate Set Index for queryname sorted input to MarkDuplicates

kachulis:
Duplicate Set index is broken for queryname sorted input.  This fixes, adds tests of this case, and also refactors some mark duplicates tests.

#### Content

#### Review


src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -400,23 +401,6 @@ protected int doWork() {
                     UmiUtil.setMolecularIdentifier(rec, "", MOLECULAR_IDENTIFIER_TAG, DUPLEX_UMI);
                 }
 
-                // Tag any read pair that was in a duplicate set with the duplicate set size and a representative read name
-                if (TAG_DUPLICATE_SET_MEMBERS) {
```
jamesemery, line 404: was this code really copy-pasted like this all along and nobody noticed? 😞

src/test/java/picard/sam/markduplicates/MarkDuplicatesTagRepresentativeReadIndexTest.java
```diff
@@ -36,69 +38,76 @@ protected MarkDuplicatesTagRepresentativeReadIndexTester getTester() {
...
         tester.expectedSetSizeMap.put(representativeReadName2,3);
-        tester.addMatePair("RUNID:1:1:15993:13365", 1, 485299, 485299, false, false, true, true, "44M1S", "44M1S", false, true, false, false, false, DEFAULT_BASE_QUALITY);
-        tester.expectedRepresentativeIndexMap.put(7, representativeReadIndexInFileForward2);
-        tester.expectedRepresentativeIndexMap.put(10, representativeReadIndexInFileForward2);
-        tester.expectedSetSizeMap.put("RUNID:1:1:15993:13365",3);
-        tester.addMatePair("RUNID:1:1:15993:13370", 1, 485299, 485299, false, false, true, true, "43M2S", "43M2S", false, true, false, false, false, DEFAULT_BASE_QUALITY);
-        tester.expectedRepresentativeIndexMap.put(8, representativeReadIndexInFileForward2);
-        tester.expectedRepresentativeIndexMap.put(11, representativeReadIndexInFileForward2);
-        tester.expectedSetSizeMap.put("RUNID:1:1:15993:13370",3);
+        final String duplicateSet2DuplicateReadName1 = "RUNID:1:1:15993:13365";
```
jamesemery, line 97: bleh is this the best of how our testing for this feature worked before? this seems very brittle... not really your problem to fix here though...

src/test/java/picard/sam/markduplicates/AbstractMarkDuplicatesCommandLineProgramTester.java
```diff
@@ -139,26 +145,44 @@ public void setExpectedOpticalDuplicate(final int expectedOpticalDuplicatePairs)
...
-    @Override
-    public void test() throws IOException {
-        testMetrics();
+
+    /**
+    * This method is called before iterating through output records in test method.
+    * Should be used to update any expectations to be used in implementation
+    * specific tests.
+     **/
+    void updateExpectationsHook() {
```
jamesemery, line 154: solid refactoring of the testing infrastructure.

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -384,13 +386,12 @@ protected int doWork() {
                     }
                     final boolean isInDuplicateSet = recordInFileIndex == nextRepresentativeIndex ||
                             (sortOrder == SAMFileHeader.SortOrder.queryname &&
-                                    recordInFileIndex > nextDuplicateIndex);
+                                    recordInFileIndex > nextRepresentativeIndex && rec.getReadName().equals(representativeQueryName));
```
jamesemery, line 389: can you rename this/comment the logic for "isInDuplicateSet" here to clairify this? This took a while to pick through what all of the steps are doing...

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
```diff
@@ -384,13 +386,12 @@ protected int doWork() {
...
                     if (isInDuplicateSet) {
                         if (!rec.isSecondaryOrSupplementary() && !rec.getReadUnmappedFlag()) {
-                            if (TAG_DUPLICATE_SET_MEMBERS) {
```
jamesemery, line 390: oh boy this code was not in a good state before you got to it...

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
jamesemery: maybe "Is RepresentativeReadForADuplicateSet"

jamesemery: So this essentially says that if we are in queryname sorted mode (It should be noted this won't work if we are in query grouped mode even though the logic should be the same, maybe that check should be updated), we mark both the first read and its mates with the duplicate sets but if we are coordinate sorted the mates don't get the flag? Is that asymmetrical behavior desirable?

jamesemery (commented):
Mostly looks good and I don't have very much to say except to comment that the logic for this code is very confusing and needs to be better documented/cleaned up to be at all comprehensible especially since there were two very obvious mistakes in plane site that you cleaned up over a 20 line span...

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
kachulis: Yeah this is super confusing, I actually couldn't figure out how this even works just now until I realized the variables are horribly named.

I believe that mates should also be marked with duplicate sets when the input is coordinate sorted as well (which is confirmed by our tests).  The reason is that in `addRepresentativeReadIndex` we add an entry for both reads of the fragment if `end.read1IndexInFile != end.read2IndexInFile`.  And when the ReadEnds object gets built we keep those entries the same for queryname sorted, but different for coordinate sorted (see L548 and thereabouts).  I think this really goes back to your overall point that this code has become mangled to the point obfuscation.   Which I completely agree with, but don't know that anyone has the capacity to fix, unfortunately. 

Regardless, I will try to at least make this part a little clearer

src/main/java/picard/sam/markduplicates/MarkDuplicates.java
kachulis: yeah... that does appear to be the case

jamesemery (approved):
Thank you for the clarifying comments

PR #1857, opened 2023-02-08
Delete the now redundant ClassFinder

lbergelson:
* This has moved to both htsjdk and Barclay and is therefore unnecessary to have a copy of here.
* Downstream users should replace with org.broadinstitute.barclay.argparser.ClassFinder

The barclay classfinder is effectively identical and is more likely to be well maintained.  There is also a version in htsjdk which should probably be unified somehhow.  

See also https://github.com/broadinstitute/barclay/pull/193.


lbergelson:
I'm not sure what the picard deprecation policy is / should be?   This could be deprecated instead.

droazen (approved):
👍

PR #1858, opened 2023-02-08
Remove obsolete references to Google Genomics API

lbergelson:
* The Google Genomics API was turned off years ago. This removes a global argument (GA4GH_CLIENT_SECRETS) which was used with it and several initialization steps in CommandLineProgram which are no longer meaningful.


droazen (approved):
👍

PR #1859, opened 2023-02-10
Allow empty input files for CollectAlignmentSummaryMetrics and FastqToSam

kachulis:
Currently, if input files to the tools have 0 reads, various exceptions are thrown by CollectAlignmentSummaryMetrics and FastqToSam.  This PR adds an option to allow FastqToSam to produce an empty Sam when passed an empty fastq instead of throwing an exception, and does not try to plot readlength distribution in CollectAlignmentSummaryMetrics when there are 0 reads (which currently hits an r error).

### DIFF
--- a/src/main/java/picard/analysis/CollectAlignmentSummaryMetrics.java
+++ b/src/main/java/picard/analysis/CollectAlignmentSummaryMetrics.java
@@ -190,8 +190,6 @@ public class CollectAlignmentSummaryMetrics extends SinglePassSamProgram {
 
         file.write(OUTPUT);
 
-
-
         if (HISTOGRAM_FILE != null) {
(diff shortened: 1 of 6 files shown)


tmelman (changes requested):
I recommend writing a simple test for the expected behavior with empty files.

kachulis:
@tmelman I added a few tests.  Turns out there was a bug in this seemingly trivial change, so that was a good call!

tmelman (approved):
Glad the tests revealed the bug, tests and code look good to me, ok to merge.

My one last comment would be make sure you're logging warnings when you mean warnings, and not when you want it to throw an error. But if this is meant to tolerate empty histogram outputs, then this is fine by me.

PR #1860, opened 2023-02-17
Lb add cloud tests

lbergelson:
Ignore me while testing

### DIFF
--- a/.github/workflows/cloud_tests.yml
+++ b/.github/workflows/cloud_tests.yml
@@ -38,9 +38,6 @@ jobs:
         experimental: [ false ]
       fail-fast: false
     continue-on-error: ${{ matrix.experimental }}
-    env:
-      TEST_TYPE: ${{ matrix.testType }}
-      SCALA_VERSION: ${{ matrix.scalaVersion }}
     name: Java ${{ matrix.Java }}, Barclay=${{ matrix.run_barclay_tests}} cloud tests
     steps:
       - uses: actions/checkout@v3
@@ -64,21 +61,19 @@ jobs:
         uses: google-github-actions/auth@v0
         with:
           credentials_json: ${{ secrets.GCP_CREDENTIALS }}
-          project_id: ${{ env.HELLBENDER_TEST_PROJECT }}
+          project_id: ${{ env.PICARD_TEST_PROJECT }}
           create_credentials_file: true
-
-      - name: "export the credentials for GATK tests"
-        if: needs.check-secrets.outputs.google-credentials == 'true'
-        run: echo "HELLBENDER_JSON_SERVICE_ACCOUNT_KEY=${{ steps.auth.outputs.credentials_file_path }}" >> $GITHUB_ENV
 
       - name: 'Set up Cloud SDK'
         if: needs.check-secrets.outputs.google-credentials == 'true'
         uses: google-github-actions/setup-gcloud@v0
 
       - name: compile test code
+        if: needs.check-secrets.outputs.google-credentials == 'true'
         run: ./gradlew compileTestJava
 
       - name: Run tests
+        if: needs.check-secrets.outputs.google-credentials == 'true'
         env:
           TEST_TYPE: cloud
         run: |
@@ -89,3 +84,10 @@ jobs:
             echo "Running tests using the legacy Picard command line parser."
             ./gradlew jacocoTestReport
           fi
+
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: cloud-test-results-${{ matrix.Java }}-barclay-${{ matrix.run_barclay_tests}}
+          path: build/reports/tests
--- a/.github/workflows/cloud_tests.yml
+++ b/.github/workflows/cloud_tests.yml
@@ -1,0 +1,91 @@
+name: Cloud Tests
+on:
+  push:
+    branches:
+      - 'master'
+  pull_request:
+  workflow_dispatch:
+
+env:
+  PICARD_TEST_INPUTS: gs://hellbender/test/resources/
+  PICARD_TEST_STAGING: gs://hellbender-test-logs/staging/
+  PICARD_TEST_LOGS: /hellbender-test-logs/build_reports/
+  PICARD_TEST_PROJECT: broad-dsde-dev
+
+jobs:
+  ## This workaround is necessary since there is no equivalent to the old TRAVIS_SECURE_ENVIRONMENT variable that indicated
+  ## if a run was privileged and had secrets. Since the GCP credentials are necessary for all tests in order to upload their,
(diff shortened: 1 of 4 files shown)


cmnbroad:
@lbergelson ping.

lbergelson:
@cmnbroad This is ready to take a look at if you have a chance.

.github/workflows/cloud_tests.yml
```diff
@@ -0,0 +1,91 @@
...
+        if: needs.check-secrets.outputs.google-credentials == 'true'
+        uses: google-github-actions/auth@v0
+        with:
+          credentials_json: ${{ secrets.GCP_CREDENTIALS }}
+          project_id: ${{ env.HELLBENDER_TEST_PROJECT }}
+          create_credentials_file: true
+
+      - name: "export the credentials for GATK tests"
+        if: needs.check-secrets.outputs.google-credentials == 'true'
+        run: echo "HELLBENDER_JSON_SERVICE_ACCOUNT_KEY=${{ steps.auth.outputs.credentials_file_path }}" >> $GITHUB_ENV
```
droazen, line 72: HELLBENDER or PICARD? Where does this environment variable actually get used?

droazen (approved):
@lbergelson Back to you with a few minor comments -- go ahead and merge after addressing them and tests pass (unless you also want feedback from @cmnbroad , of course)

.github/workflows/cloud_tests.yml
```diff
@@ -0,0 +1,91 @@
...
+    strategy:
+      matrix:
+        java: [ 17 ]
+        run_barclay_tests: [true, false]
+        experimental: [ false ]
+      fail-fast: false
+    continue-on-error: ${{ matrix.experimental }}
+    env:
+      TEST_TYPE: ${{ matrix.testType }}
+      SCALA_VERSION: ${{ matrix.scalaVersion }}
```
cmnbroad, line 43: Is this used here ?

.github/workflows/cloud_tests.yml
```diff
@@ -0,0 +1,91 @@
...
+  push:
+    branches:
+      - 'master'
+  pull_request:
+  workflow_dispatch:
+
+env:
+  PICARD_TEST_INPUTS: gs://hellbender/test/resources/
+  PICARD_TEST_STAGING: gs://hellbender-test-logs/staging/
+  PICARD_TEST_LOGS: /hellbender-test-logs/build_reports/
```
cmnbroad, line 12: Should we have separate locations for picard ?

src/test/java/picard/util/GCloudTestUtils.java
```diff
@@ -0,0 +1,61 @@
...
+     * as {@link #getTestProject()} or the tests for requester pays access may be invalid.
+     *
+     * @return PICARD_REQUESTER_PAYS_BUCKET env. var if defined, {@value GCloudTestUtils#REQUESTER_PAYS_BUCKET_DEFAULT}.
+     */
+    public static String getRequesterPaysBucket() {
+        return getEnvironmentVariable("PICARD_REQUESTER_PAYS_BUCKET", REQUESTER_PAYS_BUCKET_DEFAULT);
+    }
+
+    private static String getEnvironmentVariable(final String variableName, final String defaultValue) {
+        final String valueFromEnvironment = System.getProperty(variableName);
```
cmnbroad, line 27: This is probably copied from somewhere else, but its  always find it a bit confusing to have a method named `getEnvironmentVariable` that actually gets a JVM property.

.github/workflows/cloud_tests.yml
lbergelson: I thought I got rid of that...

src/test/java/picard/util/GCloudTestUtils.java
cmnbroad: Actually, are those values actually env variables, or properties ?

cmnbroad (commented):
Just a couple of nits about naming, and one question about property vs env variable.

.github/workflows/cloud_tests.yml
lbergelson: We can change them if we want.  Since we're using the same service account and set of permissions I figured it was easiest to put them in the same place where permissions and bucket lifecycles are already configured.  They shouldn't conflict with each other in any way that I can think of.

.github/workflows/cloud_tests.yml
```diff
@@ -0,0 +1,91 @@
...
+          ./gradlew compileJava 
+          ./gradlew installDist
+
+      #Google Cloud stuff
+      - id: 'gcloud-auth'
+        if: needs.check-secrets.outputs.google-credentials == 'true'
+        uses: google-github-actions/auth@v0
+        with:
+          credentials_json: ${{ secrets.GCP_CREDENTIALS }}
+          project_id: ${{ env.HELLBENDER_TEST_PROJECT }}
```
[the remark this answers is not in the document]
lbergelson: Yes it should be.  I guess the project ID is redundant in this case since the service account is tied to a single project?  Slightly mysteriouos.

.github/workflows/cloud_tests.yml
lbergelson: That is a mysterious invocation that was copied in from GATK but on further inspection is probably related to some tests we don't /wont do in picard.

src/test/java/picard/util/GCloudTestUtils.java
```diff
@@ -0,0 +1,61 @@
...
+public final class GCloudTestUtils {
+    /**
+     *  This is a public requester pays bucket owned by the broad-gatk-test project.
+     *  It must be owned by a different project than the service account doing the testing or the test may fail because it can access the
+     *  file directly through alternative permissions.
+     */
+    public static final String REQUESTER_PAYS_BUCKET_DEFAULT = "gs://hellbender-requester-pays-test/";
+
+    public static final String TEST_INPUTS_DEFAULT = "gs://hellbender/test/resources/";
+    public static final String TEST_STAGING_DEFAULT = "gs://hellbender-test-logs/staging/";
```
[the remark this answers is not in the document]
lbergelson: It should not cause issues unless we decide to blow up the hellbender ones at some point in the future.  It seemed like the sanest thing to do since the various permissions and buckets are configured correctly and I'm not 100% sure I remember what all the settings had to be.

src/test/java/picard/util/GCloudTestUtils.java
lbergelson: That is confusing.  I think that's how it was named originally unless I mangled it more here.  Java exposes all the environment variables as system properties, so in this case it's both I guess.

cmnbroad: I didn't realize Java did that, but it explains a lot of confusion I have previously encountered.

Issue #1862, opened 2023-02-23
CollectAlignmentSummaryMetrics fails on some small inputs due to unsafe Histogram methods


rickymagner:
### Affected tool(s)
CollectAlignmentSummaryMetrics, and maybe others which use the Histogram class from htsjdk.

### Affected version(s)
- [X] Latest public release version [3.0.0]

### Description 
When running this tool on a small bam without good alignments (but properly formatted), the tool throws a `NullPointerException`. A more useful error message would be appreciated to help users debug what went wrong.

#### Steps to reproduce
Save the following sam file to `null.sam`:
```
@HD	VN:1.6	GO:none	SO:queryname
@RG	ID:test	SM:test_sample	LB:test_lib	PL:ILLUMINA	PU:my_platform	CN:BI	DT:2023-01-01T00:00:00-0500	DS:description
aln1	516	*	0	0	*	*	0	0	AGCTACTG	99999999	RG:Z:test
aln2	516	*	0	0	*	*	0	0	GTCAGTCA	99999999	RG:Z:test
aln3	516	*	0	0	*	*	0	0	CCTTGGAA	99999999	RG:Z:test
```
Then run
```
picard CollectAlignmentSummaryMetrics -I null.sam -R <some_ref>.fasta -O test.metrics
```
to see the error:
```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.util.Map$Entry.getValue()" because the return value of "java.util.NavigableMap.firstEntry()" is null
	at htsjdk.samtools.util.Histogram.getMin(Histogram.java:438)
	at picard.analysis.AlignmentSummaryMetricsCollector$IndividualAlignmentSummaryMetricsCollector.finish(AlignmentSummaryMetricsCollector.java:271)
	at picard.analysis.AlignmentSummaryMetricsCollector$GroupAlignmentSummaryMetricsPerUnitMetricCollector.finish(AlignmentSummaryMetricsCollector.java:185)
	at picard.metrics.MultiLevelCollector$AllReadsDistributor.finish(MultiLevelCollector.java:208)
	at picard.metrics.MultiLevelCollector.finish(MultiLevelCollector.java:324)
	at picard.analysis.CollectAlignmentSummaryMetrics.finish(CollectAlignmentSummaryMetrics.java:176)
	at picard.analysis.SinglePassSamProgram.makeItSo(SinglePassSamProgram.java:177)
[...]

PR #1871, opened 2023-04-13
Fix crash on PF_READS == 0 for AlignmentSummaryMetricsCollector

kockan:
### Description
----
Create a (mostly) empty metrics file instead of giving a stacktrace if no reads pass Illumina filters.

Related to the following issue: https://github.com/broadinstitute/picard/issues/1862
----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/analysis/AlignmentSummaryMetricsCollector.java
+++ b/src/main/java/picard/analysis/AlignmentSummaryMetricsCollector.java
@@ -261,7 +261,7 @@ public class AlignmentSummaryMetricsCollector extends SAMRecordAndReferenceMulti
         @Override
         public void finish() {
             //summarize read data
-            if (metrics.TOTAL_READS > 0) {
+            if (metrics.TOTAL_READS > 0 && metrics.PF_READS > 0) {
                 metrics.PCT_PF_READS = (double) metrics.PF_READS / (double) metrics.TOTAL_READS;
                 metrics.PCT_ADAPTER = adapterReads / (double) metrics.PF_READS;
                 metrics.MEAN_READ_LENGTH = readLengthHistogram.getMean();
--- a/src/main/java/picard/analysis/AlignmentSummaryMetricsCollector.java
+++ b/src/main/java/picard/analysis/AlignmentSummaryMetricsCollector.java
@@ -40,6 +40,7 @@ import htsjdk.samtools.reference.ReferenceSequence;
 import htsjdk.samtools.util.CoordMath;
 import htsjdk.samtools.util.Histogram;
 import htsjdk.samtools.util.SequenceUtil;
+import picard.PicardException;
 import picard.metrics.PerUnitMetricCollector;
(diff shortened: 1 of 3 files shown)


src/main/java/picard/analysis/AlignmentSummaryMetricsCollector.java
```diff
@@ -261,7 +261,7 @@ public void acceptRecord(final SAMRecordAndReference samRecordAndReference) {
         @Override
         public void finish() {
             //summarize read data
-            if (metrics.TOTAL_READS > 0) {
+            if (metrics.TOTAL_READS > 0 && metrics.PF_READS > 0) {
```
kachulis, line 264: Isn't `metrics.TOTAL_READS > 0` now redundant?

kachulis (changes requested):
Thanks @kockan this looks reasonable.  Could you just add a test demonstrating the fix?

src/main/java/picard/analysis/AlignmentSummaryMetricsCollector.java
kockan: Redundant check now removed. Test case added.

kachulis (approved):
This looks good to me, thanks @kockan!

src/main/java/picard/analysis/AlignmentSummaryMetricsCollector.java
```diff
@@ -261,7 +261,7 @@ public void acceptRecord(final SAMRecordAndReference samRecordAndReference) {
         @Override
         public void finish() {
             //summarize read data
-            if (metrics.TOTAL_READS > 0) {
+            if (metrics.PF_READS > 0) {
```
rickymagner, line 264: I think it might be a good idea to add an `else` statement here, that prints a warning if the number of `PF_READS` is 0, just in case someone actually reads the logs and is wondering why their data looks weird. I think something that explains the possibilities - either lack of reads or they were all QC'd out - might help users debugging. There are a number of other tools that use a logger to print warnings you could use as a template. (Though I'm not sure what the standard is for "where" loggers get invoked. Maybe down inside a metrics collector is too low level? I'd check the tool that calls this to see what it does, if anything.)

src/test/java/picard/analysis/CollectAlignmentSummaryMetricsTest.java
```diff
@@ -768,4 +768,37 @@ public void testReadLengthHistogram(final boolean plotChart) throws IOException
             }
         }
     }
+
+    @Test
+    public void testNoPFReads() throws IOException {
```
rickymagner, line 773: In this case, do you want an IOException if there are no PF reads? I guess it's OK to have an exception in this case, since it seems pointless to try to collect metrics on a file with no passing alignments, but maybe make it a `PicardException` with a useful error message (this could be done instead of my warning comment before).

rickymagner (changes requested):

Issue #1814, opened 2022-06-12
MD5SUM from picard and samtools do not match


fgvieira:
### Instructions

- Use a **concise** yet **descriptive** title;
- Determine whether your issue is a **bug report**, a **feature request**, or a **documentation request**;
- Choose the corresponding template block below and fill it in, replacing or deleting text in italics (surrounded by `_`) as appropriate;
- Delete the other template blocks and this header.

----

### Affected tool(s)
`CreateSequenceDictionary`

### Affected version(s)
- [x] Latest public release version [2.27.2]
- [ ] Latest development/master branch as of [date of test?]

### Description 

When creating a dictionary with  `samtools dict`, I get:
```
$ samtools dict Oegenome10scaffoldC3G.fasta
@HD     VN:1.0  SO:unsorted
@SQ     SN:scaffold1    LN:114512629    M5:f6412f880b27671e3789d5836f5803f1     UR:file:///Oegenome10scaffoldC3G.fasta
@SQ     SN:scaffold2    LN:111878775    M5:67b1df915e157c1b1486862a810eb4f6     UR:file:///Oegenome10scaffoldC3G.fasta
@SQ     SN:scaffold3    LN:97763355     M5:3d1af6b898da3f3c017bf0a6a105ccc7     UR:file:///Oegenome10scaffoldC3G.fasta
@SQ     SN:scaffold4    LN:96428841     M5:23694b6ad3ef81aa4f891e5c7bee80e2     UR:file:///Oegenome10scaffoldC3G.fasta
@SQ     SN:scaffold5    LN:96132218     M5:1a846b8a41a7d622c9af78465232c99f     UR:file:///Oegenome10scaffoldC3G.fasta
@SQ     SN:scaffold6    LN:95771753     M5:310f1513181b44d338486d331271f5aa     UR:file:///Oegenome10scaffoldC3G.fasta
@SQ     SN:scaffold7    LN:83323684     M5:08e47a0f3d60a7abcb436663fc3d52dc     UR:file:///Oegenome10scaffoldC3G.fasta
@SQ     SN:scaffold8    LN:58502465     M5:546e6a679d3e736713a3b383ab312723     UR:file:///Oegenome10scaffoldC3G.fasta
@SQ     SN:scaffold9    LN:52333144     M5:539d8ed1fcbf4a67bd26780a1ea0dc5b     UR:file:///Oegenome10scaffoldC3G.fasta
[...]

PR #1884, opened 2023-05-24
Add tests to verify sequence dictionary MD5s respect dos line endings.

cmnbroad:
Add tests to verify the issue reported in https://github.com/broadinstitute/picard/issues/1814. But, neither I nor  Chris K. can reproduce the issue - it looks like we correctly respect dos line endings, and are concordant with samtools either way. Since I've already done the work, we might well add these tests.


src/test/java/picard/sam/CreateSequenceDictionaryTest.java
```diff
@@ -212,4 +212,71 @@ public void testPunctuationInReferenceNames() {
             Assert.assertTrue(anSet.contains("2/hello"));
         }
     }
+
+    // Test that dictionary MD5s are invariant across dos/unix line endings
+    @Test
+    public void testSequenceMD5sRespectLineEndings() throws IOException {
+        // NOTE: the .fasta test files here that use windows cr/lf line endings have to be listed in
```
lbergelson, line 219: Thank you for this note.

lbergelson (approved):
@cmnbroad This seems convincing to me.  Thanks for investigating.

PR #1885, opened 2023-05-24
Enable IntelInflater

lbergelson:
Fixes a long standing bug where we were accidentally not using the intel inflator when decompressing even when it was requested.  Includes a horrible test using brittle reflection.


droazen (approved):
This is fine for now as a quick fix 👍  However, we should definitely patch HTSJDK to improve safety/consistency around updating the cached defaults so we don't run into this again at some point.

droazen:
@lbergelson Can you confirm that the test you added fails without your patch?

lbergelson:
@droazen Yes, I tested without my patch and the inflater test fails.  The deflater one passes because it was working before my patch.

Issue #1631, opened 2021-01-14
CollectVariantCallingMetrics throws NullPointerException


jezkoski:
### Tool 

I'm trying to run CollectVariantCallingMetrics on my dataset.
I have tried to run anaconda installation of picard-2.23.8 and also the latest picard from github (downloaded this month), but I have the same problem with both. (the installations are on different servers so they won't be mixed)

My command.
` java -jar ./picard/build/libs/picard.jar CollectVariantCallingMetrics \
--INPUT germline_variants.vcf.gz \
--OUTPUT germ_raw_picardmetrics \
--DBSNP /data/ref_genomes/GRCh37/dbSNP154_UCSCnames.hg19.vcf.gz`

 ### Error description

I get this error, I assume there's a problem with loading my dbSNP file?

`07:55:32.308 INFO  NativeLibraryLoader - Loading libgkl_compression.so from jar:file: /home/jkoski/picard/build/libs/picard.jar!/com/intel/gkl/native/libgkl_compression.so
[Thu Jan 14 07:55:32 UTC 2021] CollectVariantCallingMetrics INPUT=germline_variants.vcf.gz OUTPUT=germ_raw_picardmetrics DBSNP=/data/ref_genomes/GRCh37/dbSNP154_UCSCnames.hg19.vcf.gz    GVCF_INPUT=false THREAD_COUNT=1 VERBOSITY=INFO QUIET=false VALIDATION_STRINGENCY=STRICT COMPRESSION_LEVEL=5 MAX_RECORDS_IN_RAM=500000 CREATE_INDEX=false CREATE_MD5_FILE=false GA4GH_CLIENT_SECRETS=client_secrets.json USE_JDK_DEFLATER=false USE_JDK_INFLATER=false
[Thu Jan 14 07:55:32 UTC 2021] Executing as jkoski@a2 on Linux 3.10.0-1062.4.3.el7.x86_64 amd64; OpenJDK 64-Bit Server VM 1.8.0_232-b09; Deflater: Intel; Inflater: Intel; Provider GCS is not available; Picard version: 2.24.0-1-g74a85ab-SNAPSHOT`

The tool still goes through all the chromosomes and lists the process. Time for last 100,000 is 0s for all the INFO lines

`INFO    2021-01-14 09:05:17     CollectVariantCallingMetrics    Read     5,100,000 variants.  Elapsed time: 00:00:19s.  Time for last 100,000:    0s.  Last read position: chr1:20,452,846
[...]

PR #1889, opened 2023-06-10
DbSnpBitSetUtil should verify that the VCF dict should match the inpu…

nh13:
…t dict

Feel free to push directly to this, for example fixes and tests.  I make this PR for demonstration purposes for #1631

### DIFF
--- a/src/main/java/picard/util/DbSnpBitSetUtil.java
+++ b/src/main/java/picard/util/DbSnpBitSetUtil.java
@@ -216,7 +216,7 @@ public class DbSnpBitSetUtil {
         final VCFFileReader variantReader = new VCFFileReader(dbSnpFile.toPath(), intervals != null);
         
         SequenceDictionaryUtils.assertSequenceDictionariesEqual(
-            SAMSequenceDictionary,
+            sequenceDictionary,
             "Reference genome",
             variantReader.getFileHeader().getSequenceDictionary(),
             variantReader.getRawInputString()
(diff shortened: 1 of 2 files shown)


kockan:
Hi @nh13 , I made some small changes for this patch to successfully pass the checks and added some basic tests for the pass/fail cases (hg19 vs hg19 and hg19 vs hg38). Does it seem reasonable to you?

src/main/java/picard/util/DbSnpBitSetUtil.java
```diff
@@ -213,6 +214,14 @@ private static void loadVcf(final PicardHtsPath dbSnpFile,
 
         final Optional<ProgressLogger> progress = log.map(l -> new ProgressLogger(l, (int) 1e5, "Read", "variants"));
         final VCFFileReader variantReader = new VCFFileReader(dbSnpFile.toPath(), intervals != null);
+        
+        SequenceDictionaryUtils.assertSequenceDictionariesEqual(
+            variantReader.getFileHeader().getSequenceDictionary(),
+            "DBSNP_REFERENCE_VCF",
+            sequenceDictionary,
+            "INPUT_VCF"
```
nh13, line 222: I think this is fine to list the command line parameter names in an internal function

nh13:
lgtm

src/main/java/picard/util/DbSnpBitSetUtil.java
kockan: Done.

kockan (approved):
Ready to go.

Issue #1854, opened 2023-02-01
Order of rows in crosscheck metrics files are non-deterministic


kachulis:
The order of rows output by CrosscheckFingerprints seems to change from run to run on the same input data.  Not sure if this is new behavior or has always been the case.  

This isn't a massive problem, but would be nice for the order to always be the same.

lbergelson:
As always I suspect HashMap/Set

serge2016:
Is it possible to fix it?

droazen:
Reopening, as this was not completely fixed by https://github.com/broadinstitute/picard/pull/1892

droazen:
Remaining TODOs (courtesy of @lbergelson ):

```
I looked at this a bit more, it seems like there is instability built in at the start since FingerPrintChecker.fingerprintFiles() builds an unsorted map from the output of a parallel and unpredictable process. I think it's going to require a more complicated change than just changing the other sets to linked sets. Either change the inputs to return their results in order, or sort the maps according to some criteria.

@lbergelson
Member
[lbergelson](https://github.com/lbergelson) commented [on Oct 10](https://github.com/broadinstitute/picard/pull/1892#issuecomment-1755914482)
I think we either need to change it so the input maps are built as the result of Futures instead of being added to the concurrent map as the are built. The errors are actually handled this way so it wouldn't be that hard to change.

The other option would be to pick an arbitrary sort order and sort the input maps after they're built and then maintain that order.
```

PR #1892, opened 2023-06-13
Changed CrosscheckFingerprints hashing to linked hashing

tmelman:
Issue: CrosscheckFingerprints produces results in a different order each time. 

Change: Switched from HashMap and HashSet to Java's LinkedHashMap and LinkedHashSet, respectively

Intended result: Produce stable sorted output.

Issue: https://github.com/broadinstitute/picard/issues/1854

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/fingerprint/CrosscheckFingerprints.java
+++ b/src/main/java/picard/fingerprint/CrosscheckFingerprints.java
@@ -54,9 +54,8 @@ import java.nio.file.Path;
 import java.text.NumberFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


serge2016:
Great!

lbergelson:
@tmelman As a heads up, `Collectors.toSet()` is cursed by HashSet, so you may have to replace instances of stream collection with `Collectors.toCollection(LinkedHashSet::new))` or continue to suffer the horrors in non-deterministic behavior.

lbergelson:
I looked at this a bit more, it seems like there is instability built in at the start since FingerPrintChecker.fingerprintFiles() builds an unsorted map from the output of a parallel and unpredictable process.   I think it's going to require a more complicated change than just changing the other sets to linked sets.  Either change the inputs to return their results in order, or sort the maps according to some criteria.

lbergelson:
I think we either need to change it so the input maps are built as the result of Futures instead of being added to the concurrent map as the are built.  The errors are actually handled this way so it wouldn't be that hard to change.

The other option would be to pick an arbitrary sort order and sort the input maps after they're built and then maintain that order.

droazen (approved):

PR #1900, opened 2023-07-14
make cloud jar mandatory

lbergelson:
Remove the cloud jar and make google cloud dependencies part of the standard release.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,17 +58,6 @@ def ensureBuildPrerequisites(buildPrerequisitesMessage) {
 ensureBuildPrerequisites(buildPrerequisitesMessage)
 
 final htsjdkVersion = System.getProperty('htsjdk.version', '3.0.1')
-final googleNio = 'com.google.cloud:google-cloud-nio:0.123.25'
-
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -188,13 +173,6 @@ shadowJar {
     finalizedBy currentJar
 }
 
-
-task cloudJar(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar)  {
```
droazen, line 192: Should we temporarily keep the "cloudJar" task around as a direct alias for the normal build, to avoid breaking downstream users' build processes? Or is it better to force awareness of this change by removing the old build target?

droazen (approved):
@lbergelson Looks good -- back to you with one question

build.gradle
lbergelson: I'm in favor of removing it so they know something happened.

lbergelson: Especially since I don't want to make a dummy cloud.jar which will be identical to the normal picard jar, so their builds would break anyway.

Issue #1767, opened 2021-12-22
HsMetrics "coverage cap" parameter affects more than theoretical sensitivity calculations


eboyden:
### Affected tool(s)
CollectHsMetrics

### Affected version(s)
at least through 2.26.6

### Description 
The description of the `--COVERAGE_CAP` parameter indicates that it's used for Theoretical Sensitivity calculations, with a default value of 200.  But it actually affects several other metrics; these include Median coverage and Fold-80 (because the 20th percentile is also apparently capped), but not Mean coverage or Max coverage (not sure about Min coverage).

This is related to these issues:
https://github.com/broadinstitute/picard/issues/938
https://github.com/broadinstitute/picard/issues/1361
as well as this GATK inquiry:
https://gatk.broadinstitute.org/hc/en-us/community/posts/360078460592-Is-there-a-max-value-for-Picard-CollectHsMetrics-MEDIAN-TARGET-COVERAGE-

#### Steps to reproduce
Run any dataset with any coverage greater than COVERAGE_CAP.  The calculated Median coverage and Fold-80 values will be wrong, possibly severely.  The documentation of CollectHsMetrics generally, and this parameter specifically, do not indicate why these values would be wrong.

#### Expected behavior
Ideally, Median coverage and Fold-80 (and any other affected metrics) should be calculated from the actual data, not coverage-capped data; the coverage cap should only be used for the theoretical sensitivity documentation, as documented.  Alternatively, ALL metrics should be affected, including Mean and Max coverage (which may often result in a Fold-80 <1 because the Mean will be < COVERAGE_CAP but the 20th percentile may = COVERAGE_CAP).  Either way, the documentation of this parameter should be updated to reflect how the coverage cap is actually implemented and which metrics it affects.

#### Actual behavior
When real coverage exceeds the coverage cap, the Median coverage is capped but the Mean (and Max) are not, which is confusing.  Fold-80 is also typically (and confusingly) much too high, because the 20th percentile is also capped, so a real Mean coverage is divided by an artificially capped 20th percentile.

eboyden:
BTW I didn't check but I wouldn't be surprised if this also affects CollectTargetedPcrMetrics.

eboyden:
[...]

PR #1913, opened 2023-09-05
Fix CollectHSMetrics - Don't use Coverage Cap on fold score

JoeVieira:
This PR resolves issue #1767 

Per documentation COVERAGE CAP was intended to only apply to Theoretical Sensitivity calculations. But it was applied to MEDIAN_COVERAGE, which then applied it to all statistics which that is used for.
https://gatk.broadinstitute.org/hc/en-us/articles/360036856051-CollectHsMetrics-Picard-#--COVERAGE_CAP

Changes with test coverage are 
* Update Histogram directly, rather than via an intermediate data structure
* Which uses the full density passed, rather than limiting to the CAP
* Test FOLD_80 score when passing a VERY low cap. 
* Coverage MEAN, MEDIAN & FOLD should still indicate perfect coverage.

### DIFF
--- a/src/main/java/picard/analysis/directed/TargetMetricsCollector.java
+++ b/src/main/java/picard/analysis/directed/TargetMetricsCollector.java
@@ -67,6 +67,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.ArrayList;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.LongStream;
@@ -125,6 +126,10 @@ public abstract class TargetMetricsCollector<METRIC_TYPE extends MultilevelMetri
     // histogram of depths. does not include bases with quality less than MINIMUM_BASE_QUALITY (default 20)
     // give it the bin label "coverage_or_base_quality" to make clear that in the metrics file the coverage and base quality histograms share the same bin column on the left
     private final Histogram<Integer> highQualityDepthHistogram = new Histogram<>("coverage_or_base_quality", "high_quality_coverage_count");
+
+    // histogram of depths with no coverage cap imposed.
+    private final Histogram<Integer> fullHighQualityDepthHistogram = new Histogram<>("coverage_or_base_quality", "high_quality_coverage_count");
+
 
     // histogram of base qualities. includes all but quality 2 bases. we use this histogram to calculate theoretical het sensitivity.
     private final Histogram<Integer> unfilteredDepthHistogram = new Histogram<>("coverage_or_base_quality", "unfiltered_coverage_count");
@@ -699,7 +704,9 @@ public abstract class TargetMetricsCollector<METRIC_TYPE extends MultilevelMetri
 
         /** Calculates how much additional sequencing is needed to raise 80% of bases to the mean for the lane. */
         private void calculateTargetCoverageMetrics() {
-            final long[] highQualityCoverageHistogramArray = new long[coverageCap+1];
+            // use ArrayList because the length of this is defined by the coverage depth.
+            final ArrayList<Long> fullHighQualityDepthHistogramList =  new ArrayList<>();
+
             int zeroCoverageTargets = 0;
 
             // the number of bases we counted towards the depth histogram plus those that got thrown out by the coverage cap
@@ -723,7 +730,7 @@ public abstract class TargetMetricsCollector<METRIC_TYPE extends MultilevelMetri
             for (final Coverage c : this.highQualityCoverageByTarget.values()) {
                 if (!c.hasCoverage()) {
                     zeroCoverageTargets++;
-                    highQualityCoverageHistogramArray[0] += c.interval.length();
+                    fullHighQualityDepthHistogramList.set(0, (long) c.interval.length());
                     targetBases[0] += c.interval.length();
                     minDepth = 0;
                     continue;
@@ -731,7 +738,9 @@ public abstract class TargetMetricsCollector<METRIC_TYPE extends MultilevelMetri
 
                 for (final int depth : c.getDepths()) {
                     totalCoverage += depth;
-                    highQualityCoverageHistogramArray[Math.min(depth, coverageCap)]++;
+                    Long index = fullHighQualityDepthHistogramList.get(depth);
+                    index = index + 1;
+                    fullHighQualityDepthHistogramList.set(depth, index);
                     maxDepth = Math.max(maxDepth, depth);
                     minDepth = Math.min(minDepth, depth);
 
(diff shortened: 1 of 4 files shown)


JoeVieira:
@gbggrant Not sure if you can review here, or you could flag who might be best, thanks.

kockan:
Hi @JoeVieira , thanks for this PR! I don't believe the user you've tagged is working on Picard right now so we've asked for a review from someone else.

JoeVieira:
@kockan appreciated.

src/test/java/picard/analysis/directed/CollectHsMetricsTest.java
```diff
@@ -217,6 +217,41 @@ public void testHsMetricsHandlesIndelsAppropriately() throws IOException {
         Assert.assertEquals(insWithIndelHandling.PCT_USABLE_BASES_ON_TARGET,   1.0d);      // inserted bases are counted as on target
     }
 
+    @Test
+    public void testHsMetricsF80DoesNotUseCovCap() throws IOException {
+        final SAMRecordSetBuilder highCoverage = new SAMRecordSetBuilder(true, SortOrder.coordinate);
```
kachulis, line 222: given the code duplication between this, `testHsMetricsHandlesIndelsAppropriately` and `testHsMetricsHighTargetCoverage`, feels like there a prime refactoring opportunity here.

kachulis (changes requested):
Thanks for this fix @JoeVieira!  Just a few small questions/comments.  

Also, have you checked how speed and memory usage are affected by this change?  presumably they aren't significant, but since you're replacing an array being indexed into with a map just would be good to confirm that there isn't a massive performance hit.

src/main/java/picard/analysis/directed/TargetMetricsCollector.java
```diff
@@ -747,11 +751,10 @@ private void calculateTargetCoverageMetrics() {
                 throw new PicardException("the number of target bases with at least 0x coverage does not equal the number of target bases");
             }
 
-            for (int i = 0; i < highQualityCoverageHistogramArray.length; ++i) {
-                highQualityDepthHistogram.increment(i, highQualityCoverageHistogramArray[i]);
+            for (Map.Entry<Integer, Long> entry : fullHighQualityDepthHistogramMap.entrySet()) {
+                highQualityDepthHistogram.increment(entry.getKey(), entry.getValue());
```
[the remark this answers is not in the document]
JoeVieira: That's a SPECTACULAR point - honestly, I was just following the pattern that was there. I'll rework this.

src/test/java/picard/analysis/directed/CollectHsMetricsTest.java
JoeVieira: Happy to do this if you feel strongly, but I'm not sure how much more maintainable the code would be for extracting a couple lines into a common method, to be honest.

JoeVieira:
@kachulis updated based on your feedback. I also shared your concern previously about additional memory usage - but with this change, it's only possible to reduce memory usage & time.

src/test/java/picard/analysis/directed/CollectHsMetricsTest.java
kachulis: that's fair, I'm fine leaving as is.

kachulis:
I agree about the memory usage now, but I don't think it's guaranteed this will only ever lead to a speedup.  We are replacing 10's of millions of `long[]` operations  and 100's of `Histogram` operations with 10's of millions of `Histogram` operations.  If the `Histogram` operations are very slow, this could result in a slowdown.  I think the risk is quite low, but I just want to make sure this is run on some normal sized data and nothing crazy happens.

JoeVieira:
This isn't a perfect comparison, since the histogram now has ~3x as many bins in it as the bugged current version. But at least it shows that it doesn't totally blow performance up. 

For a bam with 539688 reads:

PR:
`[Thu Sep 14 15:22:40 EDT 2023] picard.analysis.directed.CollectHsMetrics done. Elapsed time: 0.21 minutes.
Runtime.totalMemory()=1090519040
java -jar picard.jar CollectHsMetrics -I S2.disarmed.sorted.bam -TI  -BI  -O   14.89s user 0.27s system 116% cpu 13.003 total
`
Current: 
`[Thu Sep 14 15:23:06 EDT 2023] picard.analysis.directed.CollectHsMetrics done. Elapsed time: 0.20 minutes.
Runtime.totalMemory()=1090519040
java -jar main.jar CollectHsMetrics -I S2.disarmed.sorted.bam -TI  -BI  -O   15.21s user 0.28s system 122% cpu 12.647 total
`

JoeVieira:
@kachulis Anything else before approving?

kachulis:
Hi @JoeVieira, sorry for the delayed response.

I think the performance test is reasonably convincing, but your comment about there being 3x as many bins in the output histogram lead me to think about some other issues as well.  Specifically, writing out the full histogram (instead of the capped histogram) is a change from previous behavior, even compared to before the initial bug was introduced.

Looking into this a bit more, there are a two issues I'm wrestling with.  The first is that as currently written this PR will introduce a behavior difference between  `highQualityDepthHistogram`, which will not be capped, and `unfilteredDepthHistogram`, which will be capped.  I think the behavior for both histograms should be the same when written out.  However, I will note that `unfilteredDepthHistogram` is used in the TheoreticalSensitivity calculation, so naively uncapping it would uncap that calculation as well.

The other issue is that (and this is a long standing issue, not something introduced in this PR), there seems to be some differences between how COVERAGE_CAP is utilized in `CollectWgsMetrics` vs in `CollectHsMetrics` (and other targeted metrics).  In wgs, the metrics are truly calculated in a capped fashion, and this is specified well in the documentation.https://github.com/broadinstitute/picard/blob/a9194bd978a9169df451139d74f2632a2a9c3f7b/src/main/java/picard/analysis/CollectWgsMetrics.java#L117)

However, for targeted metrics, the situation seems a bit messier.  It looks like, historically, target metrics were initially uncapped, and then some (such and median target coverage and the fold_x metrics) but not all became capped, and this PR would then uncap them again.  I _think_ this different behavior between wgs and targeted is reasonable, since for wgs very high coverage spikes are likely artifactual, while for targeted sequencing very high coverage is often the entire point (@yfarjoun do you remember if this was the intention?).  It is a bit problematic, IMO, that the same parameter name (`COVERAGE_CAP`) is used in different ways for similar tools, but changing a parameter name that has been in use for years is rough, and anyway that's not the responsibility of your PR.

After all that, my current feeling is we have three different places in targeted metrics that are relevant to your PR, where we need to decide whether the calculation should be coverage capped or not:
1) Theoretical Sensitivity 
2) Other coverage metrics (MEAN/MEDIAN_TARGET_COVERAGE, FOLD_80_BASE_PENALTY, etc)
3) histograms as written to metrics files

currently, 1 and 3 are capped, while 2 is a mix of capped and uncapped (although originally was all uncapped).  I think from the perspective of this PR, this best option is to leave 1 capped, make 2 uncapped, and then take your pick on 3 (I think either choice is reasonable).  I think this means you just need to align the behaviors of the two histograms, while simultaneously keeping 1 capped and 2 uncapped, and then I will feel comfortable merging this.

@yfarjoun @takutosato do either of you have any thoughts about the COVERAGE_CAP behavior in CollectHsMetrics?

JoeVieira:
@kachulis I totally agree with this - this region of the code really became tangled a few years back and keeping the output histograms uncapped seems the right position. 

Does anyone on this thread understand the intention of using the unfiltered data for the theoretical ( simply to get "all" data for simulation i presumed? ) 

I'll adjust this so we're tracking uncapped unfiltered data for output also.

JoeVieira:
I would propose 

1.) Theoretical can no be uncapped - as that defeats it's purpose.
2.) Each histogram should be outputted ( Uncapped HighQuality, Uncapped Unflitered ) 
3.) outputting the exact histograms used for the stat generation seems important as well & therefore we should also output a new Capped Unfiltered Histograph called "Theoretical Capped Unfiltered ... " 

I'm happy to update this PR to do so, if folks are in agreement.

JoeVieira:
@kachulis, @lbergelson Any thoughts here? I'm happy to update with the above proposed logic. I would love to wrap this up.

eboyden:
Bumping this - we're very eager for a fixed version of CollectHsMetrics that outputs correct median coverage and Fold-80 scores.

kockan:
Hi @eboyden , I'd like to get this merged as soon as possible as well. I was going through the previous comments and it seems like @kachulis was happy with the changes proposed by @JoeVieira , except the addition of a new capped unfiltered histogram. If @JoeVieira would like to make these changes I could ask for a quick re-review.

JoeVieira:
Hey all - I'll get this sorted & resubmitted. 

It'll be a bit bigger of a PR to make the histograms uncapped & also continue to support existing behaviours for the other areas which the unfiltered histograms get used. 

Best strategy I have to do that is pulling `calculateTheoreticalHetSensitivity` into the `HsMetricCollector` subclass so it can have different behaviour than the wgs metrics methods.

eboyden:
Appreciate everyone's help on this.  For the Picard developers: this would be outside the scope of this PR, but I suggest looking into whether CollectTargetedPcrMetrics or CollectWgsMetrics or any other Collect...Metrics tools are affected by this same bug.  Seems like at least CollectWgsMetrics might be affected, based on a brief investigation I did for the original bug report https://github.com/broadinstitute/picard/issues/1767

kockan:
Thanks a lot @eboyden ! If this is a bug that affects multiple metric collection tools, they should definitely be fixed as well. I'll relay your findings to the team and find out what can be done.

kockan:
@JoeVieira Thanks for the updates! Looks good to me, but I am not as familiar with the intricacies of this tool as @kachulis (who returns in about a week and a half), so just to be safe I'd like to wait for his final review before merging this.

meganshand (approved):
@kachulis is out on leave for a few more weeks so I took a look at this. I tried to follow the conversation and I think this PR does what was agreed upon in the comments, so 👍

Issue #1890, opened 2023-06-12
BedToIntervalList reportedly does not handle stdin properly


droazen:
We have a report from a user that `BedToIntervalList` produces an empty output when its input is supplied via `-I /dev/stdin` rather than `-I <bed_file>` (original post here: https://gatk.broadinstitute.org/hc/en-us/community/posts/16231098055835-BedToIntervalList-from-dev-stdin):

BedToIntervalList behaves differently depending on whether input comes from file or from /dev/stdin. Here is a simple example:

Create a BED file with a single entry:

```
$ echo -e "chr1\t10000\t20000" > test.bed
```

Convert this to an interval list:

```
$ picard BedToIntervalList -SD reference.dict -I test.bed -O foo.interval_list
```

Do it using input from stdin:

```
$ cat test.bed | picard BedToIntervalList -SD reference.dict -I /dev/stdin -O bar.interval_list
```

Now foo.interval_list is as expected, but bar.interval_list does not contain any intervals:

```
$ grep -v ^@ foo.interval_list | wc -l

1

$ grep -v ^@ bar.interval_list | wc -l

0
```

Is this the expected behavior? The output is not necessarily empty. I became aware of this when piping a much bigger input, and the output had fewer intervals than expected.

Log from first run (with file):

```
$ gatk BedToIntervalList -SD reference.dict -I test.bed -O foo.interval_list

Using GATK jar /home/michaelk/miniconda3/envs/test-gatk/share/gatk4-4.4.0.0-0/gatk-package-4.4.0.0-local.jar

Running:

    java -Dsamjdk.use_async_io_read_samtools=false -Dsamjdk.use_async_io_write_samtools=true -Dsamjdk.use_async_io_write_tribble=false -Dsamjdk.compression_level=2 -jar /home/michaelk/miniconda3/envs/test-gatk/share/gatk4-4.4.0.0-0/gatk-package-4.4.0.0-local.jar BedToIntervalList -SD reference.dict -I test.bed -O foo.interval_list

20:17:10.435 INFO  NativeLibraryLoader - Loading libgkl_compression.so from jar:file:/home/michaelk/miniconda3/envs/test-gatk/share/gatk4-4.4.0.0-0/gatk-package-4.4.0.0-local.jar!/com/intel/gkl/native/libgkl_compression.so

[...]

PR #1918, opened 2023-11-14
Reject piped input (/dev/stdin) for BedToIntervalList

kockan:
As the title states, reject piped input (/dev/stdin) for BedToIntervalList since it cannot process such input.

### DIFF
--- a/src/main/java/picard/util/BedToIntervalList.java
+++ b/src/main/java/picard/util/BedToIntervalList.java
@@ -24,6 +24,7 @@ import picard.PicardException;
 import picard.cmdline.CommandLineProgram;
 import picard.cmdline.StandardOptionDefinitions;
 import picard.cmdline.programgroups.IntervalsManipulationProgramGroup;
+import picard.nio.PicardHtsPath;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


kockan:
For #1890

src/main/java/picard/util/BedToIntervalList.java
```diff
@@ -116,6 +117,11 @@ protected int doWork() {
         IOUtil.assertFileIsReadable(INPUT);
         IOUtil.assertFileIsReadable(SEQUENCE_DICTIONARY);
         IOUtil.assertFileIsWritable(OUTPUT);
+
+        if(PicardHtsPath.isOther(new PicardHtsPath(INPUT))) {
```
droazen, line 121: Can you switch to checking explicitly for /dev/stdin, instead of relying on this problematic `isOther()` method?

droazen (changes requested):
One comment/request

src/main/java/picard/util/BedToIntervalList.java
kockan: Done

src/main/java/picard/util/BedToIntervalList.java
```diff
@@ -114,8 +114,13 @@ public class BedToIntervalList extends CommandLineProgram {
     @Override
     protected int doWork() {
         IOUtil.assertFileIsReadable(INPUT);
+        if(INPUT.getPath().equals("/dev/stdin")) {
+            throw new IllegalArgumentException("BedToIntervalList cannot read from /dev/stdin.");
```
cmnbroad, line 118: This is a bit of a nit, but this error message could be interpreted to mean that reading from the input is failing for some reason - I would suggest explicitly stating that reading from stdin is not supported:
```suggestion
            throw new IllegalArgumentException("BedToIntervalList does not support reading from standard input  - a file must be provided");
```

We should also add a test to verify this.

@kockan I'm happy to make these changes if you like - just let me know.

kockan: @cmnbroad Of course, please feel free to push to my branch

droazen (approved):

Issue #1834, opened 2022-10-06
Update request to SamToFastq documentation


GATKSupportTeam:
We would like the SamToFastq tool doc page to be more clear that this tool can take both coordinate and name sorted inputs. There should also be a description that the order of the fastq pairs will be different depending on your input sam sort order. This description is in the GATK forum thread.

This request was created from a contribution made by Jakub Savara on July 24, 2022 10:48 UTC.

Link: [https://gatk.broadinstitute.org/hc/en-us/community/posts/7568071901083-SamToFastq-bam-order](https://gatk.broadinstitute.org/hc/en-us/community/posts/7568071901083-SamToFastq-bam-order)

\--

SamToFastq tool:

Can we use as input coordinate and also name sorted bam?

Thanks.<br><br><i>(created from <a href='https://broadinstitute.zendesk.com/agent/tickets/291776'>Zendesk ticket #291776</a>)<br> gz#291776</i>

droazen:
@GATKSupportTeam Looking at the code for this tool, it's not clear that it will produce correct output when given coordinate-sorted input -- someone should test to make sure.

kockan:
It should work with a coordinate sorted sam/bam. We still might want to add a test for completeness.

kockan:
Closed by #1920

PR #1920, opened 2023-11-17
Updated the SamToFastq documentation

kockan:
- Updated the SamToFastq documentation to clarify that the tool works properly for both name-sorted and coordinate-sorted inputs
- Fixed some whitespace issues


kockan:
For #1834

droazen (approved):

PR #1925, opened 2023-12-04
Cloud support for BuildBamIndex outputs.

cmnbroad:
Fixes https://github.com/broadinstitute/picard/issues/1827.

Draft until #1852 goes in, since this needs the cloud test infrastructure there that enables so we can include an actual cloud test.


rickymagner:
@takutosato This is ready for a test to be added as discussed earlier

Issue #1914, opened 2023-09-29
BedToIntervalList doesn't handle empty bed region edge cases properly


rickymagner:
### Affected tool(s)
`BedToIntervalList`

### Affected version(s)
Tested on 2.27.5 but expect to be true for all later versions since this tool doesn't look like it's been touched since, based on the release notes. I ran through `gatk` as well.

### Description 
If a bed file has entries with the same start and end coordinates, this gets translated into an illegal interval_list entry by the tool.

#### Steps to reproduce
Here is `test.bed`:
```
chr1	100	100
```
and output of 
```
gatk BedToIntervalList -I test.bed -O test.interval_list --SEQUENCE_DICTIONARY my.fasta
```
is:
```
# header truncated ...
chr1    101     100     +       .
```
This breaks other tools that enforce the start to be <= the end for interval_lists, which is reasonable. The problem is the input bed region is actually an empty region and shouldn't exist in the first place, but did come across such a line in the wild.

#### Expected behavior
I would recommend essentially skipping these lines and printing out a warning to the user that such lines are being skipped, in case they actually cared about the `name` column from the bed.

#### Actual behavior
It produces an illegal interval_list line.

yfarjoun:
Interval Lists should be able to represent empty intervals, but since they use 1-based inclusive coordinates, as opposed to 0-based half-open (like bed) there needs to be a coordinate change on "start" (but not "end"). The result isn't illegal....

Which tools fail on empty intervals in an interval list?

rickymagner:
Perhaps the problem isn't this tool then but how GATK uses the `-L` flag. The example that I saw this failing on was running `gatk CountVariants -V example.vcf -L test.interval_list`, but it seems to also fail on `test.bed` as well. The error is:

```
A USER ERROR has occurred: Badly formed genome unclippedLoc: Parameters to GenomeLocParser are incorrect:The stop position 100 is less than start 101 in contig chr1
```
Or maybe it's just localized to `CountVariants`? If we're confident most tools should be able to handle this edge case properly, I'd be OK treating this as a `CountVariants` bug rather than `BedToIntervalList`.

lbergelson:
Lets add a --remove-zero-length flag to these tools.

droazen:
Consensus is that we should just add a `--keep-zero-length-intervals` flag to the tool, and skip them by default.

rickymagner:
This discussion should now be resolved by #1928

PR #1928, opened 2023-12-14
Add flag to keep zero length intervals when converting bed -> interval_list

rickymagner:
### Description

This PR is a follow-up to the discussion in #1914. In particular, it implements a boolean flag `KEEP_ZERO_LENGTH_INTERVALS` for the `BedToIntervalList` tool, and sets the default to `false`. This means that if the input bed file has length 0 intervals, e.g. `chr1    1000   1000`, they will be skipped in the `.interval_list` output file, unless the flag is manually toggled to `true`. Note this will change default behavior from previous versions, but based on the discussion there seems like a very sensible default.

A new test was added to test the functionality, and previous tests were modified to toggle this flag to true to ensure that existing tests meant to ensure length zero intervals were properly handled are still handled correctly when the flag is turned on.

----

#### Content

#### Review


src/main/java/picard/util/BedToIntervalList.java
```diff
@@ -107,9 +107,13 @@ public class BedToIntervalList extends CommandLineProgram {
     @Argument(doc = "If true, entries that are on contig-names that are missing from the provided dictionary will be dropped.")
     public boolean DROP_MISSING_CONTIGS = false;
 
+    @Argument(doc = "If true, write length zero intervals in input bed file to resulting interval list file.")
```
kockan, line 110: Should we say "zero-length intervals" instead of "length zero intervals". I know it's not critical, just for the sake of consistency

kockan (approved):

src/main/java/picard/util/BedToIntervalList.java
rickymagner: I actually like "length zero" better than "zero length" so I switched the labelings everywhere to make it consistent.

PR #1929, opened 2023-12-19
Adding http-nio as a dependency

lbergelson:
* This is a plugin which provides HttpFileSystemProvider which allows connecting to http(s) paths
  through the nio Path and Files interfaces.
* Updating htsjdk to 4.1.0 which improves support for plugins like this in Tribble.
* Adding an extremely simple test to show that http-nio is available


kockan (approved):

lbergelson:
@kockan Thank you

PR #1930, opened 2023-12-19
Update Gradle to 8.5

lbergelson:
* update to use configuration api
* remove no longer relevant bits about the parser settings
* update to automatically configure javadoc/source jars and the toolchain


kockan (approved):
Make it so

PR #1934, opened 2024-01-10
Convert to use the java-library plugin

lbergelson:
Use the 'java-library' plugin instead of the 'java' one.  This should improve our dependency resolution for downstream projects.

As part of this change, remove picard dependency on log4j and replace it with a version constraint so that transitive dependencies don't import a vulnerable version.

I don't have any automatic tooling to distinguish api from implementation dependencies so this is a best effort.  We should be ready to change any if we are exposing some that we don't expect.

Addresses the issue brought up in https://github.com/broadinstitute/picard/pull/1919, however we've made the deliberate decision to include the minimum required dependencies instead of marking them all as api.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
 }
 
 plugins {
-    id "java"
+    id "java-library"
     id 'maven-publish'
     id 'signing'
     id 'jacoco'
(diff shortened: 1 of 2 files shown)


kockan (approved):
LGTM

PR #1949, opened 2024-03-03
Cloudify RevertSam

takutosato:
### Description

RevertSam now supports cloud input and output.

----

#### Content

#### Review


takutosato:
@LeeTL1220 @droazen here it is finally.

src/main/java/picard/sam/RevertSam.java
```diff
@@ -257,10 +261,12 @@ protected String[] customCommandLineValidation() {
 
     protected int doWork() {
         IOUtil.assertFileIsReadable(INPUT.toPath());
-        ValidationUtil.assertWritable(OUTPUT, OUTPUT_BY_READGROUP);
+        if (OUTPUT != null && !PicardBucketUtils.isGcsUrl(OUTPUT)) {
```
cmnbroad, line 264: Suggest we add `isFileUrl` to PicardHstPath, and check for that here, since this won't correctly filter paths on the other (non-GCS) cloud providers.

cmnbroad (commented):
@tsato I'm not sure what state this is in, so I started reviewing, but I assume its not ready yet (there are still "tsato" comments, some unused variables, etc.) Let me know if/when its ready for review.

takutosato:
@cmnbroad all the main code should be ready for review. I've been making fixes to the test, in particular how to properly clean up the test output etc so that the output from a previous test doesn't affect the current test. It's a little more challenging when the output paths are specified in a txt file (`OUTPUT_MAP`), but I should be able to wrap up by the end of the weekend.

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +597,96 @@ public void testHardClipRoundTrip() throws Exception {
...
+    final String testReadGroupMapFile = "gs://hellbender/test/resources/picard/revertSam/test_group_map_file.txt";
+    final String testReadGroupMapFileLocal = REVERT_SAM_TEST_DATA_DIR + "test_group_map_file.txt";
+
+    private final static boolean OUTPUT_BY_READ_GROUP = true;
+
+
+    @DataProvider(name="cloudTestData")
+    public Object[][] getCloudTestData() {
+        return new Object[][] {
+                {testMedium, Optional.of(GCloudTestUtils.TEST_OUTPUT_DEFAULT + "test/reverted"), OUTPUT_BY_READ_GROUP, Optional.empty()},
```
cmnbroad, line 619: There is no guarantee that you have exclusive access to these cloud files (i.e., `GCloudTestUtils.TEST_OUTPUT_DEFAULT + "test/reverted"`) when these tests run. In fact, you're pretty much guaranteed to not have exclusive access to them, since many CI test runners can be run at the same time, so this is pretty much guaranteed to result in collisions. This may very likely be the reason you're having problems deleting them in your test code below. (Even if I run theses tests locally while a CI job is running, I think bad things will happen). So we need to find a way to make then unique, probably by dynamically creating a unique output directory or something. I'm not sure if we have infrastructure for that, so we may have to invent something.

cmnbroad (changes requested):
First round done.

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -437,9 +442,9 @@ public void testAssertAllReadGroupsMappedSuccess() {
         final SAMReadGroupRecord rg1 = new SAMReadGroupRecord("rg1");
         final SAMReadGroupRecord rg2 = new SAMReadGroupRecord("rg2");
 
-        final Map<String, File> outputMap = new HashMap<String, File>();
-        outputMap.put("rg1", new File("/foo/bar/rg1.bam"));
-        outputMap.put("rg2", new File("/foo/bar/rg2.bam"));
+        final Map<String, Path> outputMap = new HashMap<>();
+        outputMap.put("rg1", new File("/foo/bar/rg1.bam").toPath());
+        outputMap.put("rg2", new File("/foo/bar/rg2.bam").toPath());
```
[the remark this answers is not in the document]
takutosato: Great, done

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +597,96 @@ public void testHardClipRoundTrip() throws Exception {
             }
         }
     }
+
+    public static final String CURRENT_DIRECTORY = System.getProperty("user.dir");
```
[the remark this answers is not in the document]
takutosato: done

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +597,96 @@ public void testHardClipRoundTrip() throws Exception {
...
+    public static final String REVERT_SAM_TEST_DATA_DIR = "testdata/picard/sam/RevertSam/";
+
+    final String testSmall = "gs://hellbender/test/resources/picard/bam/CEUTrio.HiSeq.WGS.b37.NA12878.20.21_n100.bam";
```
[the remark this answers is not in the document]
takutosato: removed

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +597,96 @@ public void testHardClipRoundTrip() throws Exception {
...
+    final String testMedium = "gs://hellbender/test/resources/picard/bam/CEUTrio.HiSeq.WGS.b37.NA12878.20.21_n10000.bam";
+    final String testMediumCram = "gs://hellbender/test/resources/picard/bam/CEUTrio.HiSeq.WGS.b37.NA12878.20.21_n10000.cram";
```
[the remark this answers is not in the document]
takutosato: Yup, moved these up to PicardCommandLineProgramTest

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +597,96 @@ public void testHardClipRoundTrip() throws Exception {
             }
         }
     }
+
+    public static final String CURRENT_DIRECTORY = System.getProperty("user.dir");
+    public static final String REVERT_SAM_TEST_DATA_DIR = "testdata/picard/sam/RevertSam/";
```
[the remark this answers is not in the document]
takutosato: done

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +597,96 @@ public void testHardClipRoundTrip() throws Exception {
...
+
+    @DataProvider(name="cloudTestData")
+    public Object[][] getCloudTestData() {
+        return new Object[][] {
+                {testMedium, Optional.of(GCloudTestUtils.TEST_OUTPUT_DEFAULT + "test/reverted"), OUTPUT_BY_READ_GROUP, Optional.empty()},
+                {testMedium, Optional.of(GCloudTestUtils.TEST_OUTPUT_DEFAULT + "test/reverted.bam"), !OUTPUT_BY_READ_GROUP, Optional.empty()},
+                {testMedium, Optional.empty(), OUTPUT_BY_READ_GROUP, Optional.of(testReadGroupMapFile)},
+                {testMedium, Optional.empty(), OUTPUT_BY_READ_GROUP, Optional.of(testReadGroupMapFileLocal)},
+                {testMediumCram, Optional.empty(), OUTPUT_BY_READ_GROUP, Optional.of(testReadGroupMapFile)},
+                {testMediumCram, Optional.of(GCloudTestUtils.TEST_OUTPUT_DEFAULT + "test/reverted.cram"), !OUTPUT_BY_READ_GROUP, Optional.empty()},
```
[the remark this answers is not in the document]
takutosato: done

src/test/java/picard/sam/RevertSamTest.java
takutosato: Yup, as I wrote elsewhere, this did create a race condition. See the updated code, which uses temp files with unique random names and does not recycle file names.

takutosato:
@cmnbroad thanks again for reviewing. Code is ready for another round of review.

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +604,135 @@ public void testHardClipRoundTrip() throws Exception {
...
+     *
+     * Write the output map for Revert Sam based on the read groups present in the header file of the given SAM file.
+     *
+     * @param sam        The read names in the header of this sam file will be used as rows
+     * @param outputBase The directory where the per read group output files will live
+     * @return The PicardHtsPath to the tsv with each row a pair (read name, output), where output points to a temp file
+o     * created in the directory specified in outputBase.
+     */
+    private PicardHtsPath getTestReadGroupMapFile(final PicardHtsPath sam, final String outputBase) {
+        final String samExtension = sam.isBam() ? ".bam" : ".cram";
```
cmnbroad, line 641: This seems unsafe for things like a .sam - why not just use `sam.getExtension()` instead ?

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +604,135 @@ public void testHardClipRoundTrip() throws Exception {
...
+        final PicardHtsPath result = PicardBucketUtils.getTempFilePath(DEFAULT_CLOUD_TEST_OUTPUT_DIR.getURIString(), ".txt");
```
cmnbroad, line 642: I would recommend not using a variable name like `result` for something like this. Suggest calling it something like `outputMapFile`.

cmnbroad, line 642: Having said that, it is super confusing that this method has an outputBase arg, but the master file is always in the cloud, even when the outputBase is not. Is that intentional ? If so I would rename the method to something that indicates that it is cloud-specific. More likely though, the intention is that it should create the .txt file in the output base ?

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +604,135 @@ public void testHardClipRoundTrip() throws Exception {
...
+    /**
+     *
+     * Write the output map for Revert Sam based on the read groups present in the header file of the given SAM file.
+     *
+     * @param sam        The read names in the header of this sam file will be used as rows
+     * @param outputBase The directory where the per read group output files will live
+     * @return The PicardHtsPath to the tsv with each row a pair (read name, output), where output points to a temp file
+o     * created in the directory specified in outputBase.
+     */
+    private PicardHtsPath getTestReadGroupMapFile(final PicardHtsPath sam, final String outputBase) {
```
cmnbroad, line 640: If outputBase must be a directory, I would rename it to `outputDirectory`.

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +604,135 @@ public void testHardClipRoundTrip() throws Exception {
...
     }
+
+
+    private final static boolean OUTPUT_BY_READ_GROUP = true;
+    // LOCAL_OUTPUT_MAP means the output map file itself is stored locally. The output SAM files will always be in the cloud.
+    private enum OutputMapOptions { LOCAL_OUTPUT_MAP, CLOUD_OUTPUT_MAP, NO_OUTPUT_MAP }
+
+    @DataProvider(name = "cloudTestData")
+    public Object[][] getCloudTestData() {
+        final PicardHtsPath tempCramOutputPath = PicardBucketUtils.getTempFilePath(GCloudTestUtils.TEST_OUTPUT_DEFAULT_STR, "RevertSam", ".cram");
```
cmnbroad, line 615: Since this is a cloud path, I would include that in the name to make it clear, i.e. tempCloudCram or something.

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +604,135 @@ public void testHardClipRoundTrip() throws Exception {
...
+        if (outputPath != null){
+            args.add("OUTPUT=" + outputPath.getURIString());
+        }
+
+        final boolean createOutputMap = outputMapOptions != OutputMapOptions.NO_OUTPUT_MAP;
+        PicardHtsPath outputMap = null;
+        if (createOutputMap){
+            // If these conditions are met, we dynamically create the output map file based on the read groups in inputSAM
+            final String outputMapDir = outputMapOptions == OutputMapOptions.CLOUD_OUTPUT_MAP ?
+                    DEFAULT_CLOUD_TEST_OUTPUT_DIR.getURIString() :
```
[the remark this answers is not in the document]
cmnbroad: Oh I see; the name generation is buried below in the output map generation code.

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +604,135 @@ public void testHardClipRoundTrip() throws Exception {
...
+                    REVERT_SAM_LOCAL_TEST_DATA_DIR;
+            outputMap = getTestReadGroupMapFile(inputSAM, outputMapDir);
+            args.add("OUTPUT_MAP=" + outputMap.getURIString());
+        }
+
+        if (inputSAM.isCram()){
+            args.add("REFERENCE_SEQUENCE=" + HG19_CHR2021);
```
cmnbroad, line 691: The reference to use should really be part of the data provider; otherwise its hard to add new test files to this.

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +604,135 @@ public void testHardClipRoundTrip() throws Exception {
...
+    /***
+     *
+     * @param inputSAM The path to the input SAM file
+     * @param outputPath The path to the single output SAM file, when output by read group is turned off. May be null.
+     * @param outputByReadGroup Whether to output by read group
+     * @param outputMapOptions Dictates whether to dynamically create an output map and where (local or cloud)
+     */
+    @Test(dataProvider = "cloudTestData", groups = "cloud")
+    public void testCloud(final PicardHtsPath inputSAM, final PicardHtsPath outputPath, final boolean outputByReadGroup,
+                          final OutputMapOptions outputMapOptions) {
```
cmnbroad, line 669: It took me a while to figure out why all of these args are needed. It's very possible that I'm missing some nuance, but it seems like it all of this (test and provider) code would be a lot simpler if there were just 3 values in the data provider: the input path, the output path (which will either be a file or a directory), and a `writeByReadGroup` boolean (which tells you whether the output path needs to be a map (OUTPUT_MAP) or a file (OUTPUT)). Everything else in the test method could be derived from those, and then you don't need the enum.

cmnbroad (changes requested):

src/test/java/picard/util/GCloudTestUtils.java
```diff
@@ -10,11 +10,16 @@ public final class GCloudTestUtils {
...
-    public static final PicardHtsPath TEST_INPUTS_DEFAULT = new PicardHtsPath("gs://hellbender/test/resources/");
-    public static final PicardHtsPath TEST_STAGING_DEFAULT = new PicardHtsPath("gs://hellbender-test-logs/staging/");
-    public static final PicardHtsPath TEST_PROJECT_DEFAULT = new PicardHtsPath("broad-dsde-dev/");
+    public static final String TEST_INPUTS_DEFAULT_STR = "gs://hellbender/test/resources/";
+    public static final String TEST_STAGING_DEFAULT_STR = "gs://hellbender-test-logs/staging/";
+    public static final String TEST_PROJECT_DEFAULT_STR = "broad-dsde-dev/";
+    public static final String TEST_OUTPUT_DEFAULT_STR = TEST_STAGING_DEFAULT_STR + "picard/";
+    public static final PicardHtsPath TEST_INPUTS_DEFAULT = new PicardHtsPath(TEST_INPUTS_DEFAULT_STR);
+    public static final PicardHtsPath TEST_STAGING_DEFAULT = new PicardHtsPath(TEST_STAGING_DEFAULT_STR);
+    public static final PicardHtsPath TEST_PROJECT_DEFAULT = new PicardHtsPath(TEST_PROJECT_DEFAULT_STR);
```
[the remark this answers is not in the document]
takutosato: I'd be happy to remove it. @lbergelson is it ok if we removed `TEST_PROJECT_DEFAULT`and `getTestProject` from `GCloudTestUtils.java`? Looks like you added it about a year ago.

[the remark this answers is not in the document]
cmnbroad: @takutosato @lbergelson Actually, git blames me, as part of the cloudification of references PR.Either way, it looks wrong to me (having the project name might be ok, but the trailing "/" and rendering it as a URI seems totally wrong to me). If Louis confirms that, lets either fix it or remove it.

src/test/java/picard/cmdline/CommandLineProgramTest.java
```diff
@@ -21,11 +23,19 @@ public abstract class CommandLineProgramTest {
     public static final File CHR_M_REFERENCE = new File(REFERENCE_TEST_DIR,"chrM.reference.fasta");
     public static final File CHR_M_DICT = new File(REFERENCE_TEST_DIR,"chrM.reference.dict");
 
+    // These are the hg19 references with chromosome names "1" (rather than "chr1")
+    public static final PicardHtsPath HG19_REFERENCE_GCLOUD = new PicardHtsPath("gs://gcp-public-data--broad-references/hg19/v0/Homo_sapiens_assembly19.fasta");
```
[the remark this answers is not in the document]
takutosato: Removed

src/main/java/picard/util/TabbedTextFileWithHeaderParser.java
```diff
@@ -121,6 +126,10 @@ public TabbedTextFileWithHeaderParser(final File file) {
         this(new TabbedInputParser(false, file));
     }
 
+    public TabbedTextFileWithHeaderParser(final Path file) throws IOException {
```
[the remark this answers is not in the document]
takutosato: I see, that makes sense, but I think elsewhere in Picard/htsjdk, we often offer constructors both for `File` and `Path`. So it seems less than ideal to have the implicit convention that if you are working with Path, you are expected to use the constructor that takes `TabbedInputParser.`

What if I put try-catch within `TabbedTextFileWithHeaderParser(final Path file)`, so I can avoid the `throws IOException` signature?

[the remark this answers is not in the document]
takutosato: Right, I can't do this because the `this` statement needs to be the first line of the constructor…

src/main/java/picard/sam/RevertSam.java
```diff
@@ -572,26 +588,55 @@ static Map<String, File> createOutputMap(
...
+    private static void writeRow(final PrintWriter writer, final String... row) {
+        for (int j = 0; j < row.length; j++){
+            writer.write(row[j]);
+            if (j < row.length - 1) {
+                writer.write('\t');
+            } else {
+                writer.println();
+            }
+        }
+    }
```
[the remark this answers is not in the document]
takutosato: Moved `writeOutputMapToFile` and `writeRow` to `RevertSamTest` and made them private.

src/main/java/picard/sam/RevertSam.java
```diff
@@ -779,25 +824,32 @@ List<PeekableIterator<SAMRecord>> iterators() {
...
             } else {
                 validateOutputParamsNotByReadGroup(output, outputMap, errors);
             }
         }
 
-        static void validateOutputParamsByReadGroup(final File output, final File outputMap, final List<String> errors) {
+        static void validateOutputParamsByReadGroup(final IOPath output, final IOPath outputMap, final List<String> errors) {
             if (output != null) {
-                if (!Files.isDirectory(output.toPath())) {
+                if (! PicardBucketUtils.isGcsUrl(output) && !Files.isDirectory(output.toPath())) {
```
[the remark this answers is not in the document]
takutosato: Yes, I think it seems safe to skip the check altogether for cloud files. But then the behavior is unpredictable if `OUTPUT` is set to an existing file, like `test.txt`. Investigating.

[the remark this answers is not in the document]
takutosato: OK, the tool doesn't clobber the existing file if it's given as OUTPUT (the tool instead creates a directory called e.g. `path/test.txt/`, so I think just checking `Files.isDirectory` for local files is ok.

src/test/java/picard/sam/RevertSamTest.java
takutosato: done

src/test/java/picard/sam/RevertSamTest.java
takutosato: done

takutosato: There are two paths here. One is the ".txt" file (really tsv), where each row is the read group, and the associated path where that read group SAM is written. This file is a temp file, always living in the cloud (by default)

The outputBase arg is the base path for each of the SAM files listed in the .txt file above. This can be cloud or local. 

I renamed the variables to make this a little more clear.

src/test/java/picard/sam/RevertSamTest.java
takutosato: Renamed

src/test/java/picard/sam/RevertSamTest.java
takutosato: done

src/test/java/picard/sam/RevertSamTest.java
takutosato: done

src/test/java/picard/sam/RevertSamTest.java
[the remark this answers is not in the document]
takutosato: There was one more case (OUTPUT_MAP = null, OUTPUT=directory, OUTPUT_BY_READ_GROUP=true) that could potentially cause collisions across PRs. I resolved it by appending a timestamp as prefix in those cases.

src/test/java/picard/sam/RevertSamTest.java
takutosato: It might be, but I think it's a bit dangerous to rely on `Files.isDirectory(OUTPUT)`given how directories are handled in GCS. And even if we did that, I don't think the code would be that much simpler. The reason why this code is hard to parse, I think, is because RevertSam accepts a wide array of input patterns (e.g. OUTPUT may be a file or directory), so I vote that we leave it as is. What do you think?

takutosato:
@cmnbroad sorry this took a while. I've addressed all your comments.

src/test/java/picard/util/GCloudTestUtils.java
```diff
@@ -10,9 +10,12 @@ public final class GCloudTestUtils {
...
     public static final PicardHtsPath REQUESTER_PAYS_BUCKET_DEFAULT = new PicardHtsPath("gs://hellbender-requester-pays-test/");
 
-    public static final PicardHtsPath TEST_INPUTS_DEFAULT = new PicardHtsPath("gs://hellbender/test/resources/");
-    public static final PicardHtsPath TEST_STAGING_DEFAULT = new PicardHtsPath("gs://hellbender-test-logs/staging/");
-    public static final PicardHtsPath TEST_PROJECT_DEFAULT = new PicardHtsPath("broad-dsde-dev/");
+    public static final String TEST_INPUTS_DEFAULT_STR = "gs://hellbender/test/resources/";
+    public static final String TEST_STAGING_DEFAULT_STR = "gs://hellbender-test-logs/staging/";
+    public static final String TEST_PROJECT_DEFAULT_STR = "broad-dsde-dev";
+    public static final String TEST_OUTPUT_DEFAULT_STR = TEST_STAGING_DEFAULT_STR + "picard/";
+    public static final PicardHtsPath TEST_INPUTS_DEFAULT = new PicardHtsPath(TEST_INPUTS_DEFAULT_STR);
```
cmnbroad, line 17: I know some of these were just moved, but in some places, you used a `_CLOUD` or `_GCLOUD` in constant names like these, which nicely reinforces that they're remote - I'd suggest doing that for these as.

src/test/java/picard/cmdline/CommandLineProgramTest.java
```diff
@@ -21,11 +23,18 @@ public abstract class CommandLineProgramTest {
...
     public static final File CHR_M_DICT = new File(REFERENCE_TEST_DIR,"chrM.reference.dict");
 
+    // These are the hg19 references with chromosome names "1" (rather than "chr1")
+    public static final PicardHtsPath HG19_CHR2021_GCLOUD = new PicardHtsPath(GCloudTestUtils.getTestInputPath() + "picard/references/human_g1k_v37.20.21.fasta");
+    public static final PicardHtsPath HG19_CHR2021 = new PicardHtsPath("testdata/picard/reference/human_g1k_v37.20.21.fasta.gz");
+
+    public static final PicardHtsPath NA12878_MINI = new PicardHtsPath(GCloudTestUtils.getTestInputPath() + "picard/bam/CEUTrio.HiSeq.WGS.b37.NA12878.20.21_n100.bam");
+    public static final PicardHtsPath NA12878_MINI_CRAM = new PicardHtsPath(GCloudTestUtils.getTestInputPath() + "picard/bam/CEUTrio.HiSeq.WGS.b37.NA12878.20.21_n100.cram");
+    public static final PicardHtsPath NA12878_MEDIUM = new PicardHtsPath(GCloudTestUtils.getTestInputPath() + "picard/bam/CEUTrio.HiSeq.WGS.b37.NA12878.20.21_n10000.bam");
+    public static final PicardHtsPath NA12878_MEDIUM_CRAM = new PicardHtsPath(GCloudTestUtils.getTestInputPath() + "picard/bam/CEUTrio.HiSeq.WGS.b37.NA12878.20.21_n10000.cram");
```
cmnbroad, line 33: I know some of these names were just moved, but in some places, you used a `_CLOUD` (or `_GCLOUD`) in the names of constants like these, which nicely reinforces that they're remote - I'd suggest doing that for these as well.

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +606,189 @@ public void testHardClipRoundTrip() throws Exception {
...
+
+        // Add a prefix to avoid name collisions across PRs etc.
+        String prefix = null;
+        if (outputPath != null){
+            args.add("OUTPUT=" + outputPath.getURIString());
+            if (outputByReadGroup){
+                LocalDateTime now = LocalDateTime.now();
+                DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd_HH:mm:ss.SSS");
+                prefix = now.format(formatter);
+                args.add("PREFIX=" + prefix);
```
cmnbroad, line 729: Unfortunately, I think this name collision resolution strategy is a step backwards.

Instead of using DEFAULT_CLOUD_TEST_OUTPUT_DIR for the output directory in these tests, it would be much simpler to change the test code to create a new temporary directory/bucket in the staging dir, and pass that to RevertSam as the outputPath. Then you don't need to add the new tool arg PREFIX, which is really only useful for tests, or this prefix/timestamp code (which is still theoretically subject to name collision).

If there are test cases for which this (or copying the inputs - see the discussion in the engine team space) doesn't suffice for some reason, let me know and we can think about other alternatives. But I'd very much prefer to not add args to tools that are only useful for tests if we can possibly avoid it.

src/test/java/picard/sam/RevertSamTest.java
```diff
@@ -592,4 +606,189 @@ public void testHardClipRoundTrip() throws Exception {
...
+        NO_OUTPUT_MAP }
+
+
+    @DataProvider(name = "cloudTestData")
+    public Object[][] getCloudTestData() {
+        final PicardHtsPath tempCloudCram = PicardBucketUtils.getTempFilePath(DEFAULT_CLOUD_TEST_OUTPUT_DIR, "RevertSam", ".cram");
+        final PicardHtsPath tempLocalCram = PicardBucketUtils.getLocalTempFilePath("RevertSam", ".cram");
+        return new Object[][]{
+                // Output by read group without the output map, write output bams in the cloud.
+                {NA12878_MEDIUM, DEFAULT_CLOUD_TEST_OUTPUT_DIR, OUTPUT_BY_READ_GROUP, OutputMapOptions.NO_OUTPUT_MAP, null },
```
cmnbroad, line 625: For this test case, I think you can just create a temporary parent bucket in the staging area to to hold the output, rather than referencing the default staging bucket here and trying to synthesize unique names using a prefix and timestamp (see my comment where the name synthesis code is).

cmnbroad (commented):
Looking much better - just a couple of minor things left, plus the remote file name collision issue (see inline comments and the engine team slack discussion).

src/test/java/picard/util/GCloudTestUtils.java
```diff
@@ -10,9 +10,12 @@ public final class GCloudTestUtils {
...
+    public static final PicardHtsPath TEST_STAGING_DEFAULT = new PicardHtsPath(TEST_STAGING_DEFAULT_STR);
     public static final PicardHtsPath TEST_OUTPUT_DEFAULT = PicardHtsPath.resolve(TEST_STAGING_DEFAULT, "picard/");
```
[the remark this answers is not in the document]
takutosato: removed

src/main/java/picard/nio/PicardBucketUtils.java
```diff
@@ -67,11 +67,19 @@ public static PicardHtsPath getTempFilePath(final String directory, String prefi
         }
     }
 
+    public static PicardHtsPath getTempFilePath(final PicardHtsPath directory, String prefix, final String extension){
+        return getTempFilePath(directory.getURIString(), prefix, extension);
+    }
+
```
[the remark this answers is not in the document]
takutosato: Done, but let me know if you were hoping for a more detailed javadoc

src/main/java/picard/util/TabbedTextFileWithHeaderParser.java
```diff
@@ -26,7 +26,12 @@
 import htsjdk.samtools.util.CloseableIterator;
 import picard.PicardException;
 
+import java.io.BufferedInputStream;
 import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
```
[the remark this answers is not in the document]
takutosato: done

src/test/java/picard/sam/AbstractAlignmentMergerTest.java
```diff
@@ -700,7 +702,7 @@ public void testUnmapBacterialContamination() throws IOException {
         MergeBamAlignment mergeBamAlignment = new MergeBamAlignment();
 
         mergeBamAlignment.ALIGNED_BAM = Collections.singletonList(file.toPath().toFile()); // TODO update to use Path when MergeBamAlignment is updated to use Path
-        mergeBamAlignment.UNMAPPED_BAM = fileUnaligned;
+        mergeBamAlignment.UNMAPPED_BAM = fileUnaligned.toPath().toFile(); // Also need to be updated to use Path
```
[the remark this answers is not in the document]
takutosato: `fileUnaligned` is now `PicardHtsPath`, but MergeBamAlignment hasn't be cloudified yet. I have another PR coming soon for that.

src/test/java/picard/util/GCloudTestUtils.java
takutosato: done

src/test/java/picard/cmdline/CommandLineProgramTest.java
takutosato: done

src/test/java/picard/sam/RevertSamTest.java
takutosato: Yes that would be the most natural thing to do, except that there are no directories in google cloud so we can't "create" a temp directory…what's confusing is that in the browser, they do let you create a directory (that's how I created DEFAULT_CLOUD_TEST_OUTPUT_DIR). For instance `createDirectory` method in `CloudStorageFileSystemProvider.java` is a no-op. 

Come to think of it, as long as we don't check that the output directory exists (which we currently do not when the directory is in the cloud), I could probably just create an nonexistent random directory name as OUTPUT, and the files will be created under it. I will do some digging and let you know.

cmnbroad: @tsato - you can create a cloud directory (see the engine-team slack discussion from Tuesday 7/9 about this). Also, I'm currently doing a similar thing myself in a different PR so I can share that code with you if you like.

src/test/java/picard/sam/RevertSamTest.java
takutosato: I addressed this issue by "creating" a "directory" with a randomly generated path. See PicardBucketUtils::getRandomGCSDirectory()

src/test/java/picard/sam/RevertSamTest.java
takutosato: Let me know if my approach in the new commit is what you had in mind.

takutosato:
@cmnbroad I addressed your comments, removed the "prefix" with a timestamp approach for avoiding race conditions, and I also removed the enum you suspected earlier was unnecessary. Back to you.

src/main/java/picard/nio/PicardBucketUtils.java
```diff
@@ -67,19 +65,48 @@ public static PicardHtsPath getTempFilePath(final String directory, String prefi
...
     }
 
+    public static PicardHtsPath getLocalTempFilePath(final String prefix, final String extension){
+        return getTempFilePath((String) null, prefix, extension);
+    }
+
+    /**
+     * Creates a path to a "directory" on a Google Cloud System filesystem with a randomly generated URI.
+     * Since the notion of directories does not exist in GCS, it creates a path to a URI ending in "/".
+     * Calling this method will not create a directory/file on GCS. It merely returns a path to a non-directory.
```
cmnbroad, line 89: This line is super-confusing:

> It merely returns a path to a non-directory.

since this method is supposed to be creating a directory. I would suggest maybe saying "it creates an object that can be used as a parent container for other objects" or something similar instead.

src/main/java/picard/nio/PicardBucketUtils.java
```diff
@@ -67,19 +65,48 @@ public static PicardHtsPath getTempFilePath(final String directory, String prefi
         }
     }
 
-    // For local temp file, directory should be null.
+    /**
+     * This overload of getTempFilePath takes the directory of type PicardHtsPath instead of String.
+     */
```
cmnbroad, line 70: Might as well use the standard javadoc format like you did elsewhere, i.e. list the params like this (this is a random example I took from elsewhere):

/**
* Picks a random name, by putting some random letters between "prefix" and "suffix".
*
* @param stagingLocation The folder where you want the file to be. Must start with "gs://" or "hdfs://"
* @param prefix The beginning of the file name
* @param suffix The end of the file name, e.g. ".tmp"
*/

src/main/java/picard/sam/RevertSam.java
```diff
@@ -287,7 +296,12 @@ protected int doWork() {
                 defaultExtension = "." + OUTPUT_BY_READGROUP_FILE_FORMAT.toString();
             }
 
-            outputMap = createOutputMap(OUTPUT_MAP, OUTPUT, defaultExtension, inHeader.getReadGroups());
+            if (OUTPUT_MAP != null){
+                outputMap = readOutputMap(OUTPUT_MAP.toPath());
+            } else {
+                outputMap = createOutputMapFromReadGroups(inHeader.getReadGroups(), OUTPUT.toPath(), defaultExtension);
+            }
```
cmnbroad, line 303: This is much better naming, thanks. This tool is super confusing - sometimes OUTPUT_MAP is an output and sometimes its an input.

src/main/java/picard/nio/PicardBucketUtils.java
```diff
@@ -67,19 +65,48 @@ public static PicardHtsPath getTempFilePath(final String directory, String prefi
...
+    public static PicardHtsPath getTempFilePath(final PicardHtsPath directory, String prefix, final String extension){
+        return getTempFilePath(directory.getURIString(), prefix, extension);
+    }
+
+    /**
+     * Calls getTempFilePath without the prefix.
+     */
```
cmnbroad, line 77: Same comment about standard javadoc format with params.

cmnbroad (commented):
Ah - this always takes longer than I think it will. I did a quick pass and there are only a couple of very minor things left. Also, we need to discuss the auto-delete configuration for the temporary directories with Louis. I think we might need to enforce that `getRandomGCSDirectory` always creates a directory in a well-known specific gcs object that is set to auto-delete, rather than letting the caller create one anywhere. I'll bring this up for discussion with Louis on Monday at the team meeting, so lets hold off on merging until then.

src/main/java/picard/nio/PicardBucketUtils.java
takutosato: Thanks, I attempted to rewrite this.

src/main/java/picard/nio/PicardBucketUtils.java
takutosato: Right; I thought it might be a little verbose to repeat the almost identical javadoc for overladed methods. It seems like using `@see` to refer to the original with the full java may be a good solution?

https://stackoverflow.com/questions/2558695/javadoc-reuse-for-overloaded-methods

src/main/java/picard/nio/PicardBucketUtils.java
```diff
@@ -67,19 +65,63 @@ public static PicardHtsPath getTempFilePath(final String directory, String prefi
...
+     *
+     * @param parentDir The root path where the new "directory" is to live e.g. "gs://hellbender-test-logs/staging/picard/test/RevertSam/".
+     * @return A PicardHtsPath object pointing to e.g. "gs://hellbender-test-logs/staging/picard/test/RevertSam/{randomly-generated-string}/"
+     */
+    public static PicardHtsPath getRandomGCSDirectory(final PicardHtsPath parentDir){
+        ValidationUtils.validateArg(parentDir.getScheme().equals(PicardBucketUtils.GOOGLE_CLOUD_STORAGE_FILESYSTEM_SCHEME), "This method is supported only for a GCS path: " + parentDir.getURIString());
+        ValidationUtils.validateArg(parentDir.getURIString().endsWith("/"), "parentDir must end in backslash '/': " + parentDir.getURIString());
+        return PicardHtsPath.fromPath(PicardBucketUtils.randomRemotePath(parentDir.getURIString(), "", "/"));
+    }
+
```
cmnbroad, line 116: I thought we agreed in our discussion yesterday that this method should (and I thought you said that it already did) constrain directory creation to the well-known staging bucket location that has a TTL set. As far as I can tell, it lets you create a directory anywhere.

I know the analogous method in GATK also allows you to create one anywhere as well, but that is how we accumulated the 10,000+ old temporary files floating around in GCS that date back to 2015. If we don't constrain this one, we'll replicate that same problem with Picard. Is there some reason we can't constrain this to the Picard staging location ?

src/main/java/picard/nio/PicardBucketUtils.java
cmnbroad: Using the `see` notation seems fine, but please keep in mind that the ultimate goal here is to eliminate these string overloads, since they're the ones that cause problems. (In fact, I would lobby that the next cloudification PR should do exactly that - do nothing except clean up and get rid of all of these methods that take a string path or dir before we accumulate so many usages of them that removing them becomes untenable).

So given that, if we're going to use "see" references like this, it would be preferable to have the `public static PicardHtsPath getTempFilePath(final PicardHtsPath directory, String prefix, final String extension)` be the one with specific javadoc, and have the other ones `see`/refer to that one. Then when we remove the individual ones, we don't have to change anything else.

src/main/java/picard/nio/PicardBucketUtils.java
takutosato: Now creates a file under a hardcoded staging location "gs://hellbender-test-logs/staging/picard/"

src/main/java/picard/nio/PicardBucketUtils.java
takutosato: If I'm not mistaken, what we want to avoid is directly manipulating the underlying URI String of an existing PicardHtsPath object to create a new one. Given that getTempFilePath is a method that creates a new PicardHtsPath from scratch, rather than from an existing PicardHtsPath, I think String input can be the main method signature (much like the constructor for PicardHtsPath itself).

src/main/java/picard/nio/PicardBucketUtils.java
```diff
@@ -67,19 +65,64 @@ public static PicardHtsPath getTempFilePath(final String directory, String prefi
...
+     *
+     * @see #getTempFilePath(String, String, String)
+     */
     public static PicardHtsPath getTempFilePath(String directory, String extension){
         return getTempFilePath(directory, "", extension);
     }
 
+    /**
+     *
+     * Creates a temporary file on the local temporary file.
```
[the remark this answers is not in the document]
takutosato: Fixed

PR #1956, opened 2024-04-05
-fixes a bug in the liftover logic whereby when the alleles couldn't …

yfarjoun:
…be extended to the "left" because the start of the variant was already at position 1, the resulting VC was corrupt leading to the problem discussed in !1951

- added a unit test (which fails with the same error as the issue prior to the application of the changes)
- _someone_ should add an actual live (with vcf etc....)
- fixes !1951
- other tests are failing...need to look at this more carefully.

### Description

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/util/LiftoverUtils.java
+++ b/src/main/java/picard/util/LiftoverUtils.java
@@ -372,15 +372,15 @@ public class LiftoverUtils {
 
         // 1. while changes in alleles do
         while (changesInAlleles) {
+            final int oldStart = theStart;
+            final int oldEnd = theEnd;
 
-            changesInAlleles = false;
             // 2. if alleles end with the same nucleotide then
             if (alleleBasesMap.values().stream()
                     .collect(Collectors.groupingBy(a -> a[a.length - 1], Collectors.toSet()))
                     .size() == 1 && theEnd > 1) {
                 // 3. truncate rightmost nucleotide of each allele
                 alleleBasesMap.replaceAll((a, v) -> truncateBase(alleleBasesMap.get(a), true));
-                changesInAlleles = true;
                 theEnd--;
                 // 4. end if
             }
@@ -390,20 +390,29 @@ public class LiftoverUtils {
                     .map(a -> a.length)
                     .anyMatch(l -> l == 0)) {
                 // 6. extend alleles 1 nucleotide to the left
+
+
+                final byte extraBase;
+                final boolean extendLeft;
+                if (theStart > 1) {
+                    extraBase = referenceSequence.getBases()[theStart - 2];
+                    extendLeft = true;
+                    theStart--;
+                } else {
+                    extraBase = referenceSequence.getBases()[theEnd];
+                    extendLeft = false;
+                    theEnd++;
+                }
+
                 for (final Allele allele : alleleBasesMap.keySet()) {
                     // the first -1 for zero-base (getBases) versus 1-based (variant position)
                     // another   -1 to get the base prior to the location of the start of the allele
-                    final byte extraBase = (theStart > 1) ?
-                            referenceSequence.getBases()[theStart - 2] :
-                            referenceSequence.getBases()[theEnd];
-
-                    alleleBasesMap.put(allele, extendOneBase(alleleBasesMap.get(allele), extraBase));
+                    alleleBasesMap.put(allele, extendOneBase(alleleBasesMap.get(allele), extraBase, extendLeft));
                 }
-                changesInAlleles = true;
-                theStart--;
 
                 // 7. end if
             }
+            changesInAlleles = theStart!=oldStart && theEnd != oldEnd;
         }
 
         // 8. while leftmost nucleotide of each allele are the same and all alleles have length 2 or more do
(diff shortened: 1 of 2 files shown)


yfarjoun:
@rickymagner 

I think I fixed the bug...but now a bunch of other tests are failing. 

Do you have some time to review and see if my fix is wrong?

yfarjoun:
ah. I found the bug in my code.... another commit coming up....

rickymagner:
Hi @yfarjoun, thanks for looking into this! I can confirm I just built your branch and ran it on the single problematic variant from the original issue, and the liftover completed successfully. I haven't looked at your code too carefully but see there's a test for this case so it seems this would provide a generalizable fix. If you make this a full PR rather than draft someone (I or someone else) can review and try to get it merged. Thanks!

src/test/java/picard/util/LiftoverVcfTest.java
```diff
@@ -975,6 +996,31 @@ public void testLeftAlignVariants(final VariantContext source, final ReferenceSe
...
+        VariantContext vc = builder.make();
+        final List<Allele> origAlleles = new ArrayList<>(builder.getAlleles());
+
+//      This path mimics the codepath through the LiftoverUtils.reverseComplementVariantContext but without reverseComplementing
+        LiftoverUtils.leftAlignVariant(builder, vc.getStart(), vc.getEnd(), vc.getAlleles(), REFERENCE_WITH_REPEATS);
+        builder.genotypes(LiftoverUtils.fixGenotypes(builder.getGenotypes(), origAlleles, builder.getAlleles()));
+        VariantContext newVc = builder.make();
+        final String refString = StringUtil.bytesToString(REFERENCE_WITH_REPEATS.getBases(), newVc.getStart() - 1, newVc.getEnd() - newVc.getStart() + 1);
+
+        Assert.assertEquals(refString,"CGTC");
```
rickymagner, line 1020: Add a space after comma

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -372,15 +372,15 @@ protected static void leftAlignVariant(final VariantContextBuilder builder, fina
...
         while (changesInAlleles) {
+            final int oldStart = theStart;
+            final int oldEnd = theEnd;
 
-            changesInAlleles = false;
             // 2. if alleles end with the same nucleotide then
             if (alleleBasesMap.values().stream()
                     .collect(Collectors.groupingBy(a -> a[a.length - 1], Collectors.toSet()))
                     .size() == 1 && theEnd > 1) {
                 // 3. truncate rightmost nucleotide of each allele
```
rickymagner, line 382: Can this be turned into a continuation of comment 2? If I understand right, this block just trims the alleles if they end with the same string (e.g. REF = "ACG", ALT="ATG" -> REF = "AC", ALT="AT" ... maybe even put this example in comment?)

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -390,20 +390,29 @@ protected static void leftAlignVariant(final VariantContextBuilder builder, fina
                     .map(a -> a.length)
                     .anyMatch(l -> l == 0)) {
                 // 6. extend alleles 1 nucleotide to the left
```
rickymagner, line 392: This comment should also merge with 5. I'm also not sure what it means to have an "empty allele." Can this have an example in comment?

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -390,20 +390,29 @@ protected static void leftAlignVariant(final VariantContextBuilder builder, fina
...
+
+
```
rickymagner, line 394: Remove a newline

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -390,20 +390,29 @@ protected static void leftAlignVariant(final VariantContextBuilder builder, fina
...
+                    extendLeft = true;
+                    theStart--;
+                } else {
+                    extraBase = referenceSequence.getBases()[theEnd];
+                    extendLeft = false;
+                    theEnd++;
+                }
+
                 for (final Allele allele : alleleBasesMap.keySet()) {
                     // the first -1 for zero-base (getBases) versus 1-based (variant position)
```
rickymagner, line 408: These comments should be updated

src/main/java/picard/util/LiftoverUtils.java
```diff
@@ -390,20 +390,29 @@ protected static void leftAlignVariant(final VariantContextBuilder builder, fina
...
-
-                    alleleBasesMap.put(allele, extendOneBase(alleleBasesMap.get(allele), extraBase));
+                    alleleBasesMap.put(allele, extendOneBase(alleleBasesMap.get(allele), extraBase, extendLeft));
                 }
-                changesInAlleles = true;
-                theStart--;
 
                 // 7. end if
             }
+            changesInAlleles = theStart!=oldStart || theEnd != oldEnd;
```
rickymagner, line 415: This means repeat the above block of truncating matching right bases, and handling whatever empty alleles mean, whenever there's been a change in the variant. Is a `while` block the best way to handle this? Why not just truncate all matching bases on the right in one go, and then handle this other empty edge case as well?

rickymagner (commented):
Hi Yossi, thanks for this! It seems to fix the issue well. I just made some remarks about updating the comments mostly to help make clear what is happening in the code for the next person to try to debug this tool. I know it's not your fault, but since we're here it's probably worth spending a few minutes doing. I highlighting a few key places some extra comments could help a bit, I think. In particular, examples of what sort of variant triggers the blocks would be very useful. Thanks!

yfarjoun:
As the comment for the function mentioned, the code follows the paper by G. Abecsis. You can see the pseudo code (including the numbering used here) here: https://genome.sph.umich.edu/wiki/Variant_Normalization

I'm happy to make it clearer somehow, but wanted to make sure the context was understood.... :-)

rickymagner:
Ah, good to know. In that case, it makes sense to keep in that format, but maybe this should be emphasized a bit more in the top level comments. There is a reference to the paper, but it's unclear the steps in the comments are those from the paper (at least it was to me). Maybe a sentence or two up top can clarify this.

yfarjoun:
thanks for the comments, @rickymagner I think I responded to all of them.

rickymagner (approved):
Thanks for adding a lot more context in the comments! This will surely help whomever looks at this code in the future. Looks good for merging, and thanks for handling this bug.

PR #1965, opened 2024-06-10
No more header requried for ribosomal interval files in CollectRnaSeqMetrics

jourdren:
With this pull request, CollectRnaSeqMetrics can new be used with ribosomal interval list file without a SAM header. In this case, CollectRnaSeqMetrics will use the header of the BAM input file.

### Description

Creating a valid ribosomal interval list for CollectRnaSeqMetrics is a tedious task as the @SQ tags of SAM header of this file must be **rigorously** the same as in the BAM header of the input file (e.g., the **exact** order of the chromosomes/sequences must be the same; the length of the chromosomes/sequences must be same). 

With the patch in this pull request, CollectRnaSeqMetrics will have the same behavior if a header exists in the ribosomal interval list file. If no header is found in the file, the patch will use the header of the input BAM.

This new behavior will streamline the creation of ribosomal interval list files, as it can be creating with just a GTF file without the need to add the sequence lengths at the beginning from the genome FASTA, GFF3 or input BAM header.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
+++ b/src/main/java/picard/analysis/directed/RnaSeqMetricsCollector.java
@@ -9,6 +9,7 @@ import htsjdk.samtools.util.IntervalList;
 import htsjdk.samtools.util.Log;
 import htsjdk.samtools.util.OverlapDetector;
 import htsjdk.samtools.util.SequenceUtil;
+import htsjdk.tribble.IntervalList.IntervalListCodec;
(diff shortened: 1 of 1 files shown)


yfarjoun:
Given that you can create an intervallist using BedToIntervalList, I'm not sure what problem this is solving. It is however making the tool be less discerning and will lead to folks hand-crafting interval lists without dictionaries rather than making sure that they used the correct dictionary for the file. 

👎  from me.

PR #1968, opened 2024-06-24
Update http-nio to 1.1.1

droazen:
This release fixes several bugs when using HttpPath resolve methods that have been affecting users. In particular, they were causing issues when trying to construct HTTP URIs for companion files such as fasta/bam indices.


lbergelson (approved):
👍

PR #1980, opened 2024-10-13
FLOW_* parameters must be accompanied by FLOW_MODE

dror27:
This is an additional PR to close issue 1959. As requested, The code base now checks for the presence of FLOW_MODE if any other FLOW_ parameter is specified. A test has also been added.


dror27:
I have noticed that the legacyTest is failing. Will fix

ilyasoifer:
@lbergelson  - do we want to merge this?

PR #1986, opened 2024-12-17
Fix failures to allocate arrays with a very large Java heap size

EthanNelson-Moore:
### Description

The maximum length of a Java array is not exactly `Integer.MAX_VALUE`, but slightly less due to the space taken up by the object header. The exact maximum differs depending on the platform and Java version. This was already accounted for in one instance, but not others. This was causing MarkDuplicates to fail with a Java heap size of slightly less than 1TB when `MAX_RECORDS_IN_RAM` was not specified (and therefore calculated based on the amount of available memory). This commit fixes the other instances and changes the maximum size in the existing instance to `Integer.MAX_VALUE - 32` instead of `Integer.MAX_VALUE - 5` to decrease the likelihood of allocation failures on different Java versions and platforms.

----

#### Content
- [ ] Added or modified tests to cover changes and any new functionality - *It would be hard to automatically test this change since triggering the issue requires an extremely large amount of memory, which is not likely to be the case in a CI pipeline.*
- [ ] Edited the README / documentation (if applicable) - *This is a bug fix, so there is no need to update the README or documentation, only the changelog of a future release.*

#### Review

### DIFF
--- a/src/main/java/picard/sam/markduplicates/EstimateLibraryComplexity.java
+++ b/src/main/java/picard/sam/markduplicates/EstimateLibraryComplexity.java
@@ -425,7 +425,7 @@ public class EstimateLibraryComplexity extends AbstractOpticalDuplicateFinderCom
         } else {
             sizeInBytes = PairedReadSequence.getSizeInBytes();
         }
-        MAX_RECORDS_IN_RAM = (int) (Runtime.getRuntime().maxMemory() / sizeInBytes) / 2;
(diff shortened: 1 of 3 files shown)


lbergelson:
Thank you for this.  Good catch, I've never tried to run it with that much ram so I would never have hit it.

lbergelson (approved):
Sounds right.  I'm going to take it on faith that this fixes this use case because I don't have a machine available to test it easily.

PR #2000, opened 2025-02-27
make MergeVcfs (more) deterministic

kachulis:
MergeVcfs currently stores input headers in a HashSet, so that the order they are passed to be merged is non-deterministic, which can lead to the resulting merged header being non-deterministic.  This PR changes that so that (hopefully) the results of MergeVcf are always the same.


lbergelson (approved):
Better than before!

PR #2006, opened 2025-04-15
Fix a critical bug in flow base duplicate marking

ilyasoifer:
### Description

The mechanism that dumped the read ends database to the disk was broken for single ended reads and practically once the read ends were spilled to the disk, the second end was ignored. 
We never discovered this bug before because we were using MarkDuplicatesSpark from the GATK. 
I changed the mechanism of spilling both read ends to the disk in case the read end location is set as significant. 

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/sam/markduplicates/MarkDuplicates.java
+++ b/src/main/java/picard/sam/markduplicates/MarkDuplicates.java
@@ -493,6 +493,9 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram imp
             pairCodec = new ReadEndsForMarkDuplicatesCodec();
             diskCodec = new ReadEndsForMarkDuplicatesCodec();
         }
+
+        //In Illumina the two ends of the fragment are saved only if the read is paired
+        // in flow based sequencing we are interested in both ends of the read
         if (flowBasedArguments.FLOW_USE_END_IN_UNPAIRED_READS){
             fragCodec.setForceSerialization();
             pairCodec.setForceSerialization();


ilyasoifer:
@lbergelson - would it be possible to merge this? I think it was a significant error.

src/main/java/picard/sam/markduplicates/util/ReadEndsForMarkDuplicates.java
```diff
@@ -62,7 +63,7 @@ public ReadEndsForMarkDuplicates(final ReadEndsForMarkDuplicates read) {
         this.read1Coordinate = read.read1Coordinate;
         this.read2ReferenceIndex = read.read2ReferenceIndex;
         this.read2Coordinate = read.read2Coordinate;
-
+        this.read2CoordinateRequiresSerialize = read.read2CoordinateRequiresSerialize;
```
lbergelson, line 66: I might rename this to either  `serializeRead2Coordinate` or `read2CoordinateRequiresSerialization` to just be a bit more idiomatic.

src/main/java/picard/sam/markduplicates/MarkDuplicatesForFlowHelper.java
```diff
@@ -177,6 +177,7 @@ public ReadEndsForMarkDuplicates buildReadEnds(final SAMFileHeader header, final
         ends.read1Coordinate = getReadEndCoordinate(rec, !rec.getReadNegativeStrandFlag(), true, md.flowBasedArguments);
         if (md.flowBasedArguments.FLOW_USE_END_IN_UNPAIRED_READS) {
             ends.read2Coordinate = getReadEndCoordinate(rec, rec.getReadNegativeStrandFlag(), false, md.flowBasedArguments);
+            ends.read2CoordinateRequiresSerialize = true;
```
lbergelson, line 180: Can you add a comment explaining why this is necessary here for future reference?

src/main/java/picard/sam/markduplicates/util/ReadEndsForMarkDuplicatesCodec.java
```diff
@@ -54,12 +54,13 @@ public void encode(final ReadEndsForMarkDuplicates read) {
             this.out.writeShort(read.score);
             this.out.writeShort(read.libraryId);
             this.out.writeByte(read.orientation);
+            this.out.writeByte(read.read2CoordinateRequiresSerialize ? 1 : 0);
```
lbergelson, line 57: I'm not totally sure you really want to serialize this every time.  It might make more sense to add an optional constructor parameter to the codec and keep track of it that way.  I haven't tried to thread it through everywhere to see how awful that is though, so maybe this is the best way.

lbergelson (changes requested):
@ilyasoifer Thanks for this patch.  I have a few comments, a test with an example flow file would be great to help keep this from decaying in the future.  If that's not feasible let me know.

src/main/java/picard/sam/markduplicates/MarkDuplicatesForFlowHelper.java
ilyasoifer: done

lbergelson (approved):
@ilyasoifer This seems cleaner.  Thank you.  I'm assuming you tested it and it works correctly on your files?   I'm going to merge this, but without a test there's no guarantee it won't be broken sometime in the future by someone who doesn't know what it's about.

PR #2010, opened 2025-05-05
Added some extra handling for when R is not available, specifically when running in a GATK docker that does not have it

KevinCLydon:
- Modified a few tools that require R for plots to make the plots optional
- Added checks in all tools that use R to print an appropriate error message if running in the GATK docker that doesn't have R
- Filled in tests for some tools that use R but did not yet have tests


src/test/java/picard/analysis/CollectInsertSizeMetricsTest.java
```diff
@@ -222,6 +224,222 @@ else if (metrics.READ_GROUP.equals("62A79AAXX100907.3")) {
...
+                "LEVEL=READ_GROUP",
+        };
+        Assert.assertEquals(runPicardCommandLine(args), 0);
+
+        Assert.assertEquals(pdf.length(), 0);
+
+        final MetricsFile<InsertSizeMetrics, Comparable<?>> output = new MetricsFile<InsertSizeMetrics, Comparable<?>>();
+        output.read(new FileReader(outfile));
+
+        for (final InsertSizeMetrics metrics : output.getMetrics()) {
```
lbergelson, line 248: In general if this sort of block of checks is repeated between different between tests its probably best to extract it into a function.  It helps test readers a lot because it helps clarify which thing is being tested specifically.   No need to go back and change it now though.

src/test/java/picard/analysis/CollectRrbsMetricsTest.java
```diff
@@ -0,0 +1,162 @@
...
+import java.io.IOException;
+import java.io.PrintStream;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import htsjdk.samtools.metrics.MetricsFile;
+import picard.cmdline.CommandLineProgramTest;
+
+public class CollectRrbsMetricsTest extends CommandLineProgramTest {
```
lbergelson, line 16: Thanks for adding these.

lbergelson (changes requested):
@KevinCLydon This looks good to me.  I have a few minor nitpicks and then it's good to go.  Thanks for adding all of those tests for the untested classes.

src/main/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -162,7 +165,8 @@ protected String[] customCommandLineValidation() {
     /////////////////////////////////////////////////////////////////////////////
     @Override
     protected void setup(final SAMFileHeader header, final File samFile) {
-        IOUtil.assertFileIsWritable(CHART_OUTPUT);
+        if(CHART_OUTPUT != null)
```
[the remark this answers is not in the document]
lbergelson: You don't have to go back and change this... this is just like, my opinion maaan.

lbergelson (approved):
@KevinCLydon Thank you.

PR #2021, opened 2025-07-24
 Yf make revert sam robust to errant XQ and XB tags

### LINKED ISSUE #2020: Errant XQ tag confuses RevertSam
TL;DR: the dragen pipeline addes XQ tag to reads which then makes it difficult to revert using RevertSam

1. RevertSam looks for hard clipped bases in the XB tag and their qualities in the XQ tag. These tags are expected to be of a certain type (string) and if found to be of a different type RevertSam crashes. 
2. The dragen-processed crams contain the XQ tag, but it is used for something else, and contains an integer. Since the XQ tag starts with "X", accoding to the sam-spec, it can be used by users however they like. 
3. Running RevertSam on a dragen-processed cram causes a crash (reported [here](https://github.com/Illumina/DRAGMAP/issues/52) ) with the only workaround to run with RESTORE_HARDCLIPS=false. 

I think that RevertSam should be more lenient when it comes to user-defined tags. If it finds XQ and XB that are both strings, and match in length, it should attempt to restore the hard-clipped bases. Otherwise it should ignore these tags.

yfarjoun:
fixes #2020 so that RevertSamFile doesn't choke on XB or XQ even if they are not strings.

- checks that both are present before attempting to revert hard-clips
- obtains values in a try-catch block
- emits (up to 100) warning for problematic values of XB and XQ (only if both are present)
- test included

### Description

----

#### Content

#### Review


yfarjoun:
Thanks @kockan I'm no longer a maintainer, so someone else will have to do the honors...

Issue #2026, opened 2025-10-06
CollectMultipleMetrics crashes on improperly paired reads


Redmar-van-den-Berg:
# Description
It appears that CollectMultipleMetrics crashes when read pairs are mapped to multiple chromosomes. Ideally CollectMultipleMetrics would be able to handle this, since improperly paired reads a common in big sequencing projects.

# Picard version
The issue occurs in Picard 3.4.0, this is a regression from 3.3.0

# Error message
```text
/usr/local/bin/picard: line 5: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8): No such file or directory
INFO	2025-10-06 13:19:20	CollectMultipleMetrics	

********** NOTE: Picard's command line syntax is changing.
**********
********** For more information, please see:
********** 
https://github.com/broadinstitute/picard/wiki/Command-Line-Syntax-Transition-For-Users-(Pre-Transition)
**********
********** The command line looks like this in the new syntax:
**********
**********    CollectMultipleMetrics -PROGRAM null -PROGRAM CollectAlignmentSummaryMetrics -PROGRAM CollectInsertSizeMetrics -PROGRAM RnaSeqMetrics -VALIDATION_STRINGENCY LENIENT -R test/data/reference/hamlet-ref.fa -I SRR8615409/snv-indels/SRR8615409.bam -REF_FLAT test/data/reference/hamlet-ref.refFlat -EXTRA_ARGUMENT RnaSeqMetrics::--RIBOSOMAL_INTERVALS test/data/reference/ucsc_rrna.refFlat -EXTRA_ARGUMENT RnaSeqMetrics::--STRAND_SPECIFICITY NONE -USE_JDK_DEFLATER true -USE_JDK_INFLATER true -O SRR8615409/snv-indels/SRR8615409
**********


INFO	2025-10-06 13:19:20	RExecutor	Executing R script via command: Rscript /tmp/script7071702731358479342.R
ERROR	2025-10-06 13:19:20	ProcessExecutor	/bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
ERROR	2025-10-06 13:19:20	ProcessExecutor	sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
ERROR	2025-10-06 13:19:20	ProcessExecutor	sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
[...]

PR #2027, opened 2025-10-06
protect collect rnaseqmetrics from pair orientation assert

yfarjoun:
### Description

[fixes](https://github.com/broadinstitute/picard/commit/773aae04c1c75578d47463793c4cce12c989b16b) https://github.com/broadinstitute/picard/issues/2026 

Protect code from calling getPairOrientation() when references of read and mate are not equal (as that now throws an exception)

- Also added similar protection to CollectInsertSizeMetrics

- Tests added.


yfarjoun:
I can't seem to hit the "merge" button...please be my guest @kockan

PR #2030, opened 2026-01-09
Filter reads for  gc bias (mapq and interval)

bwlang:
### Description

I added a filtration option to the GC bias tool to allow users to exclude known problematic regions (e.g. [from boyle et. al.](https://www.biorxiv.org/content/10.1101/2022.11.21.517407v1.full.pdf)) and clean up "spikes" in the GC bias due to poly-G stretches mapping or adapter constructs, etc. I also added a min mapq feature for similar reasons.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/analysis/CollectGcBiasMetrics.java
+++ b/src/main/java/picard/analysis/CollectGcBiasMetrics.java
@@ -34,9 +34,6 @@ import htsjdk.samtools.util.IOUtil;
 import htsjdk.samtools.util.Interval;
 import htsjdk.samtools.util.IntervalList;
 import htsjdk.samtools.util.Log;
-import htsjdk.tribble.AbstractFeatureReader;
-import htsjdk.tribble.CloseableTribbleIterator;
-import htsjdk.tribble.FeatureReader;
 import htsjdk.tribble.bed.BEDCodec;
 import htsjdk.tribble.bed.BEDFeature;
 import org.broadinstitute.barclay.argparser.Argument;
@@ -193,13 +190,9 @@ public class CollectGcBiasMetrics extends SinglePassSamProgram {
         IntervalList intervalsToExclude = null;
         if (EXCLUDE_INTERVALS != null) {
             IOUtil.assertFileIsReadable(EXCLUDE_INTERVALS);
-            // IntervalList.fromFile works for interval_list format
-            // For BED files, we need to manually load using BEDCodec with the sequence dictionary
-            if (EXCLUDE_INTERVALS.getName().endsWith(".bed")) {
-                intervalsToExclude = loadBedFile(EXCLUDE_INTERVALS, header.getSequenceDictionary());
-            } else {
-                intervalsToExclude = IntervalList.fromFile(EXCLUDE_INTERVALS);
-            }
+            // Always buffer content and detect format by examining the first line.
+            // This works for both regular files and special files (pipes, FIFOs, process substitutions).
+            intervalsToExclude = loadIntervalsFromStream(EXCLUDE_INTERVALS, header.getSequenceDictionary());
 
             // Log information about excluded regions
             final int numExcludedRegions = intervalsToExclude.getIntervals().size();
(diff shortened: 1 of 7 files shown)


bwlang:
I've also tested with larger bam files and small intervals. The interval checking overhead is less than 1% and only when an interval argument is provided.

bwlang:
@yfarjoun : it was before your #2031 PR, but i think I followed the recommendations it outlines.

bwlang:
@yfarjoun do you have any sense of whether this is likely to be merged or needs attention?

src/main/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -221,6 +254,126 @@ private void writeResultsToFiles() {
...
+     * Uses mark/reset with an 8KB buffer for interval_list format (which has headers).
+     * For BED format, only reads one line before detecting format.
+     */
+    private static IntervalList loadIntervalsFromStream(final File file, final SAMSequenceDictionary dictionary) {
+        FormatDetectionResult detectedFormat;
+        IntervalList result;
+
+        try (final java.io.BufferedReader reader = new java.io.BufferedReader(new java.io.FileReader(file))) {
+            // Mark with 8KB buffer - enough for typical interval_list headers
+            reader.mark(8192);
```
yfarjoun: this needs to be enough for what? just the first line, no? the comment makes is sounds like you need to read the entire header.

src/main/java/picard/analysis/CollectGcBiasMetrics.java
```diff
@@ -221,6 +254,126 @@ private void writeResultsToFiles() {
...
+            throw new PicardException("Error reading intervals from " + file, e);
+        }
+
+        return result;
+    }
+
+    /**
+     * Parse BED format from a reader using BEDCodec.
+     */
+    private static IntervalList parseBedFromReader(final java.io.BufferedReader reader, final SAMSequenceDictionary dictionary)
```
yfarjoun: There already is a bed reader in BedToIntervalList (it uses htsjdk's AbstractFeatureReader). perhaps it makes sense to extract the reader from there and place it in a standalone class? otherwise we will have two readers.

src/test/java/picard/analysis/CollectGcBiasMetricsTest.java
```diff
@@ -566,4 +569,193 @@ public void testWithoutChartOutput() throws IOException {
...
+        for (final GcBiasSummaryMetrics metrics : outputMapq3.getMetrics()) {
+            if (metrics.ACCUMULATION_LEVEL.equals(ACCUMULATION_LEVEL_ALL_READS)) {
+                alignedReadsMapq3 = metrics.ALIGNED_READS;
+            }
+        }
+
+        // AlignedAdapterReads.sam has 1 read with MAPQ=0 and 1 read with MAPQ=3
+        // With MIN_MAPQ=1, we should get 1 read (MAPQ=0 filtered out, MAPQ=3 included)
+        // With MIN_MAPQ=3, we should get 1 read (only the MAPQ=3 read)
+        Assert.assertEquals(alignedReadsMapq1, 1, "Expected 1 aligned read with MAPQ >= 1");
```
yfarjoun: perhaps run it with 0, 1,3 & 4 so show that the filtering really works? (can use a data provider)

src/test/java/picard/analysis/CollectGcBiasMetricsTest.java
```diff
@@ -566,4 +569,193 @@ public void testWithoutChartOutput() throws IOException {
...
+                "CHART_OUTPUT=" + pdf1.getAbsolutePath(),
+                "SCAN_WINDOW_SIZE=100",
+                "MINIMUM_GENOME_FRACTION=1.0E-5",
+                "IS_BISULFITE_SEQUENCED=false",
+                "LEVEL=ALL_READS",
+                "ASSUME_SORTED=true",
+                "EXCLUDE_INTERVALS=" + intervalsFile.getAbsolutePath()
+        };
+        runPicardCommandLine(argsIntervalList);
+
```
yfarjoun, line 680: split this into two tests please.

src/test/java/picard/analysis/CollectGcBiasMetricsTest.java
```diff
@@ -566,4 +569,193 @@ public void testWithoutChartOutput() throws IOException {
...
+        bedFile.deleteOnExit();
+
+        // Create BED file with headers (0-based, half-open: [start, end))
+        // To exclude 330-340 (1-based inclusive), we write 329-340 in BED format
+        // Include all standard BED header types to test header parsing
+        try (final FileWriter writer = new FileWriter(bedFile)) {
+            writer.write("# This is a comment line\n");
+            writer.write("# Another comment with column descriptions: chrom, start, end\n");
+            writer.write("track name=\"Test Track\" description=\"Test intervals for GC bias\"\n");
+            writer.write("browser position chrM:1-16571\n");
```
yfarjoun: I'm not familiar with this as a valid BED format. the formalized version is here: https://samtools.github.io/hts-specs/BEDv1.pdf

yfarjoun: why generate the BED file programmatically? why not add it to an appropriate subdirectory in `testdata/picard` ? (same question for the interval list)

yfarjoun (changes requested):
Thanks for this PR @bwlang and sorry for the slow review process. 

I have some questions and suggestions, but overall, I think that this would be a welcome addition.

src/main/java/picard/analysis/CollectGcBiasMetrics.java
bwlang: whups i meant enough to get through bed headers not interval_list headers... interval_list is easy just @

src/test/java/picard/analysis/CollectGcBiasMetricsTest.java
bwlang: I've seen these in the wild... and it's documented here: https://grch37.ensembl.org/info/website/upload/bed.html
I'll add some comments to that effect unless you'd rather just reject such files.

I thought i got feedback in an earlier PR (from Nils i think) that a generated version was preferred - maybe that was just for that case. I'm fine either way. 

I'll add the bed parser and test it with a file if you think that's clearest.

yfarjoun: In the page you are referenceing that example seems to be under the "bedGraph" (not bed) format. I'd rather go by the (now) official format from hts-specs.

yfarjoun: regarding the generated vs. stored file. I don't have strong feelings. I find the stored file cleaner (no need to delete and clutter the test code with generation lines), and one can simply look at the file to understand what's in it. The downside is that it can result in a collection of odd files in the testdata directory that no-one knows exactly what test they are for.... perhaps this needs to be decided at an "official" level @jonn-smith?

bwlang: > In the page you are referenceing that example seems to be under the "bedGraph" (not bed) format. I'd rather go by the (now) official format from hts-specs.

I think the portion that describes the track lines is in the bed section, it's just that they indicate them to be "compulsory" in the bedgraph section.  I also think it's strange to have these reserved words (why not just put them behind a comment character?), but on the other hand it's annoying for users to have to munge their input files...

I'll remove track and accept only # or blank lines unless you change your mind.

bwlang: one more "bed  specification" : https://genome.ucsc.edu/FAQ/FAQformat.html#format1  (allows "track" and "browser" ) headers.

bwlang: the latest push removes the track support but keeps the "generator" pattern for bed test... let me know how you want to go there.

yfarjoun: I would rather stick to the GA4GH specs and if someone complains, we can point them there... 🤷 

since @nh13 asked for the generator pattern, I'm not going to ask you to change back.. :-)

yfarjoun (changes requested):
it seems that there's a rebase problem, or perhaps your version needs rebaseing, but I see lots of irrelevant (and new) changes...

also, I did not mean that your entire code (with the bed/interval-list sniffing) should be plopped into BedToIntervallist. what I meant is that as you can see in that class there's code for reading a bed file and generating features. I suggest that you take that (small) part, extract it from BedToIntervalList (into a new utility class) and call the appropriate method from _both_ BedToIntervalList and CollectCGBias (once you know you need to read a bed file). 

Sorry for not being clear about this the first time.

bwlang:
Ok - I removed the stray commit and extracted the bed/intervallist reading to a utility class.
@yfarjoun : what do you think?

yfarjoun:
Thanks for the work here @bwlang.

I think there's been a miscommunication about the utility class extraction. The intent of my earlier suggestion was to take the *existing* BED-reading code, which uses htsjdk's `BEDCodec` / `AbstractFeatureReader` (Tribble), and move it into a shared utility so that both `BedToIntervalList` and `CollectGcBiasMetrics` can call the same code path. What happened instead is that `BEDCodec` was dropped entirely and replaced with a hand-rolled tab-split parser in `IntervalFileReader.fromBed`.

This is concerning because:

1. **Regression risk for `BedToIntervalList`.** `BEDCodec` handles various edge cases and has been in use for years. Replacing it with a new parser as a side-effect of a GC bias PR is a big behavioral change for a widely-used tool.
2. **The new parser is less robust.** It doesn't handle things like space-separated fields, track/browser lines (which `BEDCodec` silently skips), or other quirks that real-world BED files have.

What I'd suggest instead:

- Keep `BEDCodec` / `AbstractFeatureReader` as the actual BED parser.
- The utility class (`IntervalFileReader` in your implementaiton) should wrap that: open the file with Tribble, iterate `BEDFeature`s, do the coordinate conversion (0-based half-open → 1-based closed) and validation, and return an `IntervalList`.
- `BedToIntervalList.doWork()` calls that wrapper (so its behavior is unchanged apart from being factored out).
- `CollectGcBiasMetrics.setup()` also calls that wrapper when `EXCLUDE_INTERVALS` looks like a BED file.
- The format-detection logic you wrote is fine to keep for sniffing BED vs interval_list.

One concession: I know in the earlier discussion we landed on rejecting `track`/`browser` lines to stick to the GA4GH BED spec. As `BEDCodec` silently skips those lines rather than rejecting them, I'm fine with that permissive behavior.

This way you get the shared code, and the new GCBias feature, without changing the existing tool.

Happy to discuss if this and hope that it doesn't seem to onerous.

bwlang:
@yfarjoun i finally got back to this...does this match more with what you are thinking?

bwlang:
@yfarjoun : friendly monday morning ping in case you want a quick hit of progress to start your week :)

yfarjoun:
Sorry for the silence. I was on leave for the past two weeks with minimal access my email and related accounts.

bwlang:
@yfarjoun : is it good to find a time to chat about this and resolve any remaining issues live?

tfenne:
@bwlang I took the liberty of implementing these features in Riker [here](https://github.com/fulcrumgenomics/riker/pull/31).  They differ slightly, specifically in:
- Also filters the window counts in the denominator
- Only excludes reads whose computed "bin start" is in the exclude regions (so depending on use, may want to pad the exclude region).

I can merge the change in Riker fairly quickly if it works in your hands.

PR #2034, opened 2026-01-31
switches from short to int for physical locations (avoiding overflow and inaccurate markdups)

### LINKED ISSUE #1441: MarkDuplicates Missing Optical Duplicates

bwlang:
This accounts for larger flowcells with no significant impact on memory use and fixes #1441 

I measured these by capturing maximum resident set size values on a mac like this:

/usr/bin/time -l pixi run java -Xmx512m -jar build/libs/picard.jar MarkDuplicates -I chr18_test.bam -O /dev/null -M test.out --VALIDATION_STRINGENCY LENIENT 2>&1  | grep maximum 

with 

Version | Run 1 (MB) | Run 2 (MB) | Run 3 (MB) | Avg (MB)
-- | -- | -- | -- | --
short (old) | 467.8 | 478.7 | 496.4 | 481.0
int (new) | 489.6 | 490.6 | 492.5 | 490.9

I think the short implementation was not effective with these garbage collection settings since the data started as ints before being truncated to shorts in the overrides of the setters in PhysicalLocationShort.

This seems simpler than the approach in #1442

----

#### Content

#### Review


bwlang:
@nh13  not sure if you have energy for a review ... i noticed your footprints while i was wandering around in here, might be quick for you?

nh13 (changes requested):
I think we can get rid of all instances of `PhysicalLocationShort`, and the last place I could find was `ReadNameParserTests.java:88`, which is ironically to test overflow.

I think we can also programmatically generate the BAM.  I had "AI" generate it for me, so it's probably wise to test the unit test _before_ the fix to confirm it fails.

<details>

<summary>test.java</summary>

```java
/**
 * Tests that optical duplicate detection works correctly when read coordinates
 * exceed Short.MAX_VALUE (32767). Creates 3 duplicate read pairs on tile 2214
 * with coordinates x=(32579, 32588, 32768) y=(36746, 36761, 36824).
 *
 * All three pairs are within 2500 pixels of each other. When x=32768 is
 * truncated to short, it overflows to -32768, causing the third pair to
 * appear ~65000 pixels away and be missed as an optical duplicate.
 *
 * Expected: 2 optical duplicates (with int coordinates)
 * Bug case: 1 optical duplicate (with short overflow)
 */
@Test
public void testOpticalDuplicateDetectionWithCoordinateOverflow() {
    final AbstractMarkDuplicatesCommandLineProgramTester tester = getTester();

    final int tile = 2214;
    final int[][] coordinates = {
        {32579, 36746},  // x fits in short
        {32588, 36761},  // x fits in short
        {32768, 36824}   // x=32768 overflows to -32768 as short
    };

    for (int i = 0; i < coordinates.length; i++) {
        int x = coordinates[i][0];
        int y = coordinates[i][1];
        String readName = String.format("HWUSI:1:FLOWCELL:1:%d:%d:%d", tile, x, y);
        boolean isDuplicate = (i > 0);

        tester.addMatePair(readName, 0, 100, 200, false, false, isDuplicate, isDuplicate,
                "50M", "50M", false, true, false, false, false, DEFAULT_BASE_QUALITY);
    }

    tester.addArg("OPTICAL_DUPLICATE_PIXEL_DISTANCE=2500");
    tester.setExpectedOpticalDuplicate(2);
    tester.runTest();
}
```

<details

bwlang:
@nh13 : ok - i cleaned up as you suggested and confirmed that the programmatic test failed against origin/master + test only.
ready?

nh13 (approved):
Thanks for fixing up!

PR #2035, opened 2026-02-03
Add molecular position stratification to CollectSamErrorMetrics

yfarjoun:
### Description                                                                                                        
   - Adds a new "molecular position" stratification for ReadBaseStratification                                                                                                                                       
   - Unlike cycle position, this stratification provides consistent answers for overlapping bases from paired reads                                                                                                  
   - Includes tests for the new stratification                                                                                                                                                   
                                                                                                                                                                                                                     
----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/sam/SamErrorMetric/ReadBaseStratification.java
+++ b/src/main/java/picard/sam/SamErrorMetric/ReadBaseStratification.java
@@ -77,8 +77,8 @@ public class ReadBaseStratification {
     /**
      * defaults to 2500
      **/
-    public static void setLocationBinSize(int locationBinSize) { 
-       LOCATION_BIN_SIZE = locationBinSize; 
+    public static void setLocationBinSize(int locationBinSize) {
+        LOCATION_BIN_SIZE = locationBinSize;
     }
 
     /* ******* general-use classes, for defining and creating new stratifiers ***********/
@@ -116,6 +116,7 @@ public class ReadBaseStratification {
         public Integer stratify(RecordAndOffset recordAndOffset, SAMLocusAndReference locusInfo) {
             return stratify(recordAndOffset.getRecord());
         }
+
         //Static ReadNameParser so that cache of read names/ PhysicalLocations is shared between all PositionBasedStratifiers
         static final ReadNameParser readNameParser = new ReadNameParser();
 
@@ -576,9 +577,10 @@ public class ReadBaseStratification {
 
         /**
          * Returns the number of bases associated with I and D CIGAR elements.
+         *
          * @param samRecord The read to investigate
          * @return The number of bases associated with I and D CIGAR elements, or null if the evaluation of either
-         *         operation caused an error
+         * operation caused an error
          */
         @Override
         public Integer stratify(final SAMRecord samRecord) {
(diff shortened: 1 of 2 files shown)


src/main/java/picard/sam/SamErrorMetric/ReadBaseStratification.java
```diff
@@ -839,6 +846,7 @@ enum Stratifier implements CommandLineParser.ClpEnum {
         READ_GROUP(() -> readgroupStratifier, "The read-group id of the read."),
         CYCLE(() -> baseCycleStratifier, "The machine cycle during which the base was read."),
         BINNED_CYCLE(() -> binnedReadCycleStratifier, "The binned machine cycle. Similar to CYCLE, but binned into 5 evenly spaced ranges across the size of the read.  This stratifier may produce confusing results when used on datasets with variable sized reads."),
+        MOLECULAR_POS(() -> baseMolecularPosStratifier, "The molecular position of the read-base (distance from the end of the molecule, negative if closer to read 2's 3' end)."),
```
tfenne: I think this could use some more explanation - perhaps you could give an example?  E.g. if I have a 200bp insert with 2x75bp reads, what do I get for the 1st base R1 and the first base in R2?

tfenne: Now having read the javadoc on the implementation I have three thoughts:

1. I think the name should imply more what's being done - my nomination would be some version of {molecule|insert|template} x {endproximity|enddistance|etc.} to make it clear that it's the distance to the nearest end of the molecule.
2. I think given that the values produced are not immediately obvious, I think I would increase the docs to include "Values are 1..n for bases closer to the R1 end of the molecule and -1..-n for bases closer to the R2 end of the molecule."
3. I think the docs here should also state clearly that the stratifier only works on PE data with both reads mapped in FR orientation and a reasonable insert size.

src/test/java/picard/sam/SamErrorMetric/ReadBaseStratificationTest.java
```diff
@@ -537,9 +537,210 @@ public void testGetIndelElement(final String cigar, final int offset, final Abst
...
+    }
+
+    @Test(dataProvider = "molecularPositions")
+    public void testTestMolecularPositionStratifier(int offset, boolean read1, int expected_mol_pos) {
+        final SAMSequenceRecord samSequenceRecord = new SAMSequenceRecord("chr1", 2_000);
+        final SAMFileHeader samFileHeader = new SAMFileHeader();
+        samFileHeader.addSequence(samSequenceRecord);
+        final SAMReadGroupRecord readGroupRecord = new SAMReadGroupRecord("rgID");
+        samFileHeader.addReadGroup(readGroupRecord);
+
```
tfenne:
```suggestion
```

tfenne (changes requested):

yfarjoun:
Thanks @tfenne. Back to you with changes.

src/main/java/picard/sam/SamErrorMetric/ReadBaseStratification.java
```diff
@@ -839,6 +846,9 @@ enum Stratifier implements CommandLineParser.ClpEnum {
         READ_GROUP(() -> readgroupStratifier, "The read-group id of the read."),
         CYCLE(() -> baseCycleStratifier, "The machine cycle during which the base was read."),
         BINNED_CYCLE(() -> binnedReadCycleStratifier, "The binned machine cycle. Similar to CYCLE, but binned into 5 evenly spaced ranges across the size of the read.  This stratifier may produce confusing results when used on datasets with variable sized reads."),
+        INSERT_END_DISTANCE(() -> baseInsertEndDistanceStratifier, "Distance from the nearest insert end for FR read pairs. " +
+                "Negative if closer to read 2's 3' end, e.g., 150x2 reads with 100bp overlap will have values (R1) 51, .., 100, -100, .., -51 for the overlapping region. " +
```
tfenne, line 850: Two things: 
- I think it's clearer to say "closer to the start of read 2" because I interpret the "3' end of read 2" to mean the last base in read 2, not the first
- Does this really only output values in the overlapped region?  I thought it output values for all positions so long as you had an FR insert?

tfenne (approved):

PR #2041, opened 2026-03-24
Fix to CollectSamErrorMetrics getIndelOperator function to check operator type

tfenne:
### Description

The getIndelOperator function in ReadBaseStratifier takes a `RecordAndOffset` and is intended to return either the Insertion or Deletion Cigar element found in that read's cigar at that offset, depending on whether the RecordAndOffset represents an insertion of deletion operation.

It appears to function correctly if and only if passed RecordAndOffsets that are also correct - i.e. correctly identify an insertion or deletion within a read at an accurate offset.  However, when an incorrect offset is given for an insertion (as triggered by [this bug](https://github.com/samtools/htsjdk/pull/1758) in HTSJDK, it will erroneously return any cigar element with an operator that consumes read bases.  

This PR adds a test to highlight that invalid input causes this problem, and ensure that in that case `null` is returned instead of an incorrect cigar element.
----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/sam/SamErrorMetric/ReadBaseStratification.java
+++ b/src/main/java/picard/sam/SamErrorMetric/ReadBaseStratification.java
@@ -1273,8 +1273,8 @@ public class ReadBaseStratification {
             }
 
             if (recordAndOffset.getAlignmentType() == AbstractRecordAndOffset.AlignmentType.Insertion) {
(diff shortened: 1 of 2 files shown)


yfarjoun (approved):
should this be logged? (can it be easily logged ?)

seems like if sli isn't buggy, this should never happen...

PR #2043, opened 2026-04-25
Fix tests broken under i) upcoming htsjdk 5, and ii) on non-x86 architectures.

tfenne:
IntelInflaterDeflaterLoadTest's checkIntelSupported() only skipped ppc64le, so the two tests that actually load the Intel native library failed on aarch64 (e.g. Apple Silicon). Generalize the arch check to skip on anything that isn't x86_64/amd64.  (These fail on master).

ViewSam streamed records by writing rec.getSAMString() back-to-back, relying on the trailing '\n' that older htsjdk included in that string. htsjdk 5.0 intentionally drops that newline, so adjacent records were being concatenated on a single line, producing malformed SAM. This broke ViewSamTest.testIntervals and the three PipedDataTest cases (which pipe ViewSam | {SortSam, RevertSam, MergeBamAlignment}). Fix by appending '\n' only when the returned string doesn't already end with one, so the code works against both pre- and post-5.0 htsjdk releases.

#### Content

#### Review


yfarjoun (approved):
LGTM!

nh13 (approved):
LGTM

PR #2048, opened 2026-05-16
LiftoverVcf: add optional PROGRESS_JSON NDJSON sidecar output

### LINKED ISSUE #2046: Stable structured (JSON) ProgressLogger output mode for downstream parsers
### Summary

Downstream tools that surface Picard progress today have to parse `ProgressLogger`'s human-readable log format with regex, which isn't documented as a stable API. A small refactor of the log message would silently break any such parser.

A CLI flag (e.g. `--progress-json /path/to/progress.json`) that **also** emits structured progress events to a sidecar file (one JSON object per line / NDJSON) would give downstream tools a stable contract.

### Context

We parse Picard's LiftoverVcf `ProgressLogger` output via regex in an external ancestry pipeline to surface fraction-complete progress in the UI:

```
INFO  2024-01-01 12:00:00  ProgressLogger - Processed 1,000,000 records.
    Elapsed time: 00:01:00s. ... Last read position: chr3:148,946,422
```

The parser works fine today, but there's no contract that the format won't change — and a one-character whitespace tweak in `ProgressLogger` would silently break us. JSON sidecars would let us subscribe to a stable schema instead.

### Suggested JSON shape (one event per line, NDJSON)

```json
{"timestamp": "2026-05-16T12:34:56Z", "stage": "liftover", "records_processed": 1000000, "records_rejected": 12345, "last_position": "chr3:148946422", "elapsed_seconds": 60}
```

The `records_rejected` field would pair nicely with the companion issue requesting a live rejected counter in `LiftoverVcf`.

### Implementation note (where does this land?)

`ProgressLogger` itself is defined in `samtools/htsjdk` at [`src/main/java/htsjdk/samtools/util/ProgressLogger.java`](https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/samtools/util/ProgressLogger.java), so the JSON sink probably belongs there. The `--progress-json` CLI flag wiring would live here in `picard`. Filing in this repo because the user-facing UX is Picard's; happy to be redirected to `samtools/htsjdk` if that's the preferred starting point for the conversation.

### Priority

Low — could become a PR if there's interest. The current `ProgressLogger` output is fine for humans, and downstream parsers can absorb format changes when they happen; this is about giving consumers a stable subscription point so they don't have to.

carstenerickson:
Resolves #2046.

Tooling that wants a progress signal from `LiftoverVcf` today has to regex-parse the human-readable `ProgressLogger` output — not a documented stable API. This adds an optional `PROGRESS_JSON=<file>` argument; when set, one NDJSON event is appended per ProgressLogger milestone in the read and write phases, plus a final `read_complete` / `write_complete` event so small inputs still emit something.

```json
{"timestamp":"2026-05-16T12:34:56Z","stage":"read","records_processed":1000000,"records_rejected":12345,"last_position":"chr3:148946422","elapsed_seconds":60}
```

Default is `null`; existing `ProgressLogger` output is unchanged.

**Design notes** (happy to revise):

1. **`last_position` advances on every input variant**, including ones early-`continue`'d as rejections; `ProgressLogger`'s "Last read position" only advances on successful records. Intentional — JSON consumers want to know where in the input the run is, not where the last success landed.
2. **Writer leaks on early-return paths** (`CREATE_INDEX`+`DISABLE_SORT` conflict, missing contig with `WARN_ON_MISSING_CONTIG=false`). Matches the existing pattern; `rejectedRecords`, `in`, and `acceptedRecords` also leak there. Happy to refactor to try/finally.
3. **`records_rejected: 0` in write-phase events** for schema stability. Could omit instead.
4. **Hand-rolled JSON** — no existing JSON dep in `build.gradle`. Formatter is ~20 lines and unit-tested. Happy to switch to gson/jackson.

**Tests:** `LiftoverVcfProgressJsonTest` covers the formatter and escaper (null last_position, embedded quotes, control chars, empty string). `LiftoverVcfTest.testProgressJsonSidecarIsWritten` is the end-to-end integration test; `testNoProgressJsonByDefault` confirms default-off is unchanged.

### DIFF
--- a/src/main/java/picard/util/ProgressJsonWriter.java
+++ b/src/main/java/picard/util/ProgressJsonWriter.java
@@ -1,0 +1,132 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2026 The Broad Institute
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package picard.util;
+
+import picard.PicardException;
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+import java.nio.charset.StandardCharsets;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Locale;
+
+/**
+ * Writes progress events as JSON Lines (NDJSON): one UTF-8, LF-terminated JSON object per event.
+ * Gives downstream tooling a stable parsing target instead of regex-scanning human-readable logs.
+ * Each event is flushed immediately so the file can be tailed while the producing tool runs.
+ */
+public final class ProgressJsonWriter implements AutoCloseable {
+
+    private final File file;
+    private final Writer writer;
+
+    /** Opens {@code file} for writing, truncating any existing content. */
+    public ProgressJsonWriter(final File file) {
+        this.file = file;
+        try {
+            // Plain UTF-8 stream (not IOUtil's gzip-by-extension wrapping) so the sidecar can be
+            // tailed line-by-line while the tool runs.
+            this.writer = new BufferedWriter(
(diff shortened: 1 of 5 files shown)


carstenerickson:
Self-review fixup pushed (79a26759f). Three changes:

1. **Moved the `PROGRESS_JSON` PrintWriter open past `CREATE_INDEX`/`DISABLE_SORT` mutex check and the missing-sequence-dictionary check** so the user's `PROGRESS_JSON` path is no longer truncated to empty on validation failures. Matches the pre-existing pattern for `OUTPUT`/`REJECT`.
2. **Documented overwrite behavior** in the `@Argument` doc.
3. **Strengthened the integration test** to parse the `read_complete` event and assert `records_processed` equals the input record count (3 for `testLiftoverFailingVariants.vcf`). Previously verified field presence only — a wiring bug between `total` and `formatProgressEvent` would have slipped past.

Still open and addressable in this PR if the maintainers prefer: try/finally cleanup for the writer on `EXIT_CODE_WHEN_CONTIG_NOT_IN_REFERENCE` (point #2 in the PR body). The pattern matches existing leak behavior for `in`/`rejectedRecords`/`acceptedRecords` so I left it as-is; happy to address all four together if a try-with-resources refactor is welcome.

carstenerickson:
Pushed a hardening commit (4751f122a) after a deeper self-review. Extracts the sidecar writer into a small `picard.util.ProgressJsonWriter` and fixes several robustness gaps:

- **Streaming on bad chains:** read events now fire on a fixed input-record cadence rather than only on `ProgressLogger` milestones (which skip rejected variants), so a wrong chain file shows a rising reject rate live instead of only at `read_complete`.
- **UTF-8 + LF:** output is now explicit UTF-8 with `\n` terminators (was platform-default charset via `FileWriter`, and `println`'s platform line separator), so the NDJSON is valid on every platform.
- **Resource safety:** the run is wrapped in try-with-resources, so the writer is always closed — including the `EXIT_CODE_WHEN_CONTIG_NOT_IN_REFERENCE` early return and on exceptions — with proper suppression semantics.
- **Error surfacing:** write/open failures now raise `PicardException` instead of being swallowed by `PrintWriter` (and `PicardException` rather than htsjdk `SAMException` on open).
- **DISABLE_SORT:** emits a `write_skipped` sentinel (reporting the written count) since there's no separate write phase; per-stage `records_processed` semantics are documented in the arg doc.

The non-`PROGRESS_JSON` path is unchanged (console `ProgressLogger` output is byte-identical to before). Tests: `ProgressJsonWriterTest` (13) covers the formatter/escaper and on-disk UTF-8/LF output; `LiftoverVcfTest` gains a `write_skipped` integration test and an LF assertion. All green locally (the only suite failure is the pre-existing Apple-Silicon `IntelInflaterDeflaterLoadTest`, unrelated).

PR #2049, opened 2026-06-02
MarkDuplicates: add opt-in TAG_DUPLICATE_KEY emitting canonical fragment key (kf)

tfenne:
## Summary

Adds a default-off `TAG_DUPLICATE_KEY` option to `MarkDuplicates` that writes the **canonical single-end (fragment) duplicate key** to a new `kf` tag on every primary, mapped record. The value is a printable rendering of exactly the fields MarkDuplicates keys on at the fragment level, so an external tool can recover Picard's exact duplicate sets from the output BAM without re-deriving Picard's clipping/coordinate conventions.

This is **purely diagnostic**: it does not change duplicate marking, metrics, or any existing tag.

## Motivation

Downstream tooling sometimes needs to know *why* Picard grouped reads into a duplicate set — i.e. the key it actually compared on — rather than just the resulting flags or the opaque `DI`/`DS` set indices. Re-deriving that key externally means re-implementing Picard's unclipped-5′ coordinate, strand, library, and barcode conventions exactly. Emitting the key directly removes that fragile duplication of logic.

## The `kf` value

```
lib=<libraryId>;<referenceIndex>:<unclipped5Prime>:<strand>[:<barcodes>]
```

- `<strand>` is `F` (forward) or `R` (reverse).
- `<unclipped5Prime>` is the unclipped read end for reverse-strand reads, the unclipped read start otherwise — matching `buildReadEnds`.
- When barcodes/UMIs are in use (`BARCODE_TAG`, `READ_ONE_BARCODE_TAG`, or `READ_TWO_BARCODE_TAG`), the present barcode **sequences** are appended as a single hyphen-joined positional field, in the order MarkDuplicates keys on them (top-strand-normalized UMI, then read-one, then read-two barcode).

Examples:

| Configuration | `kf` |
|---|---|
| No barcodes (default) | `lib=1;0:1000:F` |
| `BARCODE_TAG` UMI `ACGTACGT` | `lib=1;0:1000:F:ACGTACGT` |
| UMI + read-one barcode | `lib=1;0:1000:F:ACTG-TTGA` |

Two reads of a pair each carry their own fragment key, so the canonical **pair** key is just the combination of the two ends; an orphan (mate unmapped) is a duplicate of a pair iff its key matches a pair end's key.

### Readable strings, not internal hashes

MarkDuplicates compares barcodes internally as 32-bit hashes. The `kf` tag deliberately emits the **verbatim sequences** instead, so the values are human-readable and reproducible by an external tool. The two representations agree on every realistic input; only on an astronomically unlikely hash collision would they differ, and in that case `kf` is the *more precise* of the two (it never produces sets coarser than Picard's, only — in that pathological case — finer).

### Tag name

`kf` is intentionally **lowercase**. The SAM specification reserves uppercase two-letter tags for standard, predefined use and sets aside lowercase tags for locally-defined purposes, which is where this diagnostic key belongs. (It therefore does not follow the uppercase convention of `DT`/`DI`/`DS`.)

## Behavior details

- **Default off.** When `TAG_DUPLICATE_KEY=false` (the default) no `kf` tag is written and behavior is unchanged.
- Written on exactly the records that participate in fragment keying — primary and mapped (`!readUnmappedFlag && !isSecondaryOrSupplementary`), the same guard MarkDuplicates uses when building its own `ReadEnds`.
- **Not supported with `FLOW_MODE`.** Flow mode matches fragments on a flow-adjusted coordinate within an uncertainty window rather than by the exact key rendered here, so combining the two options fails command-line validation with a clear message rather than emitting a misleading key. (Flow-aware key rendering could be added later.)

## Testing

New `MarkDuplicatesTagDuplicateKeyTest` (all test data built programmatically), covering:

- kf-induced duplicate-set partition equals the `DI`-induced partition (the load-bearing correctness property).
- Non-trivial canonicalization: differing per-end orientations, FF/RR, and inter-chromosomal pairs; explicit over-merge guard for FR vs RF at the same starts.
- Orphan (mate-unmapped) key matches a pair end's key and is marked duplicate.
- UMI-aware marking: reads at the same position with different UMIs stay in separate duplicate sets.
- Barcode fields appear only when a barcode tag is set; multiple barcode sequences are hyphen-joined into a single field.
- `READ_TWO_BARCODE_TAG` path renders the read-two barcode and distinguishes duplicates.
- `TAG_DUPLICATE_KEY` + `FLOW_MODE` is rejected.
- No `kf` tag is written when the option is off.

## Notes / limitations

- Diagnostic only — no change to marking, metrics, or existing tags.
- Not supported with `FLOW_MODE` (rejected at validation).
- The emitted key uses the per-run library configuration; a library has a single UMI/barcode configuration, so the hyphen-joined barcode rendering is unambiguous in practice.

----

#### Content

#### Review

### DIFF
--- a/src/main/java/picard/sam/markduplicates/MarkDuplicates.java
+++ b/src/main/java/picard/sam/markduplicates/MarkDuplicates.java
@@ -139,6 +139,32 @@ public class MarkDuplicates extends AbstractMarkDuplicatesCommandLineProgram imp
     public static final String DUPLICATE_SET_SIZE_TAG = "DS";
 
     /**
+     * The optional attribute used to store the canonical single-end (fragment) duplicate key that
+     * MarkDuplicates uses to identify duplicates. The value is a printable rendering of the fields
+     * the fragment-level key is built from: {@code lib=<libraryId>;<referenceIndex>:<unclipped5Prime>:<strand>}.
+     * When barcodes/UMIs are in use (any of BARCODE_TAG, READ_ONE_BARCODE_TAG, READ_TWO_BARCODE_TAG), the
+     * barcode sequences that also distinguish duplicates are appended as a single hyphen-joined field, giving
+     * {@code lib=<libraryId>;<referenceIndex>:<unclipped5Prime>:<strand>:<barcodes>}, where {@code barcodes}
+     * joins the present sequences in key order (the top-strand-normalized BARCODE_TAG UMI, then the
+     * READ_ONE_BARCODE_TAG and READ_TWO_BARCODE_TAG values) with hyphens, e.g. {@code ACGT} or {@code ACGT-TTGA}.
+     * It is written, when {@link #TAG_DUPLICATE_KEY} is set, on every primary, mapped record so that an
+     * external tool can recover the exact duplicate sets Picard computed without re-deriving the key.
+     * <p>
+     * Note that this key carries the verbatim barcode sequences, whereas MarkDuplicates compares barcodes
+     * internally as 32-bit hashes ({@link java.util.Objects#hash} of the UMI, {@link String#hashCode} of the
+     * read-one/read-two barcodes). The two agree on every realistic input; only in the astronomically unlikely
+     * event of a hash collision would they differ, and then this key is the more precise one (it distinguishes
+     * sequences Picard's hashes happened to collapse). In other words the key is never coarser than Picard's
+     * duplicate sets, and finer only in that pathological case.
+     * <p>
+     * The tag name is deliberately lowercase: the SAM specification reserves uppercase two-letter tags for
+     * standard, predefined use and sets aside lowercase tags for locally-defined purposes, which is where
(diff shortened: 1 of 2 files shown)


yfarjoun (changes requested):
TL;DR: The PR says it "removes that fragile duplication of logic," but buildDuplicateKeyTag recomputes the key independently of buildReadEnds, so it actually adds a second derivation pinned only by tests.

A fix would be to extract a shared helper for the key components (coordinate/orientation/libraryId/the barcode triple) that both buildReadEnds and buildDuplicateKeyTag consume. This will also fixes a drift already in the diff: buildReadEnds keeps three distinct barcode fields, but buildDuplicateKeyTag collapses them into one hyphen-joined string **with empties dropped**, so umi="",readOne="ACGT" and umi="ACGT",readOne=absent render identically and over-merges reads which Picard would keep separate.

